PR #85, opened 2014-07-20
Data standardization: z-score and 0-1 scaling

wildart:
Interface and implementation of data standardization (normalization) techniques. A proposed interface would allow scale data and reverse it to its original state. Currently available: z-score and 0-1 forward scaling.
These are commonly used scaling techniques in data analysis and machine learning.

The idea originated in MLBase package, but it was [proposed](https://github.com/JuliaStats/MLBase.jl/pull/9) to move the code into StatsBase. So my question is whether this functionality should be integrated in StatsBase or move it into separate package.


maximsch2:
Is there an update on this issue? There is a name clash right now between deprecated `MLBase.transform` (that this PR is proposing to move to StatsBase.jl) and `MultivariateStats.transform` while there is no reason why both of them can't co-exist (MultivariateStats even depends on StatsBase already, so moving `transform` here and extending it in MultivariateStats will solve the name clash).

mkborregaard:
This looks really useful, has it been left behind?

ararslan:
Yes, I think it's safe to assume that this PR has been abandoned, given that the last activity on it was 3 years ago. It predates GitHub's addition of "allow edits by maintainers," so we can't update it ourselves. Art is quite a busy fellow it seems, and I think there are other things that take up his time when he's around. So if someone wants to pick up where he left off, they can use the commits from his fork as a starting point. It would be nice to preserve the history should one choose to do that.

mkborregaard:
Thanks for the response, I was looking for the other PR and just fell over this. It'd obviously be great if someone were to pick this up.

wildart:
Back in business. Ready to merge.

mkborregaard:
That's amazing.

nalimilan:
Why did you drop the name `standardize` in favor of `transform`? I find the latter too general to export from StatsBase. Not of fan of `DataTransform`, `indim` and `outdim` either.

Also I haven't looked at this in detail, but it seems the code could be simplified now that we have broadcasting. Wouldn't it also be possible to use a single function for vectors and matrices, using the fact that `size(::AbstractVector, 2) == 1`?

wildart:
> Why did you drop the name standardize in favor of transform? 

As I recall, it was done in favor on standard api: `fit/transform/reconstruct` (see the beginning of this PR discussion). I believe this api was introduced here to reconcile with `MultivariateStats` api.

> I find the latter too general to export from StatsBase.

I thught that `StatsBase` is exactly the place for general api interface that can be inherited in dependent packages.

nalimilan:
Thanks.

Going back to the `transform` issue: I agree that it's fine in general to define a generic method in StatsBase for use in e.g. MultivariateStats and other packages, but I find it problematic to export `transform` as it's much broader name which applies beyond statistics. For example, DataFramesMeta provides `@transform` and could export its currently internal `transform` function at some point, but it would never depend on StatsBase.

We should also take into account another factor: StatsBase is supposed to progressively disappear, by moving features either to the Statistics stdlib module, or to more specialized packages (like StatsModels). At this point, I think there are two approaches regarding new features:
- it's is a very basic/essential feature, and the PR should be filed against Julia for inclusion in Statistics
- it's more specialized, and it should go to another package than StatsBase

I don't expect Statistics to export a general name like `transform`, so I'd say we should either find another name (and possibly use it for MultivariateStats too), or find another package.

Finally, the fact that we currently don't have any features which use `transform`, `reconstruct`, `indim` and `outdim` in StatsBase also gives me a pause. What is special about these transformations? Should we add more similar features in the future, and if so in which package/module? What's the scope of the `DataTransform` abstract type? Where should these go in the documentation?

nalimilan:
Funny coincidence: see [this post](https://discourse.julialang.org/t/optional-dependencies-and-name-collisions/15997) about a naming conflict with `transform!`.

nalimilan:
Maybe we could merge this without exporting `transform`/`transform!`, and continue the discussion after that? `indim` and `outdim` could be replaced with `getproperty`.

mkborregaard:
Sounds like a good idea.

wildart:
I removed `transform/reconstruct` exports as @nalimilan suggested.

src/transformations.jl
```diff
@@ -0,0 +1,208 @@
...
+function Base.getproperty(t::ZScoreTransform, p::Symbol)
+    if p == :indim || p == :outdim
+        return t.dim
+    else
+        return getfield(t, p)
+    end
+end
+
+# fit a z-score transform
+function fit(::Type{ZScoreTransform}, X::AbstractArray{T,2}; center::Bool=true, scale::Bool=true) where T<:Real
```
nalimilan:
```suggestion
function fit(::Type{ZScoreTransform}, X::AbstractArray{<:Real,2}; center::Bool=true, scale::Bool=true)
```

Same below.

src/transformations.jl
```diff
@@ -0,0 +1,208 @@
...
+    end
+end
+
+# fit a unit transform
+function fit(::Type{UnitRangeTransform}, X::AbstractArray{T,2}; unit::Bool=true) where T<:Real
+    d, n = size(X)
+
+    tmin = zeros(T, d)
+    tmax = zeros(T, d)
+    copyto!(tmin, X[:, 1])
```
nalimilan: Use `@views` to avoid a temporary copy. But wouldn't `tmin = X[:, 1]; tmax = copy(tmin)` work?

src/transformations.jl
```diff
@@ -0,0 +1,208 @@
...
+        @inbounds for i = 1:d
+            if X[i, j] < tmin[i]
+                tmin[i] = X[i, j]
+            elseif X[i, j] > tmax[i]
+                tmax[i] = X[i, j]
+            end
+        end
+    end
+    for i = 1:d
+        @inbounds tmax[i] = 1.0 / (tmax[i] - tmin[i])
```
nalimilan:
```suggestion
        @inbounds tmax[i] = 1 / (tmax[i] - tmin[i])
```
(This can help preserving the input type.)

src/transformations.jl
```diff
@@ -0,0 +1,208 @@
...
+    else
+        broadcast!(/, x, y, tscale)
+    end
+    return x
+end
+
+"""
+    standardize(DT, X; kwargs...)
+
+Return a row-standardized matrix `X` using `DT` transformation.
```
nalimilan: Can you list the supported transformations?

src/transformations.jl
```diff
@@ -0,0 +1,208 @@
...
+
+# Example
+```julia
```
nalimilan:
```suggestion
```jldoctest
```

Above, plural "Examples".

src/transformations.jl
```diff
@@ -0,0 +1,208 @@
...
+
+# reconstruct the original data from transformed values
+reconstruct!(t::S, x::AbstractArray{T,1}) where {T<:Real, S<:DataTransform} = reconstruct!(x, t, x)
+reconstruct!(t::S, x::AbstractArray{T,2}) where {T<:Real, S<:DataTransform} = reconstruct!(x, t, x)
+
+reconstruct(t::S, y::AbstractArray{T,1}) where {T<:Real, S<:DataTransform} = reconstruct!(similar(y), t, y)
+reconstruct(t::S, y::AbstractArray{T,2}) where {T<:Real, S<:DataTransform} = reconstruct!(similar(y), t, y)
+
+# Z-score transformation
+struct ZScoreTransform{T<:Real} <: DataTransform
```
nalimilan: Can you add docstrings for the transformations? Also, these features should be mentioned and docstrings included in the manual (create a new page if needed).

src/transformations.jl
wildart: I need type parameter later in the function body.

nalimilan: OK, but my remark was more general.

src/transformations.jl
```diff
@@ -121,21 +121,20 @@ struct UnitRangeTransform{T<:Real}  <: DataTransform
...
 # fit a unit transform
-function fit(::Type{UnitRangeTransform}, X::AbstractArray{T,2}; unit::Bool=true) where T<:Real
+function fit(::Type{UnitRangeTransform}, X::AbstractArray{T,2}; unit::Bool=true) where {T<:Real}
     d, n = size(X)
 
     tmin = zeros(T, d)
-    tmax = zeros(T, d)
-    copyto!(tmin, X[:, 1])
-    copyto!(tmax, X[:, 1])
+    copyto!(tmin, @view X[:, 1])
```
nalimilan: As I noted, wouldn't `tmin = X[:, 1]` be enough?

wildart: Right, copy.

wildart:
Something funny non-related to this PR happened on one of CI test runs, otherwise ready for merge.

docs/src/transformations.md
```diff
@@ -0,0 +1,39 @@
...
+learning techniques. These techniques might perform poorly if the individual
+features do not more or less look like standard normally distributed data.
+
+Standardization can be perform by using `ZScoreTransform` in the `fit` method.
+
+```@docs
+fit(::Type{ZScoreTransform}, X::AbstractArray{<:Real,2}; center::Bool=true, scale::Bool=true)
+```
+
+## Scaling
```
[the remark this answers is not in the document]
wildart: Well, standardization and normalization are some sort of scaling.

src/StatsBase.jl
```diff
@@ -194,7 +194,13 @@ module StatsBase
     r2,
     r²,
 
-    ConvergenceException
+    ConvergenceException,
+
+    # data standardization
+    standardize,
+    DataTransform,      # the type to represent a abstract data transformation
```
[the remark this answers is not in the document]
wildart: More like `AstractDataTransform`

docs/src/transformations.md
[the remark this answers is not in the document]
wildart: I renamed this part to "normalization".

docs/src/transformations.md
```diff
@@ -0,0 +1,43 @@
...
+
+Standardization can be performed using `fit(ZScoreTransform, ...)`.
+
+```@docs
+fit(::Type{ZScoreTransform}, X::AbstractArray{<:Real,2}; center::Bool=true, scale::Bool=true)
+```
+
+## Normalization
+
+**Normalization* is an alternative data transformation which scales features
```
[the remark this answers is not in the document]
wildart: Changed to "unit range normalization"

gdkrmr:
what is ETA on this? I have some code that depends on this and this discussion started 2014.

mkborregaard:
I think it's just awaiting @matbesancon 's review , otherwise ready to go in?

nalimilan:
Thanks @wildart! Sorry for the painful review process...

wildart:
Arrived. :clap:

nalimilan:
The more I think about it, the more I feel that `standardize` isn't the best name for this function, given that it's often used to refer to Z-score normalization (see also https://github.com/JuliaStats/StatsBase.jl/pull/648.). The generic term for these operations [seems to be "normalization"](https://en.wikipedia.org/wiki/Normalization_(statistics)), so we could overload `LinearAlgebra.normalize` instead (just like we already do for histograms). We could also use `scale`, which [seems to be quite common too](https://en.wikipedia.org/wiki/Feature_scaling).

Maybe we could also avoid using `StatsBase.transform(t::AbstractDataTransform, ...)`, which is a very broad name, and use both `normalize(t::AbstractDataTransform, X::AbstractArray)` and `normalize(DT::Type, X::AbstractArray)` (the latter being a shorthand for `normalize(fit(DT, X), X)` as these aren't ambiguous.

wildart:
Agree. I'm all for `normalize`.

mkborregaard:
@wildart why does not allow z-transforming or 0-1 transforming an entire matrix, but only rowwise or colwise?

wildart:
Usually, normalization is done for a sample to adjust its various features. In such case, doing it row/columnwise is quite natural. Applying normalization to a whole sample, assuming it's represent by a matrix, suggest that all features have similar properties, thus the normalization  becomes a mere affine transformation. You can do it easily with broadcast operations.

mkborregaard:
sure, I've done it with broadcast. I have encountered this requirement a couple of times in the wild, though, and think it would be great to have for completeness. But thanks, now I understand why it isn't in the current PR.

PR #117, opened 2015-06-10
wquantile

matthieugomez:
This function implements a weighted quantile. It is a generalization of quantile type 7 as implemented in Julia base (by that I mean results are the same with w = 1). 

The function follows the notation in the R help file for quantile, which hopefully will make it easy to add more quantile types. Only non negative weights are allowed.

I've added some tests and documentation.

This is my first pull request so I may be doing some things incorrectly. In particular, I'm not sure about the order of arguments (weight vs quantiles).

### DIFF
--- a/docs/source/scalarstats.rst
+++ b/docs/source/scalarstats.rst
@@ -169,7 +169,10 @@ Quantile and Friends
     w = rand(n)
     xk = median(x, weights(w))
 
+.. function:: quantile(x, w, p)
 
+  Compute the weighted quantiles of a vector ``x`` at a specified set of probability values ``p``, using weights given by a weight vector ``w`` (of type ``WeightVec``).  Weights must not be negative. The weights and data vectors must have the same length. The quantile for :math:`p` is defined as follows.  Denoting :math:`S_k = (k-1)w_k + (n-1) \sum_{i<k}w_i`, define :math:`x_{k+1}` the smallest element of ``x`` such that :math:`S_{k+1}/S_{n}` is strictly superior to :math:`p`. The function returns :math:`(1-\gamma) x_k + \gamma x_{k+1}` with  :math:`\gamma = (pS_n- S_k)/(S_{k+1}-S_k)`. This corresponds to  R-7, Excel, SciPy-(1,1), Maple-6 when ``w`` is one (see https://en.wikipedia.org/wiki/Quantile).
+  
 Mode and Modes
 ---------------
 
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -52,6 +52,7 @@ module StatsBase
     wmean,       # weighted mean
     wmean!,      # weighted mean across dimensions with provided storage
     wmedian,     # weighted median
+    wquantile,   # weighted quantile
 
     ## moments
     skewness,       # (standardized) skewness
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -293,3 +293,84 @@ end
 
 wmedian(v::RealVector, w::RealVector) = median(v, weights(w))
 wmedian{W<:Real}(v::RealVector, w::WeightVec{W}) = median(v, w)
+
+###### Weighted quantile #####
+
+# http://stats.stackexchange.com/questions/13169/defining-quantiles-over-a-weighted-sample
+# In the non weighted version, we compute a vector of index h(N, p)
+# and take interpolation between floor and ceil of this index
+# Here there is a supplementary function from index to weighted index k -> Sk
+
+function quantile{V, W <: Real}(v::RealVector{V}, w::WeightVec{W}, p::RealVector)
+
+    # checks
+    isempty(v) && error("quantile of an empty array is undefined")
+    isempty(p) && throw(ArgumentError("empty quantile array"))
+
+    w.sum == 0 && error("weight vector cannot sum to zero")
+    length(v) == length(w) || error("data and weight vectors must be the same size, got $(length(v)) and $(length(w))")
+    for x in w.values
+        isnan(x) && error("weight vector cannot contain NaN entries")
+        x < 0 && error("weight vector cannot contain negative entries")
+    end
+
+    # full sort
+    vw = sort!(collect(zip(v, w.values)))
+
+    wsum = w.sum
+
+    # prepare percentiles
+    ppermute = sortperm(p)
+    p = p[ppermute]
+    p = bound_quantiles(p)
+
+    # prepare out vector
+    N = length(vw)
+    out = Array(typeof(zero(V)/1), length(p))
+    fill!(out, vw[end][1])
+
+    # start looping on quantiles
+    cumulative_weight, Sk, Skold =  zero(W), zero(W), zero(W)
+    vk, vkold = zero(V), zero(V)
+    k = 1
+    for i in 1:length(p)
+        h = p[i] * (N - 1) * wsum
+        if h == 0
+            # happens when N or p or wsum equal zero
+            out[ppermute[i]] = vw[1][1]     
+        else
+            while Sk <= h
+                # happens in particular when k == 1
+                vk, wk = vw[k]
+                cumulative_weight += wk
+                if k >= N 
+                    # out was initialized with maximum v
+                    return out
+                end
+                k += 1
+                Skold, vkold = Sk, vk
+                vk, wk = vw[k]
+                Sk = (k - 1) * wk + (N - 1) * cumulative_weight
+            end
+            # in particular, Sk is different from Skold
+            g = (h - Skold) / (Sk - Skold)
+            out[ppermute[i]] = vkold + g * (vk - vkold)
+        end
+    end
+    return out
+end
+
+# similarly to statistics.jl in Base
+function bound_quantiles{T <: Real}(qs::AbstractVector{T})
+    epsilon = 100 * eps()
+    if (any(qs .< -epsilon) || any(qs .> 1+epsilon))
+        throw(ArgumentError("quantiles out of [0,1] range"))
+    end
+    T[min(one(T), max(zero(T), q)) for q = qs]
+end
+
+quantile{W <: Real}(v::RealVector, w::WeightVec{W}, p::Number) = quantile(v, w, [p])[1]
+wquantile{W <: Real}(v::RealVector, w::WeightVec{W}, p::RealVector) = quantile(v, w, p)
+wquantile{W <: Real}(v::RealVector, w::WeightVec{W}, p::Number) = quantile(v, w, [p])[1]
+wquantile(v::RealVector, w::RealVector, p::RealVector) = quantile(v, weights(w), p)
+wquantile(v::RealVector, w::RealVector, p::Number) = quantile(v, weights(w), [p])[1]
--- a/test/weights.jl
+++ b/test/weights.jl
@@ -233,3 +233,104 @@ wt = [-1, -1, -1, -1, -1]
 @test_throws ErrorException median(data, weights(wt))
 wt = [-1, -1, -1, 0, 0]
 @test_throws ErrorException median(data, weights(wt))
+
+
+# Weighted quantile tests
+data = (
+    [7, 1, 2, 4, 10],
+    [7, 1, 2, 4, 10],
+    [7, 1, 2, 4, 10, 15],
+    [1, 2, 4, 7, 10, 15],
+    [0, 10, 20, 30],
+    [1, 2, 3, 4, 5],
+    [1, 2, 3, 4, 5],
+    [30, 40, 50, 60, 35],
+    [2, 0.6, 1.3, 0.3, 0.3, 1.7, 0.7, 1.7],
+    [1, 2, 2],
+    [3.7, 3.3, 3.5, 2.8],
+    [100, 125, 123, 60, 45, 56, 66],
+    [2, 2, 2, 2, 2, 2],
+    [2.3],
+    [-2, -3, 1, 2, -10],
+    [1, 2, 3, 4, 5],
+    [5, 4, 3, 2, 1],
+    [-2, 2, -1, 3, 6],
+    [-10, 1, 1, -10, -10],
+)
+wt = (
+    weights([1, 1/3, 1/3, 1/3, 1]),
+    weights([1, 1, 1, 1, 1]),
+    weights([1, 1/3, 1/3, 1/3, 1, 1]),
+    weights([1/3, 1/3, 1/3, 1, 1, 1]),
+    weights([30, 191, 9, 0]),
+    weights([10, 1, 1, 1, 9]),
+    weights([10, 1, 1, 1, 900]),
+    weights([1, 3, 5, 4, 2]),
+    weights([2, 2, 5, 1, 2, 2, 1, 6]),
+    weights([0.1, 0.1, 0.8]),
+    weights([5, 5, 4, 1]),
+    weights([30, 56, 144, 24, 55, 43, 67]),
+    weights([0.1, 0.2, 0.3, 0.4, 0.5, 0.6]),
+    weights([12]),
+    weights([7, 1, 1, 1, 6]),
+    weights([1, 0, 0, 0, 2]),
+    weights([1, 2, 3, 4, 5]),
+    weights([0.1, 0.2, 0.3, 0.2, 0.1]),
+    weights([1, 1, 1, 1, 1]),
+)
+quantile_answers = (    
+    [1.0,3.6000000000000005,6.181818181818182,8.2,10.0],
+    [1.0,2.0,4.0,7.0,10.0],
+    [1.0,4.75,8.0,10.833333333333334,15.0],
+    [1.0,4.75,8.0,10.833333333333334,15.0],
+    [0.0,6.1387900355871885,11.600000000000001,15.912500000000001,30.0],
+    [1.0,1.5365853658536586,2.5999999999999996,4.405405405405405,5.0],
+    [1.0,4.239377950569287,4.492918633712858,4.746459316856429,5.0],
+    [30.0,38.75,45.714285714285715,52.85714285714286,60.0],
+    [0.3,0.6903846153846154,1.484,1.7,2.0],
+    [1.0,2.0,2.0,2.0,2.0],
+    [2.8,3.3361111111111112,3.4611111111111112,3.581578947368421,3.7],
+    [45.0,59.88593155893536,100.08846153846153,118.62115384615385,125.0],
+    [2.0,2.0,2.0,2.0,2.0],
+    [2.3,2.3,2.3,2.3,2.3],
+    [-10.0,-5.52,-2.5882352941176467,-0.9411764705882351,2.0],
+    [1.0,1.75,4.25,4.625,5.0],
+    [1.0,1.625,2.3333333333333335,3.25,5.0],
+    [-2.0,-0.5384615384615388,1.5384615384615383,2.6999999999999997,6.0],
+    [-10.0,-10.0,-10.0,1.0,1.0]
+)
+p = [0.0, 0.25, 0.5, 0.75, 1.0]
+
+srand(10)
+for i = 1:length(data)
+    @test_approx_eq quantile(data[i], wt[i], p) quantile_answers[i]
+    for j = 1:10
+        # order of p does not matter
+        reorder = sortperm(rand(length(p)))
+        @test_approx_eq quantile(data[i], wt[i], p[reorder]) quantile_answers[i][reorder]
+    end
+    for j = 1:10
+        # order of w does not matter
+        reorder = sortperm(rand(length(data[i])))
+        @test_approx_eq quantile(data[i][reorder], weights(wt[i][reorder]), p) quantile_answers[i]
+    end
+end
+# w = 1 corresponds to base quantile
+for i = 1:length(data)
+    @test_approx_eq quantile(data[i], weights(ones(Int64, length(data[i]))), p) quantile(data[i], p)
+    for j = 1:10
+        prandom = rand(4)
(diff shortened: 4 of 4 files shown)


johnmyleswhite:
Just to clarify: you didn't read any R source code when implementing this?

docs/source/scalarstats.rst
```diff
@@ -158,6 +158,10 @@ Quantile and Friends
     w = rand(n)
     xk = median(x, weights(w))
 
+.. function:: wquantile(x, w, q)
+
+  Compute the weighted quantile of ``x``, using weights given by a weight vector ``w`` (of type ``WeightVec``).  Weights must not be negative. The weight and data vectors must have the same length `n`. Quantiles as defined as follow: assume for simplicity that weights sum to `n`. Denote :math:`x_h` the smallest element of ``x`` such that the cumulated sum up of weights up to `x_h` is strictly superior to :math:`lv * q + m`. Denote :math:`x_l` the element just before `x_h`. The sample quantile is defined as `(1-gamma) x_l + gamma x_h`.
```
nalimilan: "The weight and data vectors" -> "weights"
"as follow" -> "as follows"
"for simplicity": looks like you can always assume this, right? Better remove these words.
"cumulated sum up of weights up to" -> "cumulative sum of weights up to"
"gamma" -> "\gamma"
Also, looks like several expressions lack `:math:`.

nalimilan: Oh, and you say "quantile" (singular) while a vector of probabilities is supported. It would make sense to use the plural, and to say a word about the probabilities, following the vocabulary used for `?quantile`.

docs/source/scalarstats.rst
```diff
@@ -158,6 +158,10 @@ Quantile and Friends
     w = rand(n)
     xk = median(x, weights(w))
 
+.. function:: wquantile(x, w, q)
```
nalimilan: Arguments are not in the same order as in the code. Also, better rename `q` to `p` to follow Base.

src/weights.jl
```diff
@@ -291,3 +291,81 @@ end
 
 wmedian(v::RealVector, w::RealVector) = median(v, weights(w))
 wmedian{W<:Real}(v::RealVector, w::WeightVec{W}) = median(v, w)
+
+
+
+
+###### Weighted quantile #####
+# type 7 for now, as in base
```
nalimilan: Could you move this mention of "type 7" to the docs? Ideally, if you could give a reference about this (and alternative) definitions, that would be very helpful.

src/weights.jl
```diff
@@ -291,3 +291,81 @@ end
...
+    q = q[qpermute]
+    q = bound_quantiles(q)
+
+    if w.sum == 0
+        error("weight vector cannot sum to zero")
+    end
+    if length(v) != length(w)
+        error("data and weight vectors must be the same size")
+    end
+    @inbounds for x in w.values
```
nalimilan: I don't think `@inbounds` has any effect here (nor below).

src/weights.jl
```diff
@@ -291,3 +291,81 @@ end
...
+    # make sure the quantiles are in [0,1]
+    qpermute = sortperm(q)
+    q = q[qpermute]
+    q = bound_quantiles(q)
+
+    if w.sum == 0
+        error("weight vector cannot sum to zero")
+    end
+    if length(v) != length(w)
+        error("data and weight vectors must be the same size")
```
nalimilan: Could you add `"got $length(v) and $length(w)"`? Always useful to provide more details.

nalimilan:
Thanks! I've only had a look at most obvious things.

Indeed, @johnmyleswhite is right that you must not have taken inspiration from the R code to write this, or it can't be licensed under MIT. It's a bit hard to use their notation without reading their code, though you can always make an implementation from scratch just based on the behaviour described in R docs.

matthieugomez:
Thanks for your comments. I changed my code accordingly. 

I don't understand what coverage/coveralls is so I'm not sure why it fails.

No I did not look at the R code (which does not compute weighted quantile). 

A question about your workflow: when I modify a local copy of a package like StatsBase, how can I install this local version as a package? Also, is there a way to load every function in this local version, including non exported function?

nalimilan:
Coveralls is used to measure the portion of the code which is covered by tests. Apparently there are some issues with it at the moment.

When you modify a package in `~/.julia/Package`, it's automatically used. You need to checkout the original git branch to use the official version again. I'm not sure how to load all functions from a package; maybe simply by calling `include` on the toplevel file of the package.

docs/source/scalarstats.rst
```diff
@@ -158,6 +158,9 @@ Quantile and Friends
     w = rand(n)
     xk = median(x, weights(w))
 
+.. function:: wquantile(x, p, w)
```
nalimilan: `q` is `p` in the code now.

docs/source/scalarstats.rst
```diff
@@ -158,6 +158,9 @@ Quantile and Friends
...
+
+  Compute the weighted quantile of ``x``, using weights given by a weight vector ``w`` (of type ``WeightVec``).  Weights must not be negative. The weights and data vectors must have the same length. The quantile for :math:`p` is defined as follows. Normalizing ``w`` so that the sum of weights equals :math:`n`, the length of ``x``, and denoting :math:`x(h)` the smallest element of ``x`` such that the cumulated sum of weights up to :math:`x(h)` is strictly superior to :math:`n p + m`, the function returns :math:`(1-\gamma) x(h-1) + \gamma x(h)`, with  :math:`m=1-p` and :math:`\gamma = n p + m - \sum_{i < h} w_i`. This corresponds to the unweighted quantile function when ``w`` is 1 (type 7 in R).
```
nalimilan: You still don't use the plural, and don't specify that a vector of probabilities can be passed too.

Also, regarding the type, you could give references to other software than R. Wikipedia says this method is "R-7, Excel, SciPy-(1,1), Maple-6". https://en.wikipedia.org/wiki/Quantile

docs/source/scalarstats.rst
nalimilan: Or rather the contrary.

test/weights.jl
```diff
@@ -226,3 +226,95 @@ wt = [-1, -1, -1, -1, -1]
...
+    [-10.0,-10.0,-10.0,1.0,1.0],
+)
+
+
+
+
+# w = 1 corresponds to base quantile
+for i = 1:length(data)
+    for j = 1:100
+        q = rand(length(data[i]))
```
nalimilan: I don't think tests on random data are really appreciated around here. Getting errors at random can be confusing, especially given that testing has been quite unstable for some time. You should probably add a call to `srand`, which people can remove to test the algorithm more extensively. But others will likely comment.

src/weights.jl
```diff
@@ -291,3 +291,82 @@ end
...
+    for x in v
+        isnan(x) && return x
+    end
+
+    lv = length(v)
+    lp = length(p)
+    m = 1 - p
+    index = (lv * p + m) * w.sum / lv
+    wt = w.values
+    vpermute = sortperm(v)
```
matthieugomez: Instead of  finding the permutation that orders x, I need to return the permutation that orders (x,w). Otherwise the algorithm does not give a stable value if v/w are reordered.
What's the best way to do that?

StefanKarpinski:
@matthieugomez, can you answer the question about whether you read the R source when writing this code? If you did, then I'm afraid that iterating over this patch is a waste of time since the project can't accept it in any case.

johnmyleswhite:
> No I did not look at the R code (which does not compute weighted quantile btw).

He answer it, @StefanKarpinski

src/weights.jl
```diff
@@ -291,3 +291,82 @@ end
 
 wmedian(v::RealVector, w::RealVector) = median(v, weights(w))
 wmedian{W<:Real}(v::RealVector, w::WeightVec{W}) = median(v, w)
+
+
+
+
+###### Weighted quantile #####
+# Variables m, g, low, high, correspond respectively to m, gamma, x[j], x[j+1] in the R help file for quantile
```
nalimilan: FWIW, you can use `γ` to be closer to the math in Julia.

src/weights.jl
nalimilan: I guess you could take inspiration from `sortrows`. Adding a version of it that would return the permutation has been requested: https://github.com/JuliaLang/julia/issues/9258

Unfortunately, it would require you to create a matrix holding a copy of `v` and `w`, which would be terribly wasteful AFAICT. So maybe you would need to write a custom function based on `sortperm` directly.

matthieugomez: It seems like there are too solutions

```julia
n = 1_000_000
a = rand(1:n, n)
b = rand(1:10, n)

@time sortperm([(a[i], b[i]) for i in 1:length(a)])
#elapsed time: 9.676994989 seconds   (287943584 bytes allocated, 16.39% gc time)
@time sortperm([(sub(a,i)[1], sub(b,i)[1]) for i in 1:length(a)], order= Base.Order.Lexicographic)
#elapsed time: 9.446759537 seconds (560247620 bytes allocated, 17.83% gc time)
```

I don't really understand the difference between these two solutions. Both are very slow (sorting `a` only takes 0.237).

nalimilan: I don't think the difference in timings between the two is very significant. Anyway, these are clearly not acceptable compared to your current (unstable) solution.

When I said "custom function based on `sortperm`", I was thinking of adapting `sortperm`'s code to sort without creating an array of tuples, to avoid these allocations. Not sure how hard it would be.

I also wonder whether there might be another solution than sorting. Don't you think there could exist a rounding rule which could fix this issue? Honestly, I haven't investigated the algorithm at all.

matthieugomez: Indeed. My last commit fixes it by just checking the tricky cases, w/o performance issue.

nalimilan: Ah, great!

src/weights.jl
```diff
@@ -291,3 +291,96 @@ end
...
+            high[k] = v[vpermute[i]]
+            low[k] = v[vpermute[max(i-1, 1)]]  
+            # Below we catch the case where an element equal to v[vpermute[i]] has a weight lower than threshold. 
+            # For instance median in (1, 2, 2) (0.1, 0.8, 0.1)
+            # In this case, we just need to set low[k] = high[k]
+            if low[k] != high[k] 
+                l = 1
+                wthreshold = cumulative_weight - wt[vpermute[i-1]] - index[k] 
+                # loop over posterior elements with same value
+                while (i + l <= lv && v[vpermute[i+l]] == high[k])
```
nalimilan: The standard Julia style is to put no parentheses after `while`.

src/weights.jl
```diff
@@ -291,3 +291,96 @@ end
...
+    g = fill(0.0, lp)
+    cumulative_weight = zero(eltype(wt))
+    k = 1
+    i = 0
+    while (k <= lp) & (i <= lv)
+        if cumulative_weight > index[k]
+            g[k] = (index[k] - (cumulative_weight - wt[vpermute[i]])) * lv / w.sum
+            high[k] = v[vpermute[i]]
+            low[k] = v[vpermute[max(i-1, 1)]]  
+            # Below we catch the case where an element equal to v[vpermute[i]] has a weight lower than threshold.
```
nalimilan: Maybe you could explain what the threshold corresponds to? I mean, I can see that you check for values lower than it, but it's harder to understand what it means. :-)

src/weights.jl
```diff
@@ -291,3 +291,96 @@ end
...
+    index = (lv * p + m) * w.sum / lv
+    wt = w.values
+    vpermute = sortperm(v)
+    low = fill(v[vpermute[end]], lp)
+    high = fill(v[vpermute[end]], lp)
+    g = fill(0.0, lp)
+    cumulative_weight = zero(eltype(wt))
+    k = 1
+    i = 0
+    while (k <= lp) & (i <= lv)
```
nalimilan: `&&`intended here?

src/weights.jl
```diff
@@ -291,3 +291,96 @@ end
...
+            end
+            k += 1
+        else
+            i += 1
+            if i <= lv
+                cumulative_weight += wt[vpermute[i]]
+            end
+        end
+    end
+    out = Array(Real, lp)
```
nalimilan: You want to use a concrete type here -- I guess `eltype(v)`.

src/weights.jl
nalimilan: Did you had the problematic cases to the tests?

src/weights.jl
```diff
@@ -291,3 +291,96 @@ end
...
+    lv = length(v)
+    lp = length(p)
+    m = 1 - p
+    index = (lv * p + m) * w.sum / lv
+    wt = w.values
+    vpermute = sortperm(v)
+    low = fill(v[vpermute[end]], lp)
+    high = fill(v[vpermute[end]], lp)
+    g = fill(0.0, lp)
+    cumulative_weight = zero(eltype(wt))
```
nalimilan: I'm not sure it really makes sense to use the element type of `wt` here. If the weights were stored e.g. in `Float32` or `Int32` to save space, summing them will quickly lead to overflow. And anyway you're doing many computations/comparisons with `Float64` values, so it may well be faster to do everything in `Float64`.

But I see this is how `median` currently works. @lindahua, any comments?

src/weights.jl
matthieugomez: I don't think so. if v is an int, the quantile can be a float. So should I write ` out = Array(Float64, lp)`?

src/weights.jl
nalimilan: Actually, it just occurred to me that a perfect type for this would be `typeof(w.sum)`, as by definition it's known to be able to hold the sum of all weights.

src/weights.jl
nalimilan: Yes. In Julia, functions must generally be type-stable for performance. This means they should always return the same type given their input types. Since there's a possibility that we may need to return a float, we must always return this type.

src/weights.jl
```diff
@@ -291,3 +291,96 @@ end
...
+    p = p[ppermute]
+
+    # make sure the quantiles are in [0,1]
+    p = bound_quantiles(p)
+
+    if w.sum == 0
+        error("weight vector cannot sum to zero")
+    end
+    if length(v) != length(w)
+        error("data and weight vectors must be the same size. got $(length(v)) and $(length(w))")
```
nalimilan: Comma instead of period here.

src/weights.jl
nalimilan: Even better would be to choose the type based on that of the `(1 - g[k]) * low[k] + g[k] * high[k]` expression, which would preserve e.g. `BigFloat` or BigInt` weights (not sure it's very useful, but...). Not sure how to do that cleanly, except maybe with a comprehension.

src/weights.jl
```diff
@@ -291,3 +291,96 @@ end
...
+    end
+    lv = length(v)
+    lp = length(p)
+    m = 1 - p
+    index = (lv * p + m) * w.sum / lv
+    wt = w.values
+    vpermute = sortperm(v)
+    low = fill(v[vpermute[end]], lp)
+    high = fill(v[vpermute[end]], lp)
+    g = fill(0.0, lp)
```
nalimilan: Just like for the type of `out`, here it *might* be better to choose the type so that it can hold the results of the computations with the required precision.

src/weights.jl
```diff
@@ -291,3 +291,104 @@ end
...
+    end
+    lv = length(v)
+    lp = length(p)
+    index = p * (lv - 1) * w.sum
+    wt = w.values
+    vpermute = sortperm(v)
+    lo = fill(v[vpermute[end]], lp)
+    hi = fill(v[vpermute[end]], lp)
+    g = fill(0.0, lp)
+    cumulative_weight = convert(typeof(w.sum), 0)
```
nalimilan: You can write this as `zero(w.sum)`.

nalimilan:
The tests fail since you rebased, looks like you forgot to export the function.

matthieugomez:
I removed wquantile (instead one should use quantile(v, weight(w))). I also defined quantile for AbstractVector instead of just Real but I'm not sure this is correct.  

I also used the weighted quantile function to improve the function `describe`, adding a weighted and a detailed option.

nalimilan:
I'm all for using `quantile`, but as long as `wmedian` is exported, it makes sense to export `wquantile`. Else people might think it's not supported. Regarding the type, I think it's better to keep `RealVector`, which is used everywhere at the moment. If this needs to be changed, better do that separately for all functions in a different PR.

src/scalarstats.jl
```diff
@@ -348,14 +397,147 @@ function summarystats{T<:Real}(a::AbstractArray{T})
...
     @printf(io, "Maximum:      %.6f\n", ss.max)
+    print("\n")
+    @printf(io, "Sum of weight: %.6f\n", ss.wsum)
 end
 
-describe{T<:Real}(a::AbstractArray{T}) = show(summarystats(a))
+
+
+function describe{T<:Real}(a::AbstractArray{T}; detail = false) 
+    detail? show(detailedsummarystats(a)) : show(summarystats(a))
```
nalimilan: Please add a space before `?` (same below).

src/scalarstats.jl
```diff
@@ -348,14 +397,147 @@ function summarystats{T<:Real}(a::AbstractArray{T})
...
+    @printf(io, "p25:          %.6f\n", ss.p25)
     @printf(io, "Median:       %.6f\n", ss.median)
-    @printf(io, "3rd Quartile: %.6f\n", ss.q75)
+    @printf(io, "p75           %.6f\n", ss.p75)
+    @printf(io, "p90:          %.6f\n", ss.p90)
+    @printf(io, "p95:          %.6f\n", ss.p95)
+    @printf(io, "p99:          %.6f\n", ss.p99)
     @printf(io, "Maximum:      %.6f\n", ss.max)
+    print("\n")
+    @printf(io, "Sum of weight: %.6f\n", ss.wsum)
```
nalimilan: "weights".

src/scalarstats.jl
```diff
@@ -348,14 +397,147 @@ function summarystats{T<:Real}(a::AbstractArray{T})
...
+function Base.show(io::IO, ss::WeightedSummaryStats)
+    println(io, "Summary Stats:")
+    @printf(io, "Mean:          %.6f\n", ss.mean)
+    @printf(io, "Minimum:       %.6f\n", ss.min)
+    @printf(io, "1st Quartile:  %.6f\n", ss.p25)
+    @printf(io, "Median:        %.6f\n", ss.median)
+    @printf(io, "3rd Quartile:  %.6f\n", ss.p75)
+    @printf(io, "Maximum:       %.6f\n", ss.max)
+    print("\n")
+    @printf(io, "Sum of weight: %.6f\n", ss.wsum)
```
nalimilan: "weight"

src/scalarstats.jl
```diff
@@ -348,14 +397,147 @@ function summarystats{T<:Real}(a::AbstractArray{T})
...
     @printf(io, "Mean:         %.6f\n", ss.mean)
     @printf(io, "Minimum:      %.6f\n", ss.min)
-    @printf(io, "1st Quartile: %.6f\n", ss.q25)
+    @printf(io, "1st Quartile: %.6f\n", ss.p25)
+    @printf(io, "Median:       %.6f\n", ss.median)
+    @printf(io, "3rd Quartile: %.6f\n", ss.p75)
+    @printf(io, "Maximum:      %.6f\n", ss.max)
+end
+
+function Base.show(io::IO, ss::WeightedSummaryStats)
```
nalimilan: There's a lot of duplication here. Can't you call the unweighted version, and then output the sum of weights?

src/StatsBase.jl
```diff
@@ -37,13 +37,15 @@ module StatsBase
...
 
     ## weights
-    WeightVec,   # the type to represent a weight vector
-    weights,     # construct a weight vector
-    wsum,        # weighted sum with vector as second argument
-    wsum!,       # weighted sum across dimensions with provided storage
-    wmean,       # weighted mean
-    wmean!,      # weighted mean across dimensions with provided storage
-    wmedian,     # weighted median
+    WeightVec,     # the type to represent a weight vector
```
nalimilan: Please avoid unneeded space changes here.

matthieugomez:
Thanks for the feedback again. What's the best way to use the code for the method of one type to define the method for another type ? Define Abstract type? Write a convert function from weighted to unweighted summary ? I don't see how this would make the code simpler.

nalimilan:
I think you can define an abstract type, say `AbstractSummaryStats`, from which both types inherit. Then the unweighted version will apply to `AbstractSummaryStats`, so that you can call it without any conversion from the weighted version.

nalimilan:
Even simpler: you could merge all (non-)detailed and (un-)weighted printing methods into one method parameterized on `T <: AbstractSummaryStats`, and just do `if T == WeightedSummaryStats` before the line printing the weights and `if T == DetailedSummaryStats`  before the lines printing detailed stats.

matthieugomez:
I've simplified the `wquantile` code and added tests for the `describe` method. What would I need to do to have it merged?

andreasnoack:
You'll have to rebase. @nalimilan You've done most of the review, is this ready for merging?

aviks:
Bump?

matthieugomez:
It's been a while. Can this be merged? 
I've suppressed the WeightedSummaryStats bit at it seems better to do that in a different pull request.

johnmyleswhite:
This looks good to merge. What's going on with Travis?

andreasnoack:
There have been some certificate problems on the nightlies, but they should vbe fixed now. I've restarted the build and it seems to work.

johnmyleswhite:
Thanks for working on this, @matthieugomez.

PR #118, opened 2015-06-13
Move weights argument before mean in stdm(), varm() and moment()

nalimilan:
This is consistent with mean(), skewness() and kurtosis().

I was struck by this inconsistency while reviewing these lines: https://github.com/JuliaStats/StatsBase.jl/pull/117/files#diff-87bbeceb1714b7f9134da828801ba6d9R440

I realize I could as well have chosen to follow the convention of `stdm`, `varm` and `moment` of putting means before, but it sounded more logical to me to put the weight before, as they are more common (means are only accepted by some functions, and likely used less often). Any thoughts?

### DIFF
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -1,2 +1,14 @@
 import Base.@deprecate
 import Base.depwarn
+
+import Base.varm, Base.stdm
+@deprecate varm(v::RealArray, m::Real, wv::WeightVec) varm(v, wv, m)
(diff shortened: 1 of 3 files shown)


johnmyleswhite:
Sounds good to me.

nalimilan:
Can somebody merge this, since @johnmyleswhite said it's OK?

nalimilan:
Thanks!

PR #119, opened 2015-06-23
Coeftable

matthieugomez:
This commit changes the way coeftable are printed by making them more similar to [Python](http://blog.yhathq.com/posts/logistic-regression-and-python.html) or [Stata](http://www.ats.ucla.edu/stat/stata/output/reg_output.htm)

Coeftable type gets new fields
- Title
- vector and vectornms. These vectors can store stuff like number of obs, name of dependent variabble, R^2.

In the show method for Coeftable, these fields are printed before the table when they are non empty.

Moreover, I have changed the pvalue column to round at 1e4 always, so a value lower than 0.0005 is printed as 0.000. This rounding corresponds to the way [Stata](http://www.ats.ucla.edu/stat/stata/output/reg_output.htm),  [Python](http://blog.yhathq.com/posts/logistic-regression-and-python.html), [SAS](http://www.ats.ucla.edu/stat/sas/output/reg.htm) use when they display regression results. In my opinion, this is a better solution because it makes the table easier to read at a glance.


simonster:
Can you provide examples of the old and new output?

The current approach to printing p-values was the product of the discussion in #64. I still prefer it to the alternatives, but I may be biased.

matthieugomez:
Some images of the new behavior from the test:
screen shot 2015-06-22 at 9 56 17 pm

johnmyleswhite:
I agree with @simonster: the current format seems superior to using `0.000`.

@matthieugomez: I really appreciate the work you're doing on JuliaStats. It's great to have new contributors and I'm excited to see where you'll go as you do more work.

Given your interest in contributing, I wonder if your work might be more impactful if focused on a project that involves fewer decisions of aesthetics in the way that this change does. We badly need people to work on NullableArrays. Most importantly, you'd be operating in sync with an active project, rather than making changes to a project that no one with commit access is actively working on. If you're interested in making the transition from occasional contributor to long-term committer, I think it would help a lot to focus your work on a project that has people who can help you on a more consistent basis.

nalimilan:
I agree that p-values were better in the previous version (I wrote those lines :-). But I must say I find the new version nicer in other aspects. Though, I thought details like number of observations or degrees of freedom were more commonly put below the table? Bikeshedding is open!

matthieugomez:
I returned to the default printing behavior of p-value.

@johnmyleswhite Thanks! The issues / pull requests I've submitted recently just correspond to things I had to code when writing a personal package. I'm submitting some parts as pull requests since they could be useful to more people. Feel free to disregard / merge them in a while - I get that pretty printing is not a priority right now

nalimilan:
@matthieugomez Any interest in reviving this? I see you still carry a copy of `CoefTable` in FixedEffectsModels, which sounds too bad. Other packages would benefit from printing fit statistics above or below the model.

PR #120, opened 2015-07-02
add df_residuals

matthieugomez:
`df_residual` is defined in GLM. I think it should be defined in StatsBase for every `RegressionModel`:
- The only way (I think) to add a `df_residual` method in another package without overwriting it is to import GLM.
- GLM defines a `coeftable` function for `LinearModel`
https://github.com/JuliaStats/GLM.jl/blob/831f5a5ea3613e342e5a63cd4051e1a437b7df2e/src/lm.jl#L68
which only requires `coef`, `nobs`, `stderr` and  `df_residuals`, If `df_residuals` was defined for every RegressionModel, the `coeftable` function would work on every `RegressionModel` (and could even be moved to `StatsBase`). The bottom line is that `df_residuals` is needed to do any kind of test so it would be nice to add it generically.

I've just changed the GLM name `df_residual` to `df_residuals` since it seems more consistent with the `s` of `residuals`.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -155,7 +155,7 @@ module StatsBase
     coeftable,
     confint,
     deviance,
-    df_residuals,
+    df_residual,
     fit,
     fit!,
     fitted,
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -155,6 +155,7 @@ module StatsBase
     coeftable,
     confint,
     deviance,
+    df_residuals,
     fit,
     fit!,
     fitted,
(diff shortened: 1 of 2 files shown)


nalimilan:
Makes sense, but it should be called `df_residual`. Can you also make a PR against GLM to make it import `StatsBase.df_residual`?

matthieugomez:
OK. Out of curiosity, why `df_residual` rather than `df_residuals`? R?

nalimilan:
Well, that function does not return "residuals", it returns "residual degrees of freedom". It would even be clearer to call it `residual_df`. If we start the bikeshedding, we could also try to find something shorter, like `resdf`.

matthieugomez:
ahah ok I saw it as degree of freedom of residuals.

andreasnoack:
@matthieugomez Sorry for the delay

PR #121, opened 2015-07-07
Weighted median: Handle equal weights

julian-gehring:
Fixes the weighted median if all observations have equal weights.

A weighted median with equal weights for all observations is essentially an unweighted median. The old implementation did not cover special cases of this correctly. Here a short example:

```julia
median([2, 4], WeightVec([1, 1]))
```

used to return `2.0`, whereas it should return `3.0` (the mean of the two values). For a comparison, see also the unweighted median in Base:

```julia
median([2, 4])
```

which return `3.0` as expected.

Test cases for this are included.


nalimilan:
Thanks for catching this corner case! I've put some comments inline.

src/weights.jl
```diff
@@ -257,35 +257,37 @@ function Base.median{W<:Real}(v::RealVector, w::WeightVec{W})
...
-            i = 0
-            for (i, p) in enumerate(permute)
-                if cumulative_weight > midpoint
-                    cumulative_weight -= wt[p]
-                    break
-                end
-                cumulative_weight += wt[p]
-            end
+    if !any(mask)
+        error("no nonzero weights found")
```
johnmyleswhite: Can we avoid the double negative? I got confused at first by this. Perhaps "at least one weight must be non-zero"?

johnmyleswhite:
Made a small comment. Otherwise I'd say this is good to be merged once it's rebased. Thanks for working on this and fixing such a serious bug.

src/weights.jl
julian-gehring: Fixed with b5dbbe7a9f2b4518571e6634dea28f2e4f1a97ea

johnmyleswhite:
Squash?

julian-gehring:
Rebased in b034c12c629de7279ea32b7e538c27751e1ef23a

PR #122, opened 2015-07-14
ecdf: Sort in place

julian-gehring:
The original data is not needed after the initial sorting, so we can sort in
place. This is easier to read and more efficient.

### DIFF
--- a/src/empirical.jl
+++ b/src/empirical.jl
@@ -4,10 +4,10 @@
 ## Empirical CDF
 
 function ecdf{T<:Real}(X::RealVector{T})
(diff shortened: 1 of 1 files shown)


johnmyleswhite:
This can't be the default behavior of `ecdf`. I wouldn't be opposed to having a function called `ecdf!` that did this, but this kind of impurity is too extreme for default use cases.

julian-gehring:
Ah, I didn't consider the fact that this would affect the original variable in the workspace as well. Fully agree with your concern.

Issue #123, opened 2015-07-22
trimmean sums over unsorted vector


bjarthur:
if i understand the code correctly, this:

```
    for i = (1+nl) : (n-nh)
        @inbounds s += x[i]
    end
```

should be this:

```
    for i = (1+nl) : (n-nh)
        @inbounds s += sx[i]
    end
```

bjarthur:
to wit:

```
julia> trimmean([-100, 2, 3, 7, 200], 0.4)
4.0

julia> trimmean([2, -100, 200, 3, 7], 0.4)
34.333333333333336
```

i'm working on a PR

bjarthur:
fixed via https://github.com/JuliaStats/StatsBase.jl/pull/126

PR #126, opened 2015-07-27
trimmean now works on unsorted vectors

bjarthur:
see https://github.com/JuliaStats/StatsBase.jl/issues/123


johnmyleswhite:
Thanks for fixing this!

PR #130, opened 2015-08-04
check inputs for entropy functions

ahwillia:
At the moment, you can provide inputs to `entropy` and friends that make no sense. I made an attempt to correct this -- let me know your thoughts.

Before:

```
julia> using StatsBase
julia> entropy([0.5, 0.5],2)
1.0
julia> entropy([0.5, 0.5, -0.6],2) # negative number is silently ignored
1.0
julia> entropy([0.5, 0.5, 1.5],2) # -1.5*log2(1.5) is added...
0.12255624891826569
julia> crossentropy([1],[-1]) # a slightly vague error
ERROR: DomainError
 in crossentropy at /StatsBase/src/scalarstats.jl:306
```

After:

```
julia> entropy([0.5, 0.5, -0.6],2)
ERROR: ArgumentError("Input is not a probability vector (has negative element)")
 in entropy at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:292
 in entropy at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:302

julia> entropy([0.5, 0.5, 0.1],2)
ERROR: ArgumentError("Input is not a probability vector (does not sum to one)")
 in entropy at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:296
 in entropy at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:302

julia> entropy([0.5, 0.1, 0.1],2)
ERROR: ArgumentError("Input is not a probability vector (does not sum to one)")
 in entropy at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:296
 in entropy at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:302

julia> crossentropy([1],[-1])
ERROR: ArgumentError("Second input is not a probability vector (has negative element).")
 in crossentropy at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:320

julia> kldivergence([0.3,0.3,0.4],[0,0.5,0.5])
Inf

julia> kldivergence([0.3,0.3,0.4],[0,0.5,-0.5])
ERROR: ArgumentError("Second input is not a probability vector (has negative element)")
 in kldivergence at /Users/alex/.julia/v0.3/StatsBase/src/scalarstats.jl:351
```

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -281,12 +281,19 @@ zscore{T<:Real}(X::AbstractArray{T}, dim::Int) = ((μ, σ) = mean_and_std(X, dim
 
 function entropy{T<:Real}(p::AbstractArray{T})
     s = 0.
+    ss = 0.
     z = zero(T)
     for i = 1:length(p)
         @inbounds pi = p[i]
+        ss += pi
         if pi > z
             s += pi * log(pi)
+        elseif pi < -1e-12
+            throw(ArgumentError("Input is not a probability vector (has negative element)"))
         end
+    end
+    if abs(ss-1.0) > 1e-12
+        throw(ArgumentError("Input is not a probability vector (does not sum to one)"))
     end
     return -s
 end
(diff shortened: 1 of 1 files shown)


richardreeve:
I'd not be keen on this because I use incomplete probability distributions that sum to less than 1, and for the divergences (though I'm not sure why they're here and in Distances.jl?) I'd be doing the tests multiple times anyway so it would be pretty inefficient.

ahwillia:
@richardreeve -- Hmm, do you think we should still throw an error if any input is less than zero or if the sum is greater than one?

This seems like a question about the philosophy of JuliaStats that I'm not qualified to answer: should it be optimized purely for speed and for users that really know what they're doing, or should there some constraints in place so newcomers don't screw up. My personal feelings:

* `entropy` is likely to almost always be called on a probability vector. I'm not sure how to interpret entropy for incomplete probability distributions, but this could be my ignorance.
* I tried to check the arguments in the most efficient way possible, even though it led to slightly more verbose code. (The other option is to define `is_prob_vec(p)` to check the arguments at the top of each function, but this would mean iterating through the arrays twice).
* Advanced users could write new functions that don't check the inputs if they have particular applications or concerns about speed.

But again, I'm not qualified to answer this.

richardreeve:
I'm not the right person to convince to be honest, as I'm not involved in JuliaStats, but I agree that I have no interpretation for negative probabilities, although it turns out I do have uses for pretty much any other values. At the moment I don't use this package (though I was intending to), so my opinion perhaps shouldn't hold that much weight. The silent dropping of negatives is obviously because the author is assuming inputs are valid, and so non-positive values must be zero.

Checking every input versus trusting the user is a philosophical issue so long as it gives the right answer for valid values, so I guess that's above my pay grade again. Sorry not to be more help.

PR #136, opened 2015-10-06
Fix deprecation message.

hayd:


andreasnoack:
Thanks. Have you tried just removing the type assert?

hayd:
Right, it should be unnecessary, the other option is `UTF8String[..]`. I can post a fix either way.

andreasnoack:
I took a look at the history here. I don't think the assert has to be there, so I think we should just remove it.

hayd:
removed type assert, and tests still passing!

simonster:
This function apparently has [no test coverage](https://coveralls.io/builds/3758850/source?filename=src%2Fstatmodels.jl) so it would still be good to test manually (or even better, add a test).

hayd:
Added some tests, which I think do cover this.

test/statquiz.jl
```diff
@@ -23,9 +23,9 @@ print("Test rounding: ")
 @test [@sprintf("%1.0f", x) for x in nasty[:round]] == ["1","2","3","4","5","6","7","8","9"]
 println("OK")
 print("Test math: ")
-@test int(2.6*7 - 0.2) == 18
-@test 2 - int(exp(log(sqrt(2)*sqrt(2)))) == 0
-@test int(3 - exp(log(sqrt(2)*sqrt(2)))) == 1
+@test round(Int, 2.6*7 - 0.2) == 18
+@test 2 - round(Int, exp(log(sqrt(2)*sqrt(2)))) == 0
```
hayd: oh maths.

andreasnoack:
Actually, `statquiz.jl` is not run right now and therefore the the line in focus is still not covered. It seems to reveal some problems when I try to include `statquiz.jl`, but not in your change. The reason it is not tested is probably that it has dependencies, but I think we can just add `GLM` and `DataFrames` in `.travis.yml`. It would be good to also have these test.

andreasnoack:
I'll merge this and then try to include `statquiz.jl` in the Travis tests.

PR #139, opened 2015-10-07
Run statquiz as part fo tests

andreasnoack:
The quiz has is also testing functions in `GLM.jl` and `DataFrames.jl` so this might break the `StatsBase.jl` tests because of downstream errors. However, since we don't have an umbrella stats package and all the packages here are part of `JuliaStats` I think we run the quiz here to make sure that we test as much as possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,12 @@ julia:
   - nightly
 notifications:
   email: false
-script:
+before_install:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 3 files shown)


andreasnoack:
Does anybody know why Travis is installing a version of `DataFrames` from Oct 31, 2013 on nightlies, but now 0.3 and 0.4?

hayd:
cc @wildart

simonster:
Probably because of JuliaLang/julia#13375. If Julia thinks it has StatsBase 0.0.0 then it can't install the latest DataFrames.

andreasnoack:
Okay. I'll restart the tests when the fix has been merged.

andreasnoack:
Finally green! @simonster, thanks for fixing the issue here.

PR #140, opened 2015-10-07
Handle NaN in formatting of p-values

julian-gehring:
Displays the value instead of throwing an error expception if p-value is a NaN. A NaN is a valid value that can occur e.g. due to estimate and standard error both being 0.

Here a short example (please note that different platforms give me different results here, the results here were produced with julia 3.11 from juliabox):

```julia
using GLM
using DataFrames

df = DataFrame(x = [1, 2, 3], y = [2, 4, 6])

ls = lm(y ~ x, df); ## perfect fit, with slope 2 and intercept 0
coeftable(ls)
```

The p-value for the estimated intercept will be `NaN`. Before, displaying the coefficient table used to fail, since the value was outside the allowed range `[0, 1]`.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -40,6 +40,9 @@ end
 
 ## format numbers in the p-value column
 function format_pvc(pv::Number)
+    if isnan(pv)
+        return @sprintf("%d", pv)
+    end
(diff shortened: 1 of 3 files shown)


andreasnoack:
LGTM

PR #141, opened 2015-10-09
Update to changes in Base's cov and cor

andreasnoack:
These changes are necessary after https://github.com/JuliaLang/julia/pull/13465 where the keyword arguments for `mean` and `vardim` were removed. Notice that the tests there won't pass before https://github.com/JuliaLang/julia/pull/13505 has been merged. I got some of the deprecation definitions wrong in the first attempt.

### DIFF
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -25,26 +25,60 @@ end
 
 scattermat_zm(x::DenseMatrix, vardim::Int) = Base.unscaled_covzm(x, vardim)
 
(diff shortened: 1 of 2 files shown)


tkelman:
The new syntax should be added to Compat, this won't be the only use case.

andreasnoack:
I wouldn't be so sure about that, but they are in progress anyway. I think few people uses a custom mean or MLE corrections when computing covariances. I tried to search through julia repositories on GitHub, but unfortunately the search doesn't allow `=` in the search expression so it's quite difficult for seach for the use of keyword arguments.

PR #144, opened 2015-10-19
for v0.4: import histrange from base, remove midpoints

### LINKED ISSUE #103: Warning: using StatsBase.histrange in module Main conflicts with an existing identifier.
```
julia> using Data<tab><tab>
DataArrays     DataFrames      DataStructures
julia> using DataFrames    # I typed F and hit tab to complete
Warning: using StatsBase.histrange in module Main conflicts with an existing identifier.
Warning: using StatsBase.midpoints in module Main conflicts with an existing identifier.

julia> VERSION
v"0.3.3"
```
Now, what's really interesting is this:
```
julia> using DataFrames   # I typed the entire thing in after restarting Julia

julia>
```

That is, the warnings only occur when I use tabs to see completion options. 100% reproducible. This occurs with DataArrays and DataStructures as well. (See https://github.com/JuliaStats/DataFrames.jl/issues/740 for the original issue.)

Ken-B:
These are minor changes for julia v0.4 compatibility, fixes #103.

Julia Base now has histrange, although with different arguments, so we should extend this. 

I've removed the midpoints as it is identical as in Base.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -1,4 +1,4 @@
-import Base: show, ==, push!, append!
+import Base: show, ==, push!, append!, histrange
 
 ## nice-valued ranges for histograms
 function histrange{T<:AbstractFloat}(v::AbstractArray{T}, n::Integer, closed::Symbol)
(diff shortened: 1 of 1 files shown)


andreasnoack:
Sorry for the slow response. I missed it.

PR #146, opened 2015-11-11
WIP: Add fit statistics functions and document existing ones

nalimilan:
Here's a first proposal of API to provide fit statistics (https://github.com/JuliaStats/GLM.jl/issues/74). This adds `loglikelihood`, `nullloglikelihood`, `nulldeviance`, `df`, `AIC`, `AICc`, `BIC`, and `R²`. I'll make a PR with the actual implementation in GLM.jl shortly.

I marked this as work in progress since I think at least these points must be sorted out:
- [x] R² is implemented for all statistical models, using the McFadden variant which has the advantage of being equivalent to the classic R² for linear models. But it could be argued that for models other than linear, the user should have to choose the variant explicitly, to force really thinking about it and not using it blindly. UPDATE: I now require users to specify the variant they want.
- [x] I'm computing the R² using the deviance and not the log-likelihood, as I've not been able to find out how to do the latter (even with know to be correct values checked against R or Stata). This isn't really an issue in practice, but for documentation purposes I would have found it clearer, and I would still like to understand this point. Help welcome! UPDATE: relying on the deviance is not an issue, I've simply documented it.
- [x] The `df` function currently returns the number of consumed degrees of freedom, *including the dispersion parameter* when it needs to be estimated (notably for the gaussian family). But in many cases you don't want it to be taken into account. What's the best solution, including it only when a second argument is set to `true`? Or creating another function? UPDATE: for now I've decided to keep `df` as-is.


nalimilan:
Is anybody interested in reviewing or at least discussing the general API issues I've raised above? It's too bad that this essential package doesn't get more attention -- I wish I can help improving this state.

nalimilan:
@dmbates I see that in MixedModels you suggest using an `objective` function instead of `deviance`. Do you think I should add this one here?

andreasnoack:
I'd also like to hear @dmbates on this.

What is the reasoning behind the `null*` methods?

nalimilan:
> What is the reasoning behind the null* methods?

They allow defining the R² function only once in StatsBase, supporting multiple variants, instead of repeating it in packages. The null deviance is also usually printed among other fit statistics.

But we could retain only one of `nulldeviance`/`nullloglikelihood` if that sounds too much.

andreasnoack:
But the cost is that you'll have to define `nullloglikelihood` instead and I'm not sure how useful that function will be in general.

Also, it could be argued if we should promote R² in general and in particular outside the linear model.

It makes good sense to me to define the information criteria here, but I'm not sure I can give useful answer to your question about `df`.

nalimilan:
> But the cost is that you'll have to define nullloglikelihood instead and I'm not sure how useful that function will be in general.
>
> Also, it could be argued if we should promote R² in general and in particular outside the linear model.

I agree with you that the R² might not be the best fit statistic for models other than linear, but my experience from R is that people end up implementing it in packages, each with its own limitations. Model implementations do not need to provide `nulldeviance` if they don't think this measure makes sense for them. And providing a function to compute it isn't equivalent to promoting it IMHO. Requiring the user to choose explicitly a variant would be a way to force users to think twice about it.

dmbates:
The reason for changing from the name `deviance` to `objective` in MixedModels is exactly because of the null deviance.  I don't know of a way of defining it for models with random effects.  The idea of null deviance is that it is the lowest -2loglikelihood for all models in the family.  In most model types you can define it by evalating -2loglikelihood with the fitted values equal to the observed values and any scale parameters set to unity.  I don't think that would apply to a mixed-effects model, because the loglikelihood involves the random effects and the covariance of the random effects in non-trivial ways.

nalimilan:
@dmbates I see. So should I add that `objective` function in this PR? I guess we should we have `deviance = objective` as a fallback (in particular for GLMs)?

dmbates:
@nalimilan I would suggest the opposite default that `objective = deviance`.  The sense of `objective` is that the model is fit by optimizing the objective.  So MLE's are fit by optimizing the loglikelihood over the parameter space.  We may want to go further and define whether the optimization is minimizing or maximizing.  There could be two approaches
 - pick one of minimize/maximize and define objective to be the objective with only one direction of optimization.  That is, if the choice is to minimize then the objective for the MLE's would be -loglikelihood.
- define immutable types `minobjective{T}` and `maxobjective{T}` and have `objective` methods return one or the other.

nalimilan:
@dmbates Thinking about it again, I think I'd prefer to report a deviance for mixed effects models, even if there isn't any clear reference model. Indeed, the definition is still the same, and deviance is only used for comparisons of different models. That's also how this is reported in other programs, while `objective` is quite vague.

I'm a bit surprised that you say the concept of deviance doesn't make sense for mixed effects models, since your own code and writings mention it. Don't you think it can be useful, even without a baseline value?

nalimilan:
Any other remarks about what is OK and what should be changed in order to merge?

nalimilan:
I've updated the branch (see "UPDATE" in the description above). Can we merge this, or make the needed changes in order to do so?

andreasnoack:
@dmbates Do you have any further comments here? If not, I think it can be merged.

nalimilan:
Good to go then?

andreasnoack:
Yes. I think so. The tests will cause segfault until the we get new Mac build bots, because the nightlies won't get updated. We'll just have to be careful testing locally for a while.

nalimilan:
Thanks! How can we proceed for the corresponding PRs in DataFrames.jl and GLM.jl? I guess I should wait for a new release to be tagged (else tests will fail)?

andreasnoack:
See https://github.com/JuliaLang/METADATA.jl/pull/4690

PR #149, opened 2016-02-16
inline documentation

abhijithch:
Added basic inline documentation for most of the exported APIs.

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -9,7 +9,16 @@
 typealias IntUnitRange{T<:Integer} UnitRange{T}
 
 #### functions for counting a single list of integers (1D)
+"""
+    addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
 
+### Args:
+* `r`: Accumulating array.
+* `x`: An `IntegerArray` for which the counts are computed.
+* `levels`: The range of values to check for occurence in `x`. 
+
+This function adds the counts of values in `levels` that appear in `x` to an accumulating array `r` or a map.
+"""
 function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
 	# add counts of integers from x to r
 
@@ -29,6 +38,17 @@ function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
 	return r
 end
 
+"""
+    addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange, wv::weightVector)
+
+### Args:
+* `r`: Accumulating array.
+* `x`: An `IntegerArray` for which the counts are computed.
+* `levels`: The range of values to check for occurence in `x`. 
+* `wv`: An optional parameter, which is the weight vector.
+
+This function adds the counts of values in `levels` that appear in `x` to an accumulating array `r`.
+"""
 function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange, wv::WeightVec)
 	k = length(levels)
 	length(r) == k || raise_dimerror()
@@ -60,6 +80,18 @@ proportions(x::IntegerArray, k::Integer) = proportions(x, 1:k)
 proportions(x::IntegerArray, k::Integer, wv::WeightVec) = proportions(x, 1:k, wv)
 proportions(x::IntegerArray) = proportions(x, span(x))
 proportions(x::IntegerArray, wv::WeightVec) = proportions(x, span(x), wv)
+
+"""
+    proportions(x::IntegerArray, levels::IntUnitRange, wv::WeightVec)
+
+### Args:
+* `x`: An `IntegerArray`.
+* `levels`: An integer unti range.
+* `wv`: An optional weight vector.
+
+Compute the proportions of values in a:b with respect to x.
+"""
+function proportion end
 
 #### functions for counting a single list of integers (2D)
 
@@ -199,6 +231,24 @@ end
 
 countmap{T}(x::AbstractArray{T}) = addcounts!(Dict{T,Int}(), x)
 countmap{T,W}(x::AbstractArray{T}, wv::WeightVec{W}) = addcounts!(Dict{T,W}(), x, wv)
+"""
+    countmap(x::AbstractArray{T}, wv::WeightVec{W})
+
+### Args:
+* `x`: An `AbstractArray`.
+* `wv`; An optional weight vector.
+
+Return a dictionary that maps distinct values in x to their counts (or total weights).
+"""
 
 proportionmap(x::AbstractArray) = _normalize_countmap(countmap(x), length(x))
 proportionmap(x::AbstractArray, wv::WeightVec) = _normalize_countmap(countmap(x, wv), sum(wv))
+"""
+    proportionmap(x::AbstractArray{T}, wv::WeightVec{W})
+
+### Args:
+* `x`: An `AbstractArray`.
+* `wv`; An optional weight vector.
+
+Return a dictionary that maps distinct values in x to their proportions.
+"""
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -81,4 +81,31 @@ else
         m = mean(x, wv, vardim)
         return m, Base.cov(x, wv, vardim)
     end
+    """
+        mean_and_cov(x::DenseMatrix, wv::WeightVec, vardim::Int)
+
+    ### Args:
+    * `x`: A 2-d array of type `DenseMatrix` to compute the mean and covariance of.
+    * `wv`: An optional weight vector of type `WeightVec`.
+    * `vardim`: An optional parameter specifying the dimension along which to compute the mean.
+
+    Jointly compute the mean and covariance of input matrix `x`.
+    """
+    function mean_and_cov end
+
+    """
+        scattermat(x::DenseMatrix, mean, vardim::Int=1)
+
+    ### Args:
+    * `x`: A `DenseMatrix`
+    * `mean`: Pre-computed mean vector. Default value of mean is set to nothing, which        indicates that the function would compute the mean internally. One can also set mean t    o 0, which indicates that the input `X` has already been centralized. Otherwise, the s    upplied mean will be subtracted from `X`.
+    
+    Compute scatter matrix for the variables contained in X.
+
+    A scatter matrix can be considered as a unnormalized version of the covariance matrix.
+
+    By default, it considers each column as a variable (i.e each row as an observation), a    nd subtract the mean from each vector. One may change this default behavior by setting    the keyword arguments.
+    """
+    function scattermat end
+    
 end
--- a/src/deviation.jl
+++ b/src/deviation.jl
@@ -1,7 +1,11 @@
 # Computing deviation in a variety of ways
 
 ## count the number of equal/non-equal pairs
+"""
+    counteq(a::AbstractArray, b::AbstractArray)
 
+Count the number of equal pairs of elements in a and b, i.e countnz(a .== b).
+"""
 function counteq(a::AbstractArray, b::AbstractArray)
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
@@ -14,6 +18,11 @@ function counteq(a::AbstractArray, b::AbstractArray)
     return c
 end
 
+"""
+    countne(a::AbstractArray, b::AbstractArray)
+
+Count the number of non-equal pairs of elements in a and b, i.e countnz(a .!= b).
+"""
 function countne(a::AbstractArray, b::AbstractArray)
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
@@ -27,6 +36,11 @@ function countne(a::AbstractArray, b::AbstractArray)
 end
 
 # squared L2 distance
+"""
+    sqL2dist(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the Squared L2 distance between `a` and `b`.
+"""
 function sqL2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
@@ -38,9 +52,19 @@ function sqL2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
 end
 
 # L2 distance
+"""
+    L2dist(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the L2 distance between `a` and `b`.
+"""
 L2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = sqrt(sqL2dist(a, b))
 
 # L1 distance
+"""
+    L1dist(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the L1 distance between `a` and `b`.
+"""
 function L1dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
@@ -52,6 +76,11 @@ function L1dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
 end
 
 # Linf distance
+"""
+    Linfdist(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the Linf distance between `a` and `b`.
+"""
 function Linfdist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
@@ -66,6 +95,11 @@ function Linfdist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
 end
 
 # Generalized KL-divergence
+"""
+    gkldiv(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the generalized Kullback-Leibler divergence between two arrays `a` and `b`.
+"""
 function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
     n = length(a)
     r = 0.0
@@ -82,15 +116,35 @@ function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
 end
 
 # MeanAD: mean absolute deviation
+"""
+    meanad(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the mean absolute deviation between `a` and `b`, i.e. `mean(abs(a - b))`.
+"""
 meanad{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = L1dist(a, b) / length(a)
 
 # MaxAD: maximum absolute deviation
+"""
+    maxad(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the maximum absolute deviation between `a` and `b`, i.e. `maximum(abs(a - b))`.
+"""
 maxad{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = Linfdist(a, b)
 
 # MSD: mean squared deviation
+"""
+    maxad(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the mean squared deviation between a and b, i.e. mean(abs2(a - b)).
+"""
 msd{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = sqL2dist(a, b) / length(a)
 
 # RMSD: root mean squared deviation
+"""
+    rmsd(a::AbstractArray{T}, b::AbstractArray{T}; normalize::Bool=false)
+
+Computes the root mean squared deviation, i.e. sqrt(msd(a, b)). The keyword argument normalize is default to false. If it is set to true, the result is normalized by (maximum(a) - minimum(a).
+"""
 function rmsd{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}; normalize::Bool=false)
 	v = sqrt(msd(a, b))
 	if normalize
@@ -101,6 +155,11 @@ function rmsd{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}; normalize::Bo
 end
 
 # PSNR: peak signal-to-noise ratio
+"""
+    psnr(a::AbstractArray{T}, b::AbstractArray{T}, maxv::Real)
+
+Computes peak signal-to-noise ratio, i.e. 10 * log10(maxv^2 / msd(a, b)).
+"""
 function psnr{T<:Real}(a::AbstractArray{T}, b::AbstractArray{T}, maxv::Real)
     20. * log10(maxv) - 10. * log10(msd(a, b))
 end
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -12,6 +12,11 @@ function norepeat(a::AbstractArray)
 end
 
 # run-length encoding
+"""
+    rle(v::Vector)
+
+Run-length encoding of `v`. It returns `(vals, lens)`, a sequence of values and their corresponding chunk length.
+"""
 function rle{T}(v::Vector{T})
     n = length(v)
     vals = T[]
@@ -42,6 +47,11 @@ function rle{T}(v::Vector{T})
 end
 
 # inverse run-length encoding
+"""
+    inverse_rle(vals::AbstractVector, lens::AbstractVector)
+
+Inversed run-length encoding. It takes the results of rle and reconstructs the original sequence.
+"""
 function inverse_rle{T}(vals::AbstractVector{T}, lens::IntegerVector)
     m = length(vals)
     length(lens) == m || raise_dimerror()
@@ -61,7 +71,11 @@ end
 
 
 # findat (get positions (within a) for elements in b)
+"""
+    indexmap(x)
 
+Construct a dictionary that maps each distinct value in `x` to its first index.
+"""
 function indexmap{T}(a::AbstractArray{T})
     d = Dict{T,Int}()
     for i = 1 : length(a)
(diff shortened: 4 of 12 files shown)


src/counts.jl
```diff
@@ -29,6 +38,17 @@ function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
 	return r
 end
 
+"""
+    addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange, wv::weightVector)
```
nalimilan: I think you should merge this with the previous definition, with the following signature:
```julia
addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange[, wv::weightVector])
```

src/counts.jl
```diff
@@ -9,7 +9,16 @@
 typealias IntUnitRange{T<:Integer} UnitRange{T}
 
 #### functions for counting a single list of integers (1D)
+"""
+    addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
 
+### Args:
+* `r`: Accumulating array.
+* `x`: An `IntegerArray` for which the counts are computed.
```
nalimilan: `IntegerArray` is redundant with the signature here.

src/counts.jl
```diff
@@ -9,7 +9,16 @@
...
+* `levels`: The range of values to check for occurence in `x`.
```
nalimilan: "occurrence"

src/counts.jl
```diff
@@ -9,7 +9,16 @@
 typealias IntUnitRange{T<:Integer} UnitRange{T}
 
 #### functions for counting a single list of integers (1D)
+"""
+    addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
 
+### Args:
+* `r`: Accumulating array.
```
nalimilan: Maybe you could be more explicit about what "accumulating means"?

src/counts.jl
```diff
@@ -9,7 +9,16 @@
 typealias IntUnitRange{T<:Integer} UnitRange{T}
 
 #### functions for counting a single list of integers (1D)
+"""
+    addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
```
nalimilan: `IntegerArray` and `IntUnitRange` are unexported type aliases. It would probably be better to show the actual types:
```julia
   addcounts!{T<:Integer, U<:Integer}(r::AbstractArray, x::AbstractArray{T}, levels::UnitRange{U}) 
```
or maybe (but this isn't correct Julia syntax):
```julia
   addcounts!(r::AbstractArray, x::AbstractArray{<:Integer}, levels::UnitRange{<:Integer}) 
```

src/counts.jl
```diff
@@ -9,7 +9,16 @@
...
 
+### Args:
+* `r`: Accumulating array.
+* `x`: An `IntegerArray` for which the counts are computed.
+* `levels`: The range of values to check for occurence in `x`. 
+
+This function adds the counts of values in `levels` that appear in `x` to an accumulating array `r` or a map.
```
nalimilan: Do you mean "to an accumulating array or a map `r`"? I think you should either be more explicit about what a map is, or remain vague by mentioning only "array".

src/counts.jl
```diff
@@ -61,6 +81,18 @@ proportions(x::IntegerArray, k::Integer, wv::WeightVec) = proportions(x, 1:k, wv
...
 
+"""
+    proportions(x::IntegerArray, levels::IntUnitRange, wv::WeightVec)
+
+### Args:
+* `x`: An `IntegerArray`.
+* `levels`: An integer unti range.
+* `wv`: An optional weight vector.
+
+Compute the proportions of values in a:b with respect to x.
```
nalimilan: Severals problems here:
- what is `a:b`?
- "with respect to x" isn't obvious
- you should put optional argument between `[ ]` (BTW, `levels` is optional too)
- typo in "unti"
- see also the points raised above

nalimilan: Ah, and always use ` around names of Julia objects.

src/counts.jl
```diff
@@ -199,6 +231,24 @@ end
 
 countmap{T}(x::AbstractArray{T}) = addcounts!(Dict{T,Int}(), x)
 countmap{T,W}(x::AbstractArray{T}, wv::WeightVec{W}) = addcounts!(Dict{T,W}(), x, wv)
+"""
+    countmap(x::AbstractArray{T}, wv::WeightVec{W})
```
nalimilan: `T` and `W` are not useful here.

src/counts.jl
```diff
@@ -199,6 +231,24 @@ end
...
+
+### Args:
+* `x`: An `AbstractArray`.
+* `wv`; An optional weight vector.
```
nalimilan: ; instead of :

src/counts.jl
```diff
@@ -199,6 +231,24 @@ end
 
 countmap{T}(x::AbstractArray{T}) = addcounts!(Dict{T,Int}(), x)
 countmap{T,W}(x::AbstractArray{T}, wv::WeightVec{W}) = addcounts!(Dict{T,W}(), x, wv)
+"""
+    countmap(x::AbstractArray{T}, wv::WeightVec{W})
+
+### Args:
+* `x`: An `AbstractArray`.
```
nalimilan: Repeating the type info from the signature isn't useful. You should say what's the meaning of this argument instead.

nalimilan: Actually, this probably means that giving a list of arguments isn't really useful. In most cases, you already mention them in the description of the function. I think you'd better ensure you mention all of them in that description, and get rid of the arguments list altogether.

src/cov.jl
```diff
@@ -81,4 +81,31 @@ else
...
+    Jointly compute the mean and covariance of input matrix `x`.
+    """
+    function mean_and_cov end
+
+    """
+        scattermat(x::DenseMatrix, mean, vardim::Int=1)
+
+    ### Args:
+    * `x`: A `DenseMatrix`
+    * `mean`: Pre-computed mean vector. Default value of mean is set to nothing, which        indicates that the function would compute the mean internally. One can also set mean t    o 0, which indicates that the input `X` has already been centralized. Otherwise, the s    upplied mean will be subtracted from `X`.
```
nalimilan: Spacing bug here and below.

src/deviation.jl
```diff
@@ -1,7 +1,11 @@
 # Computing deviation in a variety of ways
 
 ## count the number of equal/non-equal pairs
+"""
+    counteq(a::AbstractArray, b::AbstractArray)
 
+Count the number of equal pairs of elements in a and b, i.e countnz(a .== b).
```
nalimilan: "i.e."

Also could be more explicit, like "i.e. equivalent to `countnz(...)` but faster".

src/deviation.jl
```diff
@@ -27,6 +36,11 @@ function countne(a::AbstractArray, b::AbstractArray)
 end
 
 # squared L2 distance
+"""
+    sqL2dist(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the Squared L2 distance between `a` and `b`.
```
nalimilan: "squared"

Could you give the definition of this distance too (i.e. `sum((a - b).^2)`), and mention that this function is more efficient.

src/deviation.jl
```diff
@@ -38,9 +52,19 @@ function sqL2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
 end
 
 # L2 distance
+"""
+    L2dist(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the L2 distance between `a` and `b`.
```
nalimilan: Same here and below.

src/deviation.jl
```diff
@@ -82,15 +116,35 @@ function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
 end
 
 # MeanAD: mean absolute deviation
+"""
+    meanad(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the mean absolute deviation between `a` and `b`, i.e. `mean(abs(a - b))`.
```
nalimilan: I these docstring and those below are really the model we should follow: without a redundant list of arguments, they give all the information one needs.

src/deviation.jl
```diff
@@ -1,7 +1,11 @@
 # Computing deviation in a variety of ways
 
 ## count the number of equal/non-equal pairs
```
nalimilan: Just remove these comments when their are redundant with the new docstrings.

src/deviation.jl
```diff
@@ -82,15 +116,35 @@ function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
...
+"""
+    maxad(a::AbstractArray{T}, b::AbstractArray{T})
+
+Computes the maximum absolute deviation between `a` and `b`, i.e. `maximum(abs(a - b))`.
+"""
 maxad{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = Linfdist(a, b)
 
 # MSD: mean squared deviation
+"""
+    maxad(a::AbstractArray{T}, b::AbstractArray{T})
```
nalimilan: "msd"

src/deviation.jl
```diff
@@ -82,15 +116,35 @@ function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
...
+
+Computes the mean squared deviation between a and b, i.e. mean(abs2(a - b)).
```
nalimilan: Missing ` around a and b.

src/deviation.jl
```diff
@@ -82,15 +116,35 @@ function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
...
+"""
 msd{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = sqL2dist(a, b) / length(a)
 
 # RMSD: root mean squared deviation
+"""
+    rmsd(a::AbstractArray{T}, b::AbstractArray{T}; normalize::Bool=false)
+
+Computes the root mean squared deviation, i.e. sqrt(msd(a, b)). The keyword argument normalize is default to false. If it is set to true, the result is normalized by (maximum(a) - minimum(a).
```
nalimilan: "The keyword argument normalize is default to false." is redundant with the signature. Just say:
```
If `normalize` is `true`, the result is divided by `maximum(a) - minimum(a)`.
```

src/deviation.jl
```diff
@@ -101,6 +155,11 @@ function rmsd{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}; normalize::Bo
 end
 
 # PSNR: peak signal-to-noise ratio
+"""
+    psnr(a::AbstractArray{T}, b::AbstractArray{T}, maxv::Real)
+
+Computes peak signal-to-noise ratio, i.e. 10 * log10(maxv^2 / msd(a, b)).
```
nalimilan: Missing ` around formula.

nalimilan:
Thanks, this is very much appreciated. I've made a lot of comments in the first half of the changes, some of which probably apply to the whole PR. I'll continue reviewing the rest soon.

src/weights.jl
```diff
@@ -291,6 +304,13 @@ function Base.median{W<:Real}(v::RealVector, w::WeightVec{W})
     end
 end
 
+"""
```
nalimilan: This is already documented in Base, I don't think we need it here.

src/weights.jl
```diff
@@ -9,6 +9,11 @@ end
 WeightVec{Vec<:RealVector,W<:Real}(vs::Vec,wsum::W) = WeightVec{W,Vec}(vs, wsum)
 WeightVec(vs::RealVector) = WeightVec(vs, sum(vs))
 
+"""
+    weights(vs)
+
+This function constructs the weight vector.
```
nalimilan: "constructs a weight vector from a numeric vector or array."

src/misc.jl
```diff
@@ -86,6 +105,11 @@ function levelsmap{T}(a::AbstractArray{T})
     return d
 end
 
+"""
+    findat!(r::IntegerArray, a::AbstractArray, b::AbstractArray)
+
+For each element in `b`, find its first index in `a`. If the value does not appear in `a`, the corresponding index is `0`. Write the results of findat(a, x) to a pre-allocated array `r`.
```
nalimilan: I would add "and store the result in `r`" at the end of the first sentence, and remove the last sentence.

nalimilan: Also, you should merge this docstring with the following one, so you can say "optionally store...".

src/misc.jl
```diff
@@ -104,6 +133,18 @@ findat(a::AbstractArray, b::AbstractArray) = findat!(Array(Int, size(b)), a, b)
...
+    indicatormat(x::IntegerArray, k::Integer; sparse::Bool=false)
+
+### Args:
+* `x`: A, `IntegerArray`.
+* `k`: An `Integer`
+* `sparse`: A keyword argument, of `Bool` type.
+
+Construct a boolean matrix `r` of size `(k, length(x))` such that `r[x[i], i] = true` and all other elements are set to false.
+
+The keyword argument sparse controls whether to construct a sparse matrix. By default, it is `false`.
```
nalimilan: Missing ` around "sparse". Again, I don't think mentioning "keyword argument" is useful here.

src/misc.jl
```diff
@@ -104,6 +133,18 @@ findat(a::AbstractArray, b::AbstractArray) = findat!(Array(Int, size(b)), a, b)
...
 
+"""
+    indicatormat(x::IntegerArray, k::Integer; sparse::Bool=false)
+
+### Args:
+* `x`: A, `IntegerArray`.
+* `k`: An `Integer`
+* `sparse`: A keyword argument, of `Bool` type.
+
+Construct a boolean matrix `r` of size `(k, length(x))` such that `r[x[i], i] = true` and all other elements are set to false.
```
nalimilan: Could be useful to give a literary explanation. Something like this maybe?

Construct a boolean matrix with `k` rows and one column for each element in `x`, such that in each column the row index of the `true` element is given by the corresponding element from `x`. That is, the resulting matrix  `r` is such that `r[x[i], i] = true` and all other elements are `false`.

src/moments.jl
```diff
@@ -79,6 +79,29 @@ mean_and_std(A::RealArray, dim::Int) = (m = mean(A, dim); (m, stdm(A, m, dim)))
...
+    mean_and_var(x::RealArray, wv::WeightVec, dim::Int)
+
+### Args:
+* `x`: A vector of type `RealArray` to compute theh mean and variance of.
+* `wv`: An optional weight vector of type `WeightVec`.
+* `dim`: An optional `Int` type parameter which specifies the dimension.
+
+Jointly compute the mean and variance of x.
+"""
+function mean_and_var end
```
nalimilan: Why not add to docstring to the actual definition above? Same for `mean_and_std`, and (I think) in other places too.

src/moments.jl
```diff
@@ -183,6 +206,17 @@ end
...
+    moment(v::RealArray, k::Int, wv::WeightVec, m::Real)
+
+### Args:
+* `v`: The vector of type `RealArray`.
+* `k`: The order of central moment.
+* `wv`: A weights vector which is optional.
+* `m`: The centre value, which is optional.
+
+Compute the k-th order central moment of the values in x. It is the sample mean of (x - mean(x)).^k. One can optionally supply the center m, and/or a weight vector of type WeightVec
+"""
```
nalimilan: Move the docstring right above the second definition of `moment`.

nalimilan: Same problem with other docstrings below.

src/rankcorr.jl
```diff
@@ -18,6 +18,16 @@ corspearman(x::RealVector, Y::RealMatrix) = cor(tiedrank(x), mapslices(tiedrank,
...
 corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, 1); cor(Z, Z))
 
+"""
+    corspearman(x, y)
+
+### Args:
+* `x`: Can be any one of `RealVector` or `RealMatrix`.
+* `y`: Can be any one of `RealVector` or `RealMatrix`.
+
+Spearman’s rank correlation. Here, `x` and `y` can be either real vectors or matrices. When `x` and `y` are matrices, it computes the correlations between their columns (pairwisely).
```
nalimilan: I don't think "pairwisely" is an English word. :-)

I also don't think you should repeat "`x` and `y` can be either real vectors or matrices": just say what happens for each type. And what happens when passing a vector and a matrix?

src/ranking.jl
```diff
@@ -88,6 +98,11 @@ function denserank!(rks::RealArray, x::RealArray, p::IntegerArray)
     return rks
 end
 
+"""
+    denserank(x::RealArray
```
nalimilan: Missing parenthesis.

src/sampling.jl
```diff
@@ -80,6 +80,11 @@ function samplepair(n::Int)
     return (i1, ifelse(i2 == i1, n, i2))
 end
 
+"""
+    samplepair(a::AbstractArray)
+
+Pick two elements at distinct positions from a, and return them as a pair. This algorithm consumes exactly two random numbers.
```
nalimilan: "*Randomly* pick" (or something similar).

src/sampling.jl
```diff
@@ -330,7 +364,11 @@ end
 #  Weighted sampling
 #
 ################################################################
+"""
+    sample(wv::WeightVec)
 
+Draw an integer in 1:length(wv) with probabilities proportional to the weights given in wv. Here, wv should be a weight vector of type WeightVec.
```
nalimilan: "Here, wv should be a weight vector of type WeightVec." isn't useful.

src/scalarstats.jl
```diff
@@ -146,6 +162,28 @@ function modes{T}(a::AbstractArray{T})
...
+"""
+    mode{T}(a::AbstractArray{T}, rgn::UnitRange{T})
+
+### Args:
+* `a`: An arbitrary array of type `AbstractArray`.
+* `rgn`: A range of integer values, which is optional.
+
+Compute the first mode of `a`, over the range `rng` if specified.
+"""
+function mode end
```
nalimilan: Move these to the relevant method definitions.

src/scalarstats.jl
```diff
@@ -146,6 +162,28 @@ function modes{T}(a::AbstractArray{T})
...
 end
 
+"""
+    mode{T}(a::AbstractArray{T}, rgn::UnitRange{T})
+
+### Args:
+* `a`: An arbitrary array of type `AbstractArray`.
+* `rgn`: A range of integer values, which is optional.
+
+Compute the first mode of `a`, over the range `rng` if specified.
```
nalimilan: "rgn". By the way, it would be cool to rename this argument to `r` both in the docs and in the code, to avoid the confusion with RNG.

src/scalarstats.jl
```diff
@@ -153,9 +191,27 @@ end
...
 #############################
 
+"""
+    percentile(v::AbstractArray{T}, p)
+
+### Args:
+* `v`: An `AbstracArray`.
+* `p`: Percentage value.
+
+Computes the quantile using percentage (instead of fraction) as argument.
```
nalimilan: "quantiles", as p can be an array.

src/scalarstats.jl
```diff
@@ -166,19 +222,51 @@ nquantile{T<:Real}(v::AbstractArray{T}, n::Integer) = quantile(v, (0:n)/n)
...
+function variation end
 
 # Standard error of the mean: std(a) / sqrt(len)
+"""
+    sem(a::AbstractArray{T})
+
+### Args:
+* `a`: An array of type `AbstractArray`.
+
+Computes the standard error of the mean, i.e. sqrt(var / n).
```
nalimilan: Could as well say `sqrt(var(a) / length(a))`.

src/scalarstats.jl
```diff
@@ -166,19 +222,51 @@ nquantile{T<:Real}(v::AbstractArray{T}, n::Integer) = quantile(v, (0:n)/n)
...
 # Median absolute deviation
 mad{T<:Real}(v::AbstractArray{T}, args...;arg...) = mad!(copy(v), args...;arg...)
 mad{T<:Real}(v::Range{T}, args...;arg...) = mad!([v;], args...;arg...)
-
+"""
+    mad(x, center;constant=1.4826)
+### Args:
+* `x`: An array of type `AbstractArray`.
+* `center`: Optional parameter which specifies the center of `x`.
+* `constant`: Optional paramter, Constant scale factor, default is 1.4826.
```
nalimilan: Typo here and in the same place above, but I'd rather remove these argument lists anyway.

src/scalarstats.jl
```diff
@@ -271,6 +367,29 @@ end
...
 
+"""
+    zscore(X::AbstractArray, μ::AbstractArray, σ::AbstractArray)
+
+### Args:
+* `X`: The input matrix to compute the Z-score.
+* `μ`: The mean, this could be a scalar or a an array.
+* `σ`: The standard deviation, this could be a scalar or a an array.
+
+Compute the Z-scores, given the mean μ and standard deviation σ, which is defined as (x - μ) / σ. This function returns an array Z of the same size as X. Here, μ and σ should be both scalars or both arrays.
```
nalimilan: "which are defined".

src/scalarstats.jl
```diff
@@ -271,6 +367,29 @@ end
...
+"""
+function zscore end
+
+"""
+    zscore(X::AbstractArray, μ::AbstractArray, σ::AbstractArray)
```
nalimilan: "zscore!"

src/scalarstats.jl
nalimilan: Also applies below.

src/scalarstats.jl
```diff
@@ -271,6 +367,29 @@ end
...
+
+### Args:
+* `X`: The input matrix to compute the Z-score.
+* `μ`: The mean, this could be a scalar or a an array.
+* `σ`: The standard deviation, this could be a scalar or a an array.
+
+Compute the Z-scores inline, given the mean μ and standard deviation σ, which is defined as (x - μ) / σ. Here, μ and σ should be both scalars or both arrays.
```
nalimilan: "in-place" instead of "inline". Would likely be clearer by also saying "storing them directly in `X`".

src/scalarstats.jl
```diff
@@ -294,6 +413,16 @@ end
 function entropy{T<:Real}(p::AbstractArray{T}, b::Real)
     return entropy(p) / log(b)
 end
+"""
+    entropy(p::AbstractArray{T}, b::Real)
```
nalimilan: You need to document the default value for `b` (which corresponds to the method without this argument). Same for `crossentropy` and `kldivergence`.

src/scalarstats.jl
```diff
@@ -330,6 +469,16 @@ end
...
 end
+"""
+    kldivergence(p::AbstractArray{T}, q::AbstractArray{T}, b::Real)
+
+### Args:
+* `p`: First probability vector.
+* `q`: Second probability vector.
+* `b`: Base of logarithm, an optional parameter.
+
+Computes the Kullback-Leiver divergence between the two probability vectors.
```
nalimilan: "Leibler"

src/scalarstats.jl
```diff
@@ -346,6 +495,21 @@ immutable SummaryStats{T<:AbstractFloat}
     max::T
 end
 
+"""
+    summarystats(a::AbstractArray)
+
+Compute a set of statistics over x and return a struct of type SummaryStats defined as below:
```
nalimilan: x -> a

src/signalcorr.jl
```diff
@@ -87,6 +87,36 @@ end
...
 
+"""
+    autocov(x, lags, demean)
+
+### Args:
+* `x`: Input `Vector` or `Matrix` to compute the autocovariance of. 
+* `lags`: Specifies the lags at which the autocovariance of `x` is copmuted.
+* `demean`: A `Bool` type variable, this is a named paramter. 
+
+Compute auto-covariance of `x` at specified lags. If `x` is a vector, it returns a vector of the same length of lags. If `x` is a matrix, it returns a matrix of size `(length(lags), size(x,2))`, where each column in the result corresponding to a column in `x`.
```
nalimilan: "autocovariance", without a dash.
"same length as"
"corresponding" -> corresponds

src/signalcorr.jl
```diff
@@ -87,6 +87,36 @@ end
...
+
+Here, demean is a keyword argument (default value is true), which means that the function will subtract each `x` from its mean before computing the results. Otherwise, `x` is considered as having been centered.
```
nalimilan: "demean is a keyword argument (default value is true)," won't be needed if you add this information to the signature.

I also think "subtract each `x` from its mean" should be "subtract its mean from each column in `x` (i.e. center it)".

src/signalcorr.jl
```diff
@@ -87,6 +87,36 @@ end
...
+
+the computed autocovariance is written back to `r`.
```
nalimilan: This information should appear in the first sentence. Also, missing capital, and "back" doesn't make sense here.

src/signalcorr.jl
```diff
@@ -248,6 +307,33 @@ end
...
+"""
+    crosscov(x, y, lags, demean)
+
+### Args:
+* `x`: Input `Vector` or `Matrix` to compute the cross-covariance of.
+* `y`: Input `Vector` or `Matrix` to compute the cross-covariance of.
+* `lags`: Specifies the lags at which the cross-covariance is computed.
+* `demean`: A `Bool` type variable, this is a named paramter. 
+
+Compute cross-covariance of `x` at specified lags. If both `x` and `y` are vectors, it returns a vector of th same length of lags. Otherwise, it computes cross covariances between each pairs of columns in x and y.
```
nalimilan: th -> the (plus other corrections from above)

src/scalarstats.jl
```diff
@@ -27,7 +37,13 @@ function harmmean(a::RealArray)
     return n / s
 end
 
+
 # Trimmed mean
+"""
+    trimmean(a::RealArray)
+
```
simonbyrne: This function takes 2 arguments

ararslan:
Superseded by #172? cc @nalimilan

nalimilan:
Yeah, I guess. Though if you find a good idea to take from this PR which isn't already applied...

PR #150, opened 2016-02-21
Bump version requirement to 0.4 and remove evaluate function.

andreasnoack:
This eliminates a ton of deprecation warnings for call on 0.5.

cc: @nalimilan Maybe these caused the failures you saw in #146

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -141,7 +141,7 @@ end
         @nloops N i d->(d>1? (1:size(A,d)) : (1:1)) d->(j_d = sizeR_d==1 ? 1 : i_d) begin
             @inbounds r = (@nref N R j)
             for i_1 = 1:sizA1
-                @inbounds r += evaluate(f, (@nref N A i)) * w[i_1]
+                @inbounds r += f(@nref N A i) * w[i_1]
             end
             @inbounds (@nref N R j) = r
         end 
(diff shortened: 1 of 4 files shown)


src/weights.jl
```diff
@@ -141,7 +141,7 @@ end
         @nloops N i d->(d>1? (1:size(A,d)) : (1:1)) d->(j_d = sizeR_d==1 ? 1 : i_d) begin
             @inbounds r = (@nref N R j)
             for i_1 = 1:sizA1
-                @inbounds r += evaluate(f, (@nref N A i)) * w[i_1]
+                @inbounds r += f((@nref N A i)) * w[i_1]
```
nalimilan: Can't you get rid of the double parentheses here (and below)?

nalimilan:
Thanks, but unfortunately there's still a strange segfault on 0.5... But the nightlies are old enough that it may have been fixed since. Tests don't crash when I run them on my machine.

Anyway this looks good to merge.

andreasnoack:
I can reproduce the segfault locally if I run the tests with `--inline=no`.

andreasnoack:
As mentioned, I've isolated the segfault to the changes in Julia. I'll merge this to clean up all the  warnings. Soon, but probably after the tests can run again, we should tag a new minor version because the version requirement for Julia is bumped.

PR #152, opened 2016-03-14
Added tests for weights.jl and misc.jl

Dawny33:
Added tests for the following:

In the `weights.jl` tests:

- `weights` for `weights(vs::RealArray) = WeightVec(vec(vs))`
- `Base.sum(v::BitArray, w::WeightVec) = wsum(v, values(w))`
- `Base.sum(v::SparseMatrixCSC, w::WeightVec) = wsum(v, values(w))`
- `wsum(v::AbstractArray, w::AbstractVector) = dot(vec(v), w)`

In the `misc.jl` file:

- Added tests for `levelsmap`

### DIFF
--- a/test/misc.jl
+++ b/test/misc.jl
@@ -15,6 +15,14 @@ vals, lens = rle(z)
 @test lens == [2, 2, 1, 1, 3]
 @test inverse_rle(vals, lens) == z
 
+# levelsmap
+a = [1, 1, 2, 2, 2, 3, 1, 2, 2, 3, 3, 3, 3, 2]
+b = [true, false, false, true, false, true, true, false]
+
+@test levelsmap(a) == Dict([(2, 2), (3, 3), (1, 1)])
+@test levelsmap(b) == Dict([(false, 2), (true, 1) ])
+
+
 # findat
 
 findat([2, 4, 6, 8, 10], [8, 6, 4]) == [4, 3, 2]
--- a/test/misc.jl
+++ b/test/misc.jl
@@ -19,8 +19,8 @@ vals, lens = rle(z)
 a = [1, 1, 2, 2, 2, 3, 1, 2, 2, 3, 3, 3, 3, 2]
 b = [true, false, false, true, false, true, true, false]
 
-@test levelsmap(a) == Dict([(2, 2), (3, 3), (1, 1)])
-@test levelsmap(b) == Dict([(false, 2), (true, 1) ])
+@test levelsmap(a) == Dict(2=>2, 3=>3, 1=>1)
+@test levelsmap(b) == Dict(false=>2, true=>1)
 
 
 # findat
--- a/test/weights.jl
+++ b/test/weights.jl
@@ -11,28 +11,20 @@ using Base.Test
 
 w  = [1., 2., 3.]
 wv = weights(w)
-@test eltype(wv) == Float64
-@test length(wv) == 3
+@test eltype(wv) === Float64
+@test length(wv) === 3
 @test values(wv) === w
 @test sum(wv) === 6.0
 @test !isempty(wv)
 
 b  = trues(3)
 bv = weights(b)
-@test eltype(bv) == Bool
-@test length(bv) == 3
-@test values(bv) == b
-@test sum(bv)    == 3
+@test eltype(bv) === Bool
+@test length(bv) === 3
+@test values(bv) === b
+@test sum(bv)    === 3
 @test !isempty(bv)
 
-c  = [1, 4, 3, 5]; d = [4, 7, 18, 9]; e = [1, 2, -5, 3];
-s  = sparse(c, d, e)
-cv = weights(c)
-@test eltype(cv) == Int64
-@test length(cv) == 4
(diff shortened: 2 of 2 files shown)


test/misc.jl
```diff
@@ -15,6 +15,13 @@ vals, lens = rle(z)
 @test lens == [2, 2, 1, 1, 3]
 @test inverse_rle(vals, lens) == z
 
+# levelsmap
+y = [1, 1, 2, 2, 2, 3, 1, 2, 2, 3, 3, 3, 3]
+
+@test levelsmap(y) == Dict([(2, 2), (3, 3), (1, 1)])
+@test levelsmap(z) == Dict([(false, 2), (true, 1) ])
```
nalimilan: Probably better use a new vector here instead of `z`, which is redefined in each block above. It would be clearer (you can even use a `let` block around the two instructions you added).

test/weights.jl
```diff
@@ -15,13 +17,33 @@ wv = weights(w)
...
+b  = trues(3)
+bv = weights(b)
+@test eltype(bv) == Bool
+@test length(bv) == 3
+@test values(bv) == b
+@test sum(bv)    == 3
+@test !isempty(bv)
+
+c  = [1, 4, 3, 5]; d = [4, 7, 18, 9]; e = [1, 2, -5, 3];
+s  = sparse(c, d, e)
```
nalimilan: What are `s`, `d` and `e` for?

test/weights.jl
```diff
@@ -15,13 +17,33 @@ wv = weights(w)
...
+cv = weights(c)
+@test eltype(cv) == Int64
+@test length(cv) == 4
+@test values(cv) == c
+@test sum(cv)    == 13.0
```
nalimilan: Better use `===` here and above, to check the return type too (should be `13` here, not `13.0`).

nalimilan:
Thanks!

test/misc.jl
Dawny33: I did had second thoughts about using `z`, but as I felt it'd be a nice example for the `levelsmap` implementation, I went ahead with it.

I'd use a different vector instead of it now.

test/weights.jl
Dawny33: `c`, `d` and `e` are vectors which I used for constructing the sparse matrix `s`.

test/weights.jl
Dawny33: Yeah, `===` makes the test more robust. I would add it to [this line](https://github.com/JuliaStats/StatsBase.jl/pull/152#discussion-diff-56038439L15) too.

Dawny33:
^ Oops!  Working on it!

Dawny33:
@nalimilan Pl review!

test/weights.jl
nalimilan: But `s` isn't used, right?

test/misc.jl
```diff
@@ -15,6 +15,14 @@ vals, lens = rle(z)
 @test lens == [2, 2, 1, 1, 3]
 @test inverse_rle(vals, lens) == z
 
+# levelsmap
+a = [1, 1, 2, 2, 2, 3, 1, 2, 2, 3, 3, 3, 3, 2]
+b = [true, false, false, true, false, true, true, false]
+
+@test levelsmap(a) == Dict([(2, 2), (3, 3), (1, 1)])
```
nalimilan: The best syntax for this is `Dict(2=>2, 3=>3, 1=>1)`.

test/weights.jl
```diff
@@ -5,23 +5,45 @@ using Base.Test
...
-@test sum(wv) == 6.0
+@test sum(wv) === 6.0
 @test !isempty(wv)
 
+b  = trues(3)
+bv = weights(b)
+@test eltype(bv) == Bool
+@test length(bv) == 3
+@test values(bv) == b
+@test sum(bv)    == 3
```
nalimilan: Use `===` here too (you could even use it for all lines here).

test/weights.jl
```diff
@@ -5,23 +5,45 @@ using Base.Test
...
 ## wsum
 
 x = [6., 8., 9.]
 w = [2., 3., 4.]
+p = [1. 2. ; 3. 4.]
+q = [1., 2., 3., 4.]
 
 @test wsum(Float64[], Float64[]) === 0.0
 @test wsum(x, w) == 72.0
+@test wsum(p, q) == 29.0
```
nalimilan: Use `===` here and above.

Dawny33:
@nalimilan Removed the sparse matrix test, as it doesn't solve the purpose.

Fixed the others by replacing the `==` with a `===` as suggested.

Changed the `Dict` format as suggested.

nalimilan:
Thanks! Other PRs like that would be welcome. ;-)

I'll leave it open for a few days to leave the occasion for others to comment, then I'll merge.

Dawny33:
@nalimilan Sure!

Meanwhile, you can have a look at this too: https://github.com/JuliaStats/StatsBase.jl/pull/153

PR #153, opened 2016-03-15
Added tests for sample pair functions

Dawny33:


nalimilan:
Looks good, but again, it would be even better to use `===`, and maybe to call `samplepair` on other cases (e.g. with `10`, and with `[1,2]` to cover both corner cases and more general ones). You can also call `srand` only once, the result will still be deterministic. Finally, I think two blank lines should be enough to separate sections -- but that's minor.

Dawny33:
@nalimilan 

- Yeah, I would add the `===` instead of `==`
- Would be adding two more testcases with the `10` and `[1, 2]` 
- I tried with using `srand` only once, and also with setting seed in `let` block, and using it once inside it. However, it wasn't deterministic. And that's why I used `srand` twice.

nalimilan:
> I tried with using srand only once, and also with setting seed in let block, and using it once inside it. However, it wasn't deterministic. And that's why I used srand twice.

Not sure what's going on, but it certainly is here:
```julia
julia> srand(1);

julia> samplepair(2)
(1,2)

julia> samplepair(10)
(7,3)

julia> samplepair([3, 4, 2, 6, 8])
(4,3)

julia> samplepair([1, 2])
(1,2)

julia> srand(1);

julia> samplepair(2)
(1,2)

julia> samplepair(10)
(7,3)

julia> samplepair([3, 4, 2, 6, 8])
(4,3)

julia> samplepair([1, 2])
(1,2)

```

Dawny33:
@nalimilan Surprising!  As the following snippet:

```
srand(1);

@test samplepair(2) === (1, 2)

# srand(1)
@test samplepair(10) === (1, 5)

# srand(1)
@test samplepair([3, 4, 2, 6, 8]) === (3, 4)

# srand(1)
@test samplepair([1, 2]) === (1, 2)
```

threw an error while `Pkg.test()`. Did I miss something?

nalimilan:
The result will be different depending on the relative position of the call to `samplepair`. `srand` defines the seed, and then you'll always get the same *sequence* of random numbers, but not *repeatedly* the same random number. So you need to update the results you test against.

Dawny33:
@nalimilan Fair enough, it worked :sparkles: 

Thanks for the help.  Pl review!

nalimilan:
Thanks!

PR #155, opened 2016-04-19
Support Integer Printing in CoefTable

panlanfeng:
Currently the `CoefTable` does not support integer and print everything as float number. So it will print the degree of freedom of a parameter as "3.00", which can be confusing. 

The easiest way to fix this is adding a new field 

``` julia
integercols::Vector{Int}
```

Then 

``` julia
 CoefTable(rand(3,4), ["Estimate", "Stderr", "df", "p"], ["b1", "b2", "b3"], 4, [3;])
```

will print 

```
      Estimate   Stderr df      p
b1     0.51669 0.813798  1 0.8844
b2   0.0686266 0.378539  1 0.2470
b3    0.381557 0.600014  1 0.2409
```

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -164,10 +164,10 @@ type CoefTable
     rownms::Vector
     pvalcol::Integer
     integercols::Vector{Int}
-    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0,integercols::Vector{Int}=[0;])
+    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0,integercols::Vector{Int}=Int[])
         nr,nc = size(mat)
         0 <= pvalcol <= nc || error("pvalcol = $pvalcol should be in [0,...,$nc]")
-        all(0 .<= integercols .<= nc) || error("integercols = $integercols should be a vector with elements in [0,...,$nc]")
+        all(0 .<= integercols .<= nc) || error("integercols = $integercols should be a vector with elements in [1,...,$nc]")
         length(colnms) in [0,nc] || error("colnms should have length 0 or $nc")
         length(rownms) in [0,nr] || error("rownms should have length 0 or $nr")
         new(mat,colnms,rownms,pvalcol,integercols)
@@ -202,10 +202,8 @@ function show(io::IO, ct::CoefTable)
         end
     end
     for ic in ct.integercols
-        if ic != 0
-            for i in 1:nr
-                str[i,ic] = @sprintf("%u",mat[i,ic])
-            end
+        for i in 1:nr
+            str[i,ic] = sprint(showcompact,Int(mat[i,ic]))
         end
     end
     for j in 1:nc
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks for the PR. Looks like a good idea. Though I wonder whether we shouldn't choose a more general solution and change the `mat::Matrix` field to `mat::Vector{Vector}}`. Maybe not worth it at this stage.

src/statmodels.jl
```diff
@@ -163,12 +163,14 @@ type CoefTable
     colnms::Vector
     rownms::Vector
     pvalcol::Integer
-    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0)
+    integercols::Vector{Int}
+    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0,integercols::Vector{Int}=[0;])
```
nalimilan: Use `Int[]` instead of `[0;]`.

src/statmodels.jl
```diff
@@ -163,12 +163,14 @@ type CoefTable
...
         nr,nc = size(mat)
-        0 <= pvalcol <= nc || error("pvalcol = $pvalcol should be in 0,...,$nc]")
+        0 <= pvalcol <= nc || error("pvalcol = $pvalcol should be in [0,...,$nc]")
+        all(0 .<= integercols .<= nc) || error("integercols = $integercols should be a vector with elements in [0,...,$nc]")
```
nalimilan: Change `0` to `1`.

src/statmodels.jl
```diff
@@ -199,6 +201,13 @@ function show(io::IO, ct::CoefTable)
             str[i,pvc] = format_pvc(mat[i,pvc])
         end
     end
+    for ic in ct.integercols
+        if ic != 0
```
nalimilan: Then you don't need this check.

src/statmodels.jl
```diff
@@ -199,6 +201,13 @@ function show(io::IO, ct::CoefTable)
...
+            for i in 1:nr
+                str[i,ic] = @sprintf("%u",mat[i,ic])
```
nalimilan: `%i` is more appropriate, as the integer might be signed (though currently it doesn't make any difference). Anyway, I think a better solution is `sprint(showcompact,Int(mat[i,ic]))`: this is consistent with the code used for other columns, and if for some reason the value isn't an integer, an error will be raised, instead of rounding silently.

simonbyrne:
What if we were just to make `CoefTable` a DataFrame?

nalimilan:
Well, contrary to R, Julia DataFrames are oriented towards databases. For example, they don't support rownames. The only feature of DataFrames we would use here is the fact that they are a vector of heterogeneous columns. I don't think the dependency is worth it.

But I agree a lightweight structure to hold and print results would be useful. We've discussed this in another issue already: https://github.com/JuliaStats/StatsBase.jl/pull/138#issuecomment-190784648

panlanfeng:
Instead of using `mat::Vector{Vector}}` it may be more flexible if we use `mat::Matrix{String}`. 
I like the idea of returning an table like object beside just printing some message. Right now pasting the result into a latex file is painful and I have to write another function to do it.

nalimilan:
Works now, though with a different syntax:
```julia
julia> CoefTable(Any[rand(3), rand(3), rand(1:10, 3), rand(0.0:0.01:1.0, 3)], ["Estimate", "Stderr", "df", "p"], ["b1", "b2", "b3"])
     Estimate   Stderr df    p
b1   0.891686 0.930906  6 0.96
b2   0.496783  0.28337  7 0.91
b3   0.131123 0.546505  7 0.55
```

PR #156, opened 2016-04-22
Allow Any type in CoefTable

panlanfeng:
Allow `CoefTable` columns have different types and length.

An Example
~~~julia
srand(10)
v1 = rand(3)
v2 = ["Good", "Great"]
v3 = rand(Int8, 3)
v4 = PValue(rand(3)./10000)
m = rand(3,4)
CoefTable((v1, v2, v3, v4),
    ["Estimate", "Comments", "df", "p"],
    ["x1", "x2", "x3"])
~~~

This will print 
~~~
     Estimate Comments  df     p
x1   0.112582     Good  88 <1e-4
x2   0.368314    Great -90 <1e-4
x3   0.344454          -80 <1e-4
~~~

The previous way is still working:
~~~julia
CoefTable(rand(3,4), ["Estimate", "Stderr", "df", "p"], ["x1", "x2", "x3"], 4)
~~~

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -159,46 +159,61 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
 
 ## Nms are the coefficient names, corresponding to rows in the table
 type CoefTable
-    mat::Matrix
+    cols::Vector
     colnms::Vector
     rownms::Vector
-    pvalcol::Integer
-    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0)
-        nr,nc = size(mat)
-        0 <= pvalcol <= nc || error("pvalcol = $pvalcol should be in 0,...,$nc]")
+    function CoefTable(cols::Vector,colnms::Vector,rownms::Vector)
+        nc = length(cols)
+        nrs = map(length,cols)
+        nr = nrs[1]
         length(colnms) in [0,nc] || error("colnms should have length 0 or $nc")
         length(rownms) in [0,nr] || error("rownms should have length 0 or $nr")
-        new(mat,colnms,rownms,pvalcol)
+        all(nrs .== nr) || error("Elements of cols should have equal lengths, but got $nrs")
+        new(cols,colnms,rownms)
+    end
+    
+    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0)
+        nc = size(mat,2)
+        cols = Any[mat[:, i] for i in 1:nc]
+        if pvalcol != 0                         # format the p-values column
+            cols[pvalcol] = [PValue(cols[pvalcol][j])
+                            for j in eachindex(cols[pvalcol])]
+        end
+        CoefTable(cols,colnms,rownms)
     end
 end
 
-## format numbers in the p-value column
-function format_pvc(pv::Number)
-    if isnan(pv)
-        return @sprintf("%d", pv)
+type PValue
+    v::Number
+    function PValue(v::Number)
+        0. <= v <= 1. || isnan(v) || error("p-values must be in [0.,1.]")
+        new(v)
     end
-    0. <= pv <= 1. || error("p-values must be in [0.,1.]")
-    if pv >= 1e-4
-        return @sprintf("%.4f", pv)
+end
+
+function show(io::IO, pv::PValue)
+    v = pv.v
+    if isnan(v)
+        @printf(io,"%d", v)
+    elseif v >= 1e-4
+        @printf(io,"%.4f", v)
     else
-        return @sprintf("<1e%2.2d", ceil(Integer, max(nextfloat(log10(pv)), -99)))
+        @printf(io,"<1e%2.2d", ceil(Integer, max(nextfloat(log10(v)), -99)))
     end
 end
 
 function show(io::IO, ct::CoefTable)
-    mat = ct.mat; nr,nc = size(mat); rownms = ct.rownms; colnms = ct.colnms; pvc = ct.pvalcol
+    cols = ct.cols; rownms = ct.rownms; colnms = ct.colnms;
+    nc = length(cols)
+    nr = length(cols[1])
     if length(rownms) == 0
         rownms = [lpad("[$i]",floor(Integer, log10(nr))+3) for i in 1:nr]
     end
     rnwidth = max(4,maximum([length(nm) for nm in rownms]) + 1)
     rownms = [rpad(nm,rnwidth) for nm in rownms]
     widths = [length(cn)::Int for cn in colnms]
-    str = [sprint(showcompact,mat[i,j]) for i in 1:nr, j in 1:nc]
-    if pvc != 0                         # format the p-values column
-        for i in 1:nr
-            str[i,pvc] = format_pvc(mat[i,pvc])
-        end
-    end
+    str = ByteString[isa(cols[j][i], AbstractString) ? cols[j][i] :
+        sprint(showcompact,cols[j][i]) for i in 1:nr, j in 1:nc]
     for j in 1:nc
         for i in 1:nr
             lij = length(str[i,j])
--- a/test/statmodels.jl
+++ b/test/statmodels.jl
@@ -1,10 +1,32 @@
 using StatsBase
 using Base.Test
 
-## format_pvc: Formatting of p-values
-@test StatsBase.format_pvc(1.0) == "1.0000"
-@test StatsBase.format_pvc(1e-1) == "0.1000"
-@test StatsBase.format_pvc(1e-5) == "<1e-4"
-@test StatsBase.format_pvc(NaN) == "NaN"
-@test_throws ErrorException StatsBase.format_pvc(-0.1)
-@test_throws ErrorException StatsBase.format_pvc(1.1)
+srand(10)
+v1 = rand(3)
+v2 = ["Good", "Great", "Bad"]
+v3 = rand(Int8, 3)
+v4 = [StatsBase.PValue(rand()./10000) for i in 1:3]
+m = rand(3,4)
+@test sprint(show, CoefTable(Any[v1, v2, v3, v4],
+    ["Estimate", "Comments", "df", "p"],
+    ["x1", "x2", "x3"])) == """\
+     Estimate Comments  df     p
+x1   0.112582     Good  88 <1e-4
+x2   0.368314    Great -90 <1e-4
+x3   0.344454      Bad -80 <1e-4
+"""
+
+@test sprint(show, CoefTable(m, ["Estimate", "Stderr", "df", "p"],
+    ["x1", "x2", "x3"], 4)) == """\
+     Estimate   Stderr       df      p
+x1   0.819778 0.844007 0.923676 0.1717
+x2   0.669931  0.67919 0.066098 0.4204
+x3   0.453058  0.72525 0.999172 0.5567
+"""
+
+@test sprint(show, StatsBase.PValue(1.0)) == "1.0000"
+@test sprint(show, StatsBase.PValue(1e-1)) == "0.1000"
+@test sprint(show, StatsBase.PValue(1e-5)) == "<1e-4"
+@test sprint(show, StatsBase.PValue(NaN)) == "NaN"
+@test_throws ErrorException StatsBase.PValue(-0.1)
+@test_throws ErrorException StatsBase.PValue(1.1)
--- a/test/statquiz.jl
+++ b/test/statquiz.jl
@@ -69,7 +69,7 @@ println("\nII F")
 print("Testing regression: ")
 ctable = coeftable(lm(big ~ x, nasty))
 @test typeof(ctable) == CoefTable
-@test_approx_eq ctable.mat[:,1] [99999990, 1]
+@test_approx_eq ctable.cols[1] [99999990, 1]
 
 @test sprint(show, ctable) == """\
              Estimate Std.Error t value Pr(>|t|)


src/statmodels.jl
```diff
@@ -160,17 +160,50 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+end
+
+type PValue
+    v::Number
+    function PValue(v::Number)
+        0. <= v <= 1. || error("p-values must be in [0.,1.]")
+        new(v)
+    end
+end
+function PValue(v::Array)
```
nalimilan: This can be written more simply as `PValue(v::AbstractVector) = convert(Vector{PValue}, v)`. But do you really need this function? Why not call `convert` directly?

src/statmodels.jl
```diff
@@ -160,17 +160,50 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+    v2 = similar(v,PValue)
+    for i in eachindex(v)
+        v2[i] = PValue(v[i])
     end
+    v2
 end
+convert(::Type{PValue}, x::Number) = PValue(x)
+convert{T <: Number}(::Type{T}, x::PValue) = T(x.v)
+promote_rule{T<:Number}(::Type{T}, ::Type{PValue}) = T
```
nalimilan: This can't be right. Maybe `promote_rule{S<:Number, T<:Number}(::Type{T}, ::Type{PValue{S}}) = PValue{promote_type(S, T)}`? But do we really need this? (Same for the second conversion above.)

src/statmodels.jl
```diff
@@ -185,18 +218,35 @@ function format_pvc(pv::Number)
     end
 end
 
+function show(io::IO, pv::PValue)
```
nalimilan: Can you get rid of `format_pvc` now that you added this function? Code duplication isn't great.

src/statmodels.jl
```diff
@@ -160,17 +160,50 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+    cols::Tuple
     colnms::Vector
     rownms::Vector
-    pvalcol::Integer
-    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0)
-        nr,nc = size(mat)
-        0 <= pvalcol <= nc || error("pvalcol = $pvalcol should be in 0,...,$nc]")
+    function CoefTable(cols::Tuple,colnms::Vector,rownms::Vector)
+        nc = length(cols)
+        nrs = map(length,cols)
```
nalimilan: I would simply assert that all columns are of equal length, and then use `nr` as before. I can't see a use case for providing columns with different lengths (which is more likely to be a caller error).

src/statmodels.jl
```diff
@@ -160,17 +160,50 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
 ## Nms are the coefficient names, corresponding to rows in the table
 type CoefTable
     mat::Matrix
+    cols::Tuple
```
nalimilan: Given that other fields are vector, it would be more consistent to make this a vector too. Though all could be changed to tuples.

src/StatsBase.jl
```diff
@@ -146,6 +146,7 @@ module StatsBase
     indicatormat,   # construct indicator matrix
 
     # statistical models
+    PValue,
```
nalimilan: Probably better not export this very small type until we have a real need for it.

src/statmodels.jl
```diff
@@ -160,17 +160,50 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+        nr = max(nrs...)
         length(colnms) in [0,nc] || error("colnms should have length 0 or $nc")
         length(rownms) in [0,nr] || error("rownms should have length 0 or $nr")
-        new(mat,colnms,rownms,pvalcol)
+        if all(unique(nrs) .== nr)
+            mat = hcat(cols...)
+        else
+            mat = Matrix(nr,nc) # A better value here?
+        end
+        new(mat,cols,colnms,rownms)
```
nalimilan: I don't see the point of creating a matrix `mat` here, which is redundant with `cols`.

nalimilan:
Thanks! This looks like a nice generalization.

src/statmodels.jl
panlanfeng: `mat` is kept for compatibility so that any code trying to access the `mat` field can still work.

src/statmodels.jl
panlanfeng: It is easy to create a tuple by `(v1, v2, ...)` when calling the function `CoefTable`. 
But it is not so easy to create a vector of vectors.

src/statmodels.jl
panlanfeng: After a second thought, I agree with you.

src/statmodels.jl
panlanfeng: A such function is for the convenience of the users. When a user is trying to input a column of `PValue`, instead of writing 
~~~julia
v1 = rand(4)
convert(Vector{StatsBase.PValue}, v1)
~~~
he can simply write
~~~julia
PValue(v1)
~~~

src/statmodels.jl
nalimilan: I don't think we need to preserve compatibility here.

src/statmodels.jl
nalimilan: In Julia `[v1, v2, ...]` will no longer concatenate, and we could have a temporary convenience method if needed. Anyway, I'm all for using a tuple, but then could you change other fields to tuples too?

src/statmodels.jl
panlanfeng: I agree the `promote_rule` is weird. But this is for the use of `hcat`ing numeric vectors and `PValue` vector, so that they can still return a numeric matrix.
If we all agree it is not necessary to provide a `mat` field, I am happy to remove it and therefore it is not necessary to `hcat`.

I was hoping the `PValue` behave like a regular `Number` just with a different `show` method. But I didn't find a way to do that. Any suggestions?

src/statmodels.jl
nalimilan: But do we really expect users to create such vectors? As I said, I don't think we need to export `PValue` for now.

src/statmodels.jl
nalimilan: Let's add only the parts needed to get `CoefTable` to work. I'm not sure making `PValue` a type usable on its own is really a priority.

panlanfeng:
Now the `CoefTable` requires all vectors in the tuple `cols` have the same lengths.
The field `mat` is removed.
The remaining issue is `colnms` and `rownms` are still vectors. Changing to `Tuple` will fail `DataFrames`, see [here](https://github.com/JuliaStats/DataFrames.jl/blob/master/src/statsmodels/statsmodel.jl#L96).
Since it is easy to move to `Tuple`, we still keep `Vector` for now.

src/statmodels.jl
```diff
@@ -159,44 +159,70 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
-        nr,nc = size(mat)
-        0 <= pvalcol <= nc || error("pvalcol = $pvalcol should be in 0,...,$nc]")
+    function CoefTable(cols::Tuple,colnms::Vector,rownms::Vector)
+        nc = length(cols)
+        nrs = map(length,cols)
+        nr = max(nrs...)
         length(colnms) in [0,nc] || error("colnms should have length 0 or $nc")
         length(rownms) in [0,nr] || error("rownms should have length 0 or $nr")
-        new(mat,colnms,rownms,pvalcol)
+        all(unique(nrs) .== nr) || error("Elements of cols should have equal lengths")
```
nalimilan: I don't think you need to call `unique` here. Also, no need to take the `max`, just use the first length. Finally, it would be nice to add `", but got $nrs"` at the end of the error message.

src/statmodels.jl
```diff
@@ -159,44 +159,70 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+        new(cols,colnms,rownms)
+    end
+    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0)
```
nalimilan: Please add line breaks between functions (at least in the long syntax).

src/statmodels.jl
```diff
@@ -159,44 +159,70 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+        nr,nc = size(mat)
```
nalimilan: `nr` isn't used here, use `nc = size(mat, 2)`.

src/statmodels.jl
```diff
@@ -159,44 +159,70 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+        cols = Any[mat[:, i] for i in 1:nc]
+        if pvalcol != 0                         # format the p-values column
+            cols[pvalcol] = convert(Vector{PValue}, cols[pvalcol])
+        end
+        cols = ntuple((i)->cols[i],nc)
```
nalimilan: `tuple(cols...)` does the same. But, really, making it a vector would be simpler and more consistent with other fields (since changing the latter would break DataFrames). People can do `Any[v1, v2...]` to disable concatenation in 0.4.

src/statmodels.jl
```diff
@@ -159,44 +159,70 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
+        new(v)
     end
-    0. <= pv <= 1. || error("p-values must be in [0.,1.]")
-    if pv >= 1e-4
-        return @sprintf("%.4f", pv)
+end
+
+convert(::Type{PValue}, x::Number) = PValue(x)
+convert{T <: Number}(::Type{T}, x::PValue) = T(x.v)
+#PValue(v::AbstractArray) = convert(AbstractArray{PValue}, v)
```
nalimilan: Why keep this commented out line?

src/statmodels.jl
```diff
@@ -159,44 +159,70 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
     rnwidth = max(4,maximum([length(nm) for nm in rownms]) + 1)
     rownms = [rpad(nm,rnwidth) for nm in rownms]
     widths = [length(cn)::Int for cn in colnms]
-    str = [sprint(showcompact,mat[i,j]) for i in 1:nr, j in 1:nc]
-    if pvc != 0                         # format the p-values column
+    str = ["" for i in 1:nr, j in 1:nc]
+    for j in 1:nc
         for i in 1:nr
-            str[i,pvc] = format_pvc(mat[i,pvc])
+            if i <= nrs[j]
```
nalimilan: Just use `nr` here (defined as `nrs[1]` instead of `max`).

panlanfeng:
Thanks for the suggestions!
Now the `cols` field is a vector of `Any`.

src/statmodels.jl
```diff
@@ -159,46 +159,64 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
     colnms::Vector
     rownms::Vector
-    pvalcol::Integer
-    function CoefTable(mat::Matrix,colnms::Vector,rownms::Vector,pvalcol::Int=0)
-        nr,nc = size(mat)
-        0 <= pvalcol <= nc || error("pvalcol = $pvalcol should be in 0,...,$nc]")
+    function CoefTable(cols::Vector,colnms::Vector,rownms::Vector)
+        nc = length(cols)
+        nrs = map(length,cols)
+        nr = max(nrs...)
```
nalimilan: Better do `nr = nrs[1]`, we don't actually need the max.

src/statmodels.jl
```diff
@@ -159,46 +159,64 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
-        return @sprintf("<1e%2.2d", ceil(Integer, max(nextfloat(log10(pv)), -99)))
+        @printf(io,"<1e%2.2d", ceil(Integer, max(nextfloat(log10(v)), -99)))
     end
 end
 
 function show(io::IO, ct::CoefTable)
-    mat = ct.mat; nr,nc = size(mat); rownms = ct.rownms; colnms = ct.colnms; pvc = ct.pvalcol
+    cols = ct.cols; rownms = ct.rownms; colnms = ct.colnms;
+    nc = length(cols)
+    nrs = map(length,cols)
```
nalimilan: Same here. You don't need to extract the length of all columns, just assume they are of the same length.

src/statmodels.jl
```diff
@@ -159,46 +159,64 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
     rnwidth = max(4,maximum([length(nm) for nm in rownms]) + 1)
     rownms = [rpad(nm,rnwidth) for nm in rownms]
     widths = [length(cn)::Int for cn in colnms]
-    str = [sprint(showcompact,mat[i,j]) for i in 1:nr, j in 1:nc]
-    if pvc != 0                         # format the p-values column
-        for i in 1:nr
-            str[i,pvc] = format_pvc(mat[i,pvc])
-        end
-    end
+    str = [isa(cols[j][i], AbstractString) ? cols[j][i] : sprint(showcompact,cols[j][i]) for i in 1:nr, j in 1:nc]
```
nalimilan: Type inference gets confused by the condition: add `ByteString` as a type prefix to the comprehension. Also, break the line at 92 chars.

nalimilan:
Thanks, should be good after fixing my last comments and squashing the commits. Then I'll leave the PR open for comments for a few days, and merge if nobody objects.

nalimilan:
Ah, the `convert{T <: Number}(::Type{T}, x::PValue) = T(x.v)` method isn't tested currently. Either add a test for it or remove it. It would also be good to test `show()` on `PValue(NaN)`.

panlanfeng:
Remove the `convert` method since it is not necessary. 
And now `PValue`  can print `NaN`. Thanks for remind!

src/statmodels.jl
```diff
@@ -159,46 +159,64 @@ df_residual(obj::RegressionModel) = error("df_residual is not defined for $(type
...
 end
 
-## format numbers in the p-value column
-function format_pvc(pv::Number)
-    if isnan(pv)
-        return @sprintf("%d", pv)
+type PValue
+    v::Number
+    function PValue(v::Number)
+        if isnan(v)
```
nalimilan: How about combining the two checks into a single one?

panlanfeng:
Just rebased on the latest master.
Hope this to be merged soon. One of my package under development need to print `Int` in `CoefTable`.

KristofferC:
This @coveralls fellow is a noisy one..

nalimilan:
> This @coveralls fellow is a noisy one..

I think we can disable these messages, but I don't know how to do that nor whether I have the permissions to do so.

@panlanfeng Thanks!

PR #159, opened 2016-05-11
Preparing for 0.5. Stop using functors.

andreasnoack:
...and `scale`.


nalimilan:
LGTM. Though there still is an error about `checksize`, and lots of deprecation warnings.

andreasnoack:
That is because the tests use `DataArrays.jl` so when that is merged and tagged these should go away. Locally, with the PR version of `DataArrays.jl`, I don't see any warnings.

PR #166, opened 2016-05-28
Allow one to call `histrange` with just the endpoints & nbins

timholy:
This makes `histrange` more modular, allowing one to use just the endpoints rather than needing the whole vector.


andreasnoack:
cc: @simonbyrne

simonbyrne:
Great idea, thanks!

PR #171, opened 2016-06-10
Delete varzm now that it has been deleted from Base

andreasnoack:
There doesn't seem to be a time penalty for using `varm` with `m=0`.

Temporarily comment out `statquiz.jl` since it depends on `DataArrays` which is broken until `varzm` has been fixed/deleted

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -14,8 +14,8 @@ tests = ["weights",
          "misc",
          "sampling",
          "wsampling",
-         "statmodels",
-         "statquiz"]
+         "statmodels"]
+         # ,"statquiz"]
 
 println("Running tests:")
 
--- a/src/moments.jl
+++ b/src/moments.jl
@@ -2,36 +2,23 @@
 
 ## var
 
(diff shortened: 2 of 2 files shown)


test/runtests.jl
```diff
@@ -14,8 +14,8 @@ tests = ["weights",
          "misc",
          "sampling",
          "wsampling",
-         "statmodels",
-         "statquiz"]
+         "statmodels"]
+         # ,"statquiz"]
```
nalimilan: Why comment this out?

nalimilan: Carry on, I hadn't seen the second commit.

nalimilan:
Would probably have been better to go through a deprecation process in Base, but likely not a frequently used function.

andreasnoack:
I agree. @vtjnash's removal in https://github.com/JuliaLang/julia/pull/16764 was a bit abrupt. Initially, I tried to define the deprecation in Base, but the use of this function is so limited that I thought it was just simpler to remove it from `StatsBase` and `DataArrays`.

PR #172, opened 2016-06-10
RFC: Documentation and cleanup

ararslan:
I've been going through adding docstrings where none were present, fixing mixed indentation (tabs -> 4 spaces so everything matches), and fixing docstrings that weren't ideally formatted.

~~I haven't made it all the way through yet but I figured I'd put this up here for now and continue to add to it.~~ ~~I think I'm done?~~ I'm not quite there but I'm working on it!


ararslan:
@andreasnoack 'Twas an arduous journey but I believe I have succeeded. Want to take a look? I think I got everything.

ararslan:
Side note: It might make sense to set the coverage decrease threshold for failure in the Coveralls settings so that a decrease of -0.04% doesn't show up as a failed check...

src/counts.jl
```diff
@@ -9,53 +9,81 @@
...
-counts(x::IntegerArray, levels::IntUnitRange) = addcounts!(zeros(Int, length(levels)), x, levels)
-counts(x::IntegerArray, levels::IntUnitRange, wv::WeightVec) = addcounts!(zeros(eltype(wv), length(levels)), x, levels, wv)
+
+"""
+    counts(x, [levels,] [k,] [wv]) -> Vector
+
+Count the occurrences of the integers in `x`. The occurrences in
+`x` of the values in `levels` are counted if `levels` is given,
+or of the integers from 1 to `k` if `k` is given. A weighting
+vector `wv` can be specified to weight the counts.
```
nalimilan: Maybe be more explicit about what happens when the weights are specified (as above).

src/counts.jl
```diff
@@ -9,53 +9,81 @@
...
+        end
+    end
+    return r
 end
 
-counts(x::IntegerArray, levels::IntUnitRange) = addcounts!(zeros(Int, length(levels)), x, levels)
-counts(x::IntegerArray, levels::IntUnitRange, wv::WeightVec) = addcounts!(zeros(eltype(wv), length(levels)), x, levels, wv)
+
+"""
+    counts(x, [levels,] [k,] [wv]) -> Vector
```
nalimilan: The signature is incorrect here: `levels` and `k` are the same argument. Rather write this as two different lines:
```julia
    counts(x, [levels], [wv])::Vector
    counts(x, [k::Integer], [wv])::Vector
```

You also need to say "if `levels` is an integer range..." and "if `k` is an integer scalar`.

nalimilan: Also note the new syntax to declare return types in Julia 0.5. Better use that everywhere instead of `->`.

src/counts.jl
```diff
@@ -9,53 +9,81 @@
...
+counts(x::IntegerArray, levels::IntUnitRange, wv::WeightVec) =
+    addcounts!(zeros(eltype(wv), length(levels)), x, levels, wv)
 counts(x::IntegerArray, k::Integer) = counts(x, 1:k)
 counts(x::IntegerArray, k::Integer, wv::WeightVec) = counts(x, 1:k, wv)
 counts(x::IntegerArray) = counts(x, span(x))
 counts(x::IntegerArray, wv::WeightVec) = counts(x, span(x), wv)
 
+
+"""
+    proportions(x, [levels,] [k,], [wv]) -> Vector
```
nalimilan: Same issues here.

src/counts.jl
```diff
@@ -167,38 +207,50 @@ proportions(x::IntegerArray, y::IntegerArray, wv::WeightVec) = proportions(x, y,
...
+        cm[xi] = get(cm, xi, z) + wi
+    end
+    return cm
 end
 
+
+"""
+    countmap(x) -> Dict
+
+Return the counts of each distinct element of `x` as a dictionary.
```
nalimilan: Could be more specific: "Return a dictionary mapping each unique value in `x` to its number of occurrences".

src/cov.jl
```diff
@@ -29,6 +29,21 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
     _symmetrize!(Base.unscaled_covzm(x, _scalevars(x, values(wv), vardim), vardim))
 
 if VERSION < v"0.5.0-dev+679"
+"""
+    scattermat(x, [wv,] [mean=nothing,] [vardim=1]) -> Matrix
```
nalimilan: Add `;` before keyword arguments. Also, they don't need brackets since the default value already shows that they are optional arguments.

src/cov.jl
```diff
@@ -29,6 +29,21 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
...
+
+Compute the scatter matrix, optionally specifying a weighting vector.
```
nalimilan: Could you explain what is a "scatter matrix" in a few words?

Regarding the recommended ways of presenting keyword arguments, see point 4 at http://docs.julialang.org/en/latest/manual/documentation/.

src/cov.jl
```diff
@@ -29,6 +29,21 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
...
+The function accepts two optional keyword arguments:
+
+  • `mean`: a known mean value. By default it is set to `nothing`,
```
nalimilan: Instead of saying "it is set to `nothing", just write the `=nothing` after the argument name.

src/cov.jl
```diff
@@ -29,6 +29,21 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
...
+    which indicates that the mean is unknown, and the function will
+    compute the mean. Specifying `mean=0` indicates that the data
```
nalimilan: I'm not sure the last sentence is very useful: by definition, centered data has zero mean, and subtracting zero has no effect. You're actually documenting an implementation detail which AFAIK has no user-visible effect.

src/cov.jl
```diff
@@ -29,6 +29,21 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
...
+    are centered and hence there's no need to subtract the mean.
+  • `vardim`: the dimension of the variables. When `vardim = 1`, the
```
nalimilan: Maybe "the dimension along which variables are organized" or something like that? (Could probably find a better formulation than my proposal.)

src/cov.jl
```diff
@@ -45,6 +60,15 @@ if VERSION < v"0.5.0-dev+679"
     Base.cov(x::DenseMatrix, wv::WeightVec; mean=nothing, vardim::Int=1) =
         scale!(scattermat(x, wv; mean=mean, vardim=vardim), inv(sum(wv)))
 
+
+"""
+    mean_and_cov(x, [wv,] [vardim=1]) -> (Vector, Matrix)
```
nalimilan: Again, missing `;`, and no need for brackets around `vardim`.

src/cov.jl
```diff
@@ -45,6 +60,15 @@ if VERSION < v"0.5.0-dev+679"
...
+
+Return the means and covariance matrix as a tuple. A weighting
```
nalimilan: Singular "mean"?

src/cov.jl
```diff
@@ -45,6 +60,15 @@ if VERSION < v"0.5.0-dev+679"
...
+vector `wv` can be specified. The function accepts a single
```
nalimilan: "The function accepts a single keyword argument" isn't really useful. Go straight to the point. (BTW, the phrasing is a bit different in the previous docstring, maybe decide which one is best and stick with that.)

src/deviation.jl
```diff
@@ -26,7 +39,12 @@ function countne(a::AbstractArray, b::AbstractArray)
     return c
 end
 
-# squared L2 distance
+
+"""
+    sqL2dist(a, b) -> Float
```
nalimilan: `Float` -> `Float64`.

src/deviation.jl
```diff
@@ -26,7 +39,12 @@ function countne(a::AbstractArray, b::AbstractArray)
...
+
+Compute the squared L2 distance between two arrays.
```
nalimilan: Could you give the formula of that distance?

src/deviation.jl
```diff
@@ -37,10 +55,22 @@ function sqL2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
     return r
 end
 
+
 # L2 distance
+"""
+    L2dist(a, b) -> Float
```
nalimilan: Same here.

nalimilan: My comments also apply to all similar functions below. If you don't feel like copying the equations for all distances, just fix the `Float64` thing.

src/deviation.jl
```diff
@@ -81,26 +124,58 @@ function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
...
+    msd(a, b) -> Float
+
+Return the mean squared deviation between two arrays.
+"""
 msd{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = sqL2dist(a, b) / length(a)
 
+
 # RMSD: root mean squared deviation
+"""
+    rmsd(a, b, [normalize=<bool>]) -> Float
```
nalimilan: Same remark about keyword arguments (I won't repeat it from now on, I'm already too verbose).

src/deviation.jl
```diff
@@ -81,26 +124,58 @@ function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
...
+
+Return the root mean squared deviation between two optionally
```
nalimilan: I'd just say "between two arrays", and then mention the normalization feature; what kind of normalization is that?

src/misc.jl
```diff
@@ -12,13 +12,21 @@ function norepeat(a::AbstractArray)
 end
 
 # run-length encoding
+"""
+    rle(v) -> (Vector, Vector)
+
+Returns the run-length encoding of a vector as a tuple. The
+first element of the tuple a vector of values of the input and
```
nalimilan: Missing "is".

src/misc.jl
```diff
@@ -44,6 +52,13 @@ function rle{T}(v::Vector{T})
 end
 
 # inverse run-length encoding
+"""
+    inverse_rle(vals, lens) -> Vector
+
+Reconstruct a vector from its run length encoding. `vals` is a
```
nalimilan: "run-length".

src/misc.jl
```diff
@@ -75,6 +95,13 @@ function indexmap{T}(a::AbstractArray{T})
     return d
 end
 
+
+"""
+    levelsmap(a) -> Dict
+
+Construct a dictionary that associates the unique elements of
+`a` with the integers 1 to `length(a)`.
```
nalimilan: Looks incorrect to me. Isn't that "with their rank of appearance"? (I must say I don't understand the purpose of this function: this can easily be computed from `unique`... The package could use some cleanup.)

src/misc.jl
```diff
@@ -97,15 +124,30 @@ function findat!{T}(r::IntegerArray, a::AbstractArray{T}, b::AbstractArray{T})
     return r
 end
 
+
+"""
+    findat(a, b) -> Array
```
nalimilan: "-> Vector"

(This function should be deprecated in favor if `indexin` from Base, but that's for another PR.)

src/misc.jl
```diff
@@ -97,15 +124,30 @@ function findat!{T}(r::IntegerArray, a::AbstractArray{T}, b::AbstractArray{T})
...
 
-# x: input elements, 
+# x: input elements,
 # c: categories
 # k: the maximum integer in x
 
+"""
+    indicatormat(x, [c,] [k,] [sparse=false]) -> Matrix
+
+Construct the indicator matrix for an array `x`. The optional argument
```
nalimilan: Could you explain how this works? Are the variables to be found in rows or in columns?

Again, `c` and `k` are the same argument, only depending on the type, so you need two lines.

src/misc.jl
```diff
@@ -97,15 +124,30 @@ function findat!{T}(r::IntegerArray, a::AbstractArray{T}, b::AbstractArray{T})
...
+`k` is the maximum integer in `x` and `c` is an array of categories.
+The function accepts a single optional keyword argument `sparse` that
```
nalimilan: That sentence is redundant with the signature above.

src/rankcorr.jl
```diff
@@ -10,9 +10,18 @@
 #
 #######################################
 
+"""
+    corspearman(x[, y])
+
+Compute Spearman's rank-based correlation. If a single matrix `x` is
```
nalimilan: The standard name seems to be "Spearman's rank correlation coefficient".

src/rankcorr.jl
```diff
@@ -86,6 +95,13 @@ function corkendall!(x::RealVector, y::RealVector)
 end
 
 
+"""
+    corkendall(x[, y])
+
+Compute Kendall's rank-based correlation. If `x` is a matrix, the
```
nalimilan: "Kendall's rank correlation coefficient τ (tau)"

src/misc.jl
```diff
@@ -75,6 +95,13 @@ function indexmap{T}(a::AbstractArray{T})
     return d
 end
 
+
```
nalimilan: Are the new empty lines added on purpose here and below?

src/ranking.jl
```diff
@@ -30,15 +30,23 @@ function ordinalrank!(rks::RealArray, x::RealArray, p::IntegerArray)
     return rks
 end
 
+
+"""
+    ordinalrank(x) -> Array
+
+Return the ordinal ranking of a real-valued array. Ordinal ranking,
+aka "1234" ranking, uses the literal sorted order. All items in `x`
```
nalimilan: I don't understand "literal sorted order". I think you should make it clear that equal elements are given different (but successive) ranks based on their position in the input.

src/ranking.jl
```diff
@@ -59,15 +67,23 @@ function competerank!(rks::RealArray, x::RealArray, p::IntegerArray)
     return rks
 end
 
+
+"""
+    competerank(x) -> Array
+
+Return the standard competition ranking ("1224" ranking) of a real-valued
+array. Items that compare equal are given the same rank with a gap in the
+ranking the size of the number of tied items - 1.
```
nalimilan: Maybe "of the size"? Just an idea, I'm not a native speaker.

src/ranking.jl
nalimilan: Minor point: mentioning "1234" in the first sentence as for methods below sounds better to me.

src/cov.jl
nalimilan: BTW, the syntax for the tuple return values is `::Tuple{Vector, Matrix}`.

src/sampling.jl
```diff
@@ -74,12 +74,23 @@ end
 
 ### draw a pair of distinct integers in [1:n]
 
+"""
+    samplepair(n) -> (Int, Int)
+
+Draw a pair of distinct integers between 1 and `n`.
```
nalimilan: Maybe put "distinct" in italics (`*`)? I always miss that detail about that function.

src/sampling.jl
```diff
@@ -74,12 +74,23 @@ end
...
     i1 = randi(n)
     i2 = randi(n-1)
     return (i1, ifelse(i2 == i1, n, i2))
 end
 
+
+"""
+    samplepair(a)
+
+Draw a pair of elements from the array `a`.
```
nalimilan: Distinct in italics here too?

src/sampling.jl
```diff
@@ -276,6 +287,18 @@ end
...
+"""
+    sample!(a, x, [replace=true,] [ordered=false])
+
+Draw a random sample of elements from an array `a` and store the result
+in `x`. A polyalgorithm is used for sampling. The function accepts two
+keyword arguments:
+
+  • `replace`: Should sampling be performed with replacement? Defaults
+    to `true`.
+  • `ordered`: Should an ordered sample be taken? Defaults to `false`.
```
nalimilan: What does "ordered" imply? Just sorting the result (but more efficient)?

Also, as above, go straight to the point.

src/sampling.jl
```diff
@@ -276,6 +287,18 @@ end
 
 sample(a::AbstractArray) = a[randi(length(a))]
 
+
+"""
+    sample!(a, x, [replace=true,] [ordered=false])
+
+Draw a random sample of elements from an array `a` and store the result
```
nalimilan: So `length(x)` elements are drawn?

src/sampling.jl
```diff
@@ -316,12 +339,24 @@ function sample!(a::AbstractArray, x::AbstractArray; replace::Bool=true, ordered
     return x
 end
 
+
+"""
+    sample(a, [n,] [dims,] [replace=true,] [ordered=false]) -> Array
```
nalimilan: Need two lines for `n` and `dims` here too.

src/sampling.jl
```diff
@@ -527,22 +562,46 @@ function sample!(a::AbstractArray, wv::WeightVec, x::AbstractArray;
...
+
 
 # wsample interface
 
-wsample(w::RealVector) = sample(weights(w))
-wsample(a::AbstractArray, w::RealVector) = sample(a, weights(w))
+"""
+    wsample!(a, w, x, [replace=true,] [ordered=false])
 
+Select a weighted sample from an array `a` and store the result in `x`. The
```
nalimilan: "Select a sample from an array `a` with the probabilities for each element given by `w`"? Or something similar.

src/sampling.jl
```diff
@@ -527,22 +562,46 @@ function sample!(a::AbstractArray, wv::WeightVec, x::AbstractArray;
 end
 
 sample{T}(a::AbstractArray{T}, wv::WeightVec, n::Integer; replace::Bool=true, ordered::Bool=false) =
```
nalimilan: This function needs documentation AFAICT (the same as `wsample`).

src/sampling.jl
```diff
@@ -527,22 +562,46 @@ function sample!(a::AbstractArray, wv::WeightVec, x::AbstractArray;
...
+  • `ordered`: Should an ordered sample be taken? Defaults to `false`.
+"""
 wsample!(a::AbstractArray, w::RealVector, x::AbstractArray; replace::Bool=true, ordered::Bool=false) =
     sample!(a, weights(w), x; replace=replace, ordered=ordered)
 
+
+"""
+    wsample([a,] w, [n,] [dims,] [replace=true,] [ordered=false]) -> Array
+
+Select a random sample of the weights given in `w` if no array `a` is given.
```
nalimilan: I don't understand the meaning of the one-argument form. Maybe don't document it as it's confusing, and we'll deprecate it soon?

Same remark as above about `n`/`dims`.

src/scalarstats.jl
```diff
@@ -8,6 +8,11 @@
 #############################
 
 # Geometric mean
+"""
+    geomean(a) -> Float
```
nalimilan: Float64.

src/scalarstats.jl
```diff
@@ -18,6 +23,11 @@ function geomean(a::RealArray)
 end
 
 # Harmonic mean
+"""
+    harmmean(a) -> Array
```
nalimilan: Float64.

src/scalarstats.jl
```diff
@@ -28,6 +38,12 @@ function harmmean(a::RealArray)
 end
 
 # Trimmed mean
+"""
+    trimmean(x, p) -> Float
```
nalimilan: Float64.

src/scalarstats.jl
```diff
@@ -28,6 +38,12 @@ function harmmean(a::RealArray)
...
+
+Compute the mean of `x` after removing a proportion `p` of its highest- and
```
nalimilan: "Compute the trimmed mean of `x`, i.e...."

src/scalarstats.jl
```diff
@@ -45,6 +61,12 @@ function trimmean(x::RealArray, p::Real)
 end
 
 # compute mode, given the range of integer values
+"""
+    mode(a[, rng]) -> Int
+
+Return the first mode (most common number) of an array, optionally
+over a specified range.
```
nalimilan: "range `rng`".

src/scalarstats.jl
```diff
@@ -45,6 +61,12 @@ function trimmean(x::RealArray, p::Real)
 end
 
 # compute mode, given the range of integer values
+"""
+    mode(a[, rng]) -> Int
```
nalimilan: Integer

src/scalarstats.jl
```diff
@@ -66,6 +88,12 @@ function mode{T<:Integer}(a::AbstractArray{T}, rgn::UnitRange{T})
     return mv
 end
 
+"""
+    modes(a[, rng]) -> Vector
+
+Return all modes (most common numbers) of an array, optionally over a
```
nalimilan: So this only returns several values when counts are tied? Probably worth mentioning.

src/scalarstats.jl
```diff
@@ -153,9 +181,21 @@ end
 #
 #############################
 
+"""
+    percentile(v, p) -> Float
```
nalimilan: Float64.

src/scalarstats.jl
```diff
@@ -153,9 +181,21 @@ end
...
+
+Return the `p`th percentile of an array `v`.
```
nalimilan: Could continue saying "real-valued array" here and below as you did in other places.

src/weights.jl
```diff
@@ -370,6 +399,13 @@ function bound_quantiles{T <: Real}(qs::AbstractVector{T})
 end
 
 quantile{W <: Real}(v::RealVector, w::WeightVec{W}, p::Number) = quantile(v, w, [p])[1]
+
+
+"""
+    wquantile(v, w, p)
```
nalimilan: Looks like you didn't document the unweighted `quantile` function.

src/scalarstats.jl
```diff
@@ -153,9 +181,21 @@ end
...
+"""
 percentile{T<:Real}(v::AbstractArray{T}, p) = quantile(v, p * 0.01)
 
 quantile{T<:Real}(v::AbstractArray{T}) = quantile(v, [.0, .25, .5, .75, 1.0])
+
+"""
+    nquantile(v, n) -> Array
+
+Return the n-quantiles of an array, i.e. the values which partition `v` into
```
nalimilan: Could you say which of the quantile algorithms is used? I know there are many variants of them. (Should also be done for `quantile` and `wquantile`.)

src/scalarstats.jl
```diff
@@ -166,16 +206,40 @@ nquantile{T<:Real}(v::AbstractArray{T}, n::Integer) = quantile(v, (0:n)/n)
...
+"""
+    span(x) -> Range
+
+Return the span of an integer array, i.e. the range `minimum(x):maximum(x)`.
+"""
 span{T<:Integer}(x::AbstractArray{T}) = ((a, b) = extrema(x); a:b)
 
 # Variation coefficient: std / mean
+"""
+    variation(x[, m]) -> Float
```
nalimilan: Float64 (I stop mentioning this from now one, a mass replace should do the trick).

src/scalarstats.jl
```diff
@@ -166,16 +206,40 @@ nquantile{T<:Real}(v::AbstractArray{T}, n::Integer) = quantile(v, (0:n)/n)
...
+standard deviation to the mean.
+"""
 variation{T<:Real}(x::AbstractArray{T}, m::Real) = stdm(x, m) / m
 variation{T<:Real}(x::AbstractArray{T}) = variation(x, mean(x))
 
 # Standard error of the mean: std(a) / sqrt(len)
+"""
+    sem(a) -> Float
+
+Return the standard error of the mean of `a`.
```
nalimilan: Could you give the equation ?

src/scalarstats.jl
```diff
@@ -187,6 +251,11 @@ function mad!{T<:Real}(v::AbstractArray{T}, center::Real=median!(v); constant::R
 end
 
 # Interquartile range
+"""
+    iqr(v) -> Float
+
+Compute the interquartile range (IQR) of an array.
```
nalimilan: Definition?

src/scalarstats.jl
```diff
@@ -240,6 +309,14 @@ function _zscore_chksize(X::AbstractArray, μ::AbstractArray, σ::AbstractArray)
     end
 end
 
+
+"""
+    zscore!([Z,] X, μ, σ)
+
+Compute the z-scores of an array `X` with mean `μ` and standard deviation `σ`. If
```
nalimilan: Please define z-scores (same for function below).

src/scalarstats.jl
```diff
@@ -257,6 +334,14 @@ zscore!{T<:AbstractFloat}(X::AbstractArray{T}, μ::Real, σ::Real) = _zscore!(X,
 zscore!{T<:AbstractFloat,U<:Real,S<:Real}(X::AbstractArray{T}, μ::AbstractArray{U}, σ::AbstractArray{S}) =
     (_zscore_chksize(X, μ, σ); _zscore!(X, X, μ, σ))
 
+
+"""
+    zscore(X, [μ,] [σ]) -> Array
```
nalimilan: The comma should either be outside of brackets or inside the last pair of brackets. You should be able to do a mass replacement.

src/scalarstats.jl
```diff
@@ -257,6 +334,14 @@ zscore!{T<:AbstractFloat}(X::AbstractArray{T}, μ::Real, σ::Real) = _zscore!(X,
...
+
+Compute the z-scores of `X`, optionally specifying a precomputed mean `μ` and
+standard deviation `σ`. The mean and standard deviation can be real numbers or
+arrays of real numbers.
```
nalimilan: Explain what happens when they are arrays. Also mention this possibility for `zscores!`

src/scalarstats.jl
```diff
@@ -313,6 +411,13 @@ function crossentropy{T<:Real}(p::AbstractArray{T}, q::AbstractArray{T}, b::Real
     return crossentropy(p,q) / log(b)
 end
 
+
+"""
+    kldivergence(p, q[, b]) -> Float
+
+Compute the Kullback-Leibler divergence of `q` from `p`, optionally specifying
+a real number `b` which scales the divergence by `1/log(b)`.
```
nalimilan: "such that the divergence is scaled by `1/log(b)`".

src/scalarstats.jl
```diff
@@ -369,4 +482,10 @@ function Base.show(io::IO, ss::SummaryStats)
     @printf(io, "Maximum:      %.6f\n", ss.max)
 end
 
+
+"""
+    describe(a)
+
+Pretty-print the summary statistics provided by `summarystats`.
```
nalimilan: Maybe repeat the list of stats from `summarystats`?

src/signalcorr.jl
```diff
@@ -46,11 +46,18 @@ _autodot{T<:RealFP}(x::Vector{T}, lx::Int, l::Int) = dot(x, 1:lx-l, x, 1+l:lx)
 
 
 ## autocov
-
+"""
+    autocov!(r, x, lags[, demean=true])
+
+Compute the autocovariance of a vector or matrix `x` at `lags` and store the result
+in `r`. The function accepts an optional keyword argument `demean` that denotes whether
```
nalimilan: Remove "The function accepts..." and "the default is".

src/signalcorr.jl
```diff
@@ -403,15 +469,24 @@ function pacf_yulewalker!{T<:RealFP}(r::RealMatrix, X::Matrix{T}, lags::IntegerV
...
+that designates the estimation method. Recognized values are `:regression` (default)
+and `:yulewalker`.
+"""
 function pacf!{T<:RealFP}(r::RealMatrix, X::Matrix{T}, lags::IntegerVector; method::Symbol=:regression)
     lx = size(X, 1)
     m = length(lags)
     minlag, maxlag = extrema(lags)
     (0 <= minlag && 2maxlag < lx) || error("Invalid lag value.")
-    size(r) == (m, size(X,2)) || raise_dimerror()
+    size(r) == (m, size(X,2)) || throw(DimensionMismatch())
```
nalimilan: @andreasnoack Do you think this is OK, or was `raise_dimerror()` used to avoid a GC root in performance-sensitive functions?

src/signalcorr.jl
```diff
@@ -420,6 +495,14 @@ function pacf!{T<:RealFP}(r::RealMatrix, X::Matrix{T}, lags::IntegerVector; meth
     return r
 end
 
+
+"""
+    pacf(X, lags[, method=:regression]) -> Array
+
+Compute the partial autocorrelation function (PACF) of `X` at `lags`. The function
+accepts an optional keyword argument `method` that designates the estimation method.
+Recognized values are `:regression` (default) and `:yulewalker`.
```
nalimilan: Explain what are `:regression` and `:yulewalker`.

src/weights.jl
```diff
@@ -9,6 +9,11 @@ end
 WeightVec{Vec<:RealVector,W<:Real}(vs::Vec,wsum::W) = WeightVec{W,Vec}(vs, wsum)
 WeightVec(vs::RealVector) = WeightVec(vs, sum(vs))
 
+"""
+    weights(vs) -> WeightVec
+
+Construct a weight vector from a given array.
```
nalimilan: Could you explain what's a `WeightVec` object? Could essentially use the same docstring for `WeightVec`. I think you can find some material in the manual.

src/weights.jl
```diff
@@ -203,6 +213,14 @@ _wsum!(R::AbstractArray, A::AbstractArray, w::AbstractVector, dim::Int, init::Bo
 wsumtype{T,W}(::Type{T}, ::Type{W}) = typeof(zero(T) * zero(W) + zero(T) * zero(W))
 wsumtype{T<:BlasReal}(::Type{T}, ::Type{T}) = T
 
+
+"""
+    wsum!(R, A, w, dim[, init=true])
+
+Compute the weighted sum of `A` with weights `w` over the dimension `dim` and store
+the result in `R`. The function accepts a single optional keyword argument `init`
```
nalimilan: Remove "the function accepts...", and explain what `init` means.

nalimilan:
Thanks for doing this great work! Despite the hundreds of comments, I really appreciate it! :-)

src/scalarstats.jl
```diff
@@ -166,16 +206,40 @@ nquantile{T<:Real}(v::AbstractArray{T}, n::Integer) = quantile(v, (0:n)/n)
...
+Return the standard error of the mean of `a`.
+"""
 sem{T<:Real}(a::AbstractArray{T}) = sqrt(var(a) / length(a))
 
 # Median absolute deviation
+"""
+    mad(v, [center,] [constant=1.4826]) -> Float
+
+Compute the median absolute deviation of `v`, optionally specifying a
+precomputed median `center`. The function accepts a keyword argument
```
ararslan: @nalimilan Full disclosure I have no idea what that scaling factor is for or why that's the default.

src/weights.jl
ararslan: whoops

ararslan:
@nalimilan Wow, thanks for all the feedback! I'm glad you like what I'm doing here. I'll go through and incorporate your comments. :)

src/counts.jl
ararslan: Oh, that's a thing now? I thought that was still under discussion. Isn't the issue/PR still open on Julia?

ararslan: Looks like it still uses arrows in the latest online docs

src/scalarstats.jl
ararslan: Looks like the current online docs use `Int`

src/misc.jl
ararslan: Yeah for readability

src/cov.jl
ararslan: The current docs use this `()` notation for returning tuples. See http://docs.julialang.org/en/latest/stdlib/collections/#Base.findmax!

src/scalarstats.jl
nalimilan: Yeah, the docs haven't been updated yet, as the discussion has stalled at https://github.com/JuliaLang/julia/pull/16477. But I think it's pretty clear that we will follow the real Julia syntax where it exists. `->`/`→` can still be used to give names to return values, but not for types.

src/counts.jl
nalimilan: Yeah, the docs haven't been updated yet, as the discussion has stalled at https://github.com/JuliaLang/julia/pull/16477. But I think it's pretty clear that we will follow the real Julia syntax where it exists. `->`/`→` can still be used to give names to return values, but not for types.

src/scalarstats.jl
nalimilan: No idea about the docs, but the code clearly returns whatever integer type is used in the input.

src/cov.jl
nalimilan: As I noted above, the PR hasn't been merged yet. Anyway, the Julia docs are large and take a lot of work to fix, so that's not necessarily the best reference.

src/signalcorr.jl
ararslan: What even is `raise_dimerror()`? I couldn't find a definition of it anywhere.

src/scalarstats.jl
ararslan: @nalimilan Come to think of it, do we want the return type listed for everything? Looks like the majority of Julia documentation across various packages doesn't list it anyway; it uses `->` occasionally but more often than not omits it entirely. What would you think of that?

nalimilan: Indeed, in many cases we could as well omit it. I guess it's not really needed when a `Float64` is returned. There might be other cases too, like when an array result is clearly expected, or when an integer is the only meaningful return value. Up to you.

src/scalarstats.jl
ararslan: Not sure what you mean. There's no function called `zscores`, there's only `zscore` (singular) and its yelling version (`zscore!`), unless I'm *extremely* oblivious.

src/weights.jl
```diff
@@ -203,6 +219,13 @@ _wsum!(R::AbstractArray, A::AbstractArray, w::AbstractVector, dim::Int, init::Bo
 wsumtype{T,W}(::Type{T}, ::Type{W}) = typeof(zero(T) * zero(W) + zero(T) * zero(W))
 wsumtype{T<:BlasReal}(::Type{T}, ::Type{T}) = T
 
+
+"""
+    wsum!(R, A, w, dim; init=true)
```
ararslan: I can't for the life of me figure out what `init` actually does

src/scalarstats.jl
kmsquire: I think he's asking for an English definition of z-score(s), rather than assuming the reader/user knows what a z-score is.

nalimilan: Yes, that's what I meant.

src/weights.jl
nalimilan: This:
```julia
init && fill!(R, zero(RT))
```

Apparently `init=false` won't start from zero when summing elements. So either you have already done this, or you want to add to the existing values. I guess we can just say: "if `init=false`, the sum is added to `R` instead of starting from zero".

nalimilan: BTW, you didn't document `wsum(A, w, dim)` at the same time as `wsum(A, w)`.

src/scalarstats.jl
ararslan: Oh sorry! Misunderstood. I'll add that in my next round of updates. :smile:

src/weights.jl
ararslan: Regarding `init`: THANK YOU!

Regarding `wsum`: Oops. Will add in next round.

ararslan:
I'm getting there slowly but surely. Thanks again for all of your help and feedback, @nalimilan! Couldn't have done it without you.

src/counts.jl
```diff
@@ -9,53 +9,90 @@
 typealias IntUnitRange{T<:Integer} UnitRange{T}
 
 #### functions for counting a single list of integers (1D)
+"""
+    addcounts!(r, x, levels[, wv])
 
+Fill the array `r` with the number of occurrences in `x` of
```
nalimilan: Hmm, after reading the `init` code, I realize this function does not "fill `r`" but "adds to `r`".

src/counts.jl
```diff
@@ -9,53 +9,90 @@
...
+        end
+    end
+    return r
 end
 
-counts(x::IntegerArray, levels::IntUnitRange) = addcounts!(zeros(Int, length(levels)), x, levels)
-counts(x::IntegerArray, levels::IntUnitRange, wv::WeightVec) = addcounts!(zeros(eltype(wv), length(levels)), x, levels, wv)
+
+"""
+    counts(x, [levels,] [wv])
```
nalimilan: Inconsistent treatment of commas.

src/counts.jl
```diff
@@ -9,53 +9,90 @@
...
+
+Count the number of times that values in the range `levels` occur in
+`x`. The output is a vector of length `length(levels)`. If `levels`
+is not provided, `span(x)` is used in its place. If a weighting
```
nalimilan: You could simply write `levels=span(x)` in the signature and remove this sentence. So:
```julia
    counts(x, levels=span(x))
    counts(x, wv::WeightVec, levels=span(x))
```

This version also allows you to get if the of mention of `WeightVec` in the text.

src/counts.jl
```diff
@@ -9,53 +9,90 @@
...
+vector `wv` (of type `WeightVec`) is specified, the sum of the weights
+is used rather than the raw counts.
+
+    counts(x, k::Integer[, wv])
```
nalimilan: Move this docstring to the corresponding method.

src/counts.jl
```diff
@@ -9,53 +9,90 @@
...
+counts(x::IntegerArray, levels::IntUnitRange, wv::WeightVec) =
+    addcounts!(zeros(eltype(wv), length(levels)), x, levels, wv)
 counts(x::IntegerArray, k::Integer) = counts(x, 1:k)
 counts(x::IntegerArray, k::Integer, wv::WeightVec) = counts(x, 1:k, wv)
 counts(x::IntegerArray) = counts(x, span(x))
 counts(x::IntegerArray, wv::WeightVec) = counts(x, span(x), wv)
 
+
+"""
+    proportions(x, [levels,] [wv])
```
nalimilan: Same remarks here (I won't repeat, too many comment already ;-).

src/counts.jl
ararslan: Are you sure? Reading the code for `addcounts!` it looks to me like `r` is getting overwritten rather than added to. Maybe I'm not interpreting the code correctly.

src/counts.jl
ararslan: That to me makes it look like `levels` is a keyword argument though, which seems confusing.

src/cov.jl
```diff
@@ -29,6 +29,23 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
     _symmetrize!(Base.unscaled_covzm(x, _scalevars(x, values(wv), vardim), vardim))
 
 if VERSION < v"0.5.0-dev+679"
+"""
+    scattermat(X, [wv]; mean=nothing, vardim=1)
+
+Compute the scatter matrix, which is an unnormalized covariance matrix.
+A weighting vector `wv` (of type `WeightVec`) can be specified to weight
```
nalimilan: Same remark for `WeightsVec`.

src/cov.jl
```diff
@@ -29,6 +29,23 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
...
+the estimate.
+
+# Arguments
+* `mean`: a known mean value. By default it is set to `nothing`,
```
nalimilan: Repeat the default after the argument name, which makes the text shorter.

src/cov.jl
```diff
@@ -42,9 +59,25 @@ if VERSION < v"0.5.0-dev+679"
...
+Compute the weighted covariance matrix. By default, the covariance
+matrix is normalized by the sum of the weights. That is, `cov(X, wv)`
+is equivalent to `scattermat(X, wv) / sum(wv)`.
+"""
     Base.cov(x::DenseMatrix, wv::WeightVec; mean=nothing, vardim::Int=1) =
         scale!(scattermat(x, wv; mean=mean, vardim=vardim), inv(sum(wv)))
 
+
+"""
+    mean_and_cov(x, [wv;] vardim=1) -> (mean, cov)
```
nalimilan: `;` should be outside the brackets.

src/deviation.jl
```diff
@@ -26,7 +39,12 @@ function countne(a::AbstractArray, b::AbstractArray)
     return c
 end
 
-# squared L2 distance
+
+"""
+    sqL2dist(a, b)
+
+Compute the squared L2 distance between two arrays: `∑|a-b|^2`.
```
nalimilan: Since you use non-Julia notation, replace `^2` with `²`.

src/deviation.jl
```diff
@@ -37,10 +55,23 @@ function sqL2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
     return r
 end
 
+
 # L2 distance
+"""
+    L2dist(a, b)
+
+Compute the L2 distance between two arrays, i.e. the square root of
```
nalimilan: "the square root of the squared L2 distance" is a bit of a joke. :-) It will be shorter to give the formula.

src/counts.jl
ararslan: I've used commas in this way in other places as well. I was putting trailing commas inside brackets. In this particular instance I was trying to express that `levels` or `wv` or both could be included. Is there a better way to show that?

src/deviation.jl
```diff
@@ -51,7 +82,14 @@ function L1dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
     return r
 end
 
+
 # Linf distance
+"""
+    Linfdist(a, b)
+
+Compute the L∞ distance, also called the Chebyshev distance, between
+two arrays: `max|a-b|`.
```
nalimilan: Maybe `max(|a-b|)`? Or it that correct math notation?

nalimilan: By the way, I know double backticks can be used for LaTex notations, you might want to try that.

src/counts.jl
ararslan: There's no such thing as too many comments if it means making a PR worth merging :)

src/misc.jl
```diff
@@ -97,15 +124,40 @@ function findat!{T}(r::IntegerArray, a::AbstractArray{T}, b::AbstractArray{T})
...
 
+"""
+    indicatormat(x, k::Integer; sparse=false)
+
+Construct a boolean matrix `I` of size `(k, length(x))` such that
+`I[x[i], i] = true` and all other elements are set to `false`.
+If `sparse` is `true`, the output will be a sparse matrix, otherwise
+it will be dense (default).
+
+    indicatormat(x, c; sparse=false)
```
nalimilan: Please split the docs for each method.

src/deviation.jl
ararslan: This is the correct math notation as far as I know (at least it's how I learned it) but I'm wondering if maybe describing these functions in terms of equivalent Julia code would be better, e.g. `maximum(abs(a - b))`. What do you think?

src/misc.jl
```diff
@@ -97,15 +124,40 @@ function findat!{T}(r::IntegerArray, a::AbstractArray{T}, b::AbstractArray{T})
...
+
+Construct a boolean matrix `I` of size `(length(c), length(x))`.
+Let `ci` be the index of `x[i]` in `c`. Then `I[ci, i] = true` and
+all other elements are `false`.
+
+    indicatormat(x; sparse=false)
+
+Equivalent to `indicatormat(x, sort(unique(x)); sparse=...)`.
```
nalimilan: Replace this with `c=sort(unique(x))` in the signature for the previous method.

src/rankcorr.jl
```diff
@@ -10,9 +10,19 @@
 #
 #######################################
 
+"""
+    corspearman(x[, y])
+
+Compute Spearman's rank correlation coefficient. If a single matrix `x`
+is given, the correlation between columns of `x` is given as a matrix.
```
nalimilan: given -> returned

src/deviation.jl
ararslan: Will do, unless you think describing these with Julia code is better (see other comment). I guess this would be `sum(abs2(a - b))`.

src/rankcorr.jl
```diff
@@ -86,6 +96,13 @@ function corkendall!(x::RealVector, y::RealVector)
 end
 
 
+"""
+    corkendall(x[, y])
+
+Compute Kendall's rank correlation coefficient, τ. If `x` is a matrix, the
+correlation between columns is given as a matrix. Otherwise `x` and `y` must
```
nalimilan: Same here.

src/counts.jl
nalimilan: Well, AFAICT `r[xi - b] += 1` is the only place where `r` is set, to it's adding, not overwriting.

src/counts.jl
nalimilan: There are two common variants:
```julia
    counts(x, [levels], [wv])
    counts(x[, levels][, wv])
```

But the commas following an argument shouldn't be inside brackets.

src/counts.jl
nalimilan: Yeah, but I think this should be tackled by the help system, e.g. by printing keyword arguments in a different color. That syntax is too useful not to use it.

src/deviation.jl
nalimilan: As you prefer, as long as it's readable.

src/sampling.jl
```diff
@@ -74,12 +74,23 @@ end
 
 ### draw a pair of distinct integers in [1:n]
 
+"""
+    samplepair(n)
+
+Draw a pair of integers between 1 and `n` without replacement.
```
nalimilan: Maybe you've not revised this part yet, but I think I noted that "distinct" is the key word here (and below).

src/counts.jl
ararslan: Maybe you should open an issue (or even better a PR) on Julia asking for keyword arguments to be printed in a different color. That seems reasonable to me! :) But in the meantime would a syntax like
```julia
counts(x, levels[=span(x)])
```
be less confusing? Or maybe more confusing?

src/sampling.jl
```diff
@@ -316,12 +338,36 @@ function sample!(a::AbstractArray, x::AbstractArray; replace::Bool=true, ordered
...
 end
 
+
+"""
+    sample(a[, wv::WeightVec])
+
+Select a single random element of `a`. Weighted sampling is performed if
+a weighting vector `wv` is given.
+
+    sample(a, [wv::WeightVec,] n::Integer; replace=true, ordered=false)
```
nalimilan: Also split docstrings here.

src/sampling.jl
```diff
@@ -316,12 +338,36 @@ function sample!(a::AbstractArray, x::AbstractArray; replace::Bool=true, ordered
...
+
+Select a random, optionally weighted sample of size `n` from an array `a`
+using a polyalgorithm. `replace` dictates whether sampling is performed
+with replacement and `order` dictates whether an ordered sample, also called
+a sequential sample, should be taken.
+
+    sample(a, [wv::WeightVec,] dims; replace=true, ordered=false)
```
nalimilan: `dims::Dims` could be useful information.

src/sampling.jl
ararslan: Oh whoops! Looks like that's something I missed going through your previous comments. Sorry about that. I'll do that in the next round. Thanks for noticing!

src/sampling.jl
ararslan: That's a good point. (I actually just learned the other day what the `Dims` type is.)

src/sampling.jl
```diff
@@ -527,22 +573,52 @@ function sample!(a::AbstractArray, wv::WeightVec, x::AbstractArray;
...
+weights are given in `w`. `replace` dictates whether sampling is performed
+with replacement and `order` dictates whether an ordered sample, also called a
+sequential sample, should be taken.
+"""
 wsample!(a::AbstractArray, w::RealVector, x::AbstractArray; replace::Bool=true, ordered::Bool=false) =
     sample!(a, weights(w), x; replace=replace, ordered=ordered)
 
+
+"""
+    wsample([a,] w)
```
nalimilan: Same remarks for this block.

src/sampling.jl
```diff
@@ -527,22 +573,52 @@ function sample!(a::AbstractArray, wv::WeightVec, x::AbstractArray;
...
+
+Select a weighted random sample of size `n` from `a` with weights given in `w`
+if `a` is present, otherwise select a random sample of size `n` of the weights
+given in `w`. `replace` dictates whether sampling is performed with replacement
+and `order` dictates whether an ordered sample, also called a sequential sample,
+should be taken.
+
+    wsample([a,] w, dims; replace=true, ordered=false)
+
+Select a weighted random sample from `a` with weights given in `w` if `a` is
```
nalimilan: `n` shouldn't be mentioned here (follow structure from `sample`).

src/scalarstats.jl
```diff
@@ -45,6 +61,12 @@ function trimmean(x::RealArray, p::Real)
 end
 
 # compute mode, given the range of integer values
+"""
+    mode(a[, rng])
```
nalimilan: `rgn`

src/scalarstats.jl
```diff
@@ -240,6 +312,15 @@ function _zscore_chksize(X::AbstractArray, μ::AbstractArray, σ::AbstractArray)
     end
 end
 
+
+"""
+    zscore!([Z,] X, μ, σ)
+
+Compute the z-scores of an array `X` with mean `μ` and standard deviation `σ`.
+z-scores are the signed number of standard deviations above the mean that an
```
nalimilan: "above the mean" sounds a bit confusing here: why not "from the mean"? Could also say "z-scores are centered-reduced scores giving/equal to..." or something like that.

src/scalarstats.jl
```diff
@@ -295,6 +391,13 @@ function entropy{T<:Real}(p::AbstractArray{T}, b::Real)
     return entropy(p) / log(b)
 end
 
+
+"""
+    crossentropy(p, q[, b])
+
+Compute the cross entropy between `p` and `q`, optionally specifying a real
+number `b` such that the entropy is scaled by `1/log(b)`.
```
nalimilan: entropy -> cross-entropy (or "result")

src/signalcorr.jl
```diff
@@ -89,10 +103,17 @@ autocov{T<:Real}(x::VecOrMat{T}; demean::Bool=true) = autocov(x, default_autolag
 
 ## autocor
 
+"""
+    autocor!(r, x, lags; demean=true)
+
+Compute the autocorrelation function (ACF) of a vector or matrix `x` at `lags`
```
nalimilan: Is "function" actually correct? Just double-checking, I'm not familiar with this method.

src/signalcorr.jl
nalimilan: Ugh, you're right, it doesn't appear to be used anywhere. I guess the tests for error paths aren't complete enough to distinguish different exception types. Removing it is fine then.

src/weights.jl
```diff
@@ -6,9 +6,20 @@ immutable WeightVec{W,Vec<:RealVector}
     sum::W
 end
 
+"""
+    WeightVec(vs[, wsum])
+
+Construct a `WeightVec` with weight values `vs` and sum of weights `wsum`.
+If omitted, `wsum` is computed.
```
nalimilan: Just write `wsum=sum(vs)` in the signature and remove this sentence.

nalimilan:
Thanks, looks like we're getting close to merging!

src/deviation.jl
nalimilan: As I noted above, use the one you want.

src/counts.jl
nalimilan: Actually, I meant that I wouldn't leave a note when it happened elsewhere.

src/signalcorr.jl
ararslan: As far as I know, yes. I took a course on time series in grad school and we always referred to the autocorrelation in this way as the ACF (likewise for PACF for the partial autocorrelation). Incidentally, R's equivalent function to `autocov` is `acf`; I was actually a little surprised that Julia didn't follow this convention, particularly since `pacf` matches the corresponding function in R.

src/counts.jl
ararslan: Good call. That would also explain why the name `addcounts!` has "add" in it.

src/counts.jl
ararslan: Okay, I'll switch to one of those. Thanks!

src/signalcorr.jl
nalimilan: OK. About the naming, I don't know how other software call this, but `autocov` sounds more explicit than `acf`, and it allows to the distinction with `autocor`. Though `pacf` clearly stands out; not sure what to do about it: maybe `pautocor`? Could file an issue to keep track of this.

src/scalarstats.jl
```diff
@@ -340,7 +340,7 @@ zscore!{T<:AbstractFloat,U<:Real,S<:Real}(X::AbstractArray{T}, μ::AbstractArray
 
 
 """
-    zscore(X[, μ, σ])
+    zscore(X, [μ, σ])
```
nalimilan: `[μ], [σ])`

ararslan: I did it this way to signify that if you give one you have to give both. I saw this syntax looking through some of the docstrings for base functions. But if it's nonstandard I can go with your suggestion.

nalimilan: OK, great, I thought that was just a typo.

src/counts.jl
```diff
@@ -167,38 +217,52 @@ proportions(x::IntegerArray, y::IntegerArray, wv::WeightVec) = proportions(x, y,
...
+"""
 countmap{T}(x::AbstractArray{T}) = addcounts!(Dict{T,Int}(), x)
 countmap{T,W}(x::AbstractArray{T}, wv::WeightVec{W}) = addcounts!(Dict{T,W}(), x, wv)
 
+
+"""
+    proportionmap(x)
+
+Return a dictionary mapping each unique value in `x` to its
+proportion of `x`.
```
nalimilan: "in `x` maybe"?

src/cov.jl
```diff
@@ -29,6 +29,21 @@ scattermat_zm(x::DenseMatrix, wv::WeightVec, vardim::Int) =
     _symmetrize!(Base.unscaled_covzm(x, _scalevars(x, values(wv), vardim), vardim))
 
 if VERSION < v"0.5.0-dev+679"
```
nalimilan: Since there's a branch here, you'll have to duplicate the docs inside the other branch, or to specify only once before the branch with:
```julia
"""
docs
"""
function scattermat end


"""
docs
"""
cov
```

nalimilan: Ah, and follow the indent of the code to which you add the docstring.

src/moments.jl
```diff
@@ -4,7 +4,7 @@
 
 Base.varm(v::RealArray, wv::WeightVec, m::Real) = _moment2(v, wv, m)
 
-function Base.var(v::RealArray, wv::WeightVec; mean=nothing)
```
nalimilan: Looks like you haven't documented `var` and `std`.

src/rankcorr.jl
```diff
@@ -10,9 +10,19 @@
 #
 #######################################
 
+"""
+    corspearman(x, [y])
```
nalimilan: Actually I realize it might be more explicit to write the default as `y=x`. Same for other correlation functions.

src/sampling.jl
```diff
@@ -273,9 +284,25 @@ end
 
 
 ### Interface functions (poly-algorithms)
+"""
+    sample(a, [wv::WeightVec])
 
+Select a single random element of `a`. Weighted sampling is performed if
```
nalimilan: Instead of or in addition to "weighting sampling", maybe say something about the probability of each element being proportional to its weight? Same for functions below.

src/sampling.jl
```diff
@@ -331,6 +374,12 @@ end
 #
 ################################################################
 
+"""
+    sample(wv::WeightVec)
+
+Select a single random integer in `1:length(wv)` with probabilities
```
nalimilan: Ah, you already have a model for my last comment here.

src/sampling.jl
```diff
@@ -316,12 +343,28 @@ function sample!(a::AbstractArray, x::AbstractArray; replace::Bool=true, ordered
...
 function sample{T}(a::AbstractArray{T}, n::Integer; replace::Bool=true, ordered::Bool=false)
-    sample!(a, Array(T, n); replace=replace, ordered=ordered)
+    sample!(a, Array{T}(n); replace=replace, ordered=ordered)
 end
 
+
+"""
+    sample(a, [wv::WeightVec], dims::Dims; replace=true, ordered=false)
+
+Select a random, optionally weighted sample from an array `a` specifying
```
nalimilan: The details provided for `sample!` should be repeated here. I must say I'm still not sure what this function does (but I don't know what polyalgorithm means): does it just draw random elements like the `n` variant, and returns them in a different form? or does it draw random slices? something else?

src/scalarstats.jl
```diff
@@ -66,10 +88,17 @@ function mode{T<:Integer}(a::AbstractArray{T}, rgn::UnitRange{T})
...
 
-function modes{T<:Integer}(a::AbstractArray{T}, rgn::UnitRange{T})
-    r0 = rgn[1]
-    r1 = rgn[end]
-    n = length(rgn)
+"""
+    modes(a, [r])
+
+Return all modes (most common numbers) of an array, optionally over a
+specified range `r`. The output is always an array, even if `a` has
```
nalimilan: array -> vector. But you could even remove that sentence in favor of `::Vector` in the signature. This is one example where that information is useful.

src/scalarstats.jl
```diff
@@ -153,9 +182,21 @@ end
...
+    percentile(v, p)
+
+Return the `p`th percentile of a real-valued array `v`.
+"""
 percentile{T<:Real}(v::AbstractArray{T}, p) = quantile(v, p * 0.01)
 
 quantile{T<:Real}(v::AbstractArray{T}) = quantile(v, [.0, .25, .5, .75, 1.0])
+
+"""
+    nquantile(v, n)
```
nalimilan: `n=4` (method above)

src/scalarstats.jl
```diff
@@ -153,9 +182,21 @@ end
...
+
+Return the n-quantiles of a real-valued array, i.e. the values which
+partition `v` into `n` subsets of nearly equal size.
```
nalimilan: "By default, quartiles are computed."

src/scalarstats.jl
```diff
@@ -166,16 +207,41 @@ nquantile{T<:Real}(v::AbstractArray{T}, n::Integer) = quantile(v, (0:n)/n)
...
+"""
+    span(x)
+
+Return the span of an integer array, i.e. the range `minimum(x):maximum(x)`.
+"""
 span{T<:Integer}(x::AbstractArray{T}) = ((a, b) = extrema(x); a:b)
 
 # Variation coefficient: std / mean
+"""
+    variation(x, [m])
```
nalimilan: `m=mean(x)`

src/scalarstats.jl
```diff
@@ -166,16 +207,41 @@ nquantile{T<:Real}(v::AbstractArray{T}, n::Integer) = quantile(v, (0:n)/n)
...
+"""
+    sem(a)
+
+Return the standard error of the mean of `a`, i.e. `sqrt(var(a) / length(a))`.
+"""
 sem{T<:Real}(a::AbstractArray{T}) = sqrt(var(a) / length(a))
 
 # Median absolute deviation
+"""
+    mad(v, [center]; constant=1.4826)
```
nalimilan: `center=median(v)`

src/signalcorr.jl
```diff
@@ -77,6 +83,14 @@ function autocov!{T<:RealFP}(r::RealMatrix, x::Matrix{T}, lags::IntegerVector; d
     return r
 end
 
+
+"""
+    autocov(x, [lags]; demean=true)
+
+Compute the autocovariance of a vector or matrix `x`, optionally specifying
```
nalimilan: Should say a word about the default `lags`. I see it's computed as `0 : min(size(x,1)-1, round(Int,10*log10(size(x,1))))`, but I really don't know what's the logic behind this. Looks similar to R, though, and their docs don't explain why either. So maybe just copy that formula in a second paragraph (without the `round` to make it more readable?).

src/signalcorr.jl
```diff
@@ -420,6 +488,16 @@ function pacf!{T<:RealFP}(r::RealMatrix, X::Matrix{T}, lags::IntegerVector; meth
     return r
 end
 
+
+"""
+    pacf(X, lags; method=:regression)
+
+Compute the partial autocorrelation function (PACF) of `X` at `lags`. `method`
```
nalimilan: Should mention that `X` can be a real-valued vector or matrix. Same for some functions above.

src/weights.jl
```diff
@@ -25,6 +36,11 @@ Base.getindex(wv::WeightVec, i) = getindex(wv.values, i)
 
 ## weighted sum over vectors
 
+"""
+    wsum(v, w, [dim])
+
+Compute the weighted sum of `v` with weights `w`, optionally over the dimension `dim`.
```
nalimilan: "vector `w`".

src/weights.jl
```diff
@@ -226,6 +250,11 @@ Base.sum{T<:Number,W<:Real}(A::AbstractArray{T}, w::WeightVec{W}, dim::Int) = ws
 
 ###### Weighted means #####
 
+"""
+    wmean(v, w)
+
+Compute the weighted mean of `v` with weights `w`.
```
nalimilan: "array `v`" (same below)

src/sampling.jl
ararslan: A polyalgorithm is a set of different algorithms with a rule for choosing when to use each one. 

Experimenting with the function a bit I figured out what the `dims` argument is actually doing: It fills an array of the given dimensions with samples from the input.

src/scalarstats.jl
ararslan: I don't see what you mean. It doesn't seem to use 4 if `n` isn't provided.

src/scalarstats.jl
ararslan: As I said above, I don't think `nquantile` does that. If you just call it as `nquantile(x)`, it's a `MethodError`.

src/scalarstats.jl
nalimilan: I mean that the default value is `n=4`, even if that's implemented using a separate method (the one above). So better show this in the signature.

nalimilan: Nevermind, I confused `quantile` and `nquantile`.

src/sampling.jl
nalimilan: OK, cool. So the current docstring is correct.

nalimilan:
I think we're good at last? Just fix the last detail I spotted, and I'll merge.

nalimilan:
Perfect, thanks! If you haven't found the process too painful, you could be interested in https://github.com/JuliaStats/GLM.jl/pull/138, where it looks like I discouraged the author. ;-) You could also try to move StatsBase to Documenter.jl.

ararslan:
No, the process hasn't been painful at all! I really appreciate all of your help and feedback. I'm not familiar with Documenter but I'd be happy to look into that as well as documenting GLM.

Issue #37, opened 2014-01-18
Weighted Sampling without replacement


lindahua:
We now support non-weighted sampling (with & without replacement) + weighted sampling with replacement.

Weighted sampling without replacement is not supported yet. This is not as easy to implement. 

A naive implementation might look like

``` julia
n = length(a)     # the pool size
k = length(dst)   # the number of samples to draw

for j = 1 : k
    i = wsample(1:n, w)  # sample a value in 1:n according to weights w
    x[j] = i
    w[i] = 0     # preclude i from being sampled in future
end
```

Unfortunately, this is _mathematically incorrect_.

Suppose `π` be the weights that sum to 1, and you want to draw `x1` and `x2` such that `x1 != x2`. Then the marginal distribution of `x1` is **not** `π`.  

This can be easily illustrated by the following simple example. Suppose `π = [0.4, 0.6]`, and you want to draw `x1` and `x2` that are different. Clearly, `(x1 = 1, x2 = 2)` and `(x1 = 2, x1 = 1)` should have the same joint probability, which means that the marginal probability of `p(x1 = 1)` equals `0.5` (instead of `0.4`).

A correct way to do this is _rejection sampling_, which may look like:

``` julia
x = wsample(1:n, w, k)   # draw a sample sequence (with replacements)
while there-are-repeated-sample(x)
    x = wsample(1:n, w, k)   # sample again
end
# until we get a sample sequence without repeated elements
```

This is very inefficient, as when `k` increases, it would become very unlikely to draw a sequence of which all elements are distinct.

Any suggestions?

krlmlr:
@johnmyleswhite: Hi there ;-) I'm doing more work now on the [wrswoR package](https://github.com/krlmlr/wrswoR), and results are getting nicer. I'm wondering if it makes sense to create a proper C++ library, and simply add R and Julia bindings?

johnmyleswhite:
@krlmlr: That would be really great. If you're going that route, perhaps you might consider using C or Fortran instead, given C++'s lack of a coherent ABI on many platforms?

krlmlr:
That would be necessary for inclusion into base R, too. In C++ I can say `std::priority_queue` or `std::partial_sort` -- what would be a good C library to depend on?

johnmyleswhite:
I'm not sure: I tend to write everything from scratch when I write C code (which I do very rarely).

PR #176, opened 2016-06-20
Weighted sampling without replacement

devmotion:
Hej!

Lately I had to create Barabási-Albert random graphs for a mathematical project. For this purpose I used the [Lightgraphs.jl](https://github.com/JuliaGraphs/LightGraphs.jl) package and generalized its implementation for creating such graphs (see https://github.com/JuliaGraphs/LightGraphs.jl/issues/372 and https://github.com/JuliaGraphs/LightGraphs.jl/pull/373). Since the algorithm requires weighted sampling without replacement I discovered that its "naive" implementation in StatsBase is quite slow, see e.g [this discussion](https://github.com/JuliaGraphs/LightGraphs.jl/pull/344); therefore I stuck to the already existing sampling in the LightGraphs.jl package, even though I wasn't satisfied with it.

Meanwhile I found [this thread](https://github.com/JuliaStats/StatsBase.jl/issues/37) about weighted sampling without replacement in StatsBase and the mentioned [R package wrswoR](https://github.com/krlmlr/wrswoR). I implemented all three algorithms from [Efraimidis' paper](http://www.sciencedirect.com/science/article/pii/S002001900500298X) in pure Julia (and not in C, as suggested in the discussion about weighted sampling without replacement in StatsBase). As expected they seem to be a lot faster than the current implementation. `efraimidis_aexpj_wsample_norep!` seems to be the fastest implementation, hence I used it as the default algorithm for weighted sampling without replacement. However, it would be nice to compare and benchmark the different implementations in detail (similar to [this report](https://cran.r-project.org/web/packages/wrswoR/vignettes/wrswoR.pdf) for `wrswoR`) which I haven't done so far.

A simple comparison on my computer:

``` julia
julia> using StatsBase

julia> a = collect(1:1000000);

julia> wv = weights(ones(1000000));

julia> x = Array{Int}(10000);

julia> @time StatsBase.naive_wsample_norep!(a, wv, x);
  8.423857 seconds (6 allocations: 7.630 MB, 0.11% gc time)

julia> @time StatsBase.efraimidis_a_wsample_norep!(a, wv, x);
  0.052866 seconds (14 allocations: 15.259 MB, 8.96% gc time)

julia> @time StatsBase.efraimidis_ares_wsample_norep!(a, wv, x);
  0.035620 seconds (6 allocations: 156.469 KB)

julia> @time StatsBase.efraimidis_aexpj_wsample_norep!(a, wv, x);
  0.027651 seconds (6 allocations: 156.469 KB)
```

And with `wrswoR`:

``` R
> library(wrswoR)
> ptm <- proc.time(); invisible(sample_int_expj(1000000, 10000, rep(1, 1000000))); proc.time()-ptm
   user  system elapsed 
  0.020   0.007   0.027 
```


devmotion:
Do you need further information or tests? Or do you prefer an implementation in C?

The nightly failures are unrelated.

ararslan:
Did you by chance base any of your code off of the R implementation you linked to?

devmotion:
Since both the R implementation and my implementation are based on Efraimidis' paper the underlying algorithms are the same.

My intention was to implement the algorithms of this paper and not to copy the code of the R package, but as in the R implementation I converted the algorithms A-res and A-expj to use `randexp`.

ararslan:
Okay, I just wanted to check because this package is licensed under MIT and the R code is under GPL.

devmotion:
Anything new on this issue? Do you think there are problems with the license?

ararslan:
As long as you didn't base your code off of the R code then the licensing should be fine. You should rebase from master though.

src/sampling.jl
```diff
@@ -488,6 +487,125 @@ function naive_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
...
 
+# Weighted sampling without replacement
+#
+# Algorithm A from:
+#     Efraimidis PS, Spirakis PG (2006). "Weighted random sampling with a reservoir."
+#     Information Processing  Letters, 97 (5), 181-185. ISSN 0020-0190.
+#     doi:10.1016/j.ipl.2005.11.003.
+#     URL http://www.sciencedirect.com/science/article/pii/S002001900500298X
+#
+# Insteads of keys u^(1/w) where u = random(0,1) keys w/v where v = randexp(1) are used.
```
nalimilan: "Instead"

src/sampling.jl
```diff
@@ -488,6 +487,125 @@ function naive_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
...
+function efraimidis_a_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
+    n = length(a)
+    length(wv) == n || throw(DimensionMismatch("Inconsistent lengths."))
```
nalimilan: While you're at it, could be good to improve the error message to something like: "a and wv must be of the same length (got $n and $(length(wv)))".

src/sampling.jl
```diff
@@ -488,6 +487,125 @@ function naive_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
...
+#     Efraimidis PS, Spirakis PG (2006). "Weighted random sampling with a reservoir."
+#     Information Processing  Letters, 97 (5), 181-185. ISSN 0020-0190.
+#     doi:10.1016/j.ipl.2005.11.003.
+#     URL http://www.sciencedirect.com/science/article/pii/S002001900500298X
+#
+# Insteads of keys u^(1/w) where u = random(0,1) keys w/v where v = randexp(1) are used.
+function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
+    n = length(a)
+    length(wv) == n || throw(DimensionMismatch("Inconsistent lengths."))
+    (k = length(x)) > 0 || return x
```
nalimilan: Better define `k` on a separate line first.

src/sampling.jl
```diff
@@ -488,6 +487,125 @@ function naive_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
...
+
+    # initialize priority queue
+    h = Array{Pair{Float64,Int}}(k)
+    @inbounds for i in 1:k
+        h[i] = (wv.values[i]/randexp() => i)
+    end
+    heapify!(h)
+
+    # set threshold
+    @inbounds T = h[1].first
```
nalimilan: `T` is often used to denote a type in Julia. Unless that's the letter used in the paper, better find another one (possibly more explicit).

src/sampling.jl
```diff
@@ -488,6 +487,125 @@ function naive_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
...
+    
+    @inbounds for i in k+1:n
+        r = wv.values[i]/randexp()
+        
+        # if key `r` is larger than the minimal key
+        if r > T
+            # update priority queue
+            h[1] = (r => i)
+            Collections.percolate_down!(h, 1)
```
nalimilan: No need for `Collections.` since you called `using` above.

src/sampling.jl
```diff
@@ -488,6 +487,125 @@ function naive_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
...
+    end
+    heapify!(h)
+
+    # set threshold
+    @inbounds T = h[1].first
+    X = T*randexp()
+    
+    @inbounds for i in k+1:n
+        w = wv.values[i]
+        (X -= w) <= 0 || continue
```
nalimilan: Same here, better split this into two lines, it will be as efficient.

src/sampling.jl
```diff
@@ -518,7 +636,9 @@ function sample!(a::AbstractArray, wv::WeightVec, x::AbstractArray;
             end
         end
     else
-        naive_wsample_norep!(a, wv, x)
+        k <= n || error("Cannot draw more samples without replacement.")
```
nalimilan: Here too, printing `k` and `n` could be useful to understand the reasons for the error.

nalimilan:
Sorry this took so long. I can't tell I looked at the algorithms, but overall the code looks good to me. I wonder whether there's a reason to include all three algorithms (as well as keeping the naive one) if we only use one of them, though. At the very least, we should provide an argument to choose what method to use. This would only be useful if there are cases in which some non-default methods will be faster than the default. Is that the case?

@johnmyleswhite @simonbyrne What do you think about this?

devmotion:
Thanks, I incorporated your comments and suggestions. The variable names were chosen according to the algorithms in the paper but I changed some of them to increase the readability.
Since right now only one algorithm is used it is not necessary to include the other methods. I included them because according to the [detailed analysis of different algorithms implemented in the R package](https://cran.r-project.org/web/packages/wrswoR/vignettes/wrswoR.pdf) "each of the algorithms presented [...] has its advantages". However, a similar analysis of the Julia implementations would be needed to provide an argument for one of the other methods.

nalimilan:
Thanks. So it seems that in principle each algorithm might be useful depending on the situation. Could you add an argument to allow choosing the algorithm then?

devmotion:
Sorry for the late reply, I was very busy the last weeks. I assume some tests using [BenchmarkTools.jl](https://github.com/JuliaCI/BenchmarkTools.jl) and [some random samples](https://cran.r-project.org/web/packages/wrswoR/vignettes/wrswoR.pdf) would provide some first hints whether there are cases in which other algorithms are preferable to A-expj in Julia. However, since I've never used BenchmarkTools.jl and I leave for vacation tomorrow, I can't provide an argument at short notice. It would be great if someone could help out with setting up some basic tests.

nalimilan:
Note that there is still no argument to choose the sampling algorithm to use, which means not all the features introduced in this PR are exposed to users (though they are tested).

nalimilan:
Also, these new algorithms should be added to the list in the docs at `docs/source/sampling.rst`.

PR #178, opened 2016-06-21
Qualify all calls to view with ArrayViews

ararslan:
Now that Base exports `view`, StatsBase's ArrayViews dependency causes testing errors on the nightlies due to ambiguous references to `view`. This PR qualifies all calls to `view` as `ArrayViews.view`. It may not be an ideal long-term solution but it should fix the issue for now.


nalimilan:
@simonbyrne Do you think we could use `Base.view` as a replacement for `ArrayViews.view` here?

ararslan:
I'm not that familiar with the ArrayViews package but if `Base.view` would be a sufficient replacement for `ArrayViews.view`, I'd be interested to see whether the dependency on ArrayViews could be removed altogether.

ararslan:
Closing in favor of either #179 or #182.

PR #179, opened 2016-06-23
Remove ArrayViews dependency

ararslan:
This uses the Base functionality for array views: `sub` on 0.4 and `view` on 0.5. This change might hurt performance due to the concerns mentioned in the ArrayViews README regarding `SubArray` versus `ArrayView` as well as the fact that `Base.reshape` makes a copy and `ArrayViews.reshape_view` does not. AFAIK there's no Base analog for `reshape_view`.

This PR is kind of an experiment to see what happens and provides an alternative to #178 for fixing the current 0.5 errors caused by the `Base.view`/`ArrayViews.view` conflict.

cc @nalimilan @simonbyrne


ararslan:
Awesome, both this and #178 work! Now it's just a matter of determining which is the better choice, or whether there's an even better alternative.

ararslan:
Bump. We should get the 0.5 tests passing, especially as 0.5 is nearing a release candidate. At present, our choices are this or #182. Seems like @simonbyrne, @andreasnoack, and myself are in favor of removing the dependency on ArrayViews. @nalimilan? Anyone else want to weigh in here?

src/StatsBase.jl
```diff
@@ -2,8 +2,8 @@ __precompile__(true)
 
 module StatsBase
     using Compat
-    import Compat.String
-    using ArrayViews
+    import Compat: String, view
```
nalimilan: I think `using` is enough here, as you don't overload these functions (I know `import` was here before, and I've been guilty of this too).

nalimilan:
I have no opinion on the matter, so if other says it's OK I'm all for it. Just squash and merge now that "you've got the power". BTW, feel free to tag a new release too.

ararslan:
Speaking of which, thanks so much for the invitation to the organization, @nalimilan! I'm honored! :heart:

nalimilan:
I must say that wasn't my own decision, it was a consensus among JuliaStats members. Welcome on board!

ararslan:
Ah, well thank you all JuliaStats members! 😄

src/StatsBase.jl
ararslan: Looks like the explicit `import` actually is necessary here; it fails on 0.4 because it can't find `view`.

ararslan:
Gotta love `git reset --hard`. The CI didn't even need to run again. 😄

src/StatsBase.jl
nalimilan: Maybe I wasn't explicit enough. My proposal was `using Compat: String, view`. Or maybe I'm just wrong.

ararslan: Oh, I didn't even realize that syntax existed. Well, I just merged...

PR #180, opened 2016-06-24
Deprecate AIC, AICc, BIC in favor of aic, aicc, bic

### LINKED ISSUE #177: Naming of AIC, BIC, and AICc
The function names `AIC`, `BIC`, and `AICc` are presumably in all caps for similarity with how one would write them outside of code. However, it's a little confusing because folks familiar with Julia naming conventions might assume that these are types rather than functions, since types are typically in upper camel case and functions in lowercase.

Thus, I propose that `AIC`, `BIC`, and `AICc` be deprecated in favor of `aic`, `bic`, and `aicc`. If people think this is a good idea I'd be happy to submit a PR.

ararslan:
Fixes #177.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -47,35 +47,35 @@ fit(obj::StatisticalModel, data...) = error("fit is not defined for $(typeof(obj
 fit!(obj::StatisticalModel, data...) = error("fit! is not defined for $(typeof(obj)).")
 
 """
-    AIC(obj::StatisticalModel)
+    aic(obj::StatisticalModel)
 
 Akaike's Information Criterion, defined as `-2 log L + 2k`, with `L` the likelihood
 of the model, and `k` its number of consumed degrees of freedom (as returned by `df`).
 """
-AIC(obj::StatisticalModel) = -2loglikelihood(obj) + 2df(obj)
+aic(obj::StatisticalModel) = -2loglikelihood(obj) + 2df(obj)
 
 """
-    AICc(obj::StatisticalModel)
+    aicc(obj::StatisticalModel)
 
 Corrected Akaike's Information Criterion for small sample sizes (Hurvich and Tsai 1989),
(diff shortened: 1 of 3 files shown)


ararslan:
The travis failures are unrelated.

src/statmodels.jl
```diff
@@ -47,35 +47,35 @@ fit(obj::StatisticalModel, data...) = error("fit is not defined for $(typeof(obj
...
     n = nobs(obj)
     -2loglikelihood(obj) + 2k + 2k*(k+1)/(n-k-1)
 end
 
 """
-    BIC(obj::StatisticalModel)
+    bic(obj::StatisticalModel)
 
-Akaike's Information Criterion, defined as `-2 log L + k log n`, with `L`
+Bayesian Information Criterion, defined as `-2 log L + k log n`, with `L`
```
nalimilan: Good catch! Looks like I copied/pasted docstrings a bit too fast. :-)

nalimilan:
Should we do the same for `R2` and `adjR2`?

ararslan:
Yeah, that would make sense. If it were up to me I would have named it `rsquared` instead of `R2`, but that's beside the point. 😉

nalimilan:
Why do you prefer `rsquared`? I find it less natural and it's much longer. `adjrsquared` is very hard to read. Actually, my preferred solution is `r²` and `adjr²`, but I have added the `2` variants for cases where `²` is hard to type.

ararslan:
I think `rsquared` is more immediately descriptive than `r2` but I don't feel too strongly about it. I like `r²` but I'm honestly not sure how to type the superscript 2. Is there a LaTeX completion shortcut?

nalimilan:
`²` is on my French keyboard, and I hadn't realized that it wasn't on standard QWERTY keyboards. Anyway, the LaTeX sequence is `\^2`.

nalimilan:
Thanks!

ararslan:
My pleasure! I love helping out with this stuff. :smile:

PR #182, opened 2016-06-29
Implement v0.5 sub->view change

davidavdav:
Recent julia-v0.5 implements `view()` which replaces both `sub()` and `slice()` as far as I understand. 

This PR refactors `sub` to `Compat.view` and `view` to `ArrayViews.view`.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -3,6 +3,7 @@ __precompile__(true)
 module StatsBase
     using Compat
     import Compat.String
+    import Compat.view
     using ArrayViews
     using StatsFuns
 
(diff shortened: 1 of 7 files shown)


nalimilan:
See also https://github.com/JuliaStats/StatsBase.jl/pull/179.

Cc: @ararslan

ararslan:
See also #178 for a nearly identical approach, save for `Compat.view`.

ararslan:
Personally I'm in favor of #179, which would entirely eliminate the dependency on ArrayViews. That to me seems like the best solution going forward, especially since array views (not to be confused with `ArrayView`s) are already part of Base.

davidavdav:
I don't care which solution is chosen, I have no personal pride in this PR---it took me more time than I would have wanted, but I am happy to view this as sunken effort if the problem gets resolved another way.  (After all, I could have studied the existing PRs better).

simonbyrne:
I think we should get rid of ArrayViews: there aren't that many cases where they're faster anymore.

ararslan:
I don't have any particular attachment to my specific PRs either; I just mean I prefer the approach of 179. Regardless of what solution is chosen, your effort here is greatly appreciated, @davidavdav!

ararslan:
Closed by #179.

Issue #184, opened 2016-07-06
Change default for "closed" argument for histograms from right to left


oschulz:
Currently, histogram ranges are closed-right be default. I would argue that in the vast majority of real-life use cases, this is not what the user wants: One deals with positive-only data far more frequently than with negative-only data, and 0.0 frequently _does_ occur in real-life data. A few examples:
- Things like `a * rand()` yield values from the interval [0, a[, not ]0, a]:
  
  ```
  rand(r::AbstractRNG) = rand(r, CloseOpen)
  ```
- Instruments that measure positive quantities often output 0.0, if the input is below their resolution.
- In data analysis, when histogramming derived positive quantities (e.g. peak heights), 0.0 also often occurs when dealing with finite-resolution data (e.g. from an ADC that will output 0,1,2, ...).
- ...

It's cumbersome (and easy to forget) to always add `closed = :left`, and I fear that it will result in people overlooking the fact that their data contains zeros (I've seen many students make that mistake in general ...).

I realize that Base.hist uses closed-right intervals as well - and I think we should change that, too.

nalimilan:
The default is the same in R, so there might be a reason for it. Though in R `include.lowest=TRUE` by default, which means it works fine for positive-only data. Maybe we should add that argument?

FWIW: `hist` has been deprecated in Julia 0.5, the one in StatsBase should be used instead (and is thus the only one to fix).

oschulz:
Well, I don't think we need to be slaves to R. ;-) 

PyPlot histograms, for example, certainly include 0 on the left:

![pyplot-hist-closed-left](https://cloud.githubusercontent.com/assets/546147/16616127/7571285e-437b-11e6-9759-25f3f3ffdd0a.png)

So do other important data analysis frameworks, like [ROOT](http://root.cern.ch).

Also, it seems unlogical that Julia's default for random number generation should be CloseOpen, but the exact opposite for it's central statistics package.

simonbyrne:
Not sure what the best option is, but for some background, see https://github.com/JuliaLang/julia/pull/6601#issuecomment-41148848.

nalimilan:
Note I said "there might be a reason", but there might not. Indeed, the fact that other programs use a different default is an interesting point.

Actually, in the thread Simon linked to, I had said that I'd prefer closed on the left. I think I'd still be in favor of that, since the historical reasons related to float ranges no longer apply.

oschulz:
[...]

PR #185, opened 2016-07-07
Changed default for histogram bins from closed-right to closed-left

oschulz:
See #184.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -1,7 +1,22 @@
 import Base: show, ==, push!, append!
 
+# Mechanism for temporary deprecation of default for "closed" (because default
+# value has changed). After deprecation is lifed, remove "_check_closed_arg"
+# and all calls to it, and replace every ":default_left" with ":left". Also
+# remove "closed=:left" in tests for all lines marked "FIXME: closed".
+function _check_closed_arg(closed::Symbol, funcsym)
+    if closed == :default_left
+        Base.depwarn("Default for keyword argument \"closed\" has changed from :right to :left.", funcsym)
+        :left
+    else
+        closed
+    end
+end
+
+
 ## nice-valued ranges for histograms
-function histrange{T}(v::AbstractArray{T}, n::Integer, closed::Symbol=:left)
+function histrange{T}(v::AbstractArray{T}, n::Integer, closed::Symbol=:default_left)
+    closed = _check_closed_arg(closed,:histrange)
     F = float(T)
     nv = length(v)
     if nv == 0 && n < 0
@@ -16,7 +31,8 @@ function histrange{T}(v::AbstractArray{T}, n::Integer, closed::Symbol=:left)
     histrange(F(lo), F(hi), n, closed)
 end
 
-function histrange{F}(lo::F, hi::F, n::Integer, closed::Symbol=:left)
+function histrange{F}(lo::F, hi::F, n::Integer, closed::Symbol=:default_left)
+    closed = _check_closed_arg(closed,:histrange)
     if hi == lo
         start = F(hi)
         step = one(F)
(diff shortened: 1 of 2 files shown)


nalimilan:
(Travis failure is unrelated.)

nalimilan:
I think we should print a deprecation warning (`Base.depwarn`) when `closed` isn't specified, explaining that the default changed. Else people will get incorrect results without notice.

oschulz:
> I think we should print a deprecation warning

But how, exactly - Keyword arguments have to have a default value, after all. Is there a "standard way" to do this? Or do we just use `closed = :default_left` or so, and depwarn if it's not replaced by `:left`  or `right` by the user? The risk here is that users will see that `:default_left` in the signature, and start using it in their code.

The other thing is what we'll have to change a lot of the test code, and change it back later on, when we disable the deprecation (I assume it's only intended to stay for a limited time?).

nalimilan:
`default_left` sounds OK. People won't use it because 1) you won't show it in the docs at all, so they won't see it, and 2) it will print an annoying warning. I know that's painful, but changing behaviour silently is really too dangerous.

oschulz:
> :default_left sounds OK

@nalimilan: Ok, I'll get on it. I'll put in some comments to make it easier to remove afterwards.

oschulz:
Ok, deprecation warnings are in. Maybe we shouldn't leave them in for too long ... ;-)

oschulz:
Forgot deprecation handling in two places, fixed now.

src/hist.jl
```diff
@@ -97,20 +109,21 @@ type Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
     weights::Array{T,N}
     closed::Symbol
     function Histogram(edges::NTuple{N,AbstractArray},weights::Array{T,N},closed::Symbol)
+
```
nalimilan: Please remove that unrelated line break addition.

test/hist.jl
```diff
@@ -2,28 +2,28 @@ using StatsBase
...
-@test sum(fit(Histogram,[1,2,3]).weights) == 3
-@test fit(Histogram,Int[]).weights == Int[]
-@test fit(Histogram,[1]).weights == [1]
-@test fit(Histogram,[1,2,3],[0,2,4]) == Histogram([0,2,4],[2,1])
-@test fit(Histogram,[1,2,3],[0,2,4]) != Histogram([0,2,4],[1,1])
-@test fit(Histogram,[1,2,3],0:2:4) == Histogram(0:2:4,[2,1])
-@test all(fit(Histogram,[1:100;]/100,0.0:0.01:1.0).weights .==1)
-@test fit(Histogram,[1,1,1,1,1]).weights[1] == 5
-@test sum(fit(Histogram,(rand(100),rand(100))).weights) == 100
+@test sum(fit(Histogram,[1,2,3], closed=:left).weights) == 3  # Explicit arg "closed" during default deprecation
```
nalimilan: Looks too verbose to me. Could you simply add a `FIXME: closed` comment at the beginning of the block? A good strategy would be to add a comment at `_check_closed_arg` saying to adapt all places marked `FIXME: closed` when removing that warning, and simply marking the relevant places that way.

oschulz: Will do.

src/hist.jl
oschulz: Ooops, sorry, that somehow snuck in there ...

test/hist.jl
oschulz: Ok, done. I rebased the PR branch to keep history neat.

oschulz:
@nalimilan, are you happy with it this way?

andreasnoack:
@oschulz @nalimilan is offline the next nine days so it might take a while before he can review and I think we should wait for his comments.

oschulz:
@andreasnoack: Thanks for the heads-up. Well, a few days won't matter, people will still have to write `closed = :left` during the deprecation period anyhow. :-)

oschulz:
I lost track of this - I thought it was merged already. I just rebased on the current master - could someone review (and hopefully merge)?

oschulz:
@ararslan - while you're in review mode, could you take a look at this, too? This PR was pretty much agreed on, but then all of us just kinda lost track of it.

test/hist.jl
```diff
@@ -2,28 +2,31 @@ using StatsBase
...
+@test fit(Histogram,[1,1,1,1,1], closed=:left).weights[1] == 5  # FIXME: closed
+@test sum(fit(Histogram,(rand(100),rand(100)), closed=:left).weights) == 100  # FIXME: closed
 @test fit(Histogram,1:100,nbins=5,closed=:right).weights == [20,20,20,20,20]
 @test fit(Histogram,1:100,nbins=5,closed=:left).weights == [19,20,20,20,20,1]
 @test fit(Histogram,0:99,nbins=5,closed=:right).weights == [1,20,20,20,20,19]
 @test fit(Histogram,0:99,nbins=5,closed=:left).weights == [20,20,20,20,20]
 
-@test fit(Histogram,(1:100,1:100),nbins=5).weights == diagm([20,20,20,20,20])
-@test fit(Histogram,(1:100,1:100),nbins=(5,5)).weights == diagm([20,20,20,20,20])
+# FIXME: closed (all lines in this block):
```
ararslan: What's meant by `FIXME: closed` here and elsewhere?

oschulz: I don't remember who came up with it - these are meant to be a reminder to remove `closed = :left` and similar from the code when the deprecation period (during which closed *has* to be specified, so that people will notice the changed default) is over.

ararslan: Ah, that makes sense. It would be good to make the comment more explicit then.

oschulz:
Closed #184.

test/hist.jl
oschulz: Sure - what would you suggest? Maybe just a comment on top of the file, stating

    # FIXME: closed - comments imply that `closed = :left` arguments should be removed after
    # the deprecation period on closed (following default change from :right to :left) has ended.

Or something like that? Whatever you think will be clear to the maintainers.

ararslan: Sure, that works for me.

oschulz: Ah, now I that I look at it again - there's actually a comment on top of hist.jl that explains it in detail. :-)

oschulz: New commit - Ok like that?

ararslan: So there is! Sorry, I completely missed that. What you have here is fine then.

ararslan: > Ok like that?

Yes indeed. Thanks!

oschulz:
Hey, can I just ask - what's the status on this, concerning merging?

ararslan:
Thanks for the reminder!

PR #187, opened 2016-07-12
RFC: Implement a binomial sampler in lieu of Rmath

ararslan:
As of JuliaLang/julia#17293, `libRmath-julia` is no more in 0.5. We're using it here (and assuming it's bundled with Julia) but only for binomial sampling. So I've implemented a naïve, O(n) solution in Julia. Request for comments!


ararslan:
The tests pass on 0.4 and 0.5 on OS X, but only on 0.4 on Linux. I have no idea why it would be different on Linux, and I think the failure is unrelated...?

kmsquire:
The test failure should be checked--probably not a problem with this PR, but still might be a bug. 

For small N, this is fine. For larger N, there's a `log N` algorithm. I have an old textbook with it, although it might be online. The algorithm here depends on generating beta samples. 

(On my phone and it's late, so I'll check tomorrow if no one else does...)

ararslan:
Thanks for taking a look, @kmsquire! Still looking into the the Linux issue. Looks like the problem is from DataFrames rather than here but I'm not sure yet.

I implemented something that should hopefully be more efficient. From the same reference as the initial simple algorithm: http://luc.devroye.org/chapter_ten.pdf

tkelman:
cc @simonbyrne who I believe had been working on this somewhere? also x-ref https://github.com/dmbates/Rmath.jl/pull/5 and https://github.com/JuliaLang/Rmath-julia/pull/10 which are work in progress towards making Rmath.jl deal with the binaries

tkelman:
StatsFuns also has a number of ccall's into Rmath that will be broken on nightly now

ararslan:
Should probably switch to using Rmath-julia rather than using my garbage code here.

kmsquire:
For the short run, that probably is best, although it would be nice to have these algorithms in Julia, if only for increased flexibility: Rmath takes and returns only Float64s.

tkelman:
I have the source build basically working for linux or macs with a compiler installed, that should be good to start with. Need to change it to use JuliaLang/Rmath-julia instead of my fork

simonbyrne:
I'm on holiday at the moment, so apologies for the intermittent contact.

There is actually already code for doing this in:
https://github.com/JuliaStats/Distributions.jl/blob/f6e2918d61667f0a54c6f8661504a057098e0c81/src/samplers/binomial.jl
It isn't used anywhere, and needs some performance tuning and testing (this has been on my todo list for a long time). Do we really need this in StatsBase? I guess it could be useful for the sampling without replacement stuff.

We will still need Rmath anyway for the special functions stuff for the time being, so it's probably better to stick with that until this is up to snuff.

ararslan:
I hope you're enjoying your vacation, @simonbyrne! Thanks for taking a look.

I was actually surprised that this was in StatsBase at all. It isn't exported, but it's being used internally by `xmultinom_sample!`, which is being used by `sample!` for sequential sampling with replacement for large sample sizes.

I'm not sure how significant the performance regression would be (I assume it would indeed be slower) but we could try always using direct sampling for the with-replacement case. That would eliminate the dependency on Rmath. It seems to me that distribution-specific methods belong in Distributions and other sophisticated sampling methods could live in a dedicated sampling package.

Does always performing direct sampling when sampling with replacement seems reasonable?

ararslan:
Closing in favor of #189.

PR #189, opened 2016-07-14
Use StatsFuns rather than libRmath

ararslan:
This gets the binomial sampling (the only Rmath thing this package uses) from the most recent version of StatsFuns, which means that it doesn't require an extra dependency on Rmath.jl and should work on 0.5, where libRmath no longer exists.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,12 @@ julia:
   - nightly
 notifications:
   email: false
-before_install:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-script:
-  - julia --check-bounds=yes -e 'Pkg.clone(pwd())'
-  - julia --check-bounds=yes -e 'Pkg.test("StatsBase"; coverage=true)'
+# Work around a Travis bug
+git:
+  depth: 999999
+# Uncomment the following lines to override the default test script
+#script:
+#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+#  - julia -e 'Pkg.clone(pwd()); Pkg.build("Primes"); Pkg.test("Primes"; coverage=true)'
 after_success:
   - julia -e 'cd(Pkg.dir("StatsBase")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.4
 Compat 0.8.4
-StatsFuns 0.1.0
+StatsFuns 0.3.0
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -18,6 +18,8 @@ module StatsBase
                       softplus, invsoftplus,
                       logsumexp, softmax, softmax!
 
+    import StatsFuns: RFunctions.binomrand
+
     export
 
     ## mathfuns (TODO: removed after a certain period)
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -19,8 +19,3 @@ end
 
 randi(K::Int) = rand(RandIntSampler(K))
 randi(a::Int, b::Int) = rand(RandIntSampler(a, b))
-
-# draw a number from a binomial distribution
-
-rand_binom(n::Real, p::Real) =
-    @compat Int(ccall((:rbinom, "libRmath-julia"), Float64, (Float64, Float64), n, p))
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -57,7 +57,7 @@ function xmultinom_sample!(a::AbstractArray, x::AbstractArray)
             end
             offset = k
         else
-            m = rand_binom(rk, 1.0 / (n - i + 1))
+            m = binomrand(rk, 1.0 / (n - i + 1))
             if m > 0
                 @inbounds ai = a[i]
                 for j = 1:m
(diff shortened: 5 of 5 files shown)


andreasnoack:
Thanks for fixing this but actually I think it is wrong that we have sampling code in `StatsBase`. It belongs to `Distributions` but was added in https://github.com/JuliaStats/StatsBase.jl/commit/e8944e4009e5acd83ffcc26c1b781fdcd16dc5b5.

ararslan:
Yeah I agree for the most part. StatsBase seems like a good home for basic simple random sampling but sampling from a distribution would be better in Distributions.

ararslan:
Hey @andreasnoack, you're a master of arrays, any idea what this failure on 0.5 is about?
```
BoundsError: attempt to access 2×1 Array{Int8,2} at index [Bool[false; true],Colon()]
```
When I do what it's unhappy about on 0.5, it works fine:
```julia
julia> x = Array{Int8}(2,1)
2×1 Array{Int8,2}:
 0
 0

julia> x[Bool[false; true], Colon()]
1×1 Array{Int8,2}:
 0
```

andreasnoack:
The Boolean part is a `Matrix` but that is not obvious from the printing since pasting the output produces a `Vector`.
```jl
julia> Array(Int8, 2, 1)[[false;true]'',:]
ERROR: BoundsError: attempt to access 2×1 Array{Int8,2} at index [Bool[false; true],Colon()]
 in throw_boundserror(::Array{Int8,2}, ::Tuple{Array{Bool,2},Colon}) at ./abstractarray.jl:223
 in checkbounds(::Array{Int8,2}, ::Array{Bool,2}, ::Colon) at ./abstractarray.jl:232
 in _getindex(::Base.LinearFast, ::Array{Int8,2}, ::Array{Bool,2}, ::Colon) at ./multidimensional.jl:207
 in getindex(::Array{Int8,2}, ::Array{Bool,2}, ::Colon) at ./abstractarray.jl:602
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Expr) at ./REPL.jl:3
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

ararslan:
Great, thanks so much for the explanation! I was able to track down the source of the error, and it's DataFrames' fault, not StatsBase's; [this](https://github.com/JuliaStats/DataFrames.jl/blob/master/src/statsmodels/formula.jl#L304) is returning a matrix.

Okay to merge here and fix DataFrames separately?

andreasnoack:
I'd prefer to have all the `rand`/`sample` stuff in `Distributions`. That was the original idea and it provides a good separation between the packages.

We might also end up with confused dependency structure. As mentioned in a different thread, I'd really like the `StatsFuns` stuff to return to `Distributions` because we are confusing users by having the `StatsFuns` package. If `StatsFuns` is merged into `Distributions` then `StatsBase` and `Distributions` would depend on each other.

ararslan:
Yeah, I've noticed that the dependencies across JuliaStats packages are a little weird. For example, StatsBase has a test dependency on DataFrames and GLM and we end up performing tests here that really belong in those respective packages.

If we move the sampling stuff out of here, why would StatsBase and Distributions need to depend on each other? The binomial sampling is the only distribution-specific thing we have in here, and moving all of the sampling into Distributions would mean that we wouldn't need Distributions stuff in StatsBase. Or am I misunderstanding?

ararslan:
Since a 0.5 release candidate is quickly approaching, would it be reasonable in the very short term to get things working on 0.5 using simple fixes like this PR so that the JuliaStats builds pass? Once things work on 0.5 then we can look at more long-term solutions like moving things between packages. Would that be alright?

andreasnoack:
We are better off making the bigger changes from a non-broken state so I've merged. Thanks again for fixing this.

I've confirmed the issue with DataFrames locally. It's weird an might be related to array changes in Base.

Regarding the dependency structure then I don't think it's an issue with test dependencies to downstream packages. If we move all the `rand` stuff out of here there won't be a problem. I thought you suggested that the samplers should stay here.

ararslan:
> I thought you suggested that the samplers should stay here.

IMO it makes sense for a basic statistics package to have a method for simple random sampling, akin to R's `sample` function, but I'm not adamant about any code staying in any particular place. Distribution-specific methods such as the binomial sampler definitely make more sense in Distributions though.

> It's weird and might be related to array changes in Base.

Yeah, seems likely.

> I don't think it's an issue with test dependencies to downstream packages.

I guess I mean that I find it odd that StatsBase is effectively testing the methods from other packages, for example the DataFrames issue we came across here probably should have been caught in a DataFrames test rather than a StatsBase test. Maybe that's reasonable though, I dunno. You're the expert, so I trust your judgement. 😄

.travis.yml
```diff
@@ -7,10 +7,12 @@ julia:
...
-script:
-  - julia --check-bounds=yes -e 'Pkg.clone(pwd())'
-  - julia --check-bounds=yes -e 'Pkg.test("StatsBase"; coverage=true)'
+# Work around a Travis bug
+git:
+  depth: 999999
+# Uncomment the following lines to override the default test script
+#script:
+#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+#  - julia -e 'Pkg.clone(pwd()); Pkg.build("Primes"); Pkg.test("Primes"; coverage=true)'
```
tkelman: Primes should be StatsBase in these comments

ararslan: HAH oops. Will fix. I guess it's obvious where I copied it from...

Issue #131, opened 2015-08-04
Adding Rényi entropies?


richardreeve:
I'm using [Rényi entropies](https://en.wikipedia.org/wiki/Rényi_entropy) for a Biodiversity package I'm developing - [Diversity.jl](https://github.com/richardreeve/Diversity.jl) - and I've just realised that of course you have standard entropies here and divergences in [Distances.jl](https://github.com/JuliaStats/Distances.jl). At the moment I have all of the entropies and divergences in my package, but they may be useful to other people, so I wondered whether you'd be interested in having the entropies onboard here and the divergences in Distances.jl if I tidied them up?

richardreeve:
This is now completed, since #190 and JuliaStats/Distances.jl#49 have been merged.

PR #190, opened 2016-07-15
Renyi entropy support

richardreeve:
I've added in support for Renyi entropies here to go with Renyi divergences / relative entropies which will go into StatsBase/Distances.jl. This follows up #131 and [StatsBase/Distances.jl:#45](https://github.com/JuliaStats/Distances.jl/issues/45).

The tests all work with v0.4, but Rmath problems in v0.5 stop them there.

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -390,6 +390,44 @@ function entropy{T<:Real}(p::AbstractArray{T}, b::Real)
     return entropy(p) / log(b)
 end
 
+"""
+    renyientropy(p, α)
+
+Compute the Rényi (generalised) entropy of order `α` of an array `p`.
+"""
+function renyientropy{T<:Real, U<:Real}(p::AbstractArray{T}, α::U)
+    α < 0 && throw(ArgumentError("Order of Rényi entropy not legal, $(α) < 0."))
+    
+    s = zero(T)
+    z = zero(T)
+    if α ≈ 0
+        for i = 1:length(p)
+            @inbounds pi = p[i]
+            if pi > z
+                s += 1
+            end
+        end
+        s = log(s)
+    elseif α ≈ 1
+        for i = 1:length(p)
+            @inbounds pi = p[i]
+            if pi > z
+                s -= pi * log(pi)
+            end
+        end
+    elseif (isinf(α))
+        s = -log(maximum(p))
+    else # a normal Rényi entropy
+        for i = 1:length(p)
+            @inbounds pi = p[i]
+            if pi > z
+                s += pi ^ α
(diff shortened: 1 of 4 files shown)


richardreeve:
This has been broken by a commit to [StatsFuns.jl](https://github.com/JuliaStats/StatsFuns.jl/commit/b8e379c6a67744b9fbcbbb11fecb2c1116767241)

ararslan:
Re StatsFuns, try rebasing your PR from the latest master here.

richardreeve:
Done, but because of #192 StatsBase fails all checks (and so does this PR as a result).

test/scalarstats.jl
```diff
@@ -71,7 +71,7 @@ z2 = [8. 2. 3. 1.; 24. 10. -1. -1.; 20. 12. 1. -2.]
 
 ##### Dispersion
 
-@test span([3, 4, 5, 6, 2]) == (2:6)
+@test StatsBase.span([3, 4, 5, 6, 2]) == (2:6)
```
ararslan: Where was `span` coming from before? Was there some kind of issue with calling it unqualified?

richardreeve: @ararslan I needed to do this because previously I was using Distributions.jl to generate better random probability distributions. Since I removed the dependency, it's no longer necessary (so I can remove it?), but it's worth noting that StatsBase.jl and Distributions.jl provide functions called span() than do different things.

src/scalarstats.jl
```diff
@@ -390,6 +390,47 @@ function entropy{T<:Real}(p::AbstractArray{T}, b::Real)
     return entropy(p) / log(b)
 end
 
+"""
+    renyientropy(p, α)
+
+Compute the Rényi (generalised) entropy of order `α` of an array `p`.
```
ararslan: > Compute the generalized Rényi entropy of order `α` of a real-valued array `p`.

Unless that isn't what you meant by "(generalized)". Also we should probably stay consistent with the American English spelling used in other places in the package, i.e. use "z" instead of "s" in words like "generalized."

richardreeve: TBH I've got so confused by s's and z's now that I can't even work out which spelling is which most of the time any longer! Anyway, corrected.

richardreeve:
@KristofferC merged the matching JuliaStats/Distances.jl#49 to include Rényi divergences, so it would make sense to merge the basic Rényi entropies now too...

ararslan:
This is great, @richardreeve. Thanks for contributing!

richardreeve:
@ararslan thanks for helping me get it together.

PR #193, opened 2016-07-18
Loosen signatures in signalcorr.jl from Vector to AbstractVector.

### LINKED ISSUE #188: autocor and autocov too restrictive types?
For instance, this does not work:
```
julia> autocor(view(a, 1:100))
ERROR: MethodError: no method matching autocor(::SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true})
Closest candidates are:
  autocor{T<:Real}(::Array{T<:Real,1}, ::AbstractArray{T<:Integer,1}; demean)
  autocor{T<:Real}(::Array{T<:Real,2}, ::AbstractArray{T<:Integer,1}; demean)
  autocor{T<:Real}(::Union{Array{T<:Real,1},Array{T<:Real,2}}; demean)
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```
Could these function be more loosely typed?

andreasnoack:
Fixes #188

### DIFF
--- a/src/signalcorr.jl
+++ b/src/signalcorr.jl
@@ -14,7 +14,7 @@
 default_laglen(lx::Int) = min(lx-1, round(Int,10*log10(lx)))
 check_lags(lx::Int, lags::AbstractVector) = (maximum(lags) < lx || error("lags must be less than the sample length."))
 
-function demean_col!{T<:RealFP}(z::Vector{T}, x::Matrix{T}, j::Int, demean::Bool)
+function demean_col!{T<:RealFP}(z::AbstractVector{T}, x::AbstractMatrix{T}, j::Int, demean::Bool)
     m = size(x, 1)
     @assert m == length(z)
     b = m * (j-1)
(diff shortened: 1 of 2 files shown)


ararslan:
I saw this earlier when I restarted a Travis build and it appears not to have gone away. (I was secretly hoping it was magically resolve itself somehow.)
```
ERROR: unsatisfiable package requirements detected: no feasible version could be found for package: ArrayViews
```
The package doesn't even use ArrayViews anymore, so why this error ಠ_ಠ

andreasnoack:
@tkelman Do you have any idea about what is going on here? My own package manager also behaved crazy after the latest stats changes. It kept complaning that my Julia version was incompatible with GLM. Eventually, I gave up on fixing it and used the nuke option.

tkelman:
Not without the output of `Pkg.status()`. I can't seem to reproduce what you're seeing here.

ararslan:
And of course now that we've complained about issues, the issues are mysteriously resolved.

PR #198, opened 2016-07-28
Restrict the mad() API and be more careful with types

### LINKED ISSUE #197: mad() is broken -- should promote integers to floats before calling mad!()
The mad function should promote integers to floats before calling mad!

```
julia> mad([1,2,3,4])
ERROR: InexactError()
 in mad! at /home/daniel/.julia/v0.4/StatsBase/src/scalarstats.jl:184
 in mad! at /home/daniel/.julia/v0.4/StatsBase/src/scalarstats.jl:183
 in mad at /home/daniel/.julia/v0.4/StatsBase/src/scalarstats.jl:179
```

ararslan:
Fixes #197.

cc @andreasnoack

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -239,7 +239,9 @@ sem{T<:Real}(a::AbstractArray{T}) = sqrt(var(a) / length(a))
 Compute the median absolute deviation of `v`.
 """
 function mad{T<:Real}(v::AbstractArray{T})
-    m = median(v)
+    isempty(v) && throw(ArgumentError("mad is not defined for empty arrays"))
+
+    m = middle(first(v))
 
     # Using the MAD as a consistent estimator of the standard deviation requires
     # a scaling factor that depends on the underlying distribution. For normally
@@ -249,7 +251,7 @@ function mad{T<:Real}(v::AbstractArray{T})
 
     S = promote_type(T, typeof(k * (one(T) - m)))
 
-    mad!(Array{S,ndims(v)}(v), m; constant=k)
+    mad!(LinAlg.copy_oftype(v, S), m; constant=k)
 end
 
 function mad!{T<:Real}(v::AbstractArray{T}, center::Real=median!(v); constant::Real=1.4826)
@@ -405,11 +407,11 @@ Compute the Rényi (generalized) entropy of order `α` of an array `p`.
 """
 function renyientropy{T<:Real, U<:Real}(p::AbstractArray{T}, α::U)
     α < 0 && throw(ArgumentError("Order of Rényi entropy not legal, $(α) < 0."))
-    
+
     s = zero(T)
     z = zero(T)
     scale = sum(p)
-    
+
     if α ≈ 0
         for i = 1:length(p)
             @inbounds pi = p[i]
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -251,7 +251,7 @@ function mad{T<:Real}(v::AbstractArray{T})
 
     S = promote_type(T, typeof(k * (one(T) - m)))
 
-    mad!(LinAlg.copy_oftype(v, S), m; constant=k)
+    mad!(LinAlg.copy_oftype(v, S); constant=k)
 end
 
 function mad!{T<:Real}(v::AbstractArray{T}, center::Real=median!(v); constant::Real=1.4826)
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -241,20 +241,23 @@ Compute the median absolute deviation of `v`.
 function mad{T<:Real}(v::AbstractArray{T})
     isempty(v) && throw(ArgumentError("mad is not defined for empty arrays"))
 
-    m = middle(first(v))
+    S = promote_type(T, typeof(middle(first(v))))
 
-    # Using the MAD as a consistent estimator of the standard deviation requires
-    # a scaling factor that depends on the underlying distribution. For normally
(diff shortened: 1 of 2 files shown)


src/scalarstats.jl
```diff
@@ -234,15 +234,23 @@ sem{T<:Real}(a::AbstractArray{T}) = sqrt(var(a) / length(a))
...
-Compute the median absolute deviation of `v`, optionally specifying a
-precomputed median `center`. `constant` provides a scaling factor that
-defaults to 1.4826 for consistent estimation of the standard deviation
-of a normal distribution.
+Compute the median absolute deviation of `v`.
 """
-mad{T<:Real}(v::AbstractArray{T}, args...;arg...) = mad!(copy(v), args...;arg...)
-mad{T<:Real}(v::Range{T}, args...;arg...) = mad!([v;], args...;arg...)
+function mad{T<:Real}(v::AbstractArray{T})
+    m = median(v)
```
andreasnoack: Don't compute these here. Instead call `mad!` without extra arguments. That will be the most efficient thing to do. If e.g. `T==Float64` then `Array{S,ndims(v)}(v)` is a noop and `v` would be modified so you probably want `LinAlg.copy_oftype`. We might want to consider exporting that function.

ararslan: > Don't compute this here. Instead call `mad!` without extra arguments.

Don't I need the median so that I can construct `S` to pass it to `mad!`?

> We might want to consider exporting that function.

Which, `StatsBase.mad!` or `LinAlg.copy_oftype`? Exporting the latter sounds like a good idea.

andreasnoack: I see. We should try to come up with a promote rule that doesn't require the computation of the median. Maybe use the [middle](https://github.com/JuliaLang/julia/blob/master/base/statistics.jl#L457-L462) function since that will be the one that will promote the elements in `median`.

I was thinking of `copy_oftype` because if we strat using it systematically outside Base then it's a indication that it ought to be exported.

src/scalarstats.jl
```diff
@@ -234,15 +234,25 @@ sem{T<:Real}(a::AbstractArray{T}) = sqrt(var(a) / length(a))
...
+    # Using the MAD as a consistent estimator of the standard deviation requires
+    # a scaling factor that depends on the underlying distribution. For normally
+    # distributed data, k is chosen as 1 / quantile(Normal(), 3/4) ≈ 1.4826,
+    # which is computed manually here.
+    k = 1 / (-sqrt(2 * one(T)) * erfcinv(3 * one(T) / 2))
+
+    S = promote_type(T, typeof(k * (one(T) - m)))
+
+    mad!(LinAlg.copy_oftype(v, S), m; constant=k)
+end
```
ararslan: @andreasnoack Is this what you had in mind?

src/scalarstats.jl
```diff
@@ -234,15 +234,25 @@ sem{T<:Real}(a::AbstractArray{T}) = sqrt(var(a) / length(a))
...
+    # Using the MAD as a consistent estimator of the standard deviation requires
+    # a scaling factor that depends on the underlying distribution. For normally
+    # distributed data, k is chosen as 1 / quantile(Normal(), 3/4) ≈ 1.4826,
+    # which is computed manually here.
+    k = 1 / (-sqrt(2 * one(T)) * erfcinv(3 * one(T) / 2))
+
+    S = promote_type(T, typeof(k * (one(T) - m)))
+
+    mad!(LinAlg.copy_oftype(v, S); constant=k)
+end
```
ararslan: @andreasnoack Would you mind taking another look to make sure I've done this the right way?

src/scalarstats.jl
```diff
@@ -234,15 +234,25 @@ sem{T<:Real}(a::AbstractArray{T}) = sqrt(var(a) / length(a))
...
+function mad{T<:Real}(v::AbstractArray{T})
+    isempty(v) && throw(ArgumentError("mad is not defined for empty arrays"))
+
+    m = middle(first(v))
+
+    # Using the MAD as a consistent estimator of the standard deviation requires
+    # a scaling factor that depends on the underlying distribution. For normally
+    # distributed data, k is chosen as 1 / quantile(Normal(), 3/4) ≈ 1.4826,
+    # which is computed manually here.
+    k = 1 / (-sqrt(2 * one(T)) * erfcinv(3 * one(T) / 2))
```
andreasnoack: I think this constant should be replacing the constant in `mad!`. Then the call here in `mad` can just be `mad!(LinAlg.copy_oftype(v, S))`

ararslan: So in the definition of `mad!` set the default value of `constant` to `1 / (-sqrt...)`?

ararslan: Don't I need `k` to be able to determine the type `S`?

ararslan: We need Noack Man

@andreasnoack

andreasnoack: No, `m` is sufficient since you only need to determine the vector type. The constant is multiplied after the calculation of the median.

ararslan: Ah okay, I see now. Thanks! I think I've got it together now. Mind taking one last look?

src/scalarstats.jl
milktrader: This change is not reflected in the documentation. I'd like to change the center to `mean` but kept running into MethodErrors. There is also not a path for those of us that would like to use `mean` as the center.

Issue #132, opened 2015-08-20
mean(trim()) instead of trimmean(), to permit var(trim())


bjarthur:
given that one might want to compute other statistics, besides just the mean, on trimmed data (e.g. the variance), would it make sense to provide, instead of trimmean(), a function, which could be called trim(), that just trimmed the input for the user to then do what they wanted?

relatedly, a close alternative to trimmed estimators are those that are winsorized--  where, instead of removing the extreme values, they are clamped.  i would find it useful if StatsBase.jl provided a winsor() function similar to the proposed trim() above.

i have code to do both, and could submit a PR.

nalimilan:
That's an interesting design issue. To make `mean(trim())` be efficient, it would be necessary to avoid making a copy of the data when calling `trim()`. In other words, `trim()` would have to return a special view of the original data, which would skip trimmed values. It would probably be easy to return `NaN` for those values, which would allow indexing that object using the indexes from the original data, making it completely array-like. Same strategy for `winsor()`.

Now, is that really useful? How many functions would be used in combination with `trim`? Wouldn't it be simpler to add a `trim` argument to them?

dcarrera:
I personally wouldn't use a `trim()` but I would use a `winsor()` because you can use the Winsorized variance to get the variance of the trimmed mean. Wilcox (2010) gives the following formula:

```
   Variance of trimmean(xs,p)  =  var(winsor(xs, p)) / ( (1 - p) * length(xs) )
```

nalimilan:
Thanks to @dcarrera and @bjarthur we now have generic `trim` and `winsor` functions returning modified copies of the input vector, and `trimvar` to compute the variance of the trimmed mean.

As a further improvement, it would be interesting to have non-copying functions for these features. They could return a special iterator which would skip/replace extreme elements. This should be significantly faster for operations like computing the trimmed mean, where you discard the trimmed vector immediately.

PR #199, opened 2016-07-28
Make the trimmean function consistent with R.

dcarrera:
The trimmed mean should remove an equal number of elements from the high- and low-value ends.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -136,6 +136,13 @@ module StatsBase
     hist,
     # histrange,
     midpoints,
+    
+    ## robust
+    trim,           # trimmed set
+    trim!,          # trimmed set
+    winsor,         # Winsorized set
+    winsor!,        # Winsorized set
+    trimvar,        # variance of the mean of a trimmed set
 
     ## misc
     rle,            # run-length encoding
@@ -194,6 +201,7 @@ module StatsBase
     include("empirical.jl")
     include("hist.jl")
     include("misc.jl")
+    include("robust.jl")
 
     include("sampling.jl")
     include("statmodels.jl")
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -2,6 +2,8 @@ import Base.@deprecate
 import Base.depwarn
 
 import Base.varm, Base.stdm
+@deprecate trimmean(x::RealArray, p::Real)
+
 @deprecate varm(v::RealArray, m::Real, wv::WeightVec) varm(v, wv, m)
 @deprecate varm(A::RealArray, M::RealArray, wv::WeightVec, dim::Int) varm(v, wv, m, dim)
 @deprecate stdm(v::RealArray, m::Real, wv::WeightVec) stdm(v, wv, m)
--- a/src/robust.jl
+++ b/src/robust.jl
@@ -1,0 +1,98 @@
+# Robust Statistics
+
+#############################
+#
+#   Trimming outliers
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x, p=0.2)
+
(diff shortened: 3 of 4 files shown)


nalimilan:
The old definition of `p` was consistent with MATLAB:
http://fr.mathworks.com/help/stats/trimmean.html

But the new definition is consistent with R, SciPy and Stata:
https://stat.ethz.ch/R-manual/R-devel/library/base/html/mean.html
https://docs.scipy.org/doc/scipy-0.17.0/reference/generated/scipy.stats.trim_mean.html#scipy.stats.trim_mean
http://fmwww.bc.edu/repec/bocode/t/trimmean.html

Note that this PR changes the meaning of the argument, so it cannot be done without a deprecation period/warning. I can't see a good way of doing this except by renaming the function and deprecating the old one.

ararslan:
Since this changes the output you'll have to modify the tests as well

dcarrera:
@nalimilan  If we are going to rename the function, this could be an opportunity to also resolve issue https://github.com/JuliaStats/StatsBase.jl/issues/132

So I could add a deprecation warning to `trimmean()` and create new functions called `trim()` and `winsor()`. In that issue you suggested adding new arguments to the `mean()` function. I'm happy with that too, but I don't know how to add an argument to an existing function. Here is an implementation of `trim()` and `winsor()`:

```julia
"""
trim(x, p)

Remove a proportion `p` of its highest elements, and `p` of its lowest elements (default: p = 0.2).
See winsor() if you want to compute the variance of mean(trim(x,p)).

Example: trim([1,2,3,4,5] , 0.2) == [2,3,4]
"""
function trim(x, p=0.2)
    n = length(x)
    n > 0 || error("x can not be empty.")
    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
    g = floor(Int, n * p)
    
    return sort(x)[g+1:n-g]
end

"""
winsor(x, p)

Replace  the proportion `p` of the lowest elements with the next-lowest element, and the
proportion `p` of the highest elements with the next highest elements (default: p = 0.2).
This function is typically used to compute the variance of mean(trim(x,p)):

Variance of mean(trim(x,p))  =  var(winsor(x, p)) / ( (1 - p) * length(x) )

Example: winsor([1,2,3,4,5] , 0.2) == [2,2,3,4,4]
"""
function winsor(x, p=0.2)
    n = length(x)
    n > 0 || error("x can not be empty.")
    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
    g = floor(Int, n * p)
    
    sx = collect(sort(x))
    sx[1:g] = sx[g+1]
    sx[n-g+1:end] = sx[n-g]
    
    return sx
end
```

How do you add a deprecation warning? Do I just add `warn("This function is deprecated. Use ... instead.")` at the top of `trimmean()` ?

ararslan:
To deprecate you add an entry in src/deprecates.jl.

In the docstrings, please indent the code signature at the top by four spaces.

nalimilan:
@dcarrera Good idea. To deprecate a function, use `depwarn`. See https://github.com/JuliaLang/julia/blob/master/base/deprecated.jl.

Regarding the implementation of `trim`, I think you can use `select` instead of `sort(x)[g+1:n-g]` to avoid sorting more elements than strictly needed. Same for `winsor` with `select!`.

Also, in docstrings, code excerpts need backticks around them. As regards code examples, see bullet 5 at http://docs.julialang.org/en/latest/manual/documentation/. Finally, as explained there, default values of keyword arguments should only appear in the signature.

dcarrera:
Thanks. I'm going to close this PR and make a new one with all the right changes.

dcarrera:
New PR available at https://github.com/JuliaStats/StatsBase.jl/pull/200

PR #200, opened 2016-07-30
Add functions to trim outliers.

dcarrera:
Add new functions to trim outliers (`trim`, `winsor`, and friends) that are guaranteed to trim equally from the high- and low-ends of the array. The new definition is consistent with R, SciPy and Stata and allows the calculation of the variance of the trimmed mean. Also deprecate `trimmean`.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -136,6 +136,13 @@ module StatsBase
     hist,
     # histrange,
     midpoints,
+    
+    ## robust
+    trim,           # trimmed set
+    trim!,          # trimmed set
+    winsor,         # Winsorized set
+    winsor!,        # Winsorized set
+    trimvar,        # variance of the mean of a trimmed set
 
     ## misc
     rle,            # run-length encoding
@@ -194,6 +201,7 @@ module StatsBase
     include("empirical.jl")
     include("hist.jl")
     include("misc.jl")
+    include("robust.jl")
 
     include("sampling.jl")
     include("statmodels.jl")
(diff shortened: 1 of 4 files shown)


dcarrera:
Apparently a check failed. How do I figure out which check failed? How do I run the checks myself before committing?

andreasnoack:
Click on details. Then you can see which line had the error. You can run the tests locally with `Pkg.test("StatsBase")`.

dcarrera:
Thanks. After I fix the error do I have to close this PR and create a new one or is there a way to "update" a PR?

andreasnoack:
No need to open a new PR. You can just push to the remote branch after you have committed your changes. Then the PR is automatically updated.

ararslan:
Just a note about referencing macros in commit messages: Unfortunately GitHub's mention syntax is the same as Julia's macro syntax, so whenever you have `@something` in a commit message, it will send a notification to a GitHub user with that username. (I feel sorry for the user compat, who I'm sure gets pinged constantly.) You could write it as `at-deprecate` or just leave off the `@`.

dcarrera:
Thanks.

I figured out that the problem is that I don't know how to use `@deprecate` and `depwarn` correctly and I can't figure out where the documentation is. Can you help me out? First I missed the second parameter in `@deprecate`. So I tried this,

```julia
@deprecate trimmean(x::RealArray, p::Real) trim(x,p)
```

Now it looks like `@deprecate` makes Julia call `trim(x,p)` directly. So I'm trying to figure out how to just add a warning. I could add a `depwarn` message inside `trimmean`; but that would be in `scalarstats.jl` instead of `deprecates.jl` so I don't know if that's how I'm supposed to do it.

dcarrera:
Ok. I think I got it working. I'll try again.

dcarrera:
Ok. Let's see if this new PR works... https://github.com/JuliaStats/StatsBase.jl/pull/201

PR #201, opened 2016-07-30
Functions for trimming outliers

dcarrera:
Add new functions to trim outliers (`trim`, `winsor`, and friends) that are guaranteed to trim equally from the high- and low-ends of the array. The new definition is consistent with R, SciPy and Stata and allows the calculation of the variance of the trimmed mean. Also deprecate `trimmean`. See also #132


andreasnoack:
Please don't open new PRs for the same feature. Just update the existing.

dcarrera:
@andreasnoack  I just saw your comment in the other PR. I didn't see it earlier. Ok. Now I know.

Can you help me understand why the coverage decreased? It seems to be complaining that the tests no longer cover `trimmean` which doesn't entirely surprise me because the `@deprecate` macro seems to be redirecting calls to `trimmean`. But I don't know what to do then. How am I supposed to deprecate a function? Am I doing it wrong?

ararslan:
The coverage drop is pretty small: -0.6%. I'd say that's nothing to worry about. Though it certainly never hurts to have more tests!

src/deprecates.jl
```diff
@@ -2,6 +2,9 @@ import Base.@deprecate
 import Base.depwarn
 
 import Base.varm, Base.stdm
+
+@deprecate trimmean(x::RealArray, p::Real) mean(trim(x,p*0.5))
```
nalimilan: Write `p/2` rather. Add space after comma.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+
+# Trimmed set
+"""
+    trim(x, p=0.2)
+
+Remove a proportion `p` of its highest elements, and `p` of its lowest elements and return the result. If you want to compute the variance of `mean(trim(x,p))`, use `trimvar`.
+
+# Example
+```julia
+julia> trim([1,2,3,4,5] , 0.2)
```
nalimilan: No space before comma. Also check below that there are spaces after commas.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+#
+#   Trimming outliers
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x, p=0.2)
+
+Remove a proportion `p` of its highest elements, and `p` of its lowest elements and return the result. If you want to compute the variance of `mean(trim(x,p))`, use `trimvar`.
```
nalimilan: Break lines at 92 chars. "its" isn't explicit: say "vector `x`" instead. Mention that the result is sorted. I would also Change the last sentence into `To compute the trimmed mean of 'x', use...; to compute the variance of this quantity, use...`.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+# Example
+```julia
+julia> trim([1,2,3,4,5] , 0.2)
+ 2
+ 3
+ 4
+```
+"""
+trim(x,p::Real=0.2) = trim!(copy(x),p)
+trim!(x,p::Real=0.2) = trim!(collect(x), p)
```
nalimilan: I don't think you need this method. People should call `collect` themselves if they want to pass an immutable array (including a range).

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+
+"""
+    trim!(x, p=0.2)
+
+A variant of `trim` that modifies vector `x` in place. See also `trimvar`.
+"""
+function trim!(x::RealArray, p::Real=0.2)
```
nalimilan: `RealArray` -> `AbstractVector`. Also add this restriction to `trim`.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+ 4
+```
+"""
+trim(x,p::Real=0.2) = trim!(copy(x),p)
+trim!(x,p::Real=0.2) = trim!(collect(x), p)
+
+"""
+    trim!(x, p=0.2)
+
+A variant of `trim` that modifies vector `x` in place. See also `trimvar`.
```
nalimilan: More precisely, which partially sorts `x` in place. But the result is a copy. Don't mention `trimvar` here since it's mentioned in the docstring for `trim`.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
+    g = floor(Int, n * p)
+    
+    return select!(x, (g+1):(n-g))
+end
+
+"""
+    winsor(x, p=0.2)
+
+Replace  the proportion `p` of the lowest elements with the next-lowest element, and the proportion `p` of the highest elements with the next highest elements. This function is typically used to compute the variance of `mean(trim(x,p))`.
```
nalimilan: Remove double spaces, cut line at 92 chars.

This function is typically used to compute the Winsorized mean AFAIK. People will use `trimvar` to compute the variance of the trimmed mean, so mention that function.

nalimilan: Also, inconsistency in dashed between "next-lowest" and "next highest". More fundamentally, shouldn't it be "previous-highest"? You should also mention "of vector `x`" somewhere.

nalimilan: Ah, and "replace" could be confusing as to whether it's in place or not. Maybe better say "return a copy of vector `x` with...".

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+julia> winsor([1,2,3,4,5] , 0.2)
+ 2
+ 2
+ 3
+ 4
+ 4
+```
+"""
+winsor(x,p::Real=0.2) = winsor!(copy(x),p)
+winsor!(x,p::Real=0.2) = winsor!(collect(x), p)
```
nalimilan: Same remark as for `trim` as regards `collect`. Same for signatures above and below.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+
+A variant of `trim` that modifies vector `x` in place. See also `trimvar`.
+"""
+function trim!(x::RealArray, p::Real=0.2)
+    n = length(x)
+    n > 0 || error("x can not be empty.")
+    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
+    g = floor(Int, n * p)
+    
+    return select!(x, (g+1):(n-g))
```
nalimilan: I think it would be faster to use the same strategy as in `winsor` by calling `select!` twice, for lowest and highest elements, and then indexing into the result.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+Compute the variance of `mean(trim(x,p))` using the Winsorized variance, as described in Wilcox (2010).
+"""
+function trimvar(x::RealArray,p::Real=0.2)
+    n = length(x)
+    n > 0 || error("x can not be empty.")
+    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
+    
+    return var(winsor(x,p)) / (n * (1 - 2p)^2)
+end
+
```
nalimilan: Remove additional empty lines.

src/robust.jl
```diff
@@ -0,0 +1,98 @@
...
+#
+#   Other
+#
+#############################
+
+# Variance of a trimmed set.
+"""
+    trimvar(x,p=0.2)
+
+Compute the variance of `mean(trim(x,p))` using the Winsorized variance, as described in Wilcox (2010).
```
nalimilan: "of the trimmed mean of `x` (returned by `mean(trim(x, p))`) using..."

test/robust.jl
```diff
@@ -0,0 +1,25 @@
...
+### Trimming outliers
+
+@test trim([1,2,3,4,5,6,7,8]) == [2,3,4,5,6,7]
+@test trim([1,2,3,4,5,6,7,8], 0.1) == [1,2,3,4,5,6,7,8]
+@test trim([1,2,3,4,5,6,7,8], 0.2) == [2,3,4,5,6,7]
+@test trim([1,2,3,4,5,6,7,8,9], 0.4) == [4,5,6]
+
+@test trim!([1,2,3,4,5,6,7,8,9], 0.4) == [4,5,6]
+
+
```
nalimilan: Remove the second line break.

test/robust.jl
```diff
@@ -0,0 +1,25 @@
...
+@test winsor([1,2,3,4,5,6,7,8]) == [2,2,3,4,5,6,7,7]
+@test winsor([1,2,3,4,5,6,7,8], 0.1)   == [1,2,3,4,5,6,7,8]
```
nalimilan: If you align `==`, do it everywhere.

test/robust.jl
```diff
@@ -0,0 +1,25 @@
...
+using Base.Test
+
+### Trimming outliers
+
+@test trim([1,2,3,4,5,6,7,8]) == [2,3,4,5,6,7]
+@test trim([1,2,3,4,5,6,7,8], 0.1) == [1,2,3,4,5,6,7,8]
+@test trim([1,2,3,4,5,6,7,8], 0.2) == [2,3,4,5,6,7]
+@test trim([1,2,3,4,5,6,7,8,9], 0.4) == [4,5,6]
+
+@test trim!([1,2,3,4,5,6,7,8,9], 0.4) == [4,5,6]
```
nalimilan: Test `trim!` for all variants as for `trim`.

nalimilan:
Could you also add references to `trim` and `winsor` in the docstring for `mean`?

dcarrera:
@nalimilan Thanks for the review. I've made all the changes except for updating the docstring for `mean`. How do I do that? AFAICT `mean` is part of the Julia standard library.

dcarrera:
Someone on the julia-users list suggested that I delete the code for the deprecated function. Should I do that?

nalimilan:
> @nalimilan Thanks for the review. I've made all the changes except for updating the docstring for mean. How do I do that? AFAICT mean is part of the Julia standard library.

Ah, indeed, leave it that way then. Though you could probably add an explanation to the JuliaStats manual.

> Someone on the julia-users list suggested that I delete the code for the deprecated function. Should I do that?

Yes, it won't be used anymore since `@deprecated` calls the new function automatically.

I will be offline for two weeks, so somebody else will have to continue the review or the PR will have to wait. Sorry for that.

dcarrera:
What the heck? How could the test coverage decrease on functions that I didn't even touch? The coverage should be higher now. I added tests.

@nalimilan Thanks for all the help. Maybe someone else can commit the PR if it's ready... @ararslan ?

ararslan:
Coverage is a myserious and fickle beast; I wouldn't worry too much about changes less than a percent. 😉

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+    trim(x, p=0.2)
+
+Remove a proportion `p` of the highest elements, and `p` of the lowest
+elements of vector `x` and return the result. To compute the trimmed
+mean of 'x', use `mean(trim(x, p))`; to compute the variance of this
+quantity, use `trimvar(x, p)`.
+
+
+If you want to compute
+the variance of `mean(trim(x,p))`, use `trimvar`.
```
ararslan: You say the same thing a few lines up, so you can just remove these two lines.

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x, p=0.2)
+
+Remove a proportion `p` of the highest elements, and `p` of the lowest
+elements of vector `x` and return the result. To compute the trimmed
+mean of 'x', use `mean(trim(x, p))`; to compute the variance of this
```
ararslan: `x` should be in backticks instead of single quotes

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+"""
+    trim!(x, p=0.2)
+
+A variant of `trim` that partially sorts `x` in place while removing
+the proportion `p` of the highest and lowest elements of `x`.
+"""
+function trim!(x::AbstractVector, p::Real=0.2)
+    n = length(x)
+    n > 0 || error("x can not be empty.")
+    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
```
ararslan: Should use explicitly typed exceptions rather than `error`. In this case as well as the line above it I'd use `throw(ArgumentError(...))` with the text.

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+    g = floor(Int, n * p)
+    
+    select!(x, 1:g)
+    select!(x, (n-g+1):n)
+    
+    return x[(g+1):(n-g)]
```
ararslan: Does this function actually modify the array in place? I thought slices made copies.

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+
+"""
+    winsor!(x, p=0.2)
+
+A variant of `winsor` that modifies vector `x` in place.
+"""
+function winsor!(x::AbstractVector, p::Real=0.2)
+    n = length(x)
+    n > 0 || error("x can not be empty.")
+    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
```
ararslan: Same comment regarding typed errors

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+
+#############################
+#
+#   Trimming outliers
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x, p=0.2)
```
ararslan: What's the motivation for a default of 0.02? If it were me I'd omit a default and require the user to always specify the proportion to remove from the tails.

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+"""
+    trimvar(x, p=0.2)
+
+Compute the variance of the trimmed mean of `x`. This function uses
+the Winsorized variance, as described in Wilcox (2010).
+"""
+function trimvar(x::RealArray, p::Real=0.2)
+    n = length(x)
+    n > 0 || error("x can not be empty.")
+    0 <= p < 0.5 || error("p must be in 0 <= p < 0.5.")
```
ararslan: typed errors

src/robust.jl
```diff
@@ -0,0 +1,110 @@
...
+#############################
+
+# Variance of a trimmed set.
+"""
+    trimvar(x, p=0.2)
+
+Compute the variance of the trimmed mean of `x`. This function uses
+the Winsorized variance, as described in Wilcox (2010).
+"""
+function trimvar(x::RealArray, p::Real=0.2)
```
ararslan: Why require a different type for the input here than the others? You have `AbstractVector` above and `RealArray` here.

src/robust.jl
KristofferC: Should use `deleteat!` probably.

andreasnoack: @ararslan The calls to `select!` modify `x`.

Another option could be to return a view?

ararslan: @andreasnoack Sure, but since slice at the end is returning a copy of part of the modified `x`, isn't it?

andreasnoack: I see. I thought your comment was because of the `!` part in the name but I agree that the docstring could give the impressions that the input `x` is actually modified to the trimmed version which is not the case.

test/robust.jl
```diff
@@ -0,0 +1,29 @@
...
+@test winsor([1,2,3,4,5,6,7,8])         == [2,2,3,4,5,6,7,7]
+@test winsor([1,2,3,4,5,6,7,8], 0.1)    == [1,2,3,4,5,6,7,8]
+@test winsor([1,2,3,4,5,6,7,8], 0.2)    == [2,2,3,4,5,6,7,7]
+@test winsor([1,2,3,4,5,6,7,8,9], 0.4)  == [4,4,4,4,5,6,6,6,6]
+
+@test winsor!([1,2,3,4,5,6,7,8])        == [2,2,3,4,5,6,7,7]
+@test winsor!([1,2,3,4,5,6,7,8], 0.1)   == [1,2,3,4,5,6,7,8]
+@test winsor!([1,2,3,4,5,6,7,8], 0.2)   == [2,2,3,4,5,6,7,7]
+@test winsor!([1,2,3,4,5,6,7,8,9], 0.4) == [4,4,4,4,5,6,6,6,6]
+
```
ararslan: Should probably add tests that trigger the error handling in your function. For that you can use `@test_throws`.

dcarrera:
@ararslan Thanks. Quick responses:

- Ooops. I should have deleted those lines.
- Fixed.
- Ok. I now use `throw(ArgumentError(...))` everywhere.
- My mistake. The previous implementation of `trim!` did sort `x`, but now it doesn't. Fixed.
- The default of `p=0.2` is recommended by Rand Wilcox (e.g. 2010). For distributions with thin tails, lower `p` is more efficient, but for fat tails higher `p` protects more against against outliers; Wilcox says that `p=0.2` is a good default that works well in most cases.
- My bad. Now it's `AbstractVector`.
- Ok. Added tests that trigger error handling.

dcarrera:
@ararslan ping?

src/robust.jl
KristofferC: Fix this to not return a copy?

src/robust.jl
```diff
@@ -0,0 +1,106 @@
...
+
+Compute the variance of the trimmed mean of `x`. This function uses
+the Winsorized variance, as described in Wilcox (2010).
+"""
+function trimvar(x::AbstractVector, p::Real=0.2)
+    n = length(x)
+    n > 0 || throw(ArgumentError("x can not be empty."))
+    0 <= p < 0.5 || throw(ArgumentError("p must be in 0 <= p < 0.5."))
+    
+    return var(winsor(x,p)) / (n * (1 - 2p)^2)
```
KristofferC:
```jl
t = var(winsor(x ,p))
scale!(t,  1/ (n * (1 - 2p)^2))
return t
```

to avoid a temporary?

ararslan:
I think it would be good to amend the `trim` function to use `deleteat!` as @KristofferC suggested. That way the input is modified as described and no copies or unnecessary allocations(?) are created.

dcarrera:
Ok. I think I got it.

dcarrera:
@ararslan ping? I think I got it now.

src/robust.jl
```diff
@@ -0,0 +1,108 @@
...
+by `trimvar` to compute the variance of the trimmed mean of `x`.
+
+# Example
+```julia
+julia> winsor([1,2,3,4,5], 0.2)
+ 2
+ 2
+ 3
+ 4
+ 4
```
ararslan: This is a pretty minor point, but it'd be nice to exactly emulate the output of `winsor` at the REPL. In particular, this is missing the `Array{Whatever,1}` stuff that gets printed before the list of elements.

src/robust.jl
```diff
@@ -0,0 +1,108 @@
...
+function winsor!(x::AbstractVector, p::Real=0.2)
+    n = length(x)
+    n > 0 || throw(ArgumentError("x can not be empty."))
+    0 <= p < 0.5 || throw(ArgumentError("p must be in 0 <= p < 0.5."))
+    g = floor(Int, n * p)
+    
+    select!(x, 1:g)
+    select!(x, (n-g+1):n)
+    x[1:g] = x[g+1]
+    x[n-g+1:end] = x[n-g]
```
ararslan: As with the previous version of `trim!`, this will also make copies. To do it in place I think you can use `setindex!`.

ararslan:
Thanks for your hard work here and sorry for the delay in getting back to you.

Personally I'm not a fan of the `p=0.2` default in `trim` et al. but I'd be interested to hear what the others think about that. What say ye, @andreasnoack and @KristofferC?

Besides that and the stuff I noted, this looks good to me.

src/robust.jl
johnmyleswhite: +1 to requiring users to specify this number

src/robust.jl
ararslan: Could probably also do `view` in the `setindex!` as further means of avoiding copies.

bjarthur:
+!  thanks @dcarrera.  you beat me to it.

re. default of 0.2.  that's a pretty common value to use, so it makes sense to include it.

however, the old `trimmean` behaved differently in this regard in that it trimmed half of the specified p value from each extreme, instead of the full value.  so with p=0.2 it trimmed the upper 10% and lower 10%.  not sure which is preferred.

scipy's [winsorize](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.mstats.winsorize.html#scipy.stats.mstats.winsorize) actually takes a 2-tuple specifying the upper and lower trim fractions.  not sure we need that flexibility as doing anything other than making them equal would introduce bias.

src/robust.jl
KristofferC: I don't think this will make a copy because the RHS is a scalar.

bjarthur:
one other thing to consider:  it might be more useful to some users to specify the absolute number of values to trim, instead of the relative percentage.  so say, the 10 most extreme values on both ends, instead of say 2% on each end.

not sure what the best interface would be.  it might to prone to error, but one could consider a Float value of p less than one to be a relative percentage, and an Integer value >=1 to be an absolute number.

src/robust.jl
ararslan: Oh okay, didn't know that. Thanks @KristofferC!

ararslan:
@bjarthur I think one way to do what you're describing would be to use separate keyword arguments, which maybe would be an okay thing to do.

nalimilan:
> @bjarthur I think one way to do what you're describing would be to use separate keyword arguments, which maybe would be an okay thing to do.

Agreed, it would make sense to use a keyword argument for the proportion, so that in the future we can add support for passing a number of elements.

src/robust.jl
nalimilan: `[...] =` is lowered to `setindex!`, so these are always equivalent anyway.

bjarthur:
i don't understand how keywords could be used to specify either an absolute or relative number.  if each has a separate argument with a default, which one would be used?  perhaps you could write out the function signature for me?

nalimilan:
Both would default to `0` (or `0.0`), and the non-zero argument would be used. Passing two zeros or two non-zeros would be an error.

bjarthur:
i'm on the fence, but i think it'd be more julian to use type dispatch like this:

```
function winsor!(x::AbstractVector, p::AbstractFloat)
    0.0<p<1.0 || throw(ArgumentError)
    winsor!(x, round(Int, length(x)*p))
end

function winsor!(x::AbstractVector, p::Integer)
    0<p<length(x) || throw(ArgumentError)
    # then proceed as normal here...
end
```

nalimilan:
That could work, but doing something completely different just because you passed an `Int` rather than an `AbstractFloat` is quite unusual in Julia. I don't think you'll find another example of a numeric argument being interpreted in different ways depending on its type. A keyword is more explicit.

dcarrera:
Hello. Been busy. I'll try to progress through the latest comments now.

dcarrera:
@ararslan @bjarthur So... Have we decided that `p=0.2` is fine as a default? This is the value recommended by Wilcox (2010). That's why I set it that way.

@bjarthur I am no expert, but I have seen several papers or books define the x% trimmed mean as removing x% from the top and x% from the bottom. I have not seen any articles or book define it the other way. Also, most software (R, SciPy and Stata) uses the definition that I'm using. So... I definitely think that this is the better definition.

dcarrera:
@bjarthur @nalimilan Having the function behave in a completely different way because the value is an `Int` seems like asking for trouble. That said, an `Int` would never be a valid proportion because the proportion `p` has to be less than `0.5`. So **strictly speaking** it is not a conflict. It's just a scary idea.

Crap... why did the checks fail? All I did was change one line on the docstring.

ararslan:
The failures are related to issues with DataFrames that should be fixed in the next release.

dcarrera:
@ararslan  Thanks.

ararslan:
> Have we decided that `p=0.2` is fine as a default? This is the value recommended by Wilcox (2010). That's why I set it that way.

I'm still generally against having a default here at all; I'd prefer that users think more carefully about the proportion of their data they're excluding.

If it were me, I'd define two functions with separate keyword arguments: one for a proportion and one for a count of elements to remove. The proportion version would figure out the number of elements to remove then call the count version to do the removal. That way we have both methods available, and there isn't any duplication in case things have to be modified.

dcarrera:
Here is a possible implementation that allows the user to specify the number of elements `g` to trim from the top and bottom using a keyword arguments.

```
trim!(vals, g=10) --> Trim the 10 highest and 10 lowest values.
trim!(vals, p=0.3) --> Trim the 30% highest and 30% lowest values.
trim!(vals) --> Default: Trim the 20% highest and 20% lowest values.

trim!(vals, g=10, p=0.3) --> Behaves like trim!(vals, g=10)
```

```julia
function trim!(x::AbstractVector; p::Real=0.2, g::Integer=0)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))
    0 <= g < n/2 || throw(ArgumentError("g must be less than half the length of x."))
    if g != 0
        0 <= p < 0.5 || throw(ArgumentError("p must be in 0 <= p < 0.5."))
    end
    g = floor(Int, n * p)
    
    select!(x, (n-g+1):n)
    select!(x, 1:g)
    deleteat!(x, (n-g+1):n)
    deleteat!(x, 1:g)
    
    return x
end
```

dcarrera:
Here is a possible implementation that uses different keywords to let
the user specify either the proportion `p` or absolute number `g` of
elements to remove.

A slightly unusual feature is that negative values of `g` are not
illegal; they just hand over priority to `p`. This is so that a user
can write `trim(vals,g=0)` and have that behave as expected.


```
trim!(vals, g=10) --> Trim the 10 highest and 10 lowest values.
trim!(vals, p=0.3) --> Trim the 30% highest and 30% lowest values.
trim!(vals) --> Default: Trim the 20% highest and 20% lowest values.

trim!(vals, g=10, p=0.3) behaves like trim!(vals, g=10)
```

```julia
function trim!(x::AbstractVector; p::Real=0.2, g::Integer=-1)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))
    
    if g < 0
        0 <= p < 0.5 || throw(ArgumentError("p must be in 0 <= p < 0.5."))
        g = floor(Int, n * p)
    else
        0 <= g < n/2 || throw(ArgumentError("g must be in 0 <= g < length(x)/2."))
    end

    select!(x, (n-g+1):n)
    select!(x, 1:g)
    deleteat!(x, (n-g+1):n)
    deleteat!(x, 1:g)

    return x
end
```

ararslan:
I think if we're going to use keyword arguments, they should have more descriptive names than `p` and `g`, maybe something like `proportion` and `count`. Long but immediately descriptive.

dcarrera:
@ararslan We are not allowed to have two different functions with just different keyword arguments. The function signature is determined only by the non-keyword arguments. I understand that you want to do something like this:

```julia
function trim!(x::AbstractVector; g::Integer=0)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))
    0 <= g < n/2 || throw(ArgumentError("g must be in 0 <= g < length(x)/2."))

    select!(x, (n-g+1):n)
    select!(x, 1:g)
    deleteat!(x, (n-g+1):n)
    deleteat!(x, 1:g)

    return x
end

function trim!(x::AbstractVector; p::Real=0.2)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))
    0 <= p < 0.5 || throw(ArgumentError("p must be in 0 <= p < 0.5."))
    g = floor(Int, n * p)

    return trim!(x, g)
end
```

But the second function just overwrites the first.

dcarrera:
@ararslan I like `count`. I'm too lazy to type `roportion` :-) so I would prefer `p` and `count`.

nalimilan:
I think he meant two different *functions* (not *methods*), with two different names. I think I'd still prefer a single one. That doesn't mean we can't share the underlying code by using an internal function taking only a number of cases to drop.

nalimilan:
Let's drop the default proportion too, people should really think about it twice.

dcarrera:
So... like this?

```julia
function trim!(x::AbstractVector; p::Real=0.0, count::Integer=0)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))
    
    if count == 0
        0 <= p < 0.5 || throw(ArgumentError("p must be in 0 <= p < 0.5."))
        count = floor(Int, n * p)
    else
        0 <= count < n/2 || throw(ArgumentError("count must be in between 0 and length(x)/2."))
    end

    select!(x, (n-count+1):n)
    select!(x, 1:count)
    deleteat!(x, (n-count+1):n)
    deleteat!(x, 1:count)

    return x
end
```

ararslan:
@nalimilan No, I meant methods, I just forgot that you can't do it the way I was thinking. ¯\\\_(ツ)\_/¯

How about `prop` instead of `p`?

dcarrera:
I guess I could live with `prop` if you think it's better.

nalimilan:
LGTM, after renaming to `prop`. Though I'm not sure that the non-emptiness condition is really needed. Also, "p must be in 0 <= p < 0.5." isn't correct: say something like "p must respect 0 ≤ p < 0.5."

nalimilan:
I guess `prop == 0` should also raise an error. Else, the default when no argument is passed will be to return an empty vector, which isn't very intuitive.

src/deprecates.jl
```diff
@@ -20,5 +23,4 @@ import Base.varm, Base.stdm
 @deprecate R2(obj::StatisticalModel, variant::Symbol) r2(obj, variant)
 @deprecate R²(obj::StatisticalModel, variant::Symbol) r²(obj, variant)
 @deprecate adjR2(obj::StatisticalModel, variant::Symbol) adjr2(obj, variant)
-@deprecate adjR²(obj::StatisticalModel, variant::Symbol) adjr²(obj, variant)
-
+@deprecate adjR²(obj::StatisticalModel, variant::Symbol) adjr²(obj, variant)
```
nalimilan: Unrelated change here.

dcarrera:
@nalimilan `prop = 0` means no trimming. How is that an error? It'd just return the original array unchanged. In fact, any `prop` value smaller than `1/n` would return the array unchanged. However, `prop = 0.5` would return an empty array whenever `n` is even. That's why it raises an error.

nalimilan:
Using a trimming function and asking it not to trim doesn't sound very useful. Though if you want to accept that value, just use a different default for `prop` (e.g. `NaN`), so that at least you get an error when no argument is passed at all.

dcarrera:
I just added a new commit that replaces `p` with the keyword arguments `prop` and `count`. Honestly, I don't like `count` because I don't know whether the `trimvar()` function is mathematically sound when using the `count` keyword. The theorem that I know that says how to use the variance of the Winsorized set to obtain the variance of the trimmed set is based on the `prop` definition of trimming. I did my best to do the equivalent calculation for `count` by setting `prop = count/n`, but I am not an expert in this field and I do not know if this is sound or reasonable.

dcarrera:
@nalimilan If you want to give an error by default, then ok, just make `prop = 0` an error. I would have been happier with `prop = 0.2` as the default, since that's the value that I see recommended. Instead of giving an error by default, I'd prefer to give a sensible answer by default. But oh well. At this point I just want this patch accepted. This has taken far too long.

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+
+#############################
+#
+#   Trimming outliers
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x, [prop=0.0, count=0])
```
nalimilan: No need for brackets here and below: the presence of a default value is enough.

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+    deleteat!(x, (n-count+1):n)
+    deleteat!(x, 1:count)
+
+    return x
+end
+
+"""
+    winsor(x, [prop=0.0, count=0])
+
+Return a Winsorized version of vector `x`; i.e. replace either `count` or
```
nalimilan: Instead of "version", "copy" would be more accurate.

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+#   Trimming outliers
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x, [prop=0.0, count=0])
+
+
+Remove either `count` or proportion `prop` of the highest and lowest
```
nalimilan: More accurately: "Return a copy of `x` with...".

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+#
+#   Other
+#
+#############################
+
+# Variance of a trimmed set.
+"""
+    trimvar(x, [prop=0.0, count=0])
+
+Compute the variance of the trimmed mean of `x`. This function uses
```
nalimilan: "the trimmed mean of `x` (i.e. `mean(trim(x, ...))`)".

nalimilan:
Sorry if that feels painful, but designing APIs is hard. One (summer) month for a PR isn't unusual. Just one more round and we'll be ready for merging! Thanks!

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+function trim!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
+    n = length(x)
+    n > 0 || throw(ArgumentError("x can not be empty."))
+
+    if count == 0
+        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
+        count = floor(Int, n * prop)
+    else
+        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
+    end
```
bjarthur: the `if count == 0` code block appears three times, and each is identical.  might consider pulling it out into a separate function and using `@inline` so there is no overhead.

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+
+    if count == 0
+        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
+        count = floor(Int, n * prop)
+    else
+        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
+        prop = count/n
+    end
+
+    return var(winsor(x, count=count)) / (n * (1 - 2prop)^2)
```
bjarthur: actually, in this third use of `if count==0`, why not just skip it entirely and just pass `prop` and `count` verbatim as `var(winsor(x, prop=prop , count=count))...`

src/robust.jl
bjarthur: moreover, shouldn't it also be an `ArgumentError` if both `count` *and* `prop` are non-zero?  if so, should check for this and `throw` accordingly.

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
+        count = floor(Int, n * prop)
+    else
+        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
+    end
+
+    select!(x, (n-count+1):n)
+    select!(x, 1:count)
+    deleteat!(x, (n-count+1):n)
+    deleteat!(x, 1:count)
```
bjarthur: i think it would be faster to simply `return x[(count+1):(n-count+1)]` instead of using `deleteat!`.  have you run any benchmarks?

bjarthur:
looking great @dcarrera.  the bikeshedding can seem endless at times, but in the end it makes julia better.  don't let it get you down.

src/robust.jl
nalimilan: Being faster is not the only criterion: making a copy will use more memory, which could be a problem depending on the situation.

KristofferC: We want to update the original `x`. `deleteat` on the end and front is fast.

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+#
+#   Trimming outliers
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x, [prop=0.0, count=0])
+
+
```
ararslan: One too many blank lines here

bjarthur:
have we given up on this PR for trimming and winsorizing?  would be great to finish it off and merge...

nalimilan:
@bjarthur If @dcarrera does not reply, would you be willing to open a new PR fixing the few remaining issues (just add a commit so that we attribute the work to the original author)?

nalimilan:
Closing in favor of https://github.com/JuliaStats/StatsBase.jl/pull/267.

PR #209, opened 2016-08-22
Add generalized mean function

c123w:
Thought it would be helpful to have a generalized mean function (in addition to the special cases such as harmmean that are currently in StatsBase.jl).

### DIFF
--- a/docs/source/means.rst
+++ b/docs/source/means.rst
@@ -13,7 +13,9 @@ The package provides functions to compute means of different kinds.
 
 .. function:: genmean(x, p)
 
-  Compute the generalized/power mean with exponent ``p`` of ``x``.
+  Compute the generalized/power mean of ``x`` with exponent ``p``,
+  i.e. :math:`\left( \frac{1}{n} \sum_{i=1}^n x_i^p \right)^{\frac{1}{p}}`, where ``n == length(a)``.
+  It is taken to be the geometric mean when ``p == 0``.
 
 .. function:: trimmean(x, p)
 
--- a/docs/source/means.rst
+++ b/docs/source/means.rst
@@ -11,6 +11,10 @@ The package provides functions to compute means of different kinds.
 
   Compute the harmonic mean of ``x``.
 
+.. function:: genmean(x, p)
+
+  Compute the generalized/power mean with order ``p`` of ``x``.
+
 .. function:: trimmean(x, p)
 
   Compute the trimmed mean of ``x``, with fraction ``p`` of elements ignored.
@@ -19,14 +23,14 @@ The package provides functions to compute means of different kinds.
 
   .. code-block:: julia
 
-    # computes the mean using 80% of the numbers in ``x``, 
+    # computes the mean using 80% of the numbers in ``x``,
     # ignoring 10% of the largest and 10% of the smallest values.
 
     trimmean(x, 0.2)
 
 .. function:: mean(x, w)
 
-  The ``mean`` function is also extended to accept a weight vector of type ``WeightVec`` (see :ref:`weightvec`) to compute weighted mean. 
+  The ``mean`` function is also extended to accept a weight vector of type ``WeightVec`` (see :ref:`weightvec`) to compute weighted mean.
 
   **Examples:**
 
@@ -41,4 +45,4 @@ The package provides functions to compute means of different kinds.
 
 .. function:: mean!(dst, x, w, dim)
 
-  Compute weighted means along a certain dimension, and write results to a pre-allocated destination vector ``dst``. 
+  Compute weighted means along a certain dimension, and write results to a pre-allocated destination vector ``dst``.
--- a/docs/source/means.rst
+++ b/docs/source/means.rst
@@ -14,7 +14,7 @@ The package provides functions to compute means of different kinds.
 .. function:: genmean(x, p)
 
   Compute the generalized/power mean of ``x`` with exponent ``p``,
-  i.e. :math:`\left( \frac{1}{n} \sum_{i=1}^n x_i^p \right)^{\frac{1}{p}}`, where ``n == length(x)``.
+  i.e. :math:`\left( \frac{1}{n} \sum_{i=1}^n x_i^p \right)^{\frac{1}{p}}`, where ``n = length(x)``.
   It is taken to be the geometric mean when ``p == 0``.
 
 .. function:: trimmean(x, p)
--- a/docs/source/means.rst
+++ b/docs/source/means.rst
@@ -13,7 +13,7 @@ The package provides functions to compute means of different kinds.
 
 .. function:: genmean(x, p)
 
-  Compute the generalized/power mean with order ``p`` of ``x``.
+  Compute the generalized/power mean with exponent ``p`` of ``x``.
 
 .. function:: trimmean(x, p)
 
--- a/docs/source/means.rst
+++ b/docs/source/means.rst
@@ -14,7 +14,7 @@ The package provides functions to compute means of different kinds.
 .. function:: genmean(x, p)
 
   Compute the generalized/power mean of ``x`` with exponent ``p``,
-  i.e. :math:`\left( \frac{1}{n} \sum_{i=1}^n x_i^p \right)^{\frac{1}{p}}`, where ``n == length(a)``.
+  i.e. :math:`\left( \frac{1}{n} \sum_{i=1}^n x_i^p \right)^{\frac{1}{p}}`, where ``n == length(x)``.
   It is taken to be the geometric mean when ``p == 0``.
 
 .. function:: trimmean(x, p)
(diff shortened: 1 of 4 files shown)


src/scalarstats.jl
```diff
@@ -37,6 +37,26 @@ function harmmean(a::RealArray)
     return n / s
 end
 
+# Generalized mean
+"""
+    genmean(a, p)
```
nalimilan: Please add a line break after the signature.

src/scalarstats.jl
```diff
@@ -37,6 +37,26 @@ function harmmean(a::RealArray)
...
+Return the generalized/power mean with exponent p of a real-valued array.
```
nalimilan: Backquotes around `p`.

nalimilan: Could you also give the formula?

src/scalarstats.jl
```diff
@@ -37,6 +37,26 @@ function harmmean(a::RealArray)
...
+"""
+function genmean(a::RealArray, p::Real)
+  if p<0
+    p = convert(AbstractFloat, p)# Negative int p causes domain error.
```
nalimilan: This will cause type-instability of `p`. Use a different variable name.

src/scalarstats.jl
```diff
@@ -37,6 +37,26 @@ function harmmean(a::RealArray)
...
+  end
+  if p==0.0
```
nalimilan: `elseif`.  I would even put this branch first, as it's a fast-path.

src/scalarstats.jl
```diff
@@ -37,6 +37,26 @@ function harmmean(a::RealArray)
     return n / s
 end
 
+# Generalized mean
+"""
+    genmean(a, p)
+Return the generalized/power mean with exponent p of a real-valued array.
+"""
+function genmean(a::RealArray, p::Real)
+  if p<0
```
nalimilan: Spaces around `<` (and around `==` below).

src/scalarstats.jl
```diff
@@ -37,6 +37,26 @@ function harmmean(a::RealArray)
...
+    p = convert(AbstractFloat, p)# Negative int p causes domain error.
+  end
+  if p==0.0
+    return geomean(a)
+  end
+  s = 0.0
+  n = length(a)
+  for i = 1:n
```
nalimilan: Even simpler, you can write `for x in a`.

nalimilan:
Thanks, looks like a useful addition indeed.

src/scalarstats.jl
c123w: Happy to change this, but is there a reason that geomean and harmmean don't use "for x in a"?

nalimilan: Most likely not. That shouldn't make a difference in practice. Keep it that way if you prefer consistency.

src/scalarstats.jl
nalimilan: Also, `== 0` is enough.

nalimilan:
(If you feel like it, adding support for weighted geometric/harmonic/generalized means would also be a good idea for a later PR.)

src/scalarstats.jl
nalimilan: More precisely, to ensure type-stability, you need to always use a float when `p` is an integer and `a` holds integers, even when `p` is positive. So you can get rid of that `if`, and always do `pf = convert(AbstractFloat, p)` (and only work with `pf` below).

src/scalarstats.jl
```diff
@@ -37,6 +37,25 @@ function harmmean(a::RealArray)
...
+i.e. returns ((a_1+a_2+...+a_n)/n)^(1/p), where n is the length of the array a.
+"""
+function genmean(a::RealArray, p::Real)
+  if p == 0
+    return geomean(a)
+  end
+  s = 0.0
+  n = length(a)
+  for x in a
+    @inbounds s += convert(Float64, x)^p
```
c123w: @nalimilan I did a bit of experimenting and I think it's better to have ``s += convert(Float64,  x)^p`` instead of ``s += x^pf``. More specifically, the former is no slower than the latter with float ``p``, but much faster when ``p`` is an integer (due to exponentiation by squaring).

nalimilan: Makes sense. Please keep the comment about the domain error with integers, though. Else it's hard to understand the justification for this conversion.

test/scalarstats.jl
```diff
@@ -17,6 +17,13 @@ using Base.Test
 @test_approx_eq harmmean(1:3) 3 / (1 + 1/2 + 1/3)
 @test_approx_eq harmmean([1, 2, 4]) 12 / 7
 
+## genmean
+@test_approx_eq genmean([1,1,2,3], 1) 7/4
```
nalimilan: Please add a test with floating points values (`a` and/or `p`) too.

test/scalarstats.jl
```diff
@@ -17,6 +17,13 @@ using Base.Test
...
+@test_approx_eq genmean([1,4,2], -1) 12/7
+@test_approx_eq genmean([1,1,2,3], 0) (6.0)^(1/4)
+@test_approx_eq_eps genmean([1,1,2,3], 1e-8) (6.0)^(1/4) 1e-8
```
nalimilan: What are you testing specifically here?

c123w: Trying to ensure that the algorithm is numerically stable when `p` is near to 0 (answer should be very close to the geometric mean).

nalimilan: OK. Maybe add a comment to explain this?

test/scalarstats.jl
```diff
@@ -23,8 +23,8 @@ using Base.Test
 @test_approx_eq genmean([1,1,2,3], 0) (6.0)^(1/4)
 @test_approx_eq genmean([1.2,-0.5,0], 2) sqrt(169/300)
 @test_approx_eq genmean([16/9,0.25,1.0], 1.5) (755/648)^(2/3)
-@test_approx_eq_eps genmean([1,1,2,3], -1e-8) (6.0)^(1/4) 1e-8
-@test_approx_eq_eps genmean([0.98,1.02], 1e4) 1.02 1e-4
+@test_approx_eq_eps genmean([1,1,2,3], -1e-8) (6.0)^(1/4) 1e-8# Test numerical stability for `p` close to 0 (genmean should be close to geometric mean).
```
nalimilan: Lines should be under 92 characters, so please put the comments on separate lines (also, the common convention is to add a space before `#` when on the same line).

src/scalarstats.jl
```diff
@@ -37,6 +37,27 @@ function harmmean(a::RealArray)
     return n / s
 end
 
+# Generalized mean
+"""
+    genmean(a, p)
+
+Return the generalized/power mean with exponent `p` of a real-valued array.
```
nalimilan: Comma instead of dot before "i.e.".

src/scalarstats.jl
```diff
@@ -37,6 +37,27 @@ function harmmean(a::RealArray)
...
+i.e. returns ((a_1^p+a_2^p+...+a_n^p)/n)^(1/p), where n is the length of the array a.
```
nalimilan: "returns" is redundant here (and there's a discrepancy with "return" anyway). Also add double bacticks around the equation, and single backticks around `n` and `a`. You could write the last part simply as "with `n=length(a)`".

docs/source/means.rst
```diff
@@ -11,6 +11,10 @@ The package provides functions to compute means of different kinds.
 
   Compute the harmonic mean of ``x``.
 
+.. function:: genmean(x, p)
+
+  Compute the generalized/power mean with exponent ``p`` of ``x``.
```
nalimilan: Need to update this to reflect the new docstring.

nalimilan:
Thanks! After the last round of fixes, I'll merge in a few days if nobody objects.

docs/source/means.rst
```diff
@@ -13,7 +13,9 @@ The package provides functions to compute means of different kinds.
 
 .. function:: genmean(x, p)
 
-  Compute the generalized/power mean with exponent ``p`` of ``x``.
+  Compute the generalized/power mean of ``x`` with exponent ``p``,
+  i.e. :math:`\left( \frac{1}{n} \sum_{i=1}^n x_i^p \right)^{\frac{1}{p}}`, where ``n == length(a)``.
```
nalimilan: Note you can use that LaTeX syntax even in Julia docstrings (see 6 at http://docs.julialang.org/en/stable/manual/documentation/) when using double backticks, so that there's no difference between the two versions. Also `n = length(x)` would be more logical.

src/scalarstats.jl
```diff
@@ -37,6 +37,28 @@ function harmmean(a::RealArray)
...
+    return geomean(a)
+  end
+  s = 0.0
+  n = length(a)
+  for x in a
+    #= At least one of `x` or `p` must not be an int to avoid domain errors when `p` is a negative int.
+    We choose `x` in order to exploit exponentiation by squaring when `p` is an int. =#
+    @inbounds s += convert(Float64, x)^p
+  end
+  return (s/n)^(1/p)
```
ararslan: Please indent with 4 spaces rather than 2 for consistency with the rest of the repo

nalimilan:
Thanks, and sorry for the delay!

PR #214, opened 2016-09-16
Deprecate findat() and findat!()

nalimilan:
findat() is equivalent to Base.indexin(), and adds to the long list of
find* functions for no reason.

findat!() has no equivalent yet in Base, but it's not used in StatsBase
and its definition can easily be copied to private repositories until Base
gets indexin!().

### DIFF
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -22,3 +22,5 @@ import Base.varm, Base.stdm
 @deprecate adjR2(obj::StatisticalModel, variant::Symbol) adjr2(obj, variant)
 @deprecate adjR²(obj::StatisticalModel, variant::Symbol) adjr²(obj, variant)
 
+@deprecate findat(a, x) indexin(x, a)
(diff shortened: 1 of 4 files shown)


src/deprecates.jl
```diff
@@ -22,3 +22,5 @@ import Base.varm, Base.stdm
 @deprecate adjR2(obj::StatisticalModel, variant::Symbol) adjr2(obj, variant)
 @deprecate adjR²(obj::StatisticalModel, variant::Symbol) adjr²(obj, variant)
 
+@deprecate findat(a, x) indexin(x, a)
+@deprecate findat!(r, a, x)  (y = indexin(x, a); copy!(r, y))
```
ararslan: Is there a reason to do it like this rather than `copy!(r, indexin(x, a))`?

nalimilan: None. I've updated the PR.

PR #218, opened 2016-10-07
Fix incorrect findat() deprecation

nalimilan:
indexin() returns the last matching index, while findat() returns the first one.
In practice this distinction is not always useful for callers, keep the old code
to prevent breakage and leave users time to adapt.


nalimilan:
Do you want to merge and tag a new release?

PR #222, opened 2016-10-19
Fix test error due to Vector{Any} on 0.4

andreasnoack:


tkelman:
What caused the type to change here? This seems like a regression somewhere that should be identified in more detail.

andreasnoack:
These are all over the place on 0.4 now. It's because of the dot syntax that uses `broadcast` and on 0.4 that cannot be inferred.

PR #225, opened 2016-12-02
Adds ConvergenceException representing failure to converge errors

grero:
OK, I took a stab at a pull request. Let me know if I should change anything. In particular, I'm not sure if the test I added is appropriate.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -157,6 +157,7 @@ module StatsBase
     r2,
     r²,
     model_response,
+
     ConvergenceException
 
     # source files
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -156,7 +156,8 @@ module StatsBase
     residuals,
     r2,
     r²,
-    model_response
+    model_response,
+    ConvergenceException
 
     # source files
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -235,7 +235,9 @@ function show(io::IO, ct::CoefTable)
 end
 
 """
-Exception to handle failure of convergence when fitting statistical models.
+	ConvergenceException(iters::Int)
+
+The fitting procedure failed to converge in `iters` number of iterations.
 """
 type ConvergenceException <: Exception
     iters::Int
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -233,3 +233,12 @@ function show(io::IO, ct::CoefTable)
         println(io)
     end
 end
+
+"""
+Exception to handle failure of convergence when fitting statistical models.
+"""
+type ConvergenceException <: Exception
+    iters::Int
+end
+
+Base.showerror(io::IO, ce::ConvergenceException) = print(io, "failure to converge after $(ce.iters) iterations")
--- a/test/statmodels.jl
+++ b/test/statmodels.jl
@@ -34,5 +34,5 @@ x3   0.453058  0.72525 0.999172 0.5567
 try
     throw(ConvergenceException(10))
 catch ex
-    @test sprint(Base.showerror, ex) == "failure to converge after 10 iterations"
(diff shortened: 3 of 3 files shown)


test/statmodels.jl
```diff
@@ -30,3 +30,4 @@ x3   0.453058  0.72525 0.999172 0.5567
 @test sprint(show, StatsBase.PValue(NaN)) == "NaN"
 @test_throws ErrorException StatsBase.PValue(-0.1)
 @test_throws ErrorException StatsBase.PValue(1.1)
+@test_throws ConvergenceException throw(ConvergenceException(10))
```
ararslan: A better test would probably be
```julia
try
    throw(ConvergenceException(10))
catch ex
    @test sprint(Base.showerror, ex) == "failure to converge after 10 iterations"
end
```
That ensures that the error text is as expected.

src/StatsBase.jl
```diff
@@ -156,7 +156,8 @@ module StatsBase
     residuals,
     r2,
     r²,
-    model_response
+    model_response,
+    ConvergenceException
```
nalimilan: Please add a line break to follow the convention used elsewhere in this list.

test/statmodels.jl
```diff
@@ -30,3 +30,9 @@ x3   0.453058  0.72525 0.999172 0.5567
 @test sprint(show, StatsBase.PValue(NaN)) == "NaN"
 @test_throws ErrorException StatsBase.PValue(-0.1)
 @test_throws ErrorException StatsBase.PValue(1.1)
+
+try
+    throw(ConvergenceException(10))
+catch ex
+    @test sprint(Base.showerror, ex) == "failure to converge after 10 iterations"
```
nalimilan: No need for `Base.` here.

src/StatsBase.jl
ararslan: I don't understand. This doesn't seem to go against the existing style, and he isn't removing an existing line break?

test/statmodels.jl
ararslan: That's my bad. For some reason I thought it wasn't exported.

src/StatsBase.jl
ararslan: To clarify, a line break between `model_response,` and `ConvergenceException`.

test/statmodels.jl
grero: Hm, when I remove `Base`, I get this
```julia
julia> throw(ConvergenceException(10))
ERROR: StatsBase.ConvergenceException(10)
```
whereas with `Base`, I get what I expect
```julia
julia> throw(ConvergenceException(10))
ERROR: failure to converge after 10 iterations
```

src/StatsBase.jl
grero: Happy to do it : ) Would that be because `ConvergenceException` forms a new "grouping" on its own?

test/statmodels.jl
nalimilan: You need `Base.` when defining the method, but not when calling it.

grero: Ah, sorry I was referring to the wrong piece of code. Ignore my comment please

src/StatsBase.jl
nalimilan: Either you should follow the lexical ordering, or you should create a new group. I don't have strong feelings about which choice is best.

src/statmodels.jl
```diff
@@ -233,3 +233,12 @@ function show(io::IO, ct::CoefTable)
         println(io)
     end
 end
+
+"""
+Exception to handle failure of convergence when fitting statistical models.
```
nalimilan: Sorry not to have raised this before, but I think the description should avoid repeating "exception" and should be directed towards the user who sees this exception raised. See for example `?DomainError` or `?ArgumentError`.

ararslan:
Looks good to me.

PR #227, opened 2016-12-17
added docstring for predict

mkborregaard:
Requested by @nalimilan here https://github.com/JuliaStats/GLM.jl/pull/171

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -145,11 +145,18 @@ end
 
 const adjr² = adjr2
 
+
 abstract RegressionModel <: StatisticalModel
 
 fitted(obj::RegressionModel) = error("fitted is not defined for $(typeof(obj)).")
 model_response(obj::RegressionModel) = error("model_response is not defined for $(typeof(obj)).")
 residuals(obj::RegressionModel) = error("residuals is not defined for $(typeof(obj)).")
+
+"""
+    predict(obj::RegressionModel [, newX])
+
+Form the predicted response of model `obj`, optionally from new covariate values `newX`
+"""
 predict(obj::RegressionModel) = error("predict is not defined for $(typeof(obj)).")
 predict!(obj::RegressionModel) = error("predict! is not defined for $(typeof(obj)).")
 dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(typeof(obj)).")
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -154,7 +154,9 @@ residuals(obj::RegressionModel) = error("residuals is not defined for $(typeof(o
 """
     predict(obj::RegressionModel, [newX])
 
-Form the predicted response of model `obj`, optionally from new covariate values `newX`
+Form the predicted response of model `obj`. New covariate values `newX` can be supplied as a `DataFrame`
+with the same names as the original predictors, or as an AbstractMatrix (for low-level use) with a column
+for the value of each predictor (including a `Vector` of 1s if an intercept was fitted). 
 """
 predict(obj::RegressionModel) = error("predict is not defined for $(typeof(obj)).")
 predict!(obj::RegressionModel) = error("predict! is not defined for $(typeof(obj)).")
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -154,9 +154,9 @@ residuals(obj::RegressionModel) = error("residuals is not defined for $(typeof(o
 """
     predict(obj::RegressionModel, [newX])
 
-Form the predicted response of model `obj`. New covariate values `newX` can be supplied as a `DataFrame`
-with the same names as the original predictors, or as an AbstractMatrix (for low-level use) with a column
-for the value of each predictor (including a `Vector` of 1s if an intercept was fitted). 
+Form the predicted response of model `obj`. An object with new covariate values `newX` can be supplied,
+and should have the same type and structure as that used to fit `obj`; e.g. for a GLM
+it would generally be a  `DataFrame` with the same variable names as the original predictors.
 """
 predict(obj::RegressionModel) = error("predict is not defined for $(typeof(obj)).")
 predict!(obj::RegressionModel) = error("predict! is not defined for $(typeof(obj)).")
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -155,8 +155,8 @@ residuals(obj::RegressionModel) = error("residuals is not defined for $(typeof(o
     predict(obj::RegressionModel, [newX])
 
 Form the predicted response of model `obj`. An object with new covariate values `newX` can be supplied,
-and should have the same type and structure as that used to fit `obj`; e.g. for a GLM
-it would generally be a  `DataFrame` with the same variable names as the original predictors.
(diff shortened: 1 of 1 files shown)


src/statmodels.jl
```diff
@@ -145,11 +145,18 @@ end
 
 const adjr² = adjr2
 
+
```
ararslan: Unrelated change here. Would be great if you could remove it, though it's not a huge deal.

src/statmodels.jl
```diff
@@ -145,11 +145,18 @@ end
...
 abstract RegressionModel <: StatisticalModel
 
 fitted(obj::RegressionModel) = error("fitted is not defined for $(typeof(obj)).")
 model_response(obj::RegressionModel) = error("model_response is not defined for $(typeof(obj)).")
 residuals(obj::RegressionModel) = error("residuals is not defined for $(typeof(obj)).")
+
+"""
+    predict(obj::RegressionModel [, newX])
```
ararslan: For consistency with the other docstrings, it should be `, [newX]`.

mkborregaard: sorry, fixed

ararslan: We've done the docstrings in this package slightly differently. IIRC it's not entirely consistent in Base.

mkborregaard: Cool - I already pushed the change

ararslan: Great, thanks :)

ararslan:
In what scenario would a one-argument `predict` make sense? Presumably you're always predicting a response based on some given values, which requires at least two arguments: the model from which to obtain the estimates and the data from which the response is predicted.

It would make sense to me to change the function signature to
```julia
"""
    predict(obj::RegressionModel, newX)

Predict the response of the regression model `obj` from the values in `newX`.
"""
predict(obj::RegressionModel, newX) = error(...)
```

Having `newX` be optional seems bizarre to me. What are your thoughts?

nalimilan:
Thanks. It can be useful to get the model predictions for the original dataset, e.g. when plotting a smoothed line (LOESS) -- though it's maybe redundant with `fitted` in that case. Anyway, the docstring should say what's the default for `newX`.

Would also be nice to explain what's the expected type/format for `newX`. AFAICT we support either a matrix or a data frame, right?

mkborregaard:
@ararslan Yes, as @nalimilan it is nice to just get the fitted values when not supplying newx. Also note that it is not redundaint with `fitted` now that the function has confidence intervals!

@nalimilan It is my impression that the interface is always to use a DataFrame, see the troubles I was having here: https://github.com/JuliaStats/GLM.jl/issues/165 . I think that the DataFrames module has a predict function that generates the newx matrix and calls the functions in GLM.

ararslan:
If it were me, I wouldn't specify the format of `newX` here, then any package that extends that method would document how they use the argument. We don't _use_ it here, we just _define_ it here, so it seems a little iffy to say here what the format is or should be.

nalimilan:
I don't think packages should extend this method to support other input types for `newX`: we really don't want some model families to support some input formats, and some others to support different formats. New models should only implement the low-level version taking a matrix, and StatsModels will provide the convenience method for `AbstractTable`. Currently this method is implemented in DataFrames, but this is only temporary.

So for now I would document both the matrix and data frame formats.

mkborregaard:
Just to state though that I find the current format really confusing - the user is supposed to supply a DataFrame, but the only docstring brought up by `?predict` is the GLM low-level version which is not supposed to be used by the user. I suggest having a DataFrame/(or really StatModels) version of the docstring specifying that this is the format expected to be used by the main user, and the low-level ones are for package developers.

mkborregaard:
I tried to incorporate a description of `newX` now.

src/statmodels.jl
```diff
@@ -150,6 +150,14 @@ abstract RegressionModel <: StatisticalModel
...
 model_response(obj::RegressionModel) = error("model_response is not defined for $(typeof(obj)).")
 residuals(obj::RegressionModel) = error("residuals is not defined for $(typeof(obj)).")
+
+"""
+    predict(obj::RegressionModel, [newX])
+
+Form the predicted response of model `obj`. New covariate values `newX` can be supplied as a `DataFrame`
+with the same names as the original predictors, or as an AbstractMatrix (for low-level use) with a column
+for the value of each predictor (including a `Vector` of 1s if an intercept was fitted). 
+"""
```
andreasnoack: I don't think this is the place to be specific about `newX`. I think the right place to add documentation is here https://github.com/JuliaStats/DataFrames.jl/blob/master/src/statsmodels/statsmodel.jl#L76

nalimilan: The point is, all these methods are going to be moved to StatsModels. So I don't think it's worth the trouble of adding partial docstrings everywhere just to remove them in a few weeks.

mkborregaard:
This PR was requested in another PR (on GLM) that was merged Jan 5. Should this be merged as well now?

andreasnoack:
Sorry for a negative comment but I still don't think this is the right place for the doc string. It belongs where the functionality is implemented.

mkborregaard:
No problem, I am just following up. If you can direct me to where this should go, in your view (if anywhere) I am happy to close this and make a new PR there.

nalimilan:
Let's just document here the method taking `AbstractMatrix` then, and the other one taking a data frame can go to DataFrames and StatsModels.

andreasnoack:
The method here just throws an error. The reason to have it here is to be able to import the name such that downstream packages can all use it without conflicts. If we add a docstring here, I think it should simply state that reason. Specific implementations can then add docstrings. There is already one in GLM, right?

mkborregaard:
There is one in GLM (https://github.com/JuliaStats/GLM.jl/blob/master/src/lm.jl#L161) but the problem is that it only uses AbstractMatrix (because that is what gets called here), whereas the user is supposed to call the DataFrames version (https://github.com/JuliaStats/GLM.jl/issues/165#issuecomment-266765305). The preferred DataFrames version is not documented, which is super confusing for the user, so I think the attempt was to remedy this.

nalimilan:
> Specific implementations can then add docstrings. There is already one in GLM, right?

@andreasnoack Yet we don't want all modeling packages to write their own docstrings, since they should all follow the same pattern for the most part for consistency. There should be a minimal docstring in StatsBase/StatsModels.

Then we can document the generic data frame method in DataFrames. That one cannot be documented by each modeling package, since they should only define an `AbstractMatrix` method.

andreasnoack:
I think many modeling packages would need to write their own docstring to give details about the implementation but I've been convinced that it could be helpful to have a general description of how `predict` is supposed to work here. I would like that the use of `DataFrames` was changed a bit though. The desciption here should allow for many kinds of data input where `DataFrames` and `AbstractMatrix` are just examples and that this is up to downstream packages to define these.

mkborregaard:
Great, what is the easiest way forward - can you edit the PR directly, or do you want to merge (or close) and do the edit?

mkborregaard:
OK, I have tried to change the docstring according to @andreasnoack 's suggestion

mkborregaard:
Thanks!

PR #229, opened 2017-01-19
Update to Julia 0.6

andreasnoack:
Drop 0.4 support

Disable statquiz test until DataArrays/DataFrames situation has stabilized


REQUIRE
```diff
@@ -1,2 +1,3 @@
-julia 0.4
-Compat 0.8.4
+julia 0.5
+Compat 0.13.0
```
ararslan: What's needed from a version so new?

src/StatsBase.jl
```diff
@@ -8,7 +8,11 @@ module StatsBase
     import Base: rand, rand!
     import Base.LinAlg: BlasReal, BlasFloat
     import Base.Cartesian: @nloops, @nref, @nextract
-    import Base.Collections: heapify!, heappop!, percolate_down!
+    if VERSION < v"0.6.0-dev.1950"
+        import Base.Collections: heapify!, heappop!, percolate_down!
+    else
+        import DataStructures: heapify!, heappop!, percolate_down!
+    end
```
ararslan: If we're requiring DataStructures as a dependency anyway, we can just use
```julia
using DataStructures: heapify!, heappop!, percolate_down!
```
rather than conditioning on the version since it's also available on 0.5.

src/moments.jl
```diff
@@ -63,10 +63,18 @@ function var!(R::AbstractArray, A::RealArray, wv::WeightVec, dim::Int; mean=noth
 end
 
 Base.varm(A::RealArray, wv::WeightVec, M::RealArray, dim::Int) =
-    Base.varm!(Array(Float64, Base.reduced_dims(A, dim)), A, wv, M, dim)
+    if VERSION < v"0.5.1"
```
ararslan: Any chance we can be more specific with the version requirements? (Here and elsewhere in the diff)

ararslan:
LGTM aside from the couple of things I mentioned.

REQUIRE
andreasnoack: The `.&` and `.|` stuff.

src/StatsBase.jl
andreasnoack: Good point. I'll change that.

src/moments.jl
nalimilan: The easiest and most specific approach is to use `isdefined`.

andreasnoack: I prefer explicit versions over `isdefined` because it makes it easy to clean up later the source later. I'll make them more precise

ararslan: Should also be `@static` here since it's in a function.

src/moments.jl
```diff
@@ -63,10 +63,18 @@ function var!(R::AbstractArray, A::RealArray, wv::WeightVec, dim::Int; mean=noth
 end
 
 Base.varm(A::RealArray, wv::WeightVec, M::RealArray, dim::Int) =
-    Base.varm!(Array(Float64, Base.reduced_dims(A, dim)), A, wv, M, dim)
+    @static if VERSION < v"0.5.1-pre+19"
+        Base.varm!(similar(A, Float64, Base.reduced_dims(size(A), dim)), A, wv, M, dim)
+    else
+        Base.varm!(similar(A, Float64, Base.reduced_indices(indices(A), dim)), A, wv, M, dim)
```
ararslan: Makes me wonder if it wouldn't be easier/more robust to just submit a PR to Compat to get `indices` and `reduced_indices`.

andreasnoack: It depends on how often these are used. It is typically much simpler to add a few conditions like these than adding functionality to `Compat` because it has to be more general, tested and documented. If the functionality gets added to `Compat`, we can update the code here but the priority right now is to get the stats package to work on master.

tkelman: I asked you in DataArrays, DO NOT code against unreleased commits on release-0.5 as if they are releases. breaking commits there are going to get reverted and this stuff makes testing that branch for stability impossible. stop doing this.

andreasnoack: Would `VERSION < v"0.5.1` better? I'm not going to use `isdefined`.

tkelman: if something on release-0.5 breaks a package, then release-0.5 needs to change, not the package

tkelman: ignore what's on the tip of release-0.5. if there's a breakage it won't be included in 0.5.1 release, it's a base bug.

REQUIRE
tkelman: looks like they're technically used only in the tests, but good enough

PR #230, opened 2017-01-20
Adjust VERSION condition to avoid using release branch versions

andreasnoack:


tkelman:
Thank you. Sorry if I was overly grumpy about it.

PR #233, opened 2017-02-09
fix deprecated FloatRange

wookay:
although `floatrange` has fallback soon but to hide depwarn.

https://github.com/JuliaLang/julia/issues/20506

thanks.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -73,7 +73,11 @@ function histrange{F}(lo::F, hi::F, n::Integer, closed::Symbol=:right)
             len += one(F)
         end
     end
-    FloatRange(start,step,len,divisor)
+    @static if VERSION < v"0.6.0-dev.2376" # Julia PR 18777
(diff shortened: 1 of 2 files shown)


src/hist.jl
```diff
@@ -73,7 +73,11 @@ function histrange{F}(lo::F, hi::F, n::Integer, closed::Symbol=:right)
             len += one(F)
         end
     end
-    FloatRange(start,step,len,divisor)
+    @static if VERSION < v"0.6.0-dev.2376" # Julia PR 18777
+        FloatRange(start,step,len,divisor)
+    else
+        Base.floatrange(start,step,len,divisor)
```
yuyichao: I think this should be added to `Compat`.

andreasnoack: We can update the code here when it gets in there.

tkelman: should be using steprangelen

PR #239, opened 2017-02-15
Fix #238

### LINKED ISSUE #238: Weighted sampling without replacement
Under Julia 0.5 and StatsBase 0.13.1 it seems that weighted sampling without replacement does not work correctly (unless I am not understanding how this sampling should be used).

Consider the code:
```
a = [1:10;]
wv = WeightVec([zeros(5); ones(5)])
sample(a, wv, 1, replace=false)
```
produces `[2]` (which should not happen as weight of 2 is 0.0) and moreover it is always `[2]` so actually no sampling takes place.

bkamins:
A proposal to fix #238. The original article assumes positive weights, so I propose to skip zero weights.
Additionally it is now strictly checked if there are not less positive weights in `wv` as required sample size.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -515,15 +515,22 @@ function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::Abst
 
     # initialize priority queue
     pq = Vector{Pair{Float64,Int}}(k)
-    @inbounds for i in 1:k
-        pq[i] = (wv.values[i]/randexp() => i)
+    i = 0
+    s = 0
+    @inbounds for s in 1:n
+        if wv.values[s] > 0.0
+            i += 1
+            pq[i] = (wv.values[s]/randexp() => s)
+        end
+        i >= k && break
     end
+    i < k && throw(DimensionMismatch("wv must have at least $k positive entries (got $i)"))
     heapify!(pq)
 
     # set threshold
     @inbounds threshold = pq[1].first
 
-    @inbounds for i in k+1:n
+    @inbounds for i in s+1:n
         key = wv.values[i]/randexp()
 
         # if key is larger than the threshold
@@ -561,17 +568,25 @@ function efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::WeightVec, x::Abs
 
     # initialize priority queue
     pq = Vector{Pair{Float64,Int}}(k)
-    @inbounds for i in 1:k
-        pq[i] = (wv.values[i]/randexp() => i)
+    i = 0
+    s = 0
+    @inbounds for s in 1:n
+        if wv.values[s] > 0.0
+            i += 1
+            pq[i] = (wv.values[s]/randexp() => s)
+        end
+        i >= k && break
     end
+    i < k && throw(DimensionMismatch("wv must have at least $k positive entries (got $i)"))
     heapify!(pq)
 
     # set threshold
     @inbounds threshold = pq[1].first
     X = threshold*randexp()
 
-    @inbounds for i in k+1:n
+    @inbounds for i in s+1:n
         w = wv.values[i]
+        w > 0.0 || continue
         X -= w
         X <= 0 || continue
 
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -518,20 +518,25 @@ function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::Abst
     i = 0
     s = 0
     @inbounds for s in 1:n
-        if wv.values[s] > 0.0
+        w = wv.values[s]
+        w < 0 && error("Negative weight found in weight vector at index $s")
+        if w > 0
             i += 1
-            pq[i] = (wv.values[s]/randexp() => s)
+            pq[i] = (w/randexp() => s)
         end
         i >= k && break
     end
-    i < k && throw(DimensionMismatch("wv must have at least $k positive entries (got $i)"))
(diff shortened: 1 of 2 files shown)


src/sampling.jl
```diff
@@ -515,15 +515,22 @@ function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::Abst
 
     # initialize priority queue
     pq = Vector{Pair{Float64,Int}}(k)
-    @inbounds for i in 1:k
-        pq[i] = (wv.values[i]/randexp() => i)
+    i = 0
+    s = 0
```
ararslan: `s` shouldn't be preinitialized here since it's only used in the loop

ararslan:
It would be great if you could add a test.

src/sampling.jl
```diff
@@ -515,15 +515,22 @@ function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::Abst
...
+    @inbounds for s in 1:n
+        if wv.values[s] > 0.0
```
nalimilan: `> 0` should be enough, no?

src/sampling.jl
```diff
@@ -515,15 +515,22 @@ function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::Abst
...
+            i += 1
+            pq[i] = (wv.values[s]/randexp() => s)
+        end
+        i >= k && break
     end
+    i < k && throw(DimensionMismatch("wv must have at least $k positive entries (got $i)"))
```
nalimilan: "strictly positive"

src/sampling.jl
```diff
@@ -561,17 +568,25 @@ function efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::WeightVec, x::Abs
...
     heapify!(pq)
 
     # set threshold
     @inbounds threshold = pq[1].first
     X = threshold*randexp()
 
-    @inbounds for i in k+1:n
+    @inbounds for i in s+1:n
         w = wv.values[i]
+        w > 0.0 || continue
```
nalimilan: Same here. Would be more consistent to use `> 0` since `<= 0` is used below.

nalimilan:
So now negative weights are silently ignored? Maybe better raise an error for negative weights, which do not have any meaning at the moment? Should also be mentioned in the docs.

bkamins:
I have added tests and corrected code following @nalimilan suggestions. I did not update docs as I was unsure what would be the best approach to do it (other `sample` variants do not perform checking for negative values of weights).
Maybe simply stating that in general the behavior of `sample` with negative weights is unspecified would be best?

nalimilan:
If other methods don't do the checking, I would just file an issue about this or make a PR so that all methods are consistent. Documenting an unspecified behavior doesn't sound too useful.

bkamins:
That's why I thought that not documenting it is the best option for now.

If I understand the other methods correctly, in particular `sample(wv::WeightVec)`, they do not check for negative entries, because it would have performance penalty (they use precomputed `sum` in `WeightVec` and do not always visit every element in `wv`).

And the question is do we want to pay this performance penalty (e.g. I understood that the constructor of `WeightVec` deliberately does not check for non-negativity because of performance reasons and it would be most natural to make this check in the constructor).

nalimilan:
Thanks for checking. One issue is that negative weights might make sense for some applications but not others, so checking when constructing the weights vector may not be possible. But you're right that for some functions the cost of the check might be significant; could you point me at the method which doesn't go over all individual weights?

I'm not sure what to do. I'd like to hear others' opinions about this.

bkamins:
As I have mentioned it is in `sample(wv::WeightVec)`. Here is the line:
https://github.com/JuliaStats/StatsBase.jl/blob/master/src/sampling.jl#L354
This function gets called in probably 90% of cases of usage of weighted sampling with replacement (alias method is chosen only for large tasks which are probably rare).

My opinion is that out of the options I have mentioned `WeightVec` should enforce non-negativity as this is one-time effort (no need to check it later every time `WeightVec` is used).

One option is to add an argument `nonnegative` to `WeightVec` constructor defaulting to `true`. If it is `true` then only non-negative weights are allowed. User could explicitly set this argument to `false` to allow negative weights (and documentation should warn user that most of the methods in `StatsBase` assume nonnegative weights).

What are uses of negative weights where the sum of weights is relevant? (this is what `WeightVec` effectively provides now)
The uses of negative weights I know of are in graph contexts but I would not assume that someone would use `WeightVec` there.

bkamins:
@nalimilan Some time has passed and there is no feedback. The algorithm of Efraimidis & Spirakis does not support negative weights anyway.

test/sampling.jl
```diff
@@ -149,3 +149,29 @@ check_sample_norep(a, (3, 12), 0; ordered=false)
 
 a = sample(3:12, 5; replace=false, ordered=true)
 check_sample_norep(a, (3, 12), 0; ordered=true)
+
+# test of weighted sampling without replacement
+import StatsBase: sample
```
nalimilan: This isn't needed.

bkamins:
@nalimilan fixed

nalimilan:
Thanks!

nalimilan:
BTW, if you're interested in this part of the code, it seems we forgot to mention them in the list of algorithms provided in the docs at `docs/source/sampling.rst`.

nalimilan:
Oops, we had completely forgotten about the need to find a general fix for other methods. I've found a solution which doesn't incur an unnecessary cost when constructing weights vectors if you never use them in a context that needs to check that all entries are positive. See https://github.com/JuliaStats/StatsBase.jl/pull/834.

Issue #231, opened 2017-01-24
Histogram, normalized to area 1


alanedelman:
I don't think I see an empirical pdf or a histogram normalized so that the area
is 1 ?

andreasnoack:
I don't we support this right now. See also https://github.com/JuliaStats/KernelDensity.jl

Cc: @simonbyrne.

alanedelman:
Unexcusable  :-)
Hope this can be added asap.

oschulz:
I have code ready for this - I'll put a PR together.

oschulz:
See PR #242.

oschulz:
See PR #243.

alanedelman:
thanks

andreasnoack:
Fixed by #243

avigliotti:
Hi,

I am running julia1.0 , built from source with intel mkl, tried using normalize over an histogram, but got UndefVarError, see below:

julia> versioninfo()
Julia Version 1.0.0
Commit 5d4eaca0c9* (2018-08-08 20:58 UTC)
Platform Info:
OS: Linux (x86_64-linux-gnu)
CPU: Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz
WORD_SIZE: 64
LIBM: libopenlibm
LLVM: libLLVM-6.0.0 (ORCJIT, broadwell)

julia> using StatsBase

julia> h =fit(Histogram, randn(10000), nbins=50, closed=:right);

julia> normalize(h, mode=:pdf)
ERROR: UndefVarError: normalize not defined
Stacktrace:
[1] top-level scope at none:0

although the normalize and normlalize! functions exist in

~/.julia/packages/StatsBase/NzjNi/src/hist.jl:394 and :480

any suggestion? what am I doing wrong?

oschulz:
@avigliotti , normalize is no longer in Base on Julia 1.0, but in LinearAlgebra. Your example should work if you add `using LinearAlgebra`.

avigliotti:
yes, it does, many thanks! I haven't been working with Julia since 0.4-ish, and I'm slowly keeping up.

many thanks for the quick reply!

PR #242, opened 2017-02-26
[WIP] Add Base.norm, normalize and normalize! for Histogram

oschulz:
Draft for review - still needs testing code.

Addresses #231.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.5
-Compat 0.17.0
+Compat 0.18.0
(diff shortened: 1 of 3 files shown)


oschulz:
Question: how do define the norm for histograms with potentially negative entries? While these don't occur naturally, one still encounters them from time to time the result of comparing/subtracting two histograms (e.g. signal - background). Take the abs/norm of each weight before integration? Or take abs/norm of the regular integral?

oschulz:
Replaced by #243.

Issue #241, opened 2017-02-26
Bug in Histograms with uneven bins


mkborregaard:
For histograms with uneven bins, the height of bins must be normalized so that wider bins don't get higher bars. By convention, this is done by dividing the bin heights with the bin widths, so that the area of each bin, in the units on the X axis, is equal to the number of occurrences falling into that bin (not sure how this works for multidimensional histograms).
The current method fails to do so, which is a potentially problematic bug.

mkborregaard:
Unless of course this is not a bug, but reflects that you think of StatsBase.Histogram as primarily a convenient way of counting the number of occurrences in a bin. In that case, though, I think the pitfalls with the uneven bins case should be well documented as this may cause issues downstream. E.g. for plotting packages (like Plots) that may wish to depend on Histogram for doing histogram plots. 
It might also be good to include a weigh_uneven keyword or something like that. But the correct solution here depends on your philosophy about where to split the functionality of a histogram.

oschulz:
> Unless of course this is not a bug, but reflects that you think of StatsBase.Histogram as primarily a convenient way of counting the number of occurrences in a bin

IMHO, that's exactly what it is - histograms shouldn't be normalized automatically, otherwise it would be impossible to access the counts per bin (and derive statistical uncertaincy, etc., depending on the process under study) in an analysis. Of course, when normalizing a histogram (#231), "uneven" bin widths must be taken into account.

mkborregaard:
This PR https://github.com/JuliaStats/StatsBase.jl/pull/243 provides functionality to do the normalization.

mkborregaard:
The discussion on that PR fully encompasses my concerns, so I will close this issue for clarity.

PR #243, opened 2017-02-26
Add Base.norm, normalize and normalize! for Histogram

oschulz:
Addresses #231.

Question: how do define the norm for histograms with potentially negative entries? While these don't occur naturally, one still encounters them from time to time the result of comparing/subtracting two histograms (e.g. signal - background). Take the abs/norm of each weight before integration? Or take abs/norm of the regular integral?

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -297,17 +297,22 @@ end
 
 # Normalization modes:
 #
-# * `:norm`: Normalize by norm(h). Resulting histogram has norm 1. Histograms
-#    of same data but with different binning will have different weight values
-#    at same coordinates after normalization. Operation is idempotent.
-# * `:pdf`: Normalize by sum of weights and bin sizes. Resulting histogram
-#    has norm 1 and represents a PDF. Histograms of same data but with
-#    different binning will have same weight values at same coordinates
-#    after normalization. Operation is *not* idempotent.
+# * `:norm`: Normalize by norm(h). Resulting histogram has norm 1. Operation
+#    is always idempotent. For homogeneous binning, result is the same as with
+#    mode `:pdf`. For inhomogeneous binning, histograms of same data but with
+#    different binning will have different weight values at same coordinates
+#    after normalization.
+# *  `:pdf`: Normalize by sum of weights and bin sizes. Resulting histogram
+#    has norm 1 and represents a PDF. Operation is *not* idempotent for
+#    histograms with inhomogeneous binning. For homogeneous binning, result is
+#    the same as with mode `:norm`. Histograms of same data but with different
+#    binning will always have similar weight values at same coordinates after
+#    normalization.
 # * `:density`: Normalize by bin sizes only. Resulting histogram represents
-#    count density of input and does not have norm 1. Histograms of same data
-#    but with different binning will have same weight values at same
-#    coordinates after normalization. Operation is *not* idempotent.
+#    count density of input and does not have norm 1. Operation is *never*
+#    idempotent. Histograms of same data but with different binning will
+#    always have similar weight values at same coordinates after
+#    normalization.
 #
 # aux_weights may, e.g., be estimated statistical uncertainties.
 
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -16,20 +16,6 @@ function _check_closed_arg(closed::Symbol, funcsym)
 end
 
 
-## Fast map for tuples (a lot faster than using Base.map)
-@generated function _tuple_map{N}(f, tuples::NTuple{N, Any}...)
-    M = length(typeof(tuples::NTuple).parameters)
-    result_expr = Expr(:tuple)
-    for i in 1:N
-        call_f_expr = Expr(:call,:f)
-        for j in 1:M
-            push!(call_f_expr.args, :(tuples[$j][$i]))
-        end
-        push!(result_expr.args, call_f_expr)
-    end
-    result_expr
-end
-
 ## Fast getindex function for multiple arrays, returns a tuple of array elements
 @inline Base.@propagate_inbounds @generated function _multi_getindex(i::Integer, c::AbstractArray...)
     N = length(typeof(c::Tuple).parameters)
@@ -124,9 +110,9 @@ function histrange{F}(lo::F, hi::F, n::Integer, closed::Symbol=:default_left)
 end
 
 histrange{N}(vs::NTuple{N,AbstractVector},nbins::NTuple{N,Integer},closed::Symbol) =
-    _tuple_map((v,n) -> histrange(v,n,closed),vs,nbins)
+    map((v,n) -> histrange(v,n,closed),vs,nbins)
 histrange{N}(vs::NTuple{N,AbstractVector},nbins::Integer,closed::Symbol) =
-    _tuple_map(v -> histrange(v,nbins,closed),vs)
+    map(v -> histrange(v,nbins,closed),vs)
 
 
 
@@ -148,7 +134,7 @@ type Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
                                                weights::Array{T,N}, closed::Symbol, isdensity::Bool=false)
         closed == :right || closed == :left || error("closed must :left or :right")
         isdensity && !(T <: AbstractFloat) && error("Density histogram must have float-type weights")
-        _tuple_map(x -> length(x)-1,edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
+        map(x -> length(x)-1,edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
         new{T,N,E}(edges,weights,closed,isdensity)
     end
 end
@@ -157,7 +143,7 @@ Histogram{T,N}(edges::NTuple{N,AbstractVector},weights::AbstractArray{T,N},close
     Histogram{T,N,typeof(edges)}(edges,weights,_check_closed_arg(closed,:Histogram),isdensity)
 
 Histogram{T,N}(edges::NTuple{N,AbstractVector},::Type{T},closed::Symbol=:default_left, isdensity::Bool=false) =
-    Histogram(edges,zeros(T,_tuple_map(x -> length(x)-1,edges)...),_check_closed_arg(closed,:Histogram),isdensity)
+    Histogram(edges,zeros(T,map(x -> length(x)-1,edges)...),_check_closed_arg(closed,:Histogram),isdensity)
 
 Histogram{N}(edges::NTuple{N,AbstractVector},closed::Symbol=:default_left, isdensity::Bool=false) =
     Histogram(edges,Int,_check_closed_arg(closed,:Histogram),isdensity)
@@ -178,7 +164,7 @@ end
 binindex{T,E}(h::AbstractHistogram{T,1,E}, x::Real) = binindex(h, (x,))[1]
 
 binindex{T,N,E}(h::Histogram{T,N,E}, xs::NTuple{N,Real}) =
-    _tuple_map((edge, x) -> _edge_binindex(edge, h.closed, x), h.edges, xs)
+    map((edge, x) -> _edge_binindex(edge, h.closed, x), h.edges, xs)
 
 @inline function _edge_binindex(edge::AbstractVector, closed::Symbol, x::Real)
     if closed == :right
@@ -192,7 +178,7 @@ end
 binvolume{T,E}(h::AbstractHistogram{T,1,E}, binidx::Integer) = binvolume(h, (binidx,))
 
 binvolume{T,N,E}(h::Histogram{T,N,E}, binidx::NTuple{N,Integer}) =
-    prod(_tuple_map((edge, i) -> _edge_binvolume(edge, i), h.edges, binidx))
+    prod(map((edge, i) -> _edge_binvolume(edge, i), h.edges, binidx))
 
 @inline _edge_binvolume(edge::AbstractVector, i::Integer) = edge[i+1] - edge[i]
 @inline _edge_binvolume(edge::Range, i::Integer) = step(edge)
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -16,6 +16,31 @@ function _check_closed_arg(closed::Symbol, funcsym)
 end
 
 
+## Fast map for tuples (a lot faster than using Base.map)
+@generated function _tuple_map{N}(f, tuples::NTuple{N, Any}...)
+    M = length(typeof(tuples::NTuple).parameters)
+    result_expr = Expr(:tuple)
+    for i in 1:N
+        call_f_expr = Expr(:call,:f)
+        for j in 1:M
+            push!(call_f_expr.args, :(tuples[$j][$i]))
+        end
+        push!(result_expr.args, call_f_expr)
+    end
+    result_expr
+end
+
+## Fast getindex function for multiple arrays, returns a tuple of array elements
+@inline Base.@propagate_inbounds @generated function _multi_getindex(i::Integer, c::AbstractArray...)
+    N = length(typeof(c::Tuple).parameters)
+    result_expr = Expr(:tuple)
+    for j in 1:N
+        push!(result_expr.args, :(c[$j][i]))
+    end
+    result_expr
+end
+
+
 ## nice-valued ranges for histograms
 function histrange{T}(v::AbstractArray{T}, n::Integer, closed::Symbol=:default_left)
     closed = _check_closed_arg(closed,:histrange)
@@ -99,9 +124,9 @@ function histrange{F}(lo::F, hi::F, n::Integer, closed::Symbol=:default_left)
 end
 
 histrange{N}(vs::NTuple{N,AbstractVector},nbins::NTuple{N,Integer},closed::Symbol) =
-    map((v,n) -> histrange(v,n,closed),vs,nbins)
+    _tuple_map((v,n) -> histrange(v,n,closed),vs,nbins)
 histrange{N}(vs::NTuple{N,AbstractVector},nbins::Integer,closed::Symbol) =
-    map(v -> histrange(v,nbins,closed),vs)
+    _tuple_map(v -> histrange(v,nbins,closed),vs)
 
 
 
@@ -118,22 +143,24 @@ type Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
     edges::E
     weights::Array{T,N}
     closed::Symbol
+    isdensity::Bool
     function (::Type{Histogram{T,N,E}}){T,N,E}(edges::NTuple{N,AbstractArray},
-                                               weights::Array{T,N}, closed::Symbol)
+                                               weights::Array{T,N}, closed::Symbol, isdensity::Bool=false)
         closed == :right || closed == :left || error("closed must :left or :right")
-        map(x -> length(x)-1,edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
-        new{T,N,E}(edges,weights,closed)
+        isdensity && !(T <: AbstractFloat) && error("Density histogram must have float-type weights")
+        _tuple_map(x -> length(x)-1,edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
+        new{T,N,E}(edges,weights,closed,isdensity)
     end
 end
 
-Histogram{T,N}(edges::NTuple{N,AbstractVector},weights::AbstractArray{T,N},closed::Symbol=:default_left) =
-    Histogram{T,N,typeof(edges)}(edges,weights,_check_closed_arg(closed,:Histogram))
+Histogram{T,N}(edges::NTuple{N,AbstractVector},weights::AbstractArray{T,N},closed::Symbol=:default_left, isdensity::Bool=false) =
+    Histogram{T,N,typeof(edges)}(edges,weights,_check_closed_arg(closed,:Histogram),isdensity)
 
-Histogram{T,N}(edges::NTuple{N,AbstractVector},::Type{T},closed::Symbol=:default_left) =
-    Histogram(edges,zeros(T,map(x -> length(x)-1,edges)...),_check_closed_arg(closed,:Histogram))
+Histogram{T,N}(edges::NTuple{N,AbstractVector},::Type{T},closed::Symbol=:default_left, isdensity::Bool=false) =
+    Histogram(edges,zeros(T,_tuple_map(x -> length(x)-1,edges)...),_check_closed_arg(closed,:Histogram),isdensity)
 
-Histogram{N}(edges::NTuple{N,AbstractVector},closed::Symbol=:default_left) =
-    Histogram(edges,Int,_check_closed_arg(closed,:Histogram))
+Histogram{N}(edges::NTuple{N,AbstractVector},closed::Symbol=:default_left, isdensity::Bool=false) =
+    Histogram(edges,Int,_check_closed_arg(closed,:Histogram),isdensity)
 
 function show(io::IO, h::AbstractHistogram)
     println(io, typeof(h))
@@ -147,98 +174,117 @@ end
 
 (==)(h1::Histogram,h2::Histogram) = (==)(h1.edges,h2.edges) && (==)(h1.weights,h2.weights) && (==)(h1.closed,h2.closed)
 
-# 1-dimensional
-Histogram{T}(edge::AbstractVector, weights::AbstractVector{T}, closed::Symbol=:default_left) =
-    Histogram((edge,), weights, _check_closed_arg(closed,:Histogram))
 
-Histogram{T}(edge::AbstractVector, ::Type{T}, closed::Symbol=:default_left) =
-    Histogram(edge, zeros(T,length(edge)-1), _check_closed_arg(closed,:Histogram))
+binindex{T,E}(h::AbstractHistogram{T,1,E}, x::Real) = binindex(h, (x,))[1]
 
-Histogram(edge::AbstractVector,closed::Symbol=:default_left) =
-    Histogram(edge, Int, _check_closed_arg(closed,:Histogram))
+binindex{T,N,E}(h::Histogram{T,N,E}, xs::NTuple{N,Real}) =
+    _tuple_map((edge, x) -> _edge_binindex(edge, h.closed, x), h.edges, xs)
 
-function push!{T,E}(h::Histogram{T,1,E}, x::Real,w::Real)
-    i = if h.closed == :right
-        searchsortedfirst(h.edges[1], x) - 1
+@inline function _edge_binindex(edge::AbstractVector, closed::Symbol, x::Real)
+    if closed == :right
+        searchsortedfirst(edge, x) - 1
     else
-        searchsortedlast(h.edges[1], x)
+        searchsortedlast(edge, x)
     end
-    if 1 <= i <= length(h.weights)
-        @inbounds h.weights[i] += w
-    end
-    h
 end
+
+
+binvolume{T,E}(h::AbstractHistogram{T,1,E}, binidx::Integer) = binvolume(h, (binidx,))
+
+binvolume{T,N,E}(h::Histogram{T,N,E}, binidx::NTuple{N,Integer}) =
+    prod(_tuple_map((edge, i) -> _edge_binvolume(edge, i), h.edges, binidx))
+
+@inline _edge_binvolume(edge::AbstractVector, i::Integer) = edge[i+1] - edge[i]
+@inline _edge_binvolume(edge::Range, i::Integer) = step(edge)
+
+
+# 1-dimensional
+
+Histogram{T}(edge::AbstractVector, weights::AbstractVector{T}, closed::Symbol=:default_left, isdensity::Bool=false) =
+    Histogram((edge,), weights, closed, isdensity)
+
+Histogram{T}(edge::AbstractVector, ::Type{T}, closed::Symbol=:default_left, isdensity::Bool=false) =
+    Histogram((edge,), T, closed, isdensity)
+
+Histogram(edge::AbstractVector, closed::Symbol=:default_left, isdensity::Bool=false) =
+    Histogram((edge,), closed, isdensity)
+
+
+push!{T,E}(h::AbstractHistogram{T,1,E}, x::Real, w::Real) = push!(h, (x,), w)
 push!{T,E}(h::AbstractHistogram{T,1,E}, x::Real) = push!(h,x,one(T))
+append!{T}(h::AbstractHistogram{T,1}, v::AbstractVector) = append!(h, (v,))
+append!{T}(h::AbstractHistogram{T,1}, v::AbstractVector, wv::Union{AbstractVector,WeightVec}) = append!(h, (v,), wv)
 
-function append!{T}(h::AbstractHistogram{T,1}, v::AbstractVector)
-    for x in v
-        push!(h,x)
-    end
-    h
-end
-function append!{T}(h::AbstractHistogram{T,1}, v::AbstractVector,wv::WeightVec)
-    for (x,w) in zip(v,wv.values)
-        push!(h,x,w)
-    end
-    h
-end
 
-fit(::Type{Histogram},v::AbstractVector, edg::AbstractVector; closed::Symbol=:default_left) =
-    append!(Histogram(edg,_check_closed_arg(closed,:fit)), v)
-fit(::Type{Histogram},v::AbstractVector; closed::Symbol=:default_left, nbins=sturges(length(v))) = begin
-    closed = _check_closed_arg(closed,:fit)
-    fit(Histogram, v, histrange(v,nbins,closed); closed=closed)
-end
+fit{T}(::Type{Histogram{T}},v::AbstractVector, edg::AbstractVector; closed::Symbol=:default_left) =
+    fit(Histogram{T},(v,), (edg,), closed=closed)
+fit{T}(::Type{Histogram{T}},v::AbstractVector; closed::Symbol=:default_left, nbins=sturges(length(v))) =
+    fit(Histogram{T},(v,); closed=closed, nbins=nbins)
+fit{T}(::Type{Histogram{T}},v::AbstractVector, wv::WeightVec, edg::AbstractVector; closed::Symbol=:default_left) =
+    fit(Histogram{T},(v,), wv, (edg,), closed=closed)
+fit{T}(::Type{Histogram{T}},v::AbstractVector, wv::WeightVec; closed::Symbol=:default_left, nbins=sturges(length(v))) =
+    fit(Histogram{T}, (v,), wv; closed=closed, nbins=nbins)
 
-fit{W}(::Type{Histogram},v::AbstractVector, wv::WeightVec{W}, edg::AbstractVector; closed::Symbol=:default_left) =
-    append!(Histogram(edg,W,_check_closed_arg(closed,:fit)), v, wv)
-fit(::Type{Histogram},v::AbstractVector, wv::WeightVec; closed::Symbol=:default_left, nbins=sturges(length(v))) = begin
-    closed = _check_closed_arg(closed,:fit)
-    fit(Histogram, v, wv, histrange(v,nbins,closed); closed=closed)
-end
+fit{W}(::Type{Histogram}, v::AbstractVector, wv::WeightVec{W}, args...; kwargs...) = fit(Histogram{W}, v, wv, args...; kwargs...)
+
 
 # N-dimensional
+
 function push!{T,N}(h::Histogram{T,N},xs::NTuple{N,Real},w::Real)
-    is = if h.closed == :right
-        map((edge, x) -> searchsortedfirst(edge,x) - 1, h.edges, xs)
-    else
-        map(searchsortedlast, h.edges, xs)
-    end
-    try
-        h.weights[is...] += w
-    catch e
-        isa(e,BoundsError) || rethrow(e)
+    (h.isdensity == true) && error("Density histogram must have float-type weights")
+    idx = binindex(h, xs)
+    if checkbounds(Bool, h.weights, idx...)
+        @inbounds h.weights[idx...] += w
     end
     h
 end
+
+function push!{T<:AbstractFloat,N}(h::Histogram{T,N},xs::NTuple{N,Real},w::Real)
+    idx = binindex(h, xs)
+    if checkbounds(Bool, h.weights, idx...)
+        @inbounds h.weights[idx...] += h.isdensity ? w / binvolume(h, idx) : w
+    end
+    h
+end
+
 push!{T,N}(h::AbstractHistogram{T,N},xs::NTuple{N,Real}) = push!(h,xs,one(T))
 
+
 function append!{T,N}(h::AbstractHistogram{T,N}, vs::NTuple{N,AbstractVector})
-    for xs in zip(vs...)
-        push!(h,xs)
+    @inbounds for i in eachindex(vs...)
+        xs = _multi_getindex(i, vs...)
+        push!(h, xs, one(T))
     end
     h
 end
-function append!{T,N}(h::AbstractHistogram{T,N}, vs::NTuple{N,AbstractVector},wv::WeightVec)
-    for (xs,w) in zip(zip(vs...),wv.values)
-        push!(h,xs,w)
+function append!{T,N}(h::AbstractHistogram{T,N}, vs::NTuple{N,AbstractVector}, wv::AbstractVector)
+    @inbounds for i in eachindex(wv, vs...)
+        xs = _multi_getindex(i, vs...)
+        push!(h, xs, wv[i])
     end
     h
 end
+append!{T,N}(h::AbstractHistogram{T,N}, vs::NTuple{N,AbstractVector}, wv::WeightVec) = append!(h, vs, values(wv))
 
-fit{N}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, edges::NTuple{N,AbstractVector}; closed::Symbol=:default_left) =
-    append!(Histogram(edges,_check_closed_arg(closed,:fit)), vs)
-fit{N}(::Type{Histogram}, vs::NTuple{N,AbstractVector}; closed::Symbol=:default_left, nbins=sturges(length(vs[1]))) = begin
+
+fit{T,N}(::Type{Histogram{T}}, vs::NTuple{N,AbstractVector}, edges::NTuple{N,AbstractVector}; closed::Symbol=:default_left) =
+    append!(Histogram(edges, T, _check_closed_arg(closed,:fit), false), vs)
+
+fit{T,N}(::Type{Histogram{T}}, vs::NTuple{N,AbstractVector}; closed::Symbol=:default_left, isdensity::Bool=false, nbins=sturges(length(vs[1]))) = begin
     closed = _check_closed_arg(closed,:fit)
-    fit(Histogram, vs, histrange(vs,nbins,closed); closed=closed)
+    fit(Histogram{T}, vs, histrange(vs,nbins,closed); closed=closed)
 end
 
-fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edges::NTuple{N,AbstractVector}; closed::Symbol=:default_left) =
-    append!(Histogram(edges,W,_check_closed_arg(closed,:fit)), vs, wv)
-fit{N}(::Type{Histogram},vs::NTuple{N,AbstractVector}, wv::WeightVec; closed::Symbol=:default_left, nbins=sturges(length(vs[1]))) = begin
+fit{T,N,W}(::Type{Histogram{T}}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edges::NTuple{N,AbstractVector}; closed::Symbol=:default_left) =
+    append!(Histogram(edges, T, _check_closed_arg(closed,:fit), false), vs, wv)
+
+fit{T,N}(::Type{Histogram{T}}, vs::NTuple{N,AbstractVector}, wv::WeightVec; closed::Symbol=:default_left, isdensity::Bool=false, nbins=sturges(length(vs[1]))) = begin
     closed = _check_closed_arg(closed,:fit)
-    fit(Histogram, vs, wv, histrange(vs,nbins,closed); closed=closed)
+    fit(Histogram{T}, vs, wv, histrange(vs,nbins,closed); closed=closed)
 end
+
+fit(::Type{Histogram}, args...; kwargs...) = fit(Histogram{Int}, args...; kwargs...)
+fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, args...; kwargs...) = fit(Histogram{W}, vs, wv, args...; kwargs...)
 
 
 # Get a suitable high-precision type for the norm of a histogram.
@@ -302,12 +348,12 @@ end
 
 
 """
-    normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+    normalize!{T<:AbstractFloat, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :pdf)
 
 Normalize the histogram `h` and optionally scale one or more auxiliary weight
 arrays appropriately. See description of `normalize` for details. Returns `h`.
 """
-@generated function normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+@generated function normalize!{T<:AbstractFloat, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :pdf)
     quote
         edges = h.edges
         weights = h.weights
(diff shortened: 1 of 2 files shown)


oschulz:
Something seems to be wrong still with the current implementation (for hists with inhomogeneous binning).

oschulz:
Example - same data, different binning (h1: small homogeneous bins, h2: large inhomogeneous bins):

hists-norm-modes

mkborregaard:
Dhe :density mode would also add the functionality to deal with https://github.com/JuliaStats/StatsBase.jl/issues/241 , though it would not solve the problem that end users may not realize that uneven-binned histograms are not density-normalized by default (as is convention when plotting).

oschulz:
> The :density mode would also add the functionality to deal with #241, though it would not solve the problem that end users may not realize that uneven-binned histograms are not density-normalized

I wouldn't call this a problem - different users may want different things from inhomogeneously binned histograms. When doing statistical analysis, it's very important that the bin counts are not modified by some automatic default mechanism. I any case, users that create inhomogeneously binned histograms (after all, one has to purposefully specifcy inhomogeneous edge vectors) will usually know what they want from them - and when and in which mode they need to normalize.

> What would the :norm method be used for?

This is the mathematically "clean" mode - it ensures that normalize is an idempotent operation (and therefore, it's the default), which the typical user will expect. Normalizing something that is already normalized shouldn't change anything:

    normalize(normalize(h)).weights ≈ normalize(h).weights

Also:

    normalize(normalize(h, mode = :density)).weights ≈ normalize(h, mode = :pdf).weights

So you can use mode `:norm` to get from a `:density`-normalized histogram to a `:pdf`-normalized one.

oschulz:
Testing code added - should be good to merge now.

mkborregaard:
Whether the operation is idempotent depend on the input, right? So `:pdf` is idempotent when applied to a `:density`-normalized histogram, isn't it? 
I am asking because this relates to my view expressed in #241 that density-normalized histograms are the normal when bins are uneven (I think others share this view: see eg https://en.wikipedia.org/wiki/Histogram). To conform to that, a more satisfactory resolution to #241 would have :density the default for histograms with uneven bins (this would be breaking), with the non-density normalized possible as an option for people intending to use them for further statistical analysis, and `:pdf` the default normalization. An added argument for this would be that `:norm`, the current default, has AFAICS no practical use.
I am sure we will not agree about this, so it'd be interesting to see what the StatsBase crew decides :-)

(Other than that I think this PR looks excellent and should be merged sooner rather than later).

oschulz:
> Whether the operation is idempotent depends on the input, right? So :pdf is idempotent when applied to a :density-normalized histogram, isn't it?

If the histogram has inhomogeneous bins, no, it is *not* idempotent.

> I am asking because this relates to my view expressed in #241 that density-normalized histograms are the normal

IMHO there is no "normal", when it comes to inhomogeneously binned histograms - different people will need different things. People from my field (and likely also others doing statistical analysis on histograms) would be very unhappy if histograms counts are modified in any (or hard to predict) way without explicit user request. Not only is it unexpected, it's also not reversible - so information is lost.

> have :density the default for histograms with uneven bins 

I don't think making the default edge-dependent is a good idea. It would mean that that depending on how your edges are defined, your histogram will be treated with different algorithms. This would come as a very rude surprise to most users (esp. from mathematical fields), I would expect. Also, density is the *least* desired default, because the implied meaning of "normalize" is "make it have norm 1". `:density` it's a special mode that normalize only in respect to binning - and will often be used, but it's definitely not what a user will expect to happen if they simply call `normalize(h)`.

> `:norm` the current default, has AFAICS no practical use.

Again, that may depend very much on what you use histograms for. The `:norm` mode does something like `h / norm(h)`, which is the standard mathematical definition of normalization.

mkborregaard:
> density is the least desired default

I mean density-normalized should be the default for Histogram, pdf the default for `normalize`. Normalize should be a keyword argument to fit (Histogram).

> it's also not reversible - so information is lost

My suggestion was to make it possible to pass e.g. :raw to `fit(Histogram, x, normalize = :raw)` if you needed it.

mkborregaard:
> If the histogram has inhomogeneous bins, no, it is not idempotent.

But shouldn't it be? This seems like an implementation detail. The :pdf normalization should not density-normalize bins that have already been density-normalized once.

oschulz:
>> If the histogram has inhomogeneous bins, no, it is not idempotent.
> But shouldn't it be? This seems like an implementation detail. The :pdf normalization should not density-normalize bins that have already been density-normalized once.

How would the :pdf normalization know what has been done to the histogram before? The mathematical operation is does is - for inhomogeneously binned histograms - not idempotent. This kind of histogram is simply a bit more tricky than a homogeneously binned one - there is no easy "standard" here.

mkborregaard:
I will concede that there is certainly more than one valid way of doing this :-) Your approach, of returning raw numbers, may well be best, the important thing is that this is thoroughly documented.
I did a quick mini-review on what other statistical languages do:
R: hist
Uses raw counts for equidistant bins, and your `:pdf` normalization for uneven bins, by default. Does not have your `:density` or `:norm`.

Matlab: histogram
Uses raw counts for both equidistant and uneven bins by default, but has a section in the documentation stating that for uneven bins, your `:density` normalization should be applied. Also offers the `:pdf` normalization, but not your `:norm`.

Python, numpy.histogram
Uses raw counts for both equidistant and uneven bins by default, but has two keyword arguments intended for use with uneven bins, `normed` and `density`. Setting `normed = True` is (I think) equal to your :norm, but the documentation states "Note that this latter behavior is known to be buggy with unequal bin widths" and they have deprecated the keyword. Instead they recommend `density` which is (I think) equal to your `:pdf`.

OK, I have nothing more to comment on this PR, I think it is great.

src/hist.jl
```diff
@@ -213,3 +215,125 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+            end
+        )
+        norm(s_0)
+    end
+end
+
+
+function _float_deepcopy(x)
+    y = float(x)
+    is(y, x)? deepcopy(y) : y
```
ararslan: `is` is deprecated; this should be `x === y` instead.

src/hist.jl
```diff
@@ -213,3 +215,125 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+end
+
+
+function float{T, N, E}(h::Histogram{T, N, E})
+    float_weights = float(h.weights)
+    if is(float_weights, h.weights)
```
ararslan: `float_weights === h.weights`

src/hist.jl
```diff
@@ -213,3 +215,125 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+# aux_weights may, e.g., be estimated statistical uncertainties.
+
+@generated function normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+    quote
+        edges = h.edges
+        weights = h.weights
+
+        for A in aux_weights
+            (size(A) != size(weights)) && error("aux_weights must have same size as histogram weights")
+        end
```
ararslan: `any(A -> size(A) != size(weights), aux_weights) && error(...)`

src/hist.jl
```diff
@@ -213,3 +215,125 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+            @inbounds @nloops $N i weights d->(vs_{$N-d+1} = vs_{$N-d} * (edges[d][i_d + 1] - edges[d][i_d])) begin
+                (@nref $N weights i) /= $(Symbol("vs_$N"))
+                for A in aux_weights
+                    (@nref $N A i) /= $(Symbol("vs_$N"))
+                end
+            end
+        elseif mode == :none
+            # Do nothing.
+        else
+            error("mode must be :norm, :pdf or :density")
```
ararslan: Since `mode` is a function argument, this should be `throw(ArgumentError(...))`

test/hist.jl
```diff
@@ -72,3 +72,49 @@ show_h = sprint(show, fit(Histogram,[1,2,3]))
...
+        rand(rng, 1:maximum(edges[2]), n),
+        maximum(edges[3]) * rand(rng, n)
+    )
+
+    h = fit(Histogram, data, edges, closed = :left)
+
+    weight_sum = sum(h.weights)
+    bin_vols = [ x * y * z for x in diff(edges[1]), y in diff(edges[2]), z in diff(edges[3])]
+
+    normalize(h)
```
ararslan: What is this line testing? Did you mean `normalize!(h)`?

src/hist.jl
```diff
@@ -213,3 +215,125 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+#    normalization.
+# * `:density`: Normalize by bin sizes only. Resulting histogram represents
+#    count density of input and does not have norm 1. Operation is *never*
+#    idempotent. Histograms of same data but with different binning will
+#    always have similar weight values at same coordinates after
+#    normalization.
+# *  `:none`: Leaves histogram unchanged. Useful to simplify code that has to
+#    conditionally apply different modes of normalization.
+#
+# aux_weights may, e.g., be estimated statistical uncertainties.
```
ararslan: This should be part of a docstring for the function instead of a comment

src/hist.jl
oschulz: Thanks!

src/hist.jl
oschulz: Thanks!

src/hist.jl
oschulz: Good point.

src/hist.jl
oschulz: Yep.

test/hist.jl
oschulz: Oops, good catch - some remnant dead code. Removed.

src/hist.jl
oschulz: Yes, I had kinda planned this, but wasn't sure because StatsBase isn't exactly rich in docstrings. :-)
Added now.

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+norm_type{T<:Integer}(::Type{T}) = promote_type(T, Int64)
+norm_type{T<:AbstractFloat}(::Type{T}) = promote_type(T, Float64)
+
+
+"""
+    norm(h::Histogram)
+
+Calculate the norm of histogram `h` as the norm-/abs-value of it's integral.
+"""
+
```
ararslan: Will have to remove this line for the docstring to be applied correctly (IIRC)

oschulz:
Thanks for the review, @ararslan! I've pushed two commits that (hopefully) address your remarks.

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+
+
+"""
+    normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+
+Normalize the histogram h and one or more auxiliary weight arrays. See
+description of `normalize` for details. Returns `h`.
+"""
+
```
ararslan: Same here

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+end
+
+norm_type{T<:Integer}(::Type{T}) = promote_type(T, Int64)
+norm_type{T<:AbstractFloat}(::Type{T}) = promote_type(T, Float64)
+
+
+"""
+    norm(h::Histogram)
+
+Calculate the norm of histogram `h` as the norm-/abs-value of it's integral.
```
ararslan: No apostrophe in "its." Also I think just saying "absolute value" would be clearer than "norm-/abs-value."

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+        Histogram{eltype(float_weights), N, E}(deepcopy(h.edges), float_weights, h.closed)
+    end
+end
+
+
+
+"""
+    normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+
+Normalize the histogram h and one or more auxiliary weight arrays. See
```
ararslan: `h` should be in backticks. Are the auxiliary arrays modified too? If not, I'd say "using one or more" rather than "and one or more."

src/hist.jl
oschulz: Good point.

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+        end
+        h
+    end
+end
+
+
+"""
+    normalize{T, N, E}(h::Histogram{T, N, E}; mode::Symbol = :norm)
+
+Normalize the histogram h.
```
ararslan: `h` in backticks

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+
+Valid values for `mode` are:
+
+* `:norm`: Normalize by dividing weights by norm(h). Resulting histogram has
```
ararslan:
```
`norm(h)`
```

src/hist.jl
oschulz: How about now?

src/hist.jl
oschulz: Arg, sorry.

src/hist.jl
oschulz: Thanks!

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+norm_type{T<:Integer}(::Type{T}) = promote_type(T, Int64)
+norm_type{T<:AbstractFloat}(::Type{T}) = promote_type(T, Float64)
+
+
+"""
+    norm(h::Histogram)
+
+Calculate the norm of histogram `h` as the absolute value of its integral.
+"""
+
```
ararslan: Still need to remove the blank line

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+
+
+"""
+    normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+
+Normalize the histogram `h` and optionally scales or more auxiliary weight
+arrays appropriately. See description of `normalize` for details. Returns `h`.
+"""
+
```
ararslan: This one too

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+        Histogram{eltype(float_weights), N, E}(deepcopy(h.edges), float_weights, h.closed)
+    end
+end
+
+
+
+"""
+    normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+
+Normalize the histogram `h` and optionally scales or more auxiliary weight
```
ararslan: Should be "scale" instead of "scales"

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+   normalization.
+* `:density`: Normalize by bin sizes only. Resulting histogram represents
+   count density of input and does not have norm 1. Operation is *never*
+   idempotent. Histograms of same data but with different binning will
+   always have similar weight values at same coordinates after
+   normalization.
+*  `:none`: Leaves histogram unchanged. Useful to simplify code that has to
+   conditionally apply different modes of normalization.
+"""
+
```
ararslan: here too

src/hist.jl
```diff
@@ -213,3 +215,154 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+"""
+    normalize{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+
+Normalize the histogram h and rescales or more auxiliary weight arrays at
+the same time (`aux_weights` may, e.g., contain estimated statistical
+uncertainties). The values of the auxiliary arrays are scaled by the same
+factor as the corresponding histogram weight values. Returns a tuple of the
+normalized histogram and scaled auxiliary weights.
+"""
+
```
ararslan: and here

src/hist.jl
oschulz: Ah, sorry, now I get it. :-)

src/hist.jl
oschulz: Done.

src/hist.jl
```diff
@@ -213,3 +215,150 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+            end
+        )
+        norm(s_0)
+    end
+end
+
+
+function _float_deepcopy(x)
+    y = float(x)
+    y === x? deepcopy(y) : y
```
ararslan: Does `float` ever actually alias? That is, if `eltype(x) <: AbstractFloat`, will `float(x) === x`?

src/hist.jl
```diff
@@ -213,3 +215,150 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+        Histogram{eltype(float_weights), N, E}(deepcopy(h.edges), float_weights, h.closed)
+    end
+end
+
+
+
+"""
+    normalize!{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+
+Normalize the histogram `h` and optionally scale or more auxiliary weight
```
ararslan: I assume that should be "scale _one_ or more" rather than "scale or more"?

src/hist.jl
oschulz: You find all of them, don't you! :-)

Thanks for the help, this'll be nicely polished in the end.

src/hist.jl
```diff
@@ -213,3 +215,150 @@ fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::WeightVec{W}, edge
...
+   conditionally apply different modes of normalization.
+"""
+normalize{T, N, E}(h::Histogram{T, N, E}; mode::Symbol = :norm) =
+    normalize!(_float_deepcopy(h), mode = mode)
+
+
+"""
+    normalize{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :norm)
+
+Normalize the histogram h and rescales or more auxiliary weight arrays at
```
ararslan: `h` in backticks, "rescales," "one or more"

src/hist.jl
oschulz: Yes, unfortunately is does:

```julia
A = rand(10000)
float(A) === A
```

ararslan:
Thanks for being so responsive with the feedback, and of course thanks a lot for the contribution. It would be great if you could also test `normalize!`, including ensuring that the input has been modified.

src/hist.jl
oschulz: Darn!

oschulz:
Thanks for the responsive review - this is great!

> It would be great if you could also test normalize!

You mean a full test of all modes, etc., like for `normalize`, or just to ensure it modifies the input (modes, etc. should be well covered via `normalize`).

I also should add tests for the aux arrays.

src/hist.jl
ararslan: Huh, good to know. I feel like there's a way to do exactly this using a Base function but I'm having difficult remembering what it is at the moment.

ararslan:
Since `normalize` calls `normalize!` and you're testing the modes in `normalize`, I think it's fine to simply test that `normalize!` does what you expect it to in the basic circumstance.

> I also should add tests for the aux arrays.

That'd be awesome 👍

src/hist.jl
ararslan: Aha. `map(float, x)` or `float.(x)` both return copies, so I'd go with one of those. That makes this function unnecessary.

oschulz:
Tests added - the ones for normalize with aux weights are actually a bit tricky, numerically.

src/hist.jl
oschulz: Won't save any code though, since I also need `_float_deepcopy` for `Base.float(h::Histogram)` (this should come in handy at times) - which I defined not to copy if not necessary, to be consistent with the behavior of float for arrays.

Also, float(A) may have higher performance than float.(A) - not sure.

ararslan: The inconsistent aliasing of `float` is either a bug or a feature depending on who you ask. (I tend to consider it a bug.) I think in this case we should have `float(::Histogram)` always return a copy.

ararslan: We should seek a second opinion though... @nalimilan, @andreasnoack?

oschulz: > I think in this case we should have float(::Histogram) always return a copy

I don't have a strong opinion on this - I was actually surprised that `float` aliases for arrays, myself.

oschulz: I guess the question is whether this behavior of `float` should be considered a Julia standard or not. In general, I prefer functions that either always copy data or never - much safer to use.

mkborregaard:
Hi, what's the status on this PR? @oschulz has a PR on Plots that would add StatsBase as a dependency and greatly improve the histogram capability in Plots - but it depends on this PR :-)

ararslan:
I'm waiting on feedback from another maintainer on how to approach the `float` aliasing.

Re-ping: @nalimilan, @andreasnoack

andreasnoack:
Last week was busy. I'll try to read through the issue this week.

mkborregaard:
@andreasnoack sorry for filling the 'conversation' with a long discussion of one aspect (the best default normalization). You can probably ignore this, especially if you're busy, and focus on the actual PR :-)

oschulz:
Hey, can I just ask - what's the status on this, concerning merging?

andreasnoack:
I promised to review. I'll do it now.

oschulz:
Thanks, @andreasnoack!

oschulz:
@ararslan there was a problem with a misplaced `end` in the merged - I've rebased, should be fine now.

ararslan:
Oh, sorry about that and thanks for fixing it! I tried to resolve the conflicts using the GitHub web UI and clearly I was not sufficiently diligent.

oschulz:
> @ararslan I was not sufficiently diligent

Nope - exactly the same thing happened to me when I tried the merge. :-) Had to do a one-by-one cherry-pick with manual corrections. Weird case.

ararslan:
> Part of me wants the thing returned to be something different (a new type) from the raw counts such that you know the difference.

👍

oschulz:
> Part of me wants the thing returned to be something different (a new type) from the raw counts such that you know the difference.

I don't think it should be a different type. Mathematically, a normalized histogram and a histogram created with weights that make it normalized from the beginning are the same thing, for example. Also, having a would mean that we need a type would have several drawbacks:

* There couldn't be a memory-effecient `normalize!`. If you normalize a vector, it's still a vector, right? Why treat histograms differently?

* We'd need a whole type hierarchy with `AbstractHistogram`, `NormalizedHistogram`, `PDFHistogram`, ... - I feel this would make both StatsBase and user code unnecessarily complex.

* There are other (application-specific) count-transformations that the user  may apply. Consider the (for me quite typical) example of a radiation spectrum. The original histogram would contains the number of counts in energy/wavelength bins. Then you might use a `:density` normalization to get counts per wavelength on the y-axis. Then you'd divide the y-axis by measurement time, to get something comparable to other measurements. Having a different type for each of these doesn't make sense to me.

Essentially, you would create different histogram types to represent different units on the y-axis. I don't think that makes sense. What we could do, would be to add generic parameters to add unit information for the axes (e.g. based on Unitful.jl) - but it would result in a lot more type parameters and get complicated.

oschulz:
> E.g. when a Histogram is passed to a plotting package, it would have to either :pdf or :density normalized it to be meaningful

In my experience, that's not the case: Quite often, you actually do want to plot raw counts. Normalization is not always applied before plotting (except when you have inhomogeneous binning, then plotting raw counts is only useful for cross-checks).

src/hist.jl
```diff
@@ -237,3 +239,150 @@ fit{N}(::Type{Histogram},vs::NTuple{N,AbstractVector}, wv::WeightVec; closed::Sy
...
+                s_{$N-d+1} = zero(S)
+            end,
+            d -> begin
+                s_{$N-d} += s_{$N-d+1}
+            end,
+            begin
+                $(Symbol("s_$(N)")) += (@nref $N weights i) * $(Symbol("v_$N"))
+            end
+        )
+        norm(s_0)
```
oschulz: So that the result is always positive - see my discussion at the top of the thread on the case of histograms with negative bin entries. I have to admit I'm not really clear on how this should be handled myself. We could also remove the `norm` here.

src/hist.jl
```diff
@@ -237,3 +239,150 @@ fit{N}(::Type{Histogram},vs::NTuple{N,AbstractVector}, wv::WeightVec; closed::Sy
...
+end
+
+
+function _float_deepcopy(x)
+    y = float(x)
+    y === x? deepcopy(y) : y
+end
+
+
+function float{T, N, E}(h::Histogram{T, N, E})
```
oschulz: I think it it should either

* always copy everything (as suggested by @ararslan)
* Make a full copy, if necessary, or no copy at all if weights are already float

Making a partial copy would be even more confusing, and would make it very hard to write thread-safe code (when multi-threading in Julia becomes more common in the future).

oschulz: Maybe we could do

    float(h::Histogram; always_copy::Bool = false)

or so? Default could also be `always_copy::Bool = true`.

oschulz:
>> Part of me wants the thing returned to be something different (a new type) from the raw counts such that you know the difference.
> What we could do, would be to add generic parameters to add unit information

I thought about this a bit more: I really, *really* don't think StatsBase is the place for this. StatsBase shouldn't care about units, and `normalize` should not return a different type.

What someone might want to do, is e.g. create a wrapper type `UnitfulHistogram` that combines a `Histogram` with a tuple of Uniful.js units. Then one could define `normalize` on this in a way that automatically returns a new `UnitfulHistogram` with correctly modified units. Due to the nature of Uniful.js units, it would return a different type, and a plotting package could even add correct axis labels. That would be a cool thing, for many applications - but again, I don't think StatsBase is the place for it (Base.Array doesn't have units either, right?).

This is something for more specialized (or even application-specific) packages - but if `normalize(::StatsBase.Histogram)` itself were to return a different type, such solutions would be harder to implement, not easier.

src/hist.jl
mkborregaard: IMO the difference among two histograms is not a histogram. There is no method implementing subtraction for two histograms and returning a histogram (nor should there be). Is this about a case where a user subtracts the weights of two histograms and injects the results in a histogram object?
Also, I wouldn't like to think of normalizing the result of such an operation - what if the two histograms had different sample sizes? In the event someone would do this, wouldn't they normalize the histograms first?
In short, I don't think we need to consider negative entries.

oschulz: > IMO the difference among two histograms is not a histogram

I have to disagree - in practice, subtraction of histograms is quite common, esp. for experimental data. One example is spectroscopy: It's common practice to subtract a background histogram from a signal histogram. Afterwards, most bins are (usually) positive, but if a bin has few counts, the signal value can be lower than the background value in that bin, due to statistical fluctuations - resulting in a negative bin value. Certainly this isn't very clean, mathematically/statistically, but common practice nevertheless. Any yes, people often do normalize the result in various ways (of course the integral of the difference histogram is still positive, usually).

So maybe it would be best to output `s_0` instead of `norm(s_0)`. Then a negative norm value can act as a clear warning that something is wrong.

oschulz: > Is this about a case where a user subtracts the weights of two histograms and injects the results in a histogram object?

Yes - though I think we should add support for histogram addition and subtraction to StatsBase (addition is even more common than subtraction, in my experience).

oschulz: > Is this about a case where a user subtracts the weights of two histograms and injects the results in a histogram object?

Yes - though I think we should add support for histogram addition and subtraction to StatsBase (addition is even more common than subtraction, in my experience).

src/hist.jl
andreasnoack: We don't want to make defensive copies. Copy only when it is needed or we end up like R. We don't want to add new behavior for `float` just in this package.

andreasnoack:
I agree that your `:norm` mode is just changing the scale of the y-axis but I don't that is true for the two other modes. Actually, I don't think that the non-density versions are real histograms. It might be binning or bucketing but the historical meaning of a histogram is a (graphical) density approximation. My main argument against two different types would be that it is only relevant for uneven bins which, by far, is the more common case. Another solution might be to add a field to the type with the density version of the counts. For even bins, that field could just alias the current weight array or it could be made `Nullable` such that we don't require it to be set.

mkborregaard:
I like the idea of storing the normalization information in the type. Especially given that there appear to be  many interpretations of histograms (mine aligns with @andreasnoack ). I see two good alternatives to making a new abstract type hierarchy, or duplicating the weights Vector for density values:

1. Parameterize the histogram with a symbol specifying the normalization, e.g. `Histogram{(other type parameters), :counts}`. Upsides: this would make the type information available, while still preserving all downstream code using Histograms. Downside: like with new type hierarchy, it would force `normalize` to copy its argument.
2. Introduce an extra field (e.g. `normalization::Symbol`) in the Histogram struct. Upside: Clear, simple, lightweight. Downside: would make any function using the normalization information dispatch dynamically. These are probably mainly the normalization functions themselves.

I prefer 2.

oschulz:
> I agree that your :norm mode is just changing the scale of the y-axis but I don't that is true for the two other modes.

@andreasnoack, sorry, but I have to disagree - yes, it is: Let's take a typical (for me) application: Photo/gamma-spectroscopy. You start with a measurement - that gives you a histogram with the photon energy on the x-axis, and counts on the y-axis. Often, people look at the raw counts first. Then you might want to do a `:density` normalization (very common in this application) - that gives you counts/energy (e.g. 1/kev, in proper units) on the y-axis. Now you have a histogram that normalized in respect to binning, and can be compared to other measurements with a different binning. After that, you'll typically want to divide that weights by measurement time - which gives you a y-axis in 1/kev/s. Now you can compared to other measurements independent of binning and measurement time. Finally, you may apply a `:norm`-normalization. This will then result in a PDF that describes the spectrum of that type of source, independent time, source-intensity and binning. We call all of these a histogram.

Another physics application: Analysis of a particle-collision experiment (e.g. at LHC). Things are a lot more complex, but histograms that go though different kinds of normalizations are one of the fundamental tools.

> Actually, I don't think that the non-density versions are real histograms.

Take the applications described above: Thousands of physicists call each of these (raw counts, density, PDF) a histogram.

Another application: Markov-chain monte carlo. First, you generate lots of samples. Then do marginalize via histograms. You may not want to normalize at first, to check what's going on. Later on, you will normalize. Whether normalized or not, I've heard lot's of statisticians call all of these a histogram.

>It might be binning or bucketing but the historical meaning of a histogram is a (graphical) density approximation.

It really depends on who you ask. In many scientific disciplines (e.g. the whole of particle-/astro-particle-/nuclear-physics, spectroscopy in general and a bunch of others), all of them are considered to be histograms. I don't think we should exclude large user communities by using an artificially narrow (resp. application-specific) definition of what a histogram is. We'd just end of with lots of different histogram or histogram-like classes that all do basically the same thing, but hinder interoperability.

IMHO we'll serve users a lot better if we keep StatsBase.Histogram as generic as possible (StatsBase shouldn't deal with units or specific applications, just like Base.Array doesn't), and give people the freedom to build more application-specific histograms types on top of it (in other packages). Otherwise, I have a strong feeling that StatsBase.histogram will be unusable for many disciplines in which histograms are important tools.

Julia currently doesn't have a widely used de-facto histogram class (only very few registered packages use StatsBase.Histogram). However, I believe that should change, and that StatsBase.Histogram can fill that niche - but only if it's widely applicable.

oschulz:
> @mkborregaard I like the idea of storing the normalization information in the type. 

You'll end up with a very complex type  hierarchy for little gain. There's just to many use cases for histograms (and normalizations and units) to cover in a core package like StatsBase.

If you want to track normalizations (many users won't need to), look at my proposal to have a Uniful.jl-based wrapper: That'll give you a lot more power without any type hierarchy.

mkborregaard:
Couldn't all of these use cases be catered for by both of my suggestions above?

oschulz:
@mkborregaard : Sorry, I should have read your post a bit more carefully.

> Especially given that there appear to be many interpretations of histograms (mine aligns with @andreasnoack ).

That's the point: Let's please keep things broad and flexible (as it is now), and not only cater to one interpretation.

> 1. Parameterize the histogram with a symbol specifying the normalization, e.g. Histogram{(other type parameters), :counts}. Downside: like with new type hierarchy, it would force normalize to copy its argument (i.e. invalidate normalize!).

Additional downside: People who really need real units now have to deal with both units and "normalization-mode" - but that's redundant information. And a type parameter would make code more complex that needs to deal with histograms of different normalization types (esp. if you consider additional user-specific normalization-type-symbols).

> 2. Introduce an extra field (e.g. normalization::Symbol) in the Histogram struct. Upside: Clear, simple, lightweight.

Downside: What if people want real units? Then a Symbol won't be enough. Again, you'll dicate to users how to use `Histogram`. Won't work with Unitful, just as an example.

> Downside: would make any function using the normalization information dispatch dynamically.

Forcing dynamic dispatch would be horrible, in my opinion.

I really don't understand why StatsBase.Histogram can't just stay as core of what a histogram is (which it is now): Weights plus binning information. You're right - people have different expectations of a histogram and use histograms in many different ways. Let these specialized interpretations be catered to by specialized types, optimized for their user community, built/wrapped around `StatsBase.Histogram`. Then there will still be interoperability, because each of these types will (hopefilly) be compatible with `StatsBase.Histogram`.

One example: Many people need histograms that also keep track of statistical uncertainties, and those values have to be re-scaled when a histogram is normalized. I'm *not* proposing to add this as an additional field to histogram - because other people don't need/want this. Instead, `normalize!` does optionally accept additional arrays - so that one can built an ErrorHist (or so) type without writing a lot of code.

oschulz:
My point is, we can't stuff everyone's need into `StatsBase.Histogram`, there's just too much variety. Better make it the de-facto `Base.Array`-equivalent for everyone who needs histograms - and give it all the functionality people need to easily build more specialized types and applications.

mkborregaard:
My proposal is just to store the type of the normalization in the struct. `:counts` would be the default value. Running `normalize!(:density)` would change this to `:density`. Running `normalize!(:norm)` on that would change this to `:pdf`. The user would never have to worry about it. No downstream code would have to be changed. Or am I wrong?

oschulz:
> My proposal is just to store the type of the normalization in the struct [...] The user would never have to worry about it.

He will - if he applies his own normalizations. Or what about users who read in histogram data from some data source - they would have to make sure the normalization field has the right values. I fear the end result would be that often you have a histogram with the wrong value in the "normalization" field - that's much, much worse than not having that information at all.

Also, some people may be happy with a Symbol, while others need to dispatch on that information (so they need a type parameter), and others won't want that field at all. An OO-designer would say that this is a typical use case for a decorator pattern - instead of stuffing the information into the base type.

There's a lot of operations people do with histograms, normalization is only one kind. Will we add an extra field for each of these (e.g. statistical uncertainties, units, etc.)? I really don't get why it's so important to store normalization information in a type as basic as `StatsBase.Histogram`. `Vector` doesn't keep track on whether it's been normalized either.

src/hist.jl
oschulz: > We don't want to make defensive copies.  Copy only when it is needed or we end up like R. 

Ok, then I suggest we keep the currently implemented behavior. But if we copy, it should be a *full* copy (as implemented now), not a partial one. Otherwise it'll be completely unpredictable for the user - and then users will do exactly what you don't want: Make lot's of defensive copies in their code to get thread-safety.

oschulz:
> Running normalize(:density) on something already of type :density would fail. The user would never have to worry about it. No downstream code would have to be changed. Or am I wrong?

In my experience, trying to automatically "do the right thing" without having the user think about what he's doing never ends well - especially in math/science. :-)

mkborregaard:
OK, those were my two cents - leaving this for @ararslan and @andreasnoack to decide.

oschulz:
> OK, those were my two cents - leaving this for @ararslan and @andreasnoack to decide.

Likewise.

ararslan:
I think the fact that there are so many different ways to use a histogram suggests (at least to me) that multiple types that inherit from a common abstract type (e.g. `StatsBase.AbstractHistogram`) would be the most useful, general, and ultimately Julian approach.

The analogy with `Array` isn't entirely fair, since an array is a very basic and fundamental type. One characteristic of arrays is certainly shared by histograms though: neither can serve all purposes. In the case of `Array`, most often people will want to code against `AbstractArray` to allow for user-defined subtypes, like `StaticArray`, `OffsetArray`, and others.

Using an abstraction here, you could have signatures like `normalize(hist::AbstractHistogram) -> NormalizedHistogram`, which is a no-op for `NormalizedHistogram` input and a conversion for `Histogram`s.

For densities, one should really use the KernelDensity package. Perhaps we need a conversion method in KernelDensity from a `Histogram` to a `KDE`.

ararslan:
I'd also be interested to hear from @simonbyrne, who IIRC wrote much of the histogram code here.

oschulz:
> I think the fact that there are so many different ways to use a histogram suggests (at least to me) that multiple types that inherit from a common abstract type (e.g. StatsBase.AbstractHistogram) would be the most useful, general, and ultimately Julian approach.

Not necessarily - we couldn't have `Base.normalize!` for histograms, then (forcing people to do memory allocation whether they want to or not) which doesn't seem very Julian to me. ALso, generic code that deals with histograms would often need an additional type parameter for the histogram type to avoid dynamic dispatch (or what about types that store histograms?). I don't see how that will make user's life easier.

Also, what about custom normalizations (like I described in my examples above)? Should people then do their own custom histogram classes for each of those?

oschulz:
If histogram would include normalization information (via type of via a field), how would scenarios like this be handled:

A user normalizes some histograms (to PDFs). Then they are added up in a weighted sum (e.g. to simulate something) - not an exotic use case at all. Yes, I know, `StatsBase.histogram` doesn't currently support addition and scalar multiplication, but it should, in the future (ideally via `.+` and `.*`, including `.=`). Now, what type of histogram (or what value of the normalization-field) comes out of such operations? And why would we forbid users to do it in place (which we would, when using a type hierarchy)? Even if we could come up with a well-defined behavior that serves most use-cases (I doubt we can), implementing simple things like histogram addition would get quite complicated (checking /modifying the value of the normalization field depending on it's current value, etc.).

mkborregaard:
In principle all that's needed is a field specifying whether the histogram has been density-normalized, preventing from doing it again. All other operations would perhaps not need to be saved in the object. I am not a fan of histogram addition, but that is beside the point.

oschulz:
> I am not a fan of histogram addition, but that is beside the point.

It's not beside the point, I think - because I don't really see how persistent normalization information will play well with things like addition and multiplication (people even divide histograms - there are legitimate use cases).

As for not being a fan - again, why limit histogram use to science disciplines that don't need such operations? At least in some disciplines, histogram addition is a fundamentally important operation. A few examples:

* Statistical data analysis: It's quite common to create weighted mixtures of some base histograms (e.g. results of monte-carlo analysis, usually normalized in certain respects) to compare them with data - for example to calculate a likelihood. This needs histogram addition and scalar multiplication.

* Large scale data processing: When putting lots of data through thousands of computing nodes (e.g. petabytes of LHC data), histogram addition is the reduction operation you need  for ...`). It's not a matter of preference (or liking) - and there's nothing wrong with it, mathematically/statistically. It doesn't have to be on such a large scale - you may also need it as a reduction in a `@parallel (+) for` to reduce many intermediate histograms to one.

oschulz:
I really just wanted to implement normalization for histograms (see #231 by @alanedelman) - I didn't expect it to evolve a discussion about extending the type itself, or what operations/use-cases it should support. :-)

But, may I just ask the StatBase team (@ararslan, @andreasnoack, @simonbyrne, @nalimilan, @lindahua?): I was actually planning to do another PR soon to add support for histogram addition, scalar multiplication, histogram division (see ROOT's [TH1 histogram class](https://root.cern.ch/doc/master/classTH1.html)), etc.

Would such a PR be welcome? Or would you prefer the usage scope of `StatsBase.Histogram` to stay more narrow (depending on what people consider to be a "true" histogram or not, and what people should be "allowed" to do with them)? I'd just like to know before I start coding. :-)

mkborregaard:
Let me propose something that might resolve this.

I think that the basic histogram includes histograms with counts, density-normalized ones (for uneven bins) and pdf-normalized ones. All these types are expected to be found in a package like StatsBase (as also implied by @alanedelman in the issue that started all this (https://github.com/JuliaStats/StatsBase.jl/issues/231). They can all be implemented using a subset of this PR. 
They could have a field specifying their normalization type - there is no reason for them to be distinct types, as methods in general (other than the normalization itself) should not have different behaviour for different types.

It is also clear to me that in particle physics in particular and perhaps a few other fields, histograms are used much more flexibly and in ways that we think of as untraditional. I think the best you (@oschulz) could do would be to create a new package, e.g. Histograms.jl, that extends the Histogram type, and allows for many more operations, like addition, division, norm-scaling, negative histograms. It would be much easier for you, who are the expert on that, to have full control, instead of trying to convince everyone else about functionality you know will be useful for people in your field. It would also make your contribution much more visible. And people using this functionality would quickly find and use the package.

I think some of the particle-physics-specific elements of your Plots PR could also go there as recipes - e.g. those implementing error bars on histograms. How does this sound?

mkborregaard:
Of course, this isn't really any of my business. As you said, this is all about the direction the JuliaStats team wants to take the histogram support. I don't doubt at all that all your suggested functionality is useful and valid.

oschulz:
> It is also clear to me that in particle physics in particular and perhaps a few other fields, histograms are used much more flexibly and in ways that we think of as untraditional.

I'd welcome the opionion of a few statisticians, here, regarding what's "traditional". :-)

Concerning particle physics and many related disciplines (not exactly a small community) - histograms play a *very* central role here. For example, it's probably the most important class in [ROOT](http://root.cern.ch). But a lot of the use cases I described are also common in statistical data analysis in general. I really don't think I'm talking about exotic applications, here.

> I think the best you (@oschulz) could do would be to create a new package

Sure, I've thought about that. However, as StatsBase has become such a central element in the Julia ecosystem, I felt that it was the right place for a flexible, least-biased histogram class. So my preferred approach would be to have a lean, generic, non-restrictive histogram class in StatsBase. One that can serve as a basis for as many different use cases as possible - and one that people can build upon in more specialized packages, instead of having to roll their own from scratch.

From what I can see, StatsBase.Histogram isn't used a lot, so far (at least not in registered packages) - but then, it currently doesn't provide a lot of functionality. I wanted to help change that (and sure, naturally add some functionality I need myself, along the way), without making the `Histogram` type itself more complicated.

> As you said, this is all about the direction the JuliaStats team wants to take the histogram support.

Indeed - it's their package, after all, and it kinda is @simonbyrne's type. I'd just like to know what the long-term intention is, concerning histogram functionality, so I can focus on writing code. ;-)

mkborregaard:
> One that can serve as a basis for as many different use cases as possible - and one that people can build upon in more specialized packages, instead of having to roll their own from scratch.

Yes. A solution could be for StatsBase to define AbstractHistogram with a single concrete subtype Histogram. AFAICS this also avoids the issue of 'type piracy' for packagages extending the class.

oschulz:
> A solution could be for StatsBase to define AbstractHistogram

Possibly - we'd need a set of well-defined methods to get weights, edges, etc. from it, though. Also, if the maintainers come to the conclusion that term "histogram" should be interpreted rather narrow, `AbstractHistogram` may or may not be the right super-type for histograms of the kind I described.

I'll wait until there's a decision from the maintainers on where to take this. I really didn't mean to trigger any major changes here, I just wanted to add support for two methods in Base (with plans to add things like addition and scalar multiplication later on as well). If we need some deeper changes regarding `Histogram` first, the question is who will design/implement them (and on what time scale). I'm obviously not a fan of erecting a more complex type hierarchy here - but if it is to be done, it should be very well thought out, to provide maximum benefit to the Julia ecosystem (and to be conceptually and mathematically clean).

oschulz:
@mkborregaard: I've thought a bit more about your proposal regarding storing normalization information. I don't think a different type (as @ararslan suggested) is the way to go, and I don't like storing a symbol, either. However, you're right - while there's lot's of different normalizations, density-normalization is something fundamental, and should only be done once.

So how about we just add a single field `isdensity::Bool` to `Histogram`? Then, like you described, `normalize` could be implemented so it'll always be idempotent. On the other hand, we could still have `normalize!`, and we wouldn't limit the user in any way.

However, I'll still need thing like histogram addition, scaling, etc. in the future - would this be welcome?

oschulz:
@mkborregaard: In principle all that's needed is a field specifying whether the histogram has been density-normalized, preventing from doing it again.

I've thought a bit more this: I don't think a different type (as @ararslan suggested) is the way to go, and I don't like storing a symbol either. However, you're right - while there's lot's of different normalizations, density-normalization is something fundamental, and should only be done once.

So how about we just add a single field `isdensity::Bool` to `Histogram`? Then, like you described, `normalize` could be implemented so it'll always be idempotent. On the other hand, we could still have `normalize!`, and we wouldn't limit the user in any way.

Concerning things like future support for histogram addition - would this be welcome (e.g. via supporting `broadcast` and `broadcast!`)?

mkborregaard:
I'd support this design.

oschulz:
@oschulz: >> So how about we just add a single field isdensity::Bool to Histogram? Then, like you described, normalize could be implemented so it'll always be idempotent.
@mkborregaard: > I'd support this design.

@ararslan, @andreasnoack: Would you support this? I would keep `Histogram` lean and uncomplicated but still make `normalize` "safe" by respection previous normalization. However, I'd need to extend the constructors (so that uses can specify whether their data is already a density) - before I do that, I'd like a quick yes or no (hoping for a fairly quick review and merge, after it's implemented).

ararslan:
That seems like a decent solution to me. I'd be interested to hear from @andreasnoack and @simonbyrne on it though.

oschulz:
Gentle bump - @andreasnoack, @simonbyrne, I realize you guys are busy people (who isn't), but could we just get a quick yes/no on this?

andreasnoack:
Sorry for the slow pace here but it takes a while each time I try to think this through. I'm fine with the `isdensity` field. How should `push!` behave when `isdensity`?

oschulz:
Thanks, @andreasnoack - I appreciate that you guys care a lot about keeping things high-quality!

> How should `push!` behave when `isdensity`?

Good point - hadn't thought about that. Well, density-normalization divides the histogram weights by the bin volumes. So I guess the natural behaviour would be to divide each value to push (1 or weight) by the volume of the target-bin before adding it to the bin weight.

Numerical errors aside, this should result in the same histogram no matter whether you fill it first and then density-normalize or whether you start with an empty `isdensity == true` histogram and then fill it. You could even fill, density-normalize and then push in more values.

mkborregaard:
This would only work for histograms created by the `:density`-normalization in this PR, not the more flexible normalizations suggested in this comment https://github.com/JuliaStats/StatsBase.jl/pull/243#issuecomment-290604600, though, nor for pdf-normalized histograms (which will have `isdensity == true` but nothing else to distinguish them under this design). Maybe that's OK - the user might guess that `push!`ing to a pdf-normalized histogram is asking for trouble. The alternative is to throw an exception when pushing to Histograms with `isdensity == true`, though it would work correctly under the most common scenario.

mkborregaard:
A third solution would of course be to have a sample size field :-D this, together with `isdensity` should make it possible to always recreate the original counts histogram.

oschulz:
> Maybe that's OK - the user might guess that push!ing to a pdf-normalized histogram is asking for trouble. 

I agree - we can't automatize every possible scenario, it's unavoidable that the users use their brain and understand what they want to do. :-)

> The alternative is to throw an exception when pushing to Histograms with isdensity == true

I suggest we allow it. Like I said earlier, there's many different normalizations that can be done on histograms - it would be kinda hacky to also interpret `isdensity == true` as "might be y-axis normalized".

oschulz:
So I'll go ahead and implement this, then?

oschulz:
> A third solution would of course be to have a sample size field

You mean a "number of samples" field? Some implementations (e.g. ROOT histograms) do have this - it's a two-edges sword in my experience.

> make it possible to always recreate the original counts histogram

Not when the histogram is filled with samples of weight != 1, right?

It's definitely something to think about - but I suggest we decouple it from this PR.

mkborregaard:
> Not when the histogram is filled with samples of weight != 1, right?

```
h.weights .= h.isdensity ? h.weights .* diff(h.edges) : h.weights
h.weights .= round.(h.weigths ./ sum(h.weights) .* h.N)
```
?

mkborregaard:
But the operation would make `push!` slow. Makes sense to keep it separate.
Given that both @ararslan and @andreasnoack have nodded to the `isdensity` field, you can probably just go ahead and implement?

oschulz:
Yes, I'll try to do it very quickly.

oschulz:
> @mkborregaard :  you can probably just go ahead and implement.(?)

Almost done, just needs a little more polish. Got a bit more involved.

oschulz:
It's done, `Histogram.isdensity` is added. While I was implementing it, I stumbled on quite a few instances of duplicated code, mainly in the constructors and the `push!`, `append!` and `fit` implementations. As I didn't want to make things worse (be adding redundant `isdensity` handling), I refactored a lot of that code. I hope you like the result - I think the code is also more readable now.

I added two new (currently not exported) functions `binindex` and `binvolume`. They make for better code structure, but will also come in handy for users who do histogram analysis.

I also added two internal helper functions: `_tuple_map` (we map over tuples in many places in the code) and `_multi_getindex`. Before, we used `Base.map` and `Base.zip`, which seem to be significantly slower for our use case. I'll think about making a PR to replace the current implementation of `Base.map` for tuples - the current version seems to generate inefficient code (at least according to `@code_llvm` and `@code_native`, both on Julia v0.5 and v0.6).

The performance gain is quite substantial (an order of magnitude). For the following example

```julia
using StatsBase
using BenchmarkTools

edges = (
    cumsum(rand() * rand(9)),
    cumsum(rand(1:10) * rand(1:100, 11)),
    cumsum(5 * rand() * rand(14))
)

n = 100000

data = (
    maximum(edges[1]) * (randn(n) / 6 + 0.5),
    rand(1:maximum(edges[2]), n),
    maximum(edges[3]) * rand(n)
)

@benchmark fit(Histogram, data, edges, closed=:left)
```

I get these benchmark results:

* Current StatsBase master: mean time 95.982 ms
* This PR: mean time 7.413 ms

As I had to touch the `fit` implementations anyway, I added support for `fit(Histogram{T}, ...)`. So now we can do things like `normalize!(fit(Histogram{Float64}, ...))` (which wouldn't work for an `Int`-weighted histogram). As before, `fit(Histogram, ...)` still results in an `Int` histogram.

I also structured the histogram tests with test sets, it helped a lot during testing (easier to locate the failed tests).

@andreasnoack, @ararslan, @mkborregaard - I hope you guys like it and can find some time to review (I *so* look forward to the new Plots histograms, which depend on this). I know the last commit is huge - but since the changes are very interdependent, I don't think it makes sense to split it up.

oschulz:
Ooops - looks like I got to fix some failed tests on Julia v0.6, first.

oschulz:
All fixed now.

oschulz:
Gentle bump - @ararslan, @andreasnoack, if you should find a few spare minutes during easter time ... :-)

ararslan:
Looks okay to me on first glance, but I'll give it a more thorough look soon.

oschulz:
Thanks, @ararslan!

src/hist.jl
andreasnoack: No. We shouldn't copy more than necessary during a promotion. When you need to be sure that no aliasing is happening you can copy the object as needed.

src/hist.jl
```diff
@@ -14,6 +16,31 @@ function _check_closed_arg(closed::Symbol, funcsym)
 end
 
 
+## Fast map for tuples (a lot faster than using Base.map)
+@generated function _tuple_map{N}(f, tuples::NTuple{N, Any}...)
```
andreasnoack: I don't this is the right place for this function. It should be a PR against Base and meanwhile, I think we should use the standard version and live with the slowdown. Is `broadcast` equally slow?

src/hist.jl
andreasnoack: I agree with @mkborregaard here. I think we should just leave `s0` as it is.

src/hist.jl
```diff
@@ -14,6 +16,31 @@ function _check_closed_arg(closed::Symbol, funcsym)
...
+            push!(call_f_expr.args, :(tuples[$j][$i]))
+        end
+        push!(result_expr.args, call_f_expr)
+    end
+    result_expr
+end
+
+## Fast getindex function for multiple arrays, returns a tuple of array elements
+@inline Base.@propagate_inbounds @generated function _multi_getindex(i::Integer, c::AbstractArray...)
+    N = length(typeof(c::Tuple).parameters)
```
andreasnoack: Why no just `length(c)`?

src/hist.jl
oschulz: I checked - broadcast is equally slow (I wrote a little tuple-map benchmark, independent of StatsBase). I guess you're right tough - I was thinking about the fact that a PR on this for `Base` wouldn't make it into 0.6 anymore, so ... but maybe it's really best not to put a temporary shim in here. I'll take it out.

oschulz: I took it out - surprisingly, performance in the benchmark given above is still about the same (but in my tuple-map test, there is a vast difference in performance). While I developed this, I did need _tuple_map to make histogram filling fast - but it seems like in the final implementation now, it doesn't make much of a difference. Maybe inlining behaviour somewhere - don't know. Anyhow, looks good. (I'll play a bit more with tuple mapping, independent of StatsBase, and think about a PR for Base.)

src/hist.jl
oschulz: Uh, yeah - this went through several iterations ... anyhow, I replaced it with `length(c)` as suggested. Question - should I move this into "common.jl"? This one is definitely very relevant for performance - maybe it could be used in other places in StatsBase at some point?

src/hist.jl
oschulz: Your decision - changed.

src/hist.jl
oschulz: I agree - that way, the user will know something is wrong if `norm` returns a negative value. Changed.

oschulz:
Thanks for your comments, @andreasnoack - I made the requested changes (I hope I didn't miss any). I also went trough everything again and fixed isdensity handling in a few places. Want to give it a final pass?

src/hist.jl
```diff
@@ -140,100 +155,270 @@ function show(io::IO, h::AbstractHistogram)
...
+
+function _float_deepcopy(x)
+    y = float(x)
+    y === x? deepcopy(y) : y
+end
+
+
+function float{T, N, E}(h::Histogram{T, N, E})
+    float_weights = float(h.weights)
+    if float_weights === h.weights
```
andreasnoack: Do this with a separate method for `T<:"float types"`

src/hist.jl
andreasnoack: Also handle this one with separate method instead of a branch.

src/hist.jl
oschulz: Good point - done. I got rid of `_float_deepcopy`.

oschulz:
>@andreasnoack: I think we are almost there.

Thanks!

> One thing, though, Isn't the changes to push! and append! separable from the normalization functionality? 

Not really, because `push!` and `append!` now have to take `isdensity` into account. That's the main reason I refactored that code: The former implementation had some code duplication, so I would have had to duplicate the `isdensity` handling in several places.

> If so, please move them to a new PR

Of course it would, in principle, be possible to have a separate PR that refactors `push!` and `append!` and merge it *before* this one - but it would be quite a bit of work: The code changes are all very close, it would involve a lot of manual changes/merges/re-writing (`git add -p` wouldn't work), and then some more testing to check that I didn't screw things up. I'd be very grateful if we could keep it all in this PR.

src/hist.jl
andreasnoack: My point here is that instead of the branch you can write two methods. Like
```julia
float{T<:"float types"}(h::Histogram{T}) = h
float(h::Histogram) = Histogram(float(h.weights), h.edges)
```

andreasnoack:
> Of course it would, in principle, be possible to have a separate PR

No problem. It was only if they were completely orthogonal. I've clarified my comment to the `float` function. When that is fixed, we can go ahead and merge.

src/hist.jl
oschulz: > @andreasnoack: My point here is that instead of the branch you can write two methods

I understand - I just forgot to delete the other version, somehow, sorry. Fixed now.

andreasnoack:
Thanks again for adding this functionality.

oschulz:
@andreasnoack, @ararslan, @mkborregaard: Thanks for your time, comments and ideas!

mkborregaard:
Whew, amazing work guys.
http://reactiongifs.com/?p=22839

oschulz:
I guess we can close #231 now, too?

mkborregaard:
@ararslan @andreasnoack can I ask you when you mean to make a release, and what will be the version number associated with it? I'd like to make the next release of Plots (with the PR that depends on this) coincide with this.

ararslan:
@mkborregaard v0.14.0 tagged, METADATA PR pending: https://github.com/JuliaLang/METADATA.jl/pull/9083

oschulz:
> v0.14.0 tagged, METADATA PR pending

Thanks, @ararslan!

Issue #240, opened 2017-02-20
use of non-global RNG


tawheeler:
Reproducibility with random numbers is often critical for research or work. It would be nice to support passing in your own rng:

```julia
rng = MersenneTwister(17)
sample(rng, [1,2,3])
```

This echoes https://github.com/JuliaStats/Distributions.jl/issues/436.

richardreeve:
@matbesancon This is now fixed, and also works for all distributions, not just ones in `Random`, and so can be closed. I think it was fixed earlier, but in any event post JuliaStats/Distributions.jl#830 it definitely works:
```julia
julia> Random.seed!(rng, 1);

julia> sample(rng, [1,2,3])
3

julia> Random.seed!(rng, 1);

julia> sample(rng, [1,2,3])
3

julia> Random.seed!(rng, 1);

julia> sample(rng, [1,2,3])
3
```

PR #245, opened 2017-03-15
Optionally specify RNG to use for sampling

evanfields:
Addresses #240.

I tried to follow the style already in place in `src/sampling.jl`. Passing the RNG by keyword argument was a bit clearer but seemed to impose a slight performance penaltiy.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -14,25 +14,25 @@
 #
 #   x[i] and x[j] are independent, when i != j
 #
-function direct_sample!(a::UnitRange, x::AbstractArray)
+function direct_sample!(a::UnitRange, x::AbstractArray; rng::AbstractRNG=Base.GLOBAL_RNG)
     s = RandIntSampler(length(a))
     b = a[1] - 1
     if b == 0
         for i = 1:length(x)
-            @inbounds x[i] = rand(s)
+            @inbounds x[i] = rand(s, rng)
         end
     else
         for i = 1:length(x)
-            @inbounds x[i] = b + rand(s)
+            @inbounds x[i] = b + rand(s, rng)
         end
     end
     return x
 end
 
-function direct_sample!(a::AbstractArray, x::AbstractArray)
+function direct_sample!(a::AbstractArray, x::AbstractArray; rng::AbstractRNG=Base.GLOBAL_RNG)
     s = RandIntSampler(length(a))
     for i = 1:length(x)
-        @inbounds x[i] = a[rand(s)]
+        @inbounds x[i] = a[rand(s, rng)]
     end
     return x
 end
@@ -44,9 +44,9 @@ end
 
 Draw a pair of distinct integers between 1 and `n` without replacement.
 """
-function samplepair(n::Int)
-    i1 = randi(n)
-    i2 = randi(n-1)
+function samplepair(n::Int; rng::AbstractRNG=Base.GLOBAL_RNG)
+    i1 = randi(n, rng)
+    i2 = randi(n-1, rng)
     return (i1, ifelse(i2 == i1, n, i2))
 end
 
@@ -56,8 +56,8 @@ end
 
 Draw a pair of distinct elements from the array `a` without replacement.
 """
-function samplepair(a::AbstractArray)
-    i1, i2 = samplepair(length(a))
+function samplepair(a::AbstractArray; rng::AbstractRNG=Base.GLOBAL_RNG)
+    i1, i2 = samplepair(length(a); rng=rng)
     return a[i1], a[i2]
 end
 
@@ -68,7 +68,8 @@ end
 #
 #   Reference: The Art of Computer Programming, Vol 2, 3.4.2, p.142
 #
-function knuths_sample!(a::AbstractArray, x::AbstractArray; initshuffle::Bool=true)
+function knuths_sample!(a::AbstractArray, x::AbstractArray;
+                       rng::AbstractRNG=Base.GLOBAL_RNG, initshuffle::Bool=true)
     n = length(a)
     k = length(x)
     k <= n || error("length(x) should not exceed length(a)")
@@ -79,7 +80,7 @@ function knuths_sample!(a::AbstractArray, x::AbstractArray; initshuffle::Bool=tr
     end
     if initshuffle
         @inbounds for j = 1:k
-            l = randi(j, k)
+            l = randi(j, k, rng)
             if l != j
                 t = x[j]
                 x[j] = x[l]
@@ -91,8 +92,8 @@ function knuths_sample!(a::AbstractArray, x::AbstractArray; initshuffle::Bool=tr
     # scan remaining
     s = RandIntSampler(k)
     for i = k+1:n
-        if rand() * i < k  # keep it with probability k / i
-            @inbounds x[rand(s)] = a[i]
+        if rand(rng) * i < k  # keep it with probability k / i
+            @inbounds x[rand(s, rng)] = a[i]
         end
     end
     return x
@@ -109,7 +110,7 @@ end
 #
 #   O(n) for initialization + O(k) for random shuffling
 #
-function fisher_yates_sample!(a::AbstractArray, x::AbstractArray)
+function fisher_yates_sample!(a::AbstractArray, x::AbstractArray; rng::AbstractRNG=Base.GLOBAL_RNG)
     n = length(a)
     k = length(x)
     k <= n || error("length(x) should not exceed length(a)")
@@ -120,7 +121,7 @@ function fisher_yates_sample!(a::AbstractArray, x::AbstractArray)
     end
 
     @inbounds for i = 1:k
-        j = randi(i, n)
+        j = randi(i, n, rng)
         t = inds[j]
         inds[j] = inds[i]
         inds[i] = t
@@ -136,7 +137,7 @@ end
 #   each time draw a new index, if the index has already
 #   been sampled, redraw until it draws an unsampled one.
 #
-function self_avoid_sample!(a::AbstractArray, x::AbstractArray)
+function self_avoid_sample!(a::AbstractArray, x::AbstractArray; rng::AbstractRNG=Base.GLOBAL_RNG)
     n = length(a)
     k = length(x)
     k <= n || error("length(x) should not exceed length(a)")
@@ -152,9 +153,9 @@ function self_avoid_sample!(a::AbstractArray, x::AbstractArray)
 
     # remaining
     for i = 2:k
-        idx = rand(rgen)
+        idx = rand(rgen, rng)
         while idx in s
-            idx = rand(rgen)
+            idx = rand(rgen, rng)
         end
         x[i] = a[idx]
         push!(s, idx)
@@ -183,7 +184,7 @@ end
 #
 #  Require O(n) random numbers.
 #
-function seqsample_a!(a::AbstractArray, x::AbstractArray)
+function seqsample_a!(a::AbstractArray, x::AbstractArray; rng::AbstractRNG=Base.GLOBAL_RNG)
     n = length(a)
     k = length(x)
     k <= n || error("length(x) should not exceed length(a)")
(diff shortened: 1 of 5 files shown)


src/sampling.jl
```diff
@@ -68,7 +68,8 @@ end
 #
 #   Reference: The Art of Computer Programming, Vol 2, 3.4.2, p.142
 #
-function knuths_sample!(a::AbstractArray, x::AbstractArray; initshuffle::Bool=true)
+function knuths_sample!(a::AbstractArray, x::AbstractArray, rng::AbstractRNG=Base.GLOBAL_RNG;
+                       initshuffle::Bool=true)
```
ararslan: It would be great if you could indent the lines in the function signatures such that they align after the initial `(`. That is, this would look like
```julia
function knuths_sample!(a::AbstractArray, x::AbstractArray, rng::AbstractRNG=Base.GLOBAL_RNG;
                        initshuffle::Bool=true)
```

ararslan:
Functionally these changes look good to me; my only request here is purely a stylistic change.

evanfields:
At risk of pestering, any chance of getting this merged soon? It'd be useful for some work I'm doing right now :)

andreasnoack:
Have you discussed the position of the rng argument somewhere? We should probably try to be consistent with `rand` but I think it is wrong that we have it as the first argument there.

ararslan:
> Have you discussed the position of the rng argument somewhere?

Sorry, I overlooked that completely in my review. Agreed that we should be consistent with `rand`, but I have no opinion in regards to where the positional argument should be in `rand`.

evanfields:
I also think it's wrong that regular `rand` puts the optional RNG argument first; that violates the convention that optional positional arguments go last. (You can see the effects of this in `random.jl` where the regular syntax `(..., arg = val)` isn't used.) That said, consistency seems reasonable, so tomorrow I'll push a commit duplicating the style from base Julia.

test/wsampling.jl
```diff
@@ -83,24 +87,32 @@ wv = weights([0.2, 0.8, 0.4, 0.6])
 a = zeros(Int, 3, n)
 for j = 1:n
     naive_wsample_norep!(4:7, wv, view(a,:,j))
+	@test naive_wsample_norep!(MersenneTwister(j), 4:7, wv, zeros(Int, 2)) ==
```
nalimilan: Indentation is incorrect here and below.

test/sampling.jl
```diff
@@ -13,6 +13,7 @@ x = [randi(10) for i = 1:n]
 @test isa(x, Vector{Int})
 @test extrema(x) == (1, 10)
 @test isapprox(proportions(x, 1:10), fill(0.1, 10), atol=5.0e-3)
+@test randi(MersenneTwister(1), 1000) == randi(MersenneTwister(1), 1000)
```
nalimilan: It doesn't make a lot of sense to check whether the result is equal to itself. Any other way of testing this? What do the existing tests do for the existing methods? At least, checking that the result when passing the default RNG is equal what you would get without passing it would be better than this check. (Same below.)

src/sampling.jl
```diff
@@ -507,7 +534,7 @@ end
 #     URL http://www.sciencedirect.com/science/article/pii/S002001900500298X
 #
 # Instead of keys u^(1/w) where u = random(0,1) keys w/v where v = randexp(1) are used.
-function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
+function efraimidis_ares_wsample_norep!(rng::AbstractRNG, a::AbstractArray, wv::WeightVec, x::AbstractArray)
```
nalimilan: Please break the lines that go beyond 92 chars.

test/sampling.jl
evanfields: Can you elaborate? A key (the main?) benefit of specifying the RNG is that you can get repeatable results. So this sort of check is ensuring that if you pass in the same source of randomness, you're getting back the same sample. This check did catch a bug for me - there were a few places I hadn't passed the specified RNG along to a `rand` call in a sampling method, and that caused these sorts of tests to fail.

I agree that it'd be a good idea to make sure that explicitly specifying `Base.GLOBAL_RNG` or not specifying a RNG gives the same output. Something like
```
srand(47)
x = randi(1000)
srand(47)
y = randi(Base.GLOBAL_RNG, 1000)
@test x == y
```

Which I will happily add. But that wouldn't catch the case where `randi` (or any other method) was ignoring the passed RNG and always calling `Base.GLOBAL_RNG`.

Or am I missing something obvious?

nalimilan: OK, got it. Maybe a comment would be useful? I'll leave it up to you whether it's worth adding the check using `Base.GLOBAL_RNG` too.

evanfields:
@nalimilan Thanks for the review - I think I've addressed your comments. Let me know if there's more to be done, of course.

src/sampling.jl
```diff
@@ -504,9 +508,11 @@ naive_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray) =
 #     URL http://www.sciencedirect.com/science/article/pii/S002001900500298X
 #
 # Instead of keys u^(1/w) where u = random(0,1) keys w/v where v = randexp(1) are used.
-function efraimidis_a_wsample_norep!(rng::AbstractRNG, a::AbstractArray, wv::WeightVec, x::AbstractArray)
+function efraimidis_a_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
+                                     wv::WeightVec, x::AbstractArray)
     n = length(a)
-    length(wv) == n || throw(DimensionMismatch("a and wv must be of same length (got $n and $(length(wv)))."))
+    length(wv) == n || throw(DimensionMismatch(
+	    "a and wv must be of same length (got $n and $(length(wv)))."))
```
nalimilan: This line is unrelated, please leave it as-is. (Also, in this case, the break should be done after the `||`, or the long `if.. end` form should be used.)

src/sampling.jl
```diff
@@ -507,24 +540,26 @@ end
 #     URL http://www.sciencedirect.com/science/article/pii/S002001900500298X
 #
 # Instead of keys u^(1/w) where u = random(0,1) keys w/v where v = randexp(1) are used.
-function efraimidis_ares_wsample_norep!(a::AbstractArray, wv::WeightVec, x::AbstractArray)
+function efraimidis_ares_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
+                                        wv::WeightVec, x::AbstractArray)
     n = length(a)
-    length(wv) == n || throw(DimensionMismatch("a and wv must be of same length (got $n and $(length(wv)))."))
+    length(wv) == n || throw(DimensionMismatch(
```
nalimilan: Same for this one.

test/sampling.jl
```diff
@@ -5,6 +5,26 @@ import StatsBase: norepeat, randi
...
+# a) if the same rng is passed to a sample function twice, 
+#    the results should be the same (repeatability)
+# b) not specifying a rng should be the same as specifying Base.GLOBAL_RNG
+function test_rng_use(func, non_rng_args...)
+    # some sampling methods mutate a passed array and return it
+    # so that the tests don't pass trivially, we need to copy those
+    # pre-allocated storage arrays
+
+    # repeatability
+    @test func(MersenneTwister(1), deepcopy(non_rng_args)...) ==
```
nalimilan: `copy` should be enough, right?

nalimilan:
Thanks. I've noted a few details. You'll also have to rebase against master as a recently merged PR touched that code. Sorry about that.

test/sampling.jl
evanfields: I don't think `copy` is enough. Suppose `func` is one of the sampling methods that writes into a passed array. If just `copy` is used, both the left hand side and right hand side `func` calls will be writing into the same array. So the equality test just ends up testing that an array is equal to itself. For example:

```
using Base.Test

function test_rng_use(func, non_rng_args...)
    # repeatability
    @test func(MersenneTwister(1), copy(non_rng_args)...) ==
          func(MersenneTwister(1), copy(non_rng_args)...)
end

# should fail
function ignores_rng!(r::AbstractRNG, x::Vector{Float64})
	for i in eachindex(x)
		x[i] = rand()
	end
	return x
end
ignores_rng!(x) = ignores_rng!(Base.GLOBAL_RNG, x)

test_rng_use(ignores_rng!, [1.0, 2.0])
```
run in Julia:
```
julia> include("C:\\Users\\Evan\\Desktop\\randtest.jl")
WARNING: copy(x::Tuple) is deprecated, use identity(x) instead.
 in depwarn(::String, ::Symbol) at .\deprecated.jl:64
 in copy(::Tuple{Array{Float64,1}}) at .\deprecated.jl:50
 in test_rng_use(::Function, ::Array{Float64,1}, ::Vararg{Array{Float64,1},N}) at C:\Users\Evan\Desktop\randtest.jl:5
 in include_from_node1(::String) at .\loading.jl:488
 in eval(::Module, ::Any) at .\boot.jl:234
 in eval_user_input(::Any, ::Base.REPL.REPLBackend) at .\REPL.jl:64
 in macro expansion at .\REPL.jl:95 [inlined]
 in (::Base.REPL.##3#4{Base.REPL.REPLBackend})() at .\event.jl:68
while loading C:\Users\Evan\Desktop\randtest.jl, in expression starting on line 18
Test Passed
  Expression: func(MersenneTwister(1),copy(non_rng_args)...) == func(MersenneTwister(1),copy(non_rng_args)...)
   Evaluated: [0.367674,0.409404] == [0.367674,0.409404]

julia>
```

src/sampling.jl
evanfields: Sorry for being dense. `throw(DimensionMismatch("a and wv must be of same length (got $n and $(length(wv)))."))` is 87 characters. Whether the current line is broken after `||` or a `if` block is used, that has to be preceded by 8 characters of indentation. 8 + 87 > 92, so the `throw` call must be broken across a line, right? Where is the natural place? (I'm happy to do whatever you tell me, not sure what the fix is here. Or you can push to the PR, right?)

Or define the string on a different line?
```
if length(wv) != n
    s = "a and wv must be of same length (got $n and $(length(wv)))."
    throw(DimensionMismatch(s))
end
```

nalimilan: Yeah, I don't think there's a standard solution for handling long strings. One of them is to use triple-quoted strings. Anyway since that's unrelated to this PR, leave it alone (for now at least).

test/sampling.jl
nalimilan: Sorry, I hadn't noticed this was called on the tuple of arguments rather than directly on the array. That's OK then.

evanfields:
Pretty sure I did the rebase process wrong - I'm moderately terrible at git - but stuff looks right and Travis is running...

nalimilan:
Unfortunately, you got it wrong indeed. Use `git rebase -i master` and remove manually the unwanted commits.

evanfields:
Shame and apologies. :/ There are now two copies of each commit from this branch, e.g. ` f50a88f optionally specify sampling rng via kwarg` and `0a27653 optionally specify sampling rng via kwarg`. How do I tell which is the right copy to retain?

nalimilan:
No idea, but the one above is f50a88f.`git show f50a88f` and `git show 0a27653` should allow comparing them.

evanfields:
I think this is fixed. Linear history, no duplicate commits, merges with master.

nalimilan:
Woops, I almost forgot that the signatures need to be updated in `docs/source/sampling.rst`.

src/sampling.jl
```diff
@@ -14,61 +14,64 @@
...
 end
+direct_sample!(a::AbstractArray, x::AbstractArray) = direct_sample!(Base.GLOBAL_RNG, a, x)
 
 ### draw a pair of distinct integers in [1:n]
 
 """
-    samplepair(n)
+    samplepair([rng], n)
 
 Draw a pair of distinct integers between 1 and `n` without replacement.
```
nalimilan: The docstring should mention what `rng` is (same for other functions).

nalimilan:
Thanks!

evanfields:
Thanks for all the help nalimilan and ararslan. Hopefully any future PRs will be smoother. Learning... :)

ararslan:
Thanks for the great contribution and for your perseverance!

PR #247, opened 2017-04-13
Move some `describe` code from DataFrames/Tables -> StatsBase

cjprybol:
DataFrames/Tables both implement their own functionality to extend `describe` and their functions overwrite one another and throw warnings https://github.com/JuliaData/DataTables.jl/issues/33. This moves some of the functionality:
- specify an IO source in `describe`
- return very basic info on non-numeric arrays (length, type, # unique)

to StatsBase. Next steps will be to move DataArrays functionality out of DataFrames and into DataArrays and the same for NullableArrays & DataTables.

This PR does not contain the code to count nulls and compute the % of nulls, which exists in DataFrames/Tables. I'll propose that I add that to this PR, but I thought I'd get feedback on the conservative changes first.

```julia
julia> using StatsBase
INFO: Recompiling stale cache file /Users/Cameron/.julia/lib/v0.6/StatsBase.ji for module StatsBase.

julia> describe(collect(1:10))
Summary Stats:
Mean:         5.500000
Minimum:      1.000000
1st Quartile: 3.250000
Median:       5.500000
3rd Quartile: 7.750000
Maximum:      10.000000

julia> describe(["s", "s", "s"])
Summary Stats:
Length:       3
Type:         String
Unique:       1
```

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -554,12 +554,12 @@ end
 
 function Base.show(io::IO, ss::SummaryStats)
     println(io, "Summary Stats:")
-    @printf(io, "Mean:         %.6f\n", ss.mean)
-    @printf(io, "Minimum:      %.6f\n", ss.min)
-    @printf(io, "1st Quartile: %.6f\n", ss.q25)
-    @printf(io, "Median:       %.6f\n", ss.median)
-    @printf(io, "3rd Quartile: %.6f\n", ss.q75)
-    @printf(io, "Maximum:      %.6f\n", ss.max)
+    @printf(io, "Mean:           %.6f\n", ss.mean)
+    @printf(io, "Minimum:        %.6f\n", ss.min)
+    @printf(io, "1st Quartile:   %.6f\n", ss.q25)
+    @printf(io, "Median:         %.6f\n", ss.median)
+    @printf(io, "3rd Quartile:   %.6f\n", ss.q75)
+    @printf(io, "Maximum:        %.6f\n", ss.max)
 end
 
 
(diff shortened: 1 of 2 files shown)


src/scalarstats.jl
```diff
@@ -570,4 +570,11 @@ Pretty-print the summary statistics provided by `summarystats`:
 the mean, minimum, 25th percentile, median, 75th percentile, and
 maximum.
 """
-describe{T<:Real}(a::AbstractArray{T}) = show(summarystats(a))
+describe(a::AbstractArray) = describe(STDOUT, a)
+describe{T<:Real}(io::IO, a::AbstractArray{T}) = show(io, summarystats(a))
+function describe{T<:Any}(io::IO, a::AbstractArray{T})
```
ararslan: `{T}` is the same as `{T<:Any}`, but you don't actually need the typevar in this case, since you don't care what's in `a`, just that it's an `AbstractArray`, and the typevar isn't referenced in the function body.

ararslan:
> This PR does not contain the code to count nulls and compute the % of nulls, which exists in DataFrames/Tables. I'll propose that I add that to this PR

Since there isn't even a unified way to identify nulls, I think that kind of functionality is better left to the packages that provide null functionality (i.e. DataArrays and NullableArrays or DataFrames/Tables).

cjprybol:
Sounds good to me. Thanks for the review Alex!

ararslan:
Thanks for the contribution! Could you add some tests?

cjprybol:
Happy to!

nalimilan:
Thanks for doing this. Though I'm not sure we need the basic printing for non-real arrays: if did that, it would be more consistent to also print that information (length, type, uniques) for real arrays. Unless we're certain we want this, I'd rather not have any fallback.

cjprybol:
We could keep the non-real, non-nullable describe code in Tables/Frames and move it to AbstractTables in the future? If we leave it as it currently is (a called function) then we won't fix https://github.com/JuliaData/DataTables.jl/issues/33, but if we write it inside of the Tables/Frames function then that would fix 33. We could check if `eltype !<: Nullable && eltype !<: Real` or use a try-catch where we handle the columns for which describe isn't defined. Here's a modified form of the DataTables implementation as an example.

```julia
function StatsBase.describe(io, dt::AbstractDataTable)
    for (name, col) in eachcol(dt)
        println(io, name)
        try
            describe(io, col)
        catch
            println(io, "Summary Stats:")
            println(io, "Length:       $(length(a))")
            println(io, "Type:         $(string(eltype(a)))")
            println(io, "Unique:       $(length(unique(a)))")
        end
        println(io, )
    end
end
```

we could scrap the non-real describe in the NullableArrays PR too if we put the `Number Missing` and `% Missing` fields in the try catch block as well. Open to other ideas too.

nalimilan:
OK, I'm fine with keeping the non-real methods, but then ensure that information regarding length and element type is printed for both real and non-real vectors. Only the number of unique values should be specific to non-real (just like summary stats are specific to real).

PR #248, opened 2017-04-20
Made `WeightVec` a subtype of `RealVector`

Rory-Finnegan:
Seems like `WeightVec` should support the same functionality as regular vectors.

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -1,24 +1,34 @@
 
 ###### Weight vector #####
 
-immutable WeightVec{T, V<:RealVector, S} <: RealVector{T}
-    values::V
-    sum::S
-end
+if VERSION < v"0.6.0-dev.2123"
+    immutable WeightVec{S<:Real, T<:Real, V<:RealVector} <: RealVector{T}
+        values::V
+        sum::S
+    end
 
-"""
-    WeightVec(vs, [wsum])
+    function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
+        return WeightVec{S, eltype(vs), V}(vs, s)
+    end
 
-Construct a `WeightVec` with weight values `vs` and sum of weights `wsum`.
-If omitted, `wsum` is computed.
-"""
-function WeightVec{V<:RealVector}(vs::V)
-    sum_ = sum(vs)
-    return WeightVec{eltype(vs), V, typeof(sum_)}(vs, sum_)
-end
+    function WeightVec{V<:RealVector}(vs::V)
+        sum_ = sum(vs)
+        return WeightVec{typeof(sum_), eltype(vs), V}(vs, sum_)
+    end
+else
+    immutable WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractVector{T}
+        values::V
+        sum::S
+    end
 
-function WeightVec{S, V<:RealVector}(vs::V, s::S)
-    return WeightVec{eltype(vs), V, S}(vs, s)
+    function WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}}(vs::V, s::S)
+        return WeightVec{S, T, V}(vs, s)
+    end
+
+    function WeightVec{T<:Real, V<:AbstractVector{T}}(vs::V)
+        sum_ = sum(vs)
+        return WeightVec{typeof(sum_), T, V}(vs, sum_)
+    end
 end
 
 """
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -1,9 +1,9 @@
 
 ###### Weight vector #####
 
-immutable WeightVec{W,Vec<:RealVector}
-    values::Vec
-    sum::W
+immutable WeightVec{T, V<:RealVector, S} <: RealVector{T}
+    values::V
+    sum::S
 end
 
 """
@@ -12,8 +12,14 @@ end
 Construct a `WeightVec` with weight values `vs` and sum of weights `wsum`.
 If omitted, `wsum` is computed.
 """
-WeightVec{Vec<:RealVector,W<:Real}(vs::Vec,wsum::W) = WeightVec{W,Vec}(vs, wsum)
-WeightVec(vs::RealVector) = WeightVec(vs, sum(vs))
+function WeightVec{V<:RealVector}(vs::V)
+    sum_ = sum(vs)
+    return WeightVec{eltype(vs), V, typeof(sum_)}(vs, sum_)
+end
+
+function WeightVec{S, V<:RealVector}(vs::V, s::S)
+    return WeightVec{eltype(vs), V, S}(vs, s)
+end
 
 """
     weights(vs)
@@ -30,6 +36,7 @@ sum(wv::WeightVec) = wv.sum
 isempty(wv::WeightVec) = isempty(wv.values)
 
 Base.getindex(wv::WeightVec, i) = getindex(wv.values, i)
+Base.size(wv::WeightVec) = size(wv.values)
 
 
 ##### Weighted sum #####
@@ -281,6 +288,9 @@ Base.mean{T<:Number,W<:Real}(A::AbstractArray{T}, w::WeightVec{W}, dim::Int) =
 
 
 ###### Weighted median #####
+function Base.median(v::AbstractArray, w::WeightVec)
+    throw(MethodError(median, (typeof(v), typeof(w))))
+end
 
 function Base.median{W<:Real}(v::RealVector, w::WeightVec{W})
     isempty(v) && error("median of an empty array is undefined")
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -6,29 +6,26 @@ if VERSION < v"0.6.0-dev.2123"
         values::V
         sum::S
     end
-
-    function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
-        return WeightVec{S, eltype(vs), V}(vs, s)
-    end
-
-    function WeightVec{V<:RealVector}(vs::V)
-        sum_ = sum(vs)
-        return WeightVec{typeof(sum_), eltype(vs), V}(vs, sum_)
-    end
 else
     immutable WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractVector{T}
         values::V
         sum::S
     end
+end
 
-    function WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}}(vs::V, s::S)
-        return WeightVec{S, T, V}(vs, s)
-    end
+"""
+    function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
 
-    function WeightVec{T<:Real, V<:AbstractVector{T}}(vs::V)
-        sum_ = sum(vs)
-        return WeightVec{typeof(sum_), T, V}(vs, sum_)
-    end
+Construct a `WeightVec` with weight values `vs` and sum of weights `wsum`.
+If omitted, `wsum` is computed.
+"""
+function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
+    return WeightVec{S, eltype(vs), V}(vs, s)
+end
+
+function WeightVec{V<:RealVector}(vs::V)
+    s = sum(vs)
+    return WeightVec{typeof(s), eltype(vs), V}(vs, s)
 end
 
 """
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -291,7 +291,7 @@ Base.mean{T<:Number,W<:Real}(A::AbstractArray{T}, w::WeightVec{W}, dim::Int) =
 
 ###### Weighted median #####
 function Base.median(v::AbstractArray, w::WeightVec)
-    throw(MethodError(median, (typeof(v), typeof(w))))
+    throw(MethodError(median, (v, w)))
 end
 
 function Base.median{W<:Real}(v::RealVector, w::WeightVec{W})
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -14,18 +14,13 @@ else
 end
 
 """
-    function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
+    WeightVec(vs, [wsum])
 
 Construct a `WeightVec` with weight values `vs` and sum of weights `wsum`.
 If omitted, `wsum` is computed.
 """
-function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
+function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S=sum(vs))
     return WeightVec{S, eltype(vs), V}(vs, s)
-end
-
-function WeightVec{V<:RealVector}(vs::V)
-    s = sum(vs)
-    return WeightVec{typeof(s), eltype(vs), V}(vs, s)
 end
 
 """
--- a/test/weights.jl
+++ b/test/weights.jl
@@ -5,10 +5,12 @@ import Compat: view
 
 @test isa(weights([1, 2, 3]), WeightVec{Int})
 @test isa(weights([1., 2., 3.]), WeightVec{Float64})
-
 @test isa(weights([1 2 3; 4 5 6]), WeightVec{Int})
 
+@test isa(WeightVec([1, 2, 3], 6), WeightVec{Int})
+
 @test isempty(weights(Float64[]))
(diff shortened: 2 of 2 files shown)


src/weights.jl
```diff
@@ -281,6 +288,9 @@ Base.mean{T<:Number,W<:Real}(A::AbstractArray{T}, w::WeightVec{W}, dim::Int) =
 
 
 ###### Weighted median #####
+function Base.median(v::AbstractArray, w::WeightVec)
```
ararslan: Shouldn't this method be unnecessary since all you're doing is throwing a `MethodError`? Presumably without the method, it would also be a `MethodError`.

ararslan:
It would be great if you could add tests for this.

src/weights.jl
Rory-Finnegan: I needed to added this because [this](https://github.com/JuliaStats/StatsBase.jl/blob/master/test/weights.jl#L239) test was failing with a `BoundsError` instead as it was able to dispatch to a method in base.

```julia
julia> median([4 3 2 1 0], weights(wt))
ERROR: BoundsError: attempt to access 2-element Array{Any,1} at index [3]
 in mapslices(::Base.#median!, ::Array{Int64,2}, ::StatsBase.WeightVec{Int64,Array{Int64,1},Int64}) at ./abstractarray.jl:1619
 in median(::Array{Int64,2}, ::StatsBase.WeightVec{Int64,Array{Int64,1},Int64}) at ./statistics.jl:579```

I figured it was better to manually throw a `MethodError` rather than letting it hit the `BoundsError` in that case.

Comment (ararslan) [inline src/weights.jl]: Ah, I see. This is fine then. Thanks for clarifying.

File: src/weights.jl [hunk 2]
```diff
@@ -1,9 +1,9 @@
 
 ###### Weight vector #####
 
-immutable WeightVec{W,Vec<:RealVector}
-    values::Vec
-    sum::W
+immutable WeightVec{T, V<:RealVector, S} <: RealVector{T}
```
nalimilan: Better do `WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}} <: RealVector{T}`. That should simplify constructors since the parameter could be inferred from the input arguments.

andreasnoack: Is it really useful to allow `typeof(sum)` and `eltype(values)` to be different? Having three parameters for such a simple type seems overly complicated.

nalimilan: I wondered about this too, then I realized that it *might* be useful to allow storing weights using a small type (`UInt8`, `Float16`) to save space, and yet be able to store the sum in a type that won't overflow or lose precision too easily (`Int64`, `BigInt`, `Float64`).

Rory-Finnegan: My view was that all of these could be useful in the cases like `WeightVec{Bool, NullableArray{Bool}, Int}`, but I figured you'd be more likely to want to dispatch on `V` than `S`.

Rory-Finnegan: @nalimilan Isn't `WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractVector{T}` triangular dispatch which is only supported in 0.6?

Rory-Finnegan: @nalimilan I've updated the parameterization to match your suggestion, but it needed version check depending on whether the julia version supports triangular dispatch (implemented in [this PR](https://github.com/JuliaLang/julia/pull/18457)).

src/weights.jl
```diff
@@ -1,19 +1,35 @@
...
+        values::V
+        sum::S
+    end
+
+    function WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}}(vs::V, s::S)
+        return WeightVec{S, T, V}(vs, s)
+    end
+
+    function WeightVec{T<:Real, V<:AbstractVector{T}}(vs::V)
+        sum_ = sum(vs)
```
nalimilan: The trailing underscore is a bit weird. Why not use `s` or any other simple name?

src/weights.jl
```diff
@@ -1,19 +1,35 @@
...
+        return WeightVec{typeof(sum_), eltype(vs), V}(vs, sum_)
+    end
+else
+    immutable WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractVector{T}
+        values::V
+        sum::S
+    end
+
+    function WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}}(vs::V, s::S)
```
nalimilan: You should be able to move this function and the next one out of the version-dependent block using `V<:RealVector` and replacing `T` with `eltype(vs)`. That is, just slightly adapting the existing constructors.

src/weights.jl
Rory-Finnegan: This is mostly just habit as I try to avoid using single character variable names. For math and stats code I tend to use the name of the function name with an underscore as my variable names if I can't think of anything better.

src/weights.jl
```diff
@@ -1,19 +1,32 @@
...
+else
+    immutable WeightVec{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractVector{T}
+        values::V
+        sum::S
+    end
 end
 
 """
-    WeightVec(vs, [wsum])
+    function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
```
nalimilan: This change is incorrect, the signature should just reflect how the user can call the function, not how it's implemented.

src/weights.jl
```diff
@@ -1,19 +1,32 @@
...
 
 Construct a `WeightVec` with weight values `vs` and sum of weights `wsum`.
 If omitted, `wsum` is computed.
 """
-WeightVec{Vec<:RealVector,W<:Real}(vs::Vec,wsum::W) = WeightVec{W,Vec}(vs, wsum)
-WeightVec(vs::RealVector) = WeightVec(vs, sum(vs))
+function WeightVec{S<:Real, V<:RealVector}(vs::V, s::S)
```
nalimilan: You can actually use `s::S=sum(vs)` to get rid of the next method.

Rory-Finnegan: Interesting, I didn't realize that was valid syntax in 0.5 (ie: referencing an earlier argument when setting the default value).

src/weights.jl
```diff
@@ -281,6 +290,9 @@ Base.mean{T<:Number,W<:Real}(A::AbstractArray{T}, w::WeightVec{W}, dim::Int) =
 
 
 ###### Weighted median #####
+function Base.median(v::AbstractArray, w::WeightVec)
+    throw(MethodError(median, (typeof(v), typeof(w))))
```
nalimilan: Hadn't noticed that one. Should pass `(v, w)` instead of their types. Otherwise, looks good to me.

src/weights.jl
nalimilan: That said, the current behavior isn't that different:
```julia
julia> median([4 3 2 1 0], weights([1,2,3,4,5]))
ERROR: MethodError: no method matching start(::StatsBase.WeightVec{Int64,Array{Int64,1}})
Closest candidates are:
  start(::SimpleVector) at essentials.jl:259
  start(::Base.MethodList) at reflection.jl:560
  start(::ExponentialBackOff) at error.jl:107
  ...
Stacktrace:
 [1] append_any(::StatsBase.WeightVec{Int64,Array{Int64,1}}, ::Vararg{StatsBase.WeightVec{Int64,Array{Int64,1}},N} where N) at ./essentials.jl:170
 [2] median(::Array{Int64,2}, ::StatsBase.WeightVec{Int64,Array{Int64,1}}) at ./statistics.jl:619

```

So I'm not sure this `MethodError` is really needed. We typically don't throw them for all possible cases that fail.

Rory-Finnegan:
Should I squash my commits or is a more verbose history preferred here?

ararslan:
I'm pretty sure this change is what broke DataArrays. It's throwing a `MethodError` when passed an object constructed with `weights()`.

rofinn:
Is this the error you're talking about?
```
ERROR: MethodError: Cannot `convert` an object of type DataArrays.NAtype to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
 in setindex!(::Array{Float64,1}, ::DataArrays.NAtype, ::Int64) at ./array.jl:415
 in copy!(::Array{Float64,1}, ::Tuple{DataArrays.NAtype,Float64,Float64}) at ./abstractarray.jl:487
 in vect(::DataArrays.NAtype, ::Vararg{Any,N}) at ./abstractarray.jl:27
```
Yeah, I'm not sure what the best solution is cause `DataArrays.NAtype` isn't even a subtype of `Real`. We could probably loosen the constraints on the `WeightVec` construction.

nalimilan:
Can you post the code which triggers the failure?

ararslan:
From the DataArrays tests:

```julia
da1 = DataArray(randn(128))
da2 = DataArray(randn(128))
da1[1:3:end] = NA
da2[1:2:end] = NA
mean(da1, weights(da2))
```
The last line is now a `MethodError`:
```
julia> mean(da1, weights(da2))
ERROR: MethodError: no method matching StatsBase.WeightVec(::DataArrays.DataArray{Float64,1}, ::DataArrays.NAtype)
Closest candidates are:
  StatsBase.WeightVec(::V<:(AbstractArray{T,1} where T<:Real)) where V<:(AbstractArray{T,1} where T<:Real) at /Users/alex/.julia/v0.7/StatsBase/src/weights.jl:23
  StatsBase.WeightVec(::V<:(AbstractArray{T,1} where T<:Real), ::S<:Real) where {S<:Real, V<:(AbstractArray{T,1} where T<:Real)} at /Users/alex/.julia/v0.7/StatsBase/src/weights.jl:23
Stacktrace:
 [1] weights(::DataArrays.DataArray{Float64,1}) at /Users/alex/.julia/v0.7/StatsBase/src/weights.jl:31
```

rofinn:
@ararslan That makes sense cause `DataArray.NAtype` isn't a subtype of `Real` and `sum(da2)` returns `NA`.

rofinn:
Removing the restriction `S<:Real` would probably fix this.

ararslan:
How did that ever work then?

nalimilan:
OK. I guess we could remove the `S<:Real` restriction since we don't really need it. But I'm not actually sure this is a use case we want to support: when do you want to use a missing weight? One can always use a zero weight to ignore an observation.

ararslan:
Perhaps, but I'm not sure it's worth loosening that restriction; I don't think it makes sense to have a weight vector whose sum is missing.

ararslan:
Perhaps it would make more sense to require that the input to a weight constructor be a `Base.Array` rather than any `AbstractArray{<:Real}`?

nalimilan:
Well, there's no point in restricting the type of underlying array. `DataArray{T}` is cheating by pretending that its eltype is `T` when it's actually `Union{NAtype, T}`.

Let's just remove that test from DataArrays if we don't have a use case for it.

rofinn:
@ararslan To answer your specific question, it was initially `WeightVec{W,Vec<:RealVector}` and this PR changed it to `WeightVec{S<:Real, T<:Real, V<:RealArray}` as I thought that made more sense. I tend to agree that having a weight vector whose sum is missing doesn't really make sense.

rofinn:
Seems like a better approach for that use case would be to use `dropna` before this?

ararslan:
> Seems like a better approach for that use case would be to use dropna before this?

Not necessarily, since you can `dropna` from your weight vector but you may have reason for legitimately missing values in your data vector, in which case you don't want to `dropna` from that, and then the vectors have differing lengths, which is an error.

ararslan:
> Let's just remove that test from DataArrays if we don't have a use case for it.

Okay ¯\\\_(ツ)\_/¯

ararslan:
Cool, DataArrays is all good now.

However, it appears this change still introduced ambiguities with Base methods:
```
julia> using StatsBase

julia> using Base.Test

julia> detect_ambiguities(StatsBase, Base, Core)
Skipping StatsBase.hist
Skipping StatsBase.wmean!
Skipping Base.<|
3-element Array{Tuple{Method,Method},1}:
 (stdm(v::AbstractArray{T,N} where N where T<:Real, wv::StatsBase.WeightVec, m::AbstractArray{T,N} where N where T<:Real, dim::Int64) in StatsBase at /Users/alex/.julia/v0.7/StatsBase/src/moments.jl:99, stdm(v::AbstractArray{T,N} where N where T<:Real, m::AbstractArray{T,N} where N where T<:Real, wv::StatsBase.WeightVec, dim::Int64) in StatsBase at deprecated.jl:50)
 (varm(A::AbstractArray{T,N} where N where T<:Real, wv::StatsBase.WeightVec, M::AbstractArray{T,N} where N where T<:Real, dim::Int64) in StatsBase at /Users/alex/.julia/v0.7/StatsBase/src/moments.jl:65, varm(A::AbstractArray{T,N} where N where T<:Real, M::AbstractArray{T,N} where N where T<:Real, wv::StatsBase.WeightVec, dim::Int64) in StatsBase at deprecated.jl:50)
 (stdm(v::AbstractArray{T,N} where N where T<:Real, wv::StatsBase.WeightVec, m::Real) in StatsBase at /Users/alex/.julia/v0.7/StatsBase/src/moments.jl:87, stdm(v::AbstractArray{T,N} where N where T<:Real, m::AbstractArray{T,N} where N where T<:Real, dim::Int64) in StatsBase at /Users/alex/.julia/v0.7/StatsBase/src/moments.jl:98)
```

rofinn:
Shouldn't the first two go away once those deprecations are removed? Not sure about the last one though.

ararslan:
Yeah the first two aren't too worrisome, it's mostly the last one. This previously wasn't a problem because we didn't have `WeightVec <: AbstractVector`. Not sure what to do about it now.

rofinn:
Could we do something like `Base.stdm(v::RealVector, w::AbstractWeights, m::Real)` and `Base.stdm(v::RealMatrix, m::RealArray, dim::Int)`?

ararslan:
We can't overload a Base function with Base types though; this is the sinful act of type piracy.

rofinn:
All the more reason for it not to be a Base function? ;)

rofinn:
Wait also does this even count as type piracy cause we're overloading our own behaviour and not base?

ararslan:
If we extend a Base function using Base types, regardless of how we define the behavior, we've committed ~~an act of treason~~ type piracy.

rofinn:
Well in that case we're already committing type piracy `Base.stdm(v::RealArray, m::RealArray, dim::Int)`.

ararslan:
Oh dang you're right. If we get rid of those we should be okay on ambiguities. We should commence... _the purge_.

nalimilan:
Indeed it's surprising that we ship this method. It's not even documented in StatsBase. Let's remove it, the ambiguities due to deprecations are OK (though they shouldn't be hard to fix if we wanted to).

rofinn:
I think it makes sense to either keep that method or add it to Base which already has a `varm(A::AbstractArray{T,N} where N, m::AbstractArray, region; corrected)`, but no corresponding method for `stdm`.

nalimilan:
I guess we could add it to Base. But since it's not exported, it's not really an issue whether a method is missing or not.

rofinn:
So, it is exported though. https://github.com/JuliaLang/julia/blob/98649ed087d910540bd9c311a4e650a74b5c54f6/base/exports.jl#L861

nalimilan:
Ah, cool. I extrapolated incorrectly from `covm`, which isn't exported.

ararslan:
That wouldn't help us for 0.5 and 0.6 though, since if we add the method to Base then it will only be available in 0.7+.

ararslan:
We can just define an internal function that behaves like `Base.stdm` with the appropriate methods, assuming the currently pirated method is being used in any significant way in the package. Otherwise it could just be a simple refactoring.

rofinn:
As far as i can tell it's only being used [here](https://github.com/JuliaStats/StatsBase.jl/blob/master/src/moments.jl#L127), so we could probably rename it to `_stdm` and change that one reference for now and deprecate the previous method in case anyone was using it. I still think the correct long term solution would be to move the base stats stuff into statsbase, but that sounds like a reasonable interim solution.

ararslan:
Not worth a deprecation IMO since it's a Base function.

nalimilan:
Certainly not worth a deprecation. I think the easiest solution is to define the method on older Julia versions using a `VERSION` check (or a `method_exists` one).

Issue #252, opened 2017-05-03
Return type not inferred in fit(Histogram) after v0.13.1 => v0.14.0


sbromberger:
Minimal example:
```
julia> using StatsBase, Base.Test

julia> v = rand(1:20, 200);

julia> @inferred(fit(Histogram, v))
WARNING: Default for keyword argument "closed" has changed from :right to :left.
Stacktrace:
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:64
 [2] _check_closed_arg at /Users/bromberger1/.julia/v0.6/StatsBase/src/hist.jl:11 [inlined]
 [3] #fit#66(::Symbol, ::Bool, ::Int64, ::Function, ::Type{StatsBase.Histogram{Int64,N,E} where E where N}, ::Tuple{Array{Int64,1}}) at /Users/bromberger1/.julia/v0.6/StatsBase/src/hist.jl:266
 [4] (::StatsBase.#kw##fit)(::Array{Any,1}, ::StatsBase.#fit, ::Type{StatsBase.Histogram{Int64,N,E} where E where N}, ::Tuple{Array{Int64,1}}) at ./<missing>:0
 [5] fit(::Type{StatsBase.Histogram{Int64,N,E} where E where N}, ::Array{Int64,1}) at /Users/bromberger1/.julia/v0.6/StatsBase/src/hist.jl:213
 [6] #fit#69(::Array{Any,1}, ::Function, ::Type{StatsBase.Histogram}, ::Array{Int64,1}, ::Vararg{Array{Int64,1},N} where N) at /Users/bromberger1/.julia/v0.6/StatsBase/src/hist.jl:278
 [7] fit(::Type{StatsBase.Histogram}, ::Array{Int64,1}) at /Users/bromberger1/.julia/v0.6/StatsBase/src/hist.jl:278
 [8] eval(::Module, ::Any) at ./boot.jl:235
 [9] eval_user_input(::Any, ::Base.REPL.REPLBackend) at ./REPL.jl:66
 [10] macro expansion at ./REPL.jl:97 [inlined]
 [11] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
while loading no file, in expression starting on line 0
[...]

PR #253, opened 2017-05-03
Fixes for type instabilities in histogram functions

oschulz:
Addresses #252.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -27,6 +27,14 @@ end
 end
 
 
+# Need a generated function to promote edge types, because a simple
+# promote_type(map(eltype, h.edges)...) isn't type stable (tested
+# with Julia v0.5).
+@generated function _promote_edge_types{N}(edges::NTuple{N,AbstractVector})
+    promote_type(map(eltype, edges.parameters)...)
+end
+
+
 ## nice-valued ranges for histograms
 function histrange{T}(v::AbstractArray{T}, n::Integer, closed::Symbol=:default_left)
     closed = _check_closed_arg(closed,:histrange)
@@ -134,7 +142,7 @@ type Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
                                                weights::Array{T,N}, closed::Symbol, isdensity::Bool=false)
         closed == :right || closed == :left || error("closed must :left or :right")
         isdensity && !(T <: AbstractFloat) && error("Density histogram must have float-type weights")
-        map(x -> length(x)-1,edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
+        _edges_nbins(edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
         new{T,N,E}(edges,weights,closed,isdensity)
     end
 end
@@ -143,7 +151,7 @@ Histogram{T,N}(edges::NTuple{N,AbstractVector},weights::AbstractArray{T,N},close
     Histogram{T,N,typeof(edges)}(edges,weights,_check_closed_arg(closed,:Histogram),isdensity)
 
 Histogram{T,N}(edges::NTuple{N,AbstractVector},::Type{T},closed::Symbol=:default_left, isdensity::Bool=false) =
-    Histogram(edges,zeros(T,map(x -> length(x)-1,edges)...),_check_closed_arg(closed,:Histogram),isdensity)
+    Histogram(edges,zeros(T,_edges_nbins(edges)...),_check_closed_arg(closed,:Histogram),isdensity)
 
 Histogram{N}(edges::NTuple{N,AbstractVector},closed::Symbol=:default_left, isdensity::Bool=false) =
     Histogram(edges,Int,_check_closed_arg(closed,:Histogram),isdensity)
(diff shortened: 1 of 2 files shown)


src/hist.jl
```diff
@@ -259,34 +272,42 @@ end
 append!{T,N}(h::AbstractHistogram{T,N}, vs::NTuple{N,AbstractVector}, wv::WeightVec) = append!(h, vs, values(wv))
 
 
+# Turn kwargs nbins into a type-stable tuple of integers:
+function _nbins_tuple{N}(vs::NTuple{N,AbstractVector}, nbins)
```
oschulz: Note: I tried to do this via dispatch instead of if-else, but that version made the inference engine crash on Julia v0.5 (worked on 0.6, though).

src/hist.jl
```diff
@@ -27,6 +27,14 @@ end
 end
 
 
+# Need a generated function to promote edge types, because a simple
+# promote_type(map(eltype, h.edges)...) isn't type stable (tested
+# with Julia v0.5).
+@generated function _promote_edge_types{N}(edges::NTuple{N,AbstractVector})
```
nalimilan: Have you tried with `Base.promote_eltype`?

src/hist.jl
nalimilan: Maybe it would be better to include the most natural dispatch-based code under a `if VERSION >= "0.6.0+XXXX"` block so that we can remove this hack in the future?

src/hist.jl
oschulz: Just tried with `Base.promote_eltype` - results in the same type instabilities as before.

src/hist.jl
oschulz: I wouldn't call it a hack - and it complicate things. Besides, this way we have a nice error message for the user if he tries to use something else than an integer or integer tuple for `nbins`.

oschulz: The dispatch-based version consisted of two helper functions, wasn't that pretty, either, anyhow.

src/hist.jl
nalimilan: Ah, too bad. That's surprising, since that function is used in Base.

src/hist.jl
nalimilan: OK. And wouldn't `broadcast((t, x) -> typeof(t)(x), template, nbins)` be enough to recycle an integer `nbins`?

src/hist.jl
oschulz: Is neither documented, not exported, though ...

src/hist.jl
oschulz: That would be an elegant solution. Not on Julia v0.5, though, because it doesn't support broadcast for tuples (because size isn't defined for tuples):

    ERROR: MethodError: no method matching size(::Tuple{Int64,Int64})

src/hist.jl
nalimilan: Yes, but since it is used I thought it was type-stable.

src/hist.jl
nalimilan: That's why checking the Julia version could make sense, so that the code is cleaner when we drop 0.5 support.

oschulz: Agreed - which version should I use, though? I have a `v"0.6.0-pre.alpha.206` here, and it supports broadcast for tuples. But if it should be `v"0.6.0+XXXX"` or `v"0.6.0-dev.XXXX"`, I'm not sure which number to use.

src/hist.jl
```diff
@@ -259,34 +272,48 @@ end
 append!{T,N}(h::AbstractHistogram{T,N}, vs::NTuple{N,AbstractVector}, wv::WeightVec) = append!(h, vs, values(wv))
 
 
+# Turn kwargs nbins into a type-stable tuple of integers:
+function _nbins_tuple{N}(vs::NTuple{N,AbstractVector}, nbins)
+    template = map(length, vs)
+
+    if VERSION < v"0.6.0-"
```
oschulz: Is this Ok? I don't know at what point Julia v0.6 started to support broadcast for tuples.

nalimilan: Use a version on which you know this works, for example the one you're using. What we should avoid is a too broad check which would trigger failures on old 0.6 versions.

nalimilan: Oh, and add `@static` so that the branch is resolved at compile time.

src/hist.jl
ararslan: You can find the commit that introduced broadcast for tuples and use ./contrib/commit-name.sh XXX where XXX is the commit SHA.

oschulz: Thanks for the tip with commit-name.sh, @ararslan!

oschulz: I think it's #18580

oschulz: Ok, done - I hope it's the right version, I think the current tuple broadcast (at least the first entry point, from `@less`) stems from #18580.

src/hist.jl
oschulz: I don't really understand why it's not (or why the `map` wasn't) - but I think we can live with the generated _promote_edge_types (esp. since it got rid of another generated function).

src/hist.jl
oschulz: Added @static - thanks!

oschulz:
Thanks, @nalimilan!

PR #256, opened 2017-05-08
Started work on a contributing guide.

rofinn:


CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+
+```julia
+# Yes
+Base.stdm(v::RealArray, w::AbstractWeights, m::Real; corrected::DepBool=nothing) =
+    sqrt(varm(v, w, m, corrected=depcheck(:stdm, corrected)))
+
+# No
+function Base.stdm(v::RealArray, w::AbstractWeights, m::Real; corrected::DepBool=nothing)
+    sqrt(varm(v, w, m, corrected=depcheck(:stdm, corrected)))
+end
```
ararslan: FWIW I'm actually fine with this, though I guess it's better to be consistent.

rofinn: Yeah, personally I prefer switching away from the short form once the line is too long, but most of the existing code appears to follow this style.

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+* Try to keep changes in PRs constrained if possible. This will reduce the amount of work for the reviewer and likely speed up the review process.
+* Using `git diff upstream/master` and `git diff --stat upstream/master` (assuming you've run `git remote add upstream https://github.com/JuliaStats/StatsBase.jl`) can be useful commands for comparing your branch against the upstream master before pushing.
+* Running tests locally with `Pkg.test("StatsBase")` before committing can be a helpful way of reducing noise.
+* [Coverage](https://github.com/JuliaCI/Coverage.jl) can be useful for checking if you've lowered the overall test coverage against [master](https://coveralls.io/r/JuliaStats/StatsBase.jl?branch=master)
+
+## Style
+
+### Summary
+
+* Use 4 spaces per indentation level, no tabs.
```
nalimilan: Can we just refer to the Julia style guide instead of repeating the rules here? That will prevent this one from getting outdated if the Julia one is improved.

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+
+    if mean == nothing
+        return varm(v, w, Base.mean(v, w); corrected=corrected)
+    else
+        return varm(v, w, mean; corrected=corrected)
+    end
+end
+```
+
+However, for more complicated methods it's helpful to be more explicit:
```
nalimilan: Honestly, I'm not sure everybody would agree on this one as it's pretty clear that the last line returns here. I wouldn't give this example.

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+* Include spaces after commas and around operators
+
+### Methods Definitions
+
+Short form method definitions when the method body only contains 1 statement:
+
+```julia
+# Yes
+Base.stdm(v::RealArray, w::AbstractWeights, m::Real; corrected::DepBool=nothing) =
+    sqrt(varm(v, w, m, corrected=depcheck(:stdm, corrected)))
```
nalimilan: Can you hide the `depcheck` hack? It's meant to be removed anyway, and the person who will remove it will probably forget to sync the style guide.

Same below.

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+            c += 1
+        end
+    end
+    c   # The `c` return is less obvious here
+end
+```
+
+### Documentation
+
+All exported functions should have a docstring associated with it.
```
nalimilan: should -> must :-)

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+A docstring should start with the method signature(s) followed by a description
```
nalimilan: Rather than repeating the rules here, refer to the manual: https://docs.julialang.org/en/stable/manual/documentation/

CONTRIBUTING.md
rofinn: Sure. I'm a fan of reiterating some of the big ones that don't seem likely to change.

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+## Reporting Issues
+
+* It's always helpful to start by reviewing existing issues is and if necessary referencing back to previous discussions.
+* Including minimal sample code (either to reproduce a bug or demonstrating desired behaviour) is greatly appreciated.
+* For bugs, it's helpful to checkout the latest development version (`Pkg.checkout("StatsBase")`) and post the output from `versioninfo()`.
+
+## Pull Requests (PR)s
+
+* Try to keep changes in PRs constrained if possible. This will reduce the amount of work for the reviewer and likely speed up the review process.
+* Using `git diff upstream/master` and `git diff --stat upstream/master` (assuming you've run `git remote add upstream https://github.com/JuliaStats/StatsBase.jl`) can be useful commands for comparing your branch against the upstream master before pushing.
```
ararslan: It depends on whether you're working from a fork and have your fork set to `origin`. I think it's more common to have the original repository set to be `origin` and your fork set to its own remote. As an example, in my case I use my GitHub username as the remote name for my fork.

Perhaps it's better to speak in generalities rather than specific Git commands, since everyone can have a slightly different setup and/or preferred workflow, and it can become confusing.

rofinn: Well I guess that depends on your workflow, but in either case you'll need to set the remote either for upstream or for your fork. I tend to clone my own fork rather than using `Pkg.checkout` for a few reasons (e.g., I don't accidentally do `git push origin <branch>` on a repo I don't own).

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+* Running tests locally with `Pkg.test("StatsBase")` before committing can be a helpful way of reducing noise.
+* [Coverage](https://github.com/JuliaCI/Coverage.jl) can be useful for checking if you've lowered the overall test coverage against [master](https://coveralls.io/r/JuliaStats/StatsBase.jl?branch=master)
```
rofinn: I don't think the coverage stats are visible in the PRs though.

CONTRIBUTING.md
ararslan: I agree regarding reiterating certain concrete points such as whitespace. (Partially because I care way too much about indentation.) I bet a significant number of contributors won't be bothered to follow a link, so it's a good idea to make sure they at least see a few key points.

CONTRIBUTING.md
ararslan: I avoid `Pkg` altogether when developing packages, tbh, though that's an aside.

CONTRIBUTING.md
ararslan: We have Coveralls set up as a PR status check, don't we?

rofinn: Nope, just travis.

ararslan: Oh, hm. I guess we must have turned it off at some point, though I don't know why we would have. Any idea, @nalimilan?

CONTRIBUTING.md
rofinn: This might be likely enough to change that it might be worth leaving it as a link in the summary.

CONTRIBUTING.md
ararslan: Something must be broken with Coveralls then, because
screen shot 2017-05-09 at 1 38 12 pm

CONTRIBUTING.md
nalimilan: The guidelines are also likely to be improved, so I'd prefer a conspicuously broken link than misleading outdated rules. Also I don't think the links will break that often, they are useful in many other contexts.

CONTRIBUTING.md
nalimilan: The Coveralls webhook doesn't appear to be enabled. I've tried adding it by copying what exists for CategoricalArrays, but apparently the secret token is incorrect. I don't understand why... Sometimes it appears one needs to log out and in for authentication to work. I'll try again later.

nalimilan:
Bump.

rofinn:
I'll try to make some time tomorrow to update this PR.

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Style
+
+### Summary
+
+* Use 4 spaces per indentation level, no tabs.
+* Try to adhere to the 92 character line length limit.
+* Use upper camel case convention for [modules](http://julia.readthedocs.org/en/latest/manual/modules/) and [types](http://julia.readthedocs.org/en/latest/manual/types/).
```
iblis17: these links is out-of-date now...

PR #257, opened 2017-05-08
Added a weighted `cor` method.

### LINKED ISSUE #255: Implementation of `cor(X, wv, ...)`
It might be nice if StatsBase provided a method for a weighted pearson correlation.

rofinn:
Closes #255

### DIFF
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -103,7 +103,7 @@ function Base.corm(x::DenseMatrix, mean, w::AbstractWeights, vardim::Int=1)
 end
 
 """
-    cor(X, w::AbstractWeights; mean=nothing, vardim=1)
+    cor(X, w::AbstractWeights, vardim=1)
 
 Compute the Pearson correlation matrix of `X` along the dimension
 `vardim` with a weighting `w` .
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -64,15 +64,6 @@ cov
 
 
 """
-    cor(X, w::AbstractWeights; mean=nothing, vardim=1)
-
-Compute the Pearson correlation matrix of `X` along the dimension
-`vardim` with a weighting `w` .
-"""
-cor
-
-
-"""
     mean_and_cov(x, [wv::AbstractWeights]; vardim=1, corrected=false) -> (mean, cov)
 
 Return the mean and covariance matrix as a tuple. A weighting
@@ -111,6 +102,12 @@ function Base.corm(x::DenseMatrix, mean, w::AbstractWeights, vardim::Int=1)
     Base.cov2cor!(c, s)
 end
 
+"""
+    cor(X, w::AbstractWeights; mean=nothing, vardim=1)
+
+Compute the Pearson correlation matrix of `X` along the dimension
+`vardim` with a weighting `w` .
+"""
 Base.cor(x::DenseMatrix, w::AbstractWeights, vardim::Int=1) =
     Base.corm(x, Base.mean(x, w, vardim), w, vardim)
 
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -64,6 +64,15 @@ cov
 
 
 """
+    cor(X, w::AbstractWeights; mean=nothing, vardim=1)
+
+Compute the Pearson correlation matrix of `X` along the dimension
+`vardim` with a weighting `w` .
+"""
+cor
+
+
+"""
     mean_and_cov(x, [wv::AbstractWeights]; vardim=1, corrected=false) -> (mean, cov)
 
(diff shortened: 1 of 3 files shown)


src/cov.jl
```diff
@@ -64,6 +64,15 @@ cov
 
 
 """
+    cor(X, w::AbstractWeights; mean=nothing, vardim=1)
+
+Compute the Pearson correlation matrix of `X` along the dimension
+`vardim` with a weighting `w` .
+"""
+cor
```
ararslan: Instead of having the docstring atop the bare object here, you might as we put it above the method definition that it corresponds to below, since it's documenting a particular method.

rofinn: Alright, I'm guessing we should update `cov` at some point then.

nalimilan:
Thanks. I wonder whether we should directly implement the corrected correlation: else we'll have to go through the same painful deprecation path if we add it and change the default to `corrected=true` for consistency with other functions. It's a bit weird that Base's `cor` doesn't implement the correction, but there may be good reasons for that.

It's not clear to me why the correlation is generally not corrected in statistical software, contrary to the covariance. Indeed, while R's `cor` uses the correction, Numpy's and MATLAB's `corcoeff` do not, even if Numpy's `cov` supports it (though not by default like Julia).

I've found a few references which seem to imply the correction isn't really needed:
https://stats.stackexchange.com/questions/220961/is-the-sample-correlation-coefficient-an-unbiased-estimator-of-the-population-co
https://stats.stackexchange.com/questions/148346/shrunken-r-vs-unbiased-r-estimators-of-rho

But another in-depth paper does not even mention the uncorrected estimate (and proposes much more complex corrections which are out of scope here):
https://ir.nctu.edu.tw/bitstream/11536/14082/1/000287537100002.pdf

rofinn:
> I've found a few references which seem to imply the correction isn't really needed:
https://stats.stackexchange.com/questions/220961/is-the-sample-correlation-coefficient-an-unbiased-estimator-of-the-population-co
https://stats.stackexchange.com/questions/148346/shrunken-r-vs-unbiased-r-estimators-of-rho

Yeah, I reviewed those posts when I noticed that there wasn't a corrected option in base. My inclination was that StatsBase should probably follow the behaviour of base julia and if folks want more complex bias correction maybe that should be its own package correlation package (which could also support other types of correlation... not just Pearson).

nalimilan:
Let's just wait for a bit to see whether somebody remembers why this design decision was made in Base, and merge if not.

andreasnoack:
> It's a bit weird that Base's `cor` doesn't implement the correction

Which correction?

> Indeed, while R's `cor` uses the correction

Where do you see that? It looks very similar to ours.

```julia
julia> x, y = randn(5), randn(5);

julia> R"cor($x,$y)"[1] - cor(x,y)
2.220446049250313e-16
```

I think there is no consensus on a small sample correction here because there is no unbiased estimator in the normal case.

nalimilan:
Good catch, I need to learn to use RCall more often. Unfortunately, R documents both `cor` and `cov` in the same page, which says this:
```
     The denominator n - 1 is used which gives an unbiased estimator of
     the (co)variance for i.i.d. observations.
```

Apparently this only applies to the covariance. Also, the code is in C, so I didn't bother to check it. Anyway it's good to know that it doesn't apply a correction either. So we can merge this and we'll be consistent with major existing software.

src/cov.jl
```diff
@@ -96,6 +96,20 @@ Base.covm(x::DenseMatrix, mean, w::AbstractWeights, vardim::Int=1;
...
     Base.covm(x, Base.mean(x, w, vardim), w, vardim; corrected=depcheck(:cov, corrected))
 
+function Base.corm(x::DenseMatrix, mean, w::AbstractWeights, vardim::Int=1)
+    c = Base.covm(x, mean, w, vardim; corrected=false)
+    s = Base.stdm(x, w, mean, vardim; corrected=false)
+    Base.cov2cor!(c, s)
+end
+
+"""
+    cor(X, w::AbstractWeights; mean=nothing, vardim=1)
```
nalimilan: Ah, I knew I could find something to change: remove `; mean=nothing`. :-)

andreasnoack:
Notice that sample size factors cancel in the correlation computation so adjusting with `n` or `n-1` gives the exact same correlation estimate.

src/cov.jl
rofinn: Oh darn, `vardim` also shouldn't be a keyword arg :(

nalimilan:
Ah, and you'll have to add the new docstring to the corresponding .rst file.

PR #258, opened 2017-05-09
Add ambiguity detection to the package tests

ararslan:


ararslan:
I'm not entirely sure I've done this correctly, so it would be awesome if anyone who understands this better could take a look. cc @tkelman

ararslan:
Actually, I think it should just be
```julia
@test isempty(detect_ambiguities(StatsBase, Base, Core))
```

PR #261, opened 2017-05-09
Added `cor2cov` and `cov2cor` methods.

### LINKED ISSUE #246: Correlation to Covariance
It would be useful to have a function in StatsBase that allows us to define a distribution with the correlation matrix or to go from the correlation matrix to the covariance matrix.

Matlab has a nice function for that, corr2cov:
https://www.mathworks.com/help/finance/corr2cov.html

If something like this already exists in Julia, please let me know!

rofinn:
Closes #246

### DIFF
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -120,3 +120,23 @@ function mean_and_cov(x::DenseMatrix, wv::AbstractWeights, vardim::Int=1;
     m = mean(x, wv, vardim)
     return m, Base.cov(x, wv, vardim; corrected=depcheck(:mean_and_cov, corrected))
 end
+
+"""
+    cov2cor(C, s)
+
+Compute the correlation matrix from the covariance matrix `C` and a vector of standard
+deviations `s`.
+"""
+cov2cor(C::DenseMatrix, s::AbstractArray) = Base.cov2cor!(copy(C), vec(s))
+
+"""
+    cor2cov(C, s)
+
+Compute the covariance matrix from the correlation matrix `C` and a vector of standard
+deviations `s`.
+"""
+function cor2cov(C::DenseMatrix, s::AbstractArray)
+    # This could be optimized with for loops if necessary
+    D = Diagonal(vec(s))
+    D * C * D
+end
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -125,18 +125,29 @@ end
     cov2cor(C, s)
 
 Compute the correlation matrix from the covariance matrix `C` and a vector of standard
-deviations `s`.
+deviations `s`. Use `Base.cov2cor!` for an in-place version.
 """
-cov2cor(C::AbstractMatrix, s::AbstractArray) = Base.cov2cor!(copy(C), vec(s))
+cov2cor(C::AbstractMatrix, s::AbstractArray) = Base.cov2cor!(copy(C), s)
 
 """
     cor2cov(C, s)
 
 Compute the covariance matrix from the correlation matrix `C` and a vector of standard
-deviations `s`.
+deviations `s`. Use `StatsBase.cor2cov!` for an in-place version.
 """
-function cor2cov(C::AbstractMatrix, s::AbstractArray)
-    # This could be optimized with for loops if necessary
-    D = Diagonal(vec(s))
-    D * C * D
+cor2cov(C::AbstractMatrix, s::AbstractArray) = cor2cov!(copy(C), s)
+
+"""
+    cor2cov!(C, s)
+
+Converts the correlation matrix `C` to a covariance matrix in-place using a vector of
+standard deviations `s`.
+"""
+function cor2cov!(C::AbstractMatrix, s::AbstractArray)
+    n = length(s)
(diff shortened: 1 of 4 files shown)


ararslan:
I would say that we could provide in-place methods as well, though perhaps we should wait on that until `cov2cor!` is moved out of Base.

src/cov.jl
```diff
@@ -120,3 +120,23 @@ function mean_and_cov(x::DenseMatrix, wv::AbstractWeights, vardim::Int=1;
...
+"""
+cov2cor(C::DenseMatrix, s::AbstractArray) = Base.cov2cor!(copy(C), vec(s))
+
+"""
+    cor2cov(C, s)
+
+Compute the covariance matrix from the correlation matrix `C` and a vector of standard
+deviations `s`.
+"""
+function cor2cov(C::DenseMatrix, s::AbstractArray)
```
andreasnoack: Why is `s` not restricted to be a vector?

src/cov.jl
```diff
@@ -120,3 +120,23 @@ function mean_and_cov(x::DenseMatrix, wv::AbstractWeights, vardim::Int=1;
...
     return m, Base.cov(x, wv, vardim; corrected=depcheck(:mean_and_cov, corrected))
 end
+
+"""
+    cov2cor(C, s)
+
+Compute the correlation matrix from the covariance matrix `C` and a vector of standard
+deviations `s`.
+"""
+cov2cor(C::DenseMatrix, s::AbstractArray) = Base.cov2cor!(copy(C), vec(s))
```
ararslan: `Base.cov2cor!` accepts any `AbstractMatrix`; why constrain to dense here?

andreasnoack:
True. I think it would make sense to define `cor2cov!` instead.

src/cov.jl
rofinn: This is mostly for convenience, so that you can pass `std(X, 1)` and have the `1xn` array converted to a vector rather than requiring the caller do that. I have no strong preference for this, but we do that in other places anyways.

src/cov.jl
rofinn: No particularly good reason I guess.

rofinn:
So I didn't want to implement `cor2cov!` because:

1. `cov2cor!` seems to mostly be used as a way of creating the correlation matrix internally where I don't see a general conversion function as being something that should happen in place.
2. The simple `D * C * C` implementation seems much cleaner to understand than using loops or `A_mul_B!`, which the latter tends to allocate more memory anyways.

ararslan:
> I don't see a general conversion function as being something that should happen in place.

An in-place matrix function isn't really a general conversion function as such. One of the advantages of Julia is that you're able to work with a lot of structures (of which matrices are one example) in-place, whereas in R everything is copied. Say you're working with an enormous correlation matrix, you need the covariance matrix, and you won't be needing the correlation matrix again. It makes sense to be able to reclaim the memory used by the correlation matrix so that you don't have to store two enormous matrices in memory when you only need one of them.

> `A_mul_B!`, which [...] tends to allocate more memory anyways.

That sounds like a bug to me, though @andreasnoack would know better.

andreasnoack:
> `A_mul_B!`, which [...] tends to allocate more memory anyways.

Yeah, I wasn't sure about that statement. @rofinn Could you explain?

Anyway, a loop would be the right thing to do for an in-place version.

rofinn:
> `A_mul_B!`, which [...] tends to allocate more memory anyways.

Nevermind, this was just a badly written benchmark test. If we want to provide an in-place method should we export `Base.cov2cor!` and add a docstring?

andreasnoack:
I think it is okay to leave the the in-place version unexported. The experts will find them if they need them. However, it is a good idea to base the copying versions on in-place versions.

src/cov.jl
```diff
@@ -120,3 +120,34 @@ function mean_and_cov(x::DenseMatrix, wv::AbstractWeights, vardim::Int=1;
...
+    cor2cov!(C, s)
+
+Converts the correlation matrix `C` to a covariance matrix in-place using a vector of
+standard deviations `s`.
+"""
+function cor2cov!(C::AbstractMatrix, s::AbstractArray)
+    n = length(s)
+    size(C) == (n, n) || throw(DimensionMismatch("inconsistent dimensions"))
+    for i in CartesianRange(size(C))
+        C[i] = C[i] * s[i[1]] * s[i[2]]
```
ararslan: `C[i] *= s[i[1]] * s[i[2]]`? Could also maybe `@inbounds` the loop, though I don't know whether that will make much of a difference.

ararslan:
Thanks, this is looking good. Could you add a test that tests the in-place method directly, perhaps also testing that the input has indeed been modified?

rofinn:
NOTE: The travis failure is due to problems downloading the macos julia nightly.

rofinn:
Is there anything that needs to be done for this (or maybe I missed)?

rofinn:
Bump?

ararslan:
Thanks for the reminder!

rofinn:
Awesome thanks!

PR #267, opened 2017-05-19
add trim and winsor

bjarthur:
this PR is a continuation of @dcarrera 's work at https://github.com/JuliaStats/StatsBase.jl/pull/201.  still re-reading through all the comments there to see what remains to be done.

Fixes https://github.com/JuliaStats/StatsBase.jl/issues/132.


test/robust.jl
```diff
@@ -0,0 +1,49 @@
...
+@test winsor!([1,2,3,4,5,6,7,8],   prop=0.1) == [1,2,3,4,5,6,7,8]
+@test winsor!([1,2,3,4,5,6,7,8],   prop=0.2) == [2,2,3,4,5,6,7,7]
+@test winsor!([1,2,3,4,5,6,7,8,9], prop=0.4) == [4,4,4,4,5,6,6,6,6]
+@test winsor!([1,2,3,4,5,6,7,8],   count=1)  == [2,2,3,4,5,6,7,7]
+@test winsor!([1,2,3,4,5,6,7,8,9], count=3)  == [4,4,4,4,5,6,6,6,6]
+
+@test_throws ArgumentError winsor!([])
+@test_throws ArgumentError winsor!([1,2,3,4,5], prop=0.5)
+
+### Variance
```
nalimilan: We should also have at least one test of `mean(trim(...))` just to check that the combination of both functions works.

bjarthur:
i think having two mutually exclusive keyword args is not elegant.  earlier i proposed dispatching on `Int` and `Float` instead, which is also a bit quirky.  how about this instead:  pass in a single second argument, possibly as a keyword.  if it's between 0 and 0.5 use that as the current `prop`.  if it's between 1 and n>>1, use it as the current `count`.  opinions?

nalimilan:
I don't see the problem with exclusive keyword arguments. They are quite clear and easier to document than playing with thresholds to understand what the user meant. They are also more robust in case the user makes a mistake and passes an incorrect value.

src/robust.jl
```diff
@@ -0,0 +1,133 @@
...
+    return x
+end
+
+"""
+    winsor(x, [prop=0.0, count=0])
+
+Return a Winsorized version of vector `x`; i.e. replace either `count` or
+proportion `prop` of the lowest elements of `x` with the next-lowest, and
+replace an equal number of the highest elements with the previous-highest.
+This function is used to compute the Winsorized mean of `x`. It is also
```
bjarthur: is `trimvar` the appropriate way to compute the variance of a winsorized set?  the name implies otherwise, and @dcarrera 's [comment](https://github.com/JuliaStats/StatsBase.jl/pull/201#issuecomment-243910309) describes it as computing the trimmed variance.

nalimilan: Ah, right, that's confusing. Just remove the last sentence then, we don't really care where this function is used here, we just want to know how to use it.

bjarthur:
are there plans to switch to Documenter.jl?  that's a lot easier than manually copying to RST.

nalimilan:
> are there plans to switch to Documenter.jl? that's a lot easier than manually copying to RST.

I don't think anybody has said s/he would work on it, but that's clearly the plan and help would be welcome.

bjarthur:
tests pass.  let me know if i missed anything.

src/StatsBase.jl
```diff
@@ -189,6 +195,7 @@ module StatsBase
     include("empirical.jl")
     include("hist.jl")
     include("misc.jl")
+    include("robust.jl")
```
nalimilan: Insert the line at a position consistent with that in the docs.

src/robust.jl
```diff
@@ -0,0 +1,129 @@
...
+#   Trimming outliers
+#
+#############################
+
+# Trimmed set
+"""
+    trim(x; prop=0.0, count=0)
+
+
+Remove either `count` or proportion `prop` of the highest and lowest elements
```
nalimilan: "Remove... in `x`" could be interpreted incorrectly as meaning that the original `x` is modified. Would be better to say something like "Return a copy of `x` with ... removed."

src/robust.jl
```diff
@@ -0,0 +1,129 @@
...
+in `x` and return the result.  To compute the trimmed mean of `x`, use
+`mean(trim(x))`; to compute the variance of this quantity, use `trimvar(x)`.
```
nalimilan: ``[`trimvar(x)`](@ref)``

src/robust.jl
```diff
@@ -0,0 +1,129 @@
...
+    return x
+end
+
+"""
+    winsor(x; prop=0.0, count=0)
+
+Return a Winsorized version of vector `x`; i.e. replace either `count` or
+proportion `prop` of the lowest elements of `x` with the next-lowest, and
+replace an equal number of the highest elements with the previous-highest.
+To compute the Winsorized mean of `x`, use `mean(winsor(x))`.
```
nalimilan: "(see ``[`winsor`](@ref)``)"

src/robust.jl
```diff
@@ -0,0 +1,129 @@
...
+    deleteat!(x, (n-count+1):n)
+    deleteat!(x, 1:count)
+
+    return x
+end
+
+"""
+    winsor(x; prop=0.0, count=0)
+
+Return a Winsorized version of vector `x`; i.e. replace either `count` or
```
nalimilan: Same remark about the formulation.

src/robust.jl
```diff
@@ -0,0 +1,129 @@
...
+"""
+function trim!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
+    n = length(x)
+    n > 0 || throw(ArgumentError("x can not be empty."))
+
+    if count == 0
+        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
+        count = floor(Int, n * prop)
+    else
+        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
```
nalimilan: Should check that `prop != 0`, and if not throw an `ArgumentError`. Same for `winsor`.

src/robust.jl
bjarthur: does `@ref` work on a function call, as opposed to just a function name?

src/robust.jl
bjarthur: this `@ref` would be self referential.  all other changes made in this latest push.

src/robust.jl
nalimilan: The docs seem to imply it works (but I didn't check it).

docs/source/robust.rst
```diff
@@ -0,0 +1,20 @@
+Robust Statistics
+==========
+
+.. function:: trim(x; prop=0.0, count=0)
```
nalimilan: These docstrings need updating (sigh...).

bjarthur: fixed.  test pass.

docs/source/robust.rst
```diff
@@ -0,0 +1,20 @@
...
+
+  Return a copy of ``x`` with either ``count`` or proportion ``prop`` of the highest
+  and lowest elements removed.  To compute the trimmed mean of ``x`` use
+  ``mean(trim(x))``; to compute the variance use ``trimvar(x)`` (see [``trimvar``](@ref)).
```
nalimilan: AFAIK  `@ref` won't work in RST files (re-sigh). Just drop the link for now, once we move to Documenter the docstring will be extracted automatically with its link.

bjarthur:
fixed

docs/source/robust.rst
```diff
@@ -0,0 +1,20 @@
...
+
+  Return a copy of ``x`` with either ``count`` or proportion ``prop`` of the lowest
+  elements of ``x`` replaced with the next-lowest, and an equal number of the
+  highest elements replaced with the previous-highest.  To compute the Winsorized
+  mean of ``x`` use ``mean(winsor(x))``.
+
+.. function:: trimvar(x; prop=0.0, count=0)
+
+  Compute the variance of the trimmed mean of ``x``. This function uses
+  the Winsorized variance, as described in Wilcox (2010).
```
nalimilan, line 20: Actually this is just the variance, not the variance of the (estimator of the) mean, right?

PR #270, opened 2017-05-23
Add Base.similar, merge and merge! for Histogram

oschulz:
Merging of histogram is a frequently required operation, e.g. after processing/histogramming data in parallel (multiple threads and/or processes) or to combine histogrammed data from multiple sources (e.g. different signal channels).

It seems natural to extend `Base.merge` to histograms for this.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -441,13 +441,13 @@ end
 
 
 """
-    similar(h::Histogram)
+    zero(h::Histogram)
 
 Create a new histogram with the same binning, type and shape of weights
-and the same properties (`closed` and `isdensity`) as `h`. The weights of the
-new histogram are set to zero.
+and the same properties (`closed` and `isdensity`) as `h`, with all weights
+set to zero.
 """
-Base.similar{T,N,E}(h::Histogram{T,N,E}) =
+Base.zero{T,N,E}(h::Histogram{T,N,E}) =
     Histogram{T,N,E}(deepcopy(h.edges), zero(h.weights), h.closed, h.isdensity)
 
 
@@ -479,4 +479,4 @@ have the same binning, shape of weights and properties (`closed` and
 `isdensity`). The weights of all histograms are summed up for each bin, the
 weights of the resulting histogram will have the same type as those of `h`.
 """
-Base.merge(h::Histogram, others::Histogram...) = merge!(similar(h), h, others...)
+Base.merge(h::Histogram, others::Histogram...) = merge!(zero(h), h, others...)
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -438,3 +438,45 @@ function normalize{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; m
     normalize!(h_fltcp, aux_weights_fltcp..., mode = mode)
     (h_fltcp, aux_weights_fltcp...)
 end
+
+
+"""
+    similar(h::Histogram)
+
+Create a new histogram with the same binning, type and shape of weights
+and the same properties (`closed` and `isdensity`) as `h`. The weights of the
+new histogram are set to zero.
+"""
+Base.similar{T,N,E}(h::Histogram{T,N,E}) =
+    Histogram{T,N,E}(deepcopy(h.edges), zero(h.weights), h.closed, h.isdensity)
+
+
+"""
+    merge!(target::Histogram, others::Histogram...)
+
+Update histogram `target` by merging it with the histograms `others`. See
+`merge(histogram::Histogram, others::Histogram...)` for details.
+"""
+function Base.merge!(target::Histogram, others::Histogram...)
+    for h in others
+        target.edges != h.edges && throw(ArgumentError("can't merge histograms with different binning"))
+        size(target.weights) != size(h.weights) && throw(ArgumentError("can't merge histograms with different dimensions"))
+        target.closed != h.closed && throw(ArgumentError("can't merge histograms with different closed left/right settings"))
+        target.isdensity != h.isdensity && throw(ArgumentError("can't merge histograms with different isdensity settings"))
+    end
+    for h in others
+        target.weights .+= h.weights
+    end
+    target
+end
+
+
+"""
+    merge(h::Histogram, others::Histogram...)
+
+Construct a new histogram by merging `h` with `others`. All histograms must
+have the same binning, shape of weights and properties (`closed` and
+`isdensity`). The weights of all histograms are summed up for each bin, the
+weights of the resulting histogram will have the same type as those of `h`.
+"""
+Base.merge(h::Histogram, others::Histogram...) = merge!(similar(h), h, others...)
--- a/test/hist.jl
+++ b/test/hist.jl
@@ -193,4 +193,24 @@ end
 end
 
 
+@testset "Histogram similar" begin
+    h = fit(Histogram, (rand(100), rand(100)), closed=:left)
+    h2 = similar(h)
+    @test all(x -> x≈0, h2.weights)
+    @test !(h.weights === h2.weights)
+    @test h.edges == h2.edges
+    @test h.closed == h2.closed
+    @test h.isdensity == h2.isdensity
+end
+
+
+@testset "Histogram merge" begin
+    histograms = [fit(Histogram, (rand(100), 10 * rand(100)), (0:0.1:1, 0:1:10), closed=:left) for _ in 1:10]
+    h = similar(histograms[1])
+    merge!(h, histograms ...)
+    @test h.weights == (+).((x->x.weights).(histograms)...)
+    @test merge(histograms...) == h
+end
+
+
 end # @testset "StatsBase.Histogram"
--- a/test/hist.jl
+++ b/test/hist.jl
@@ -195,7 +195,7 @@ end
 
 @testset "Histogram zero" begin
     h = fit(Histogram, (rand(100), rand(100)), closed=:left)
-    h2 = zero(h)
+    h2 = @inferred zero(h)
     @test all(x -> x≈0, h2.weights)
     @test !(h.weights === h2.weights)
     @test h.edges == h2.edges
@@ -209,7 +209,7 @@ end
     h = zero(histograms[1])
     merge!(h, histograms ...)
     @test h.weights == (+).((x->x.weights).(histograms)...)
-    @test merge(histograms...) == h
+    @test (@inferred merge(histograms...)) == h
 end
 
 
--- a/test/hist.jl
+++ b/test/hist.jl
@@ -193,9 +193,9 @@ end
 end
 
 
-@testset "Histogram similar" begin
+@testset "Histogram zero" begin
     h = fit(Histogram, (rand(100), rand(100)), closed=:left)
-    h2 = similar(h)
+    h2 = zero(h)
     @test all(x -> x≈0, h2.weights)
     @test !(h.weights === h2.weights)
     @test h.edges == h2.edges
@@ -206,7 +206,7 @@ end
 
 @testset "Histogram merge" begin
     histograms = [fit(Histogram, (rand(100), 10 * rand(100)), (0:0.1:1, 0:1:10), closed=:left) for _ in 1:10]
-    h = similar(histograms[1])
+    h = zero(histograms[1])
     merge!(h, histograms ...)
     @test h.weights == (+).((x->x.weights).(histograms)...)
     @test merge(histograms...) == h


src/hist.jl
```diff
@@ -438,3 +438,47 @@ function normalize{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; m
...
+new histogram are set to zero.
+"""
+Base.similar{T,N,E}(h::Histogram{T,N,E}) =
+    Histogram{T,N,E}(deepcopy(h.edges), zero(h.weights), h.closed, h.isdensity)
+
+
+"""
+    merge!(target::Histogram, others::Histogram...)
+
+    Update histogram `target` by merging it with the histograms `others`.
```
ararslan: The signatures should be indented but not the prose

src/hist.jl
```diff
@@ -438,3 +438,47 @@ function normalize{T, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; m
...
+"""
+    merge(h::Histogram, others::Histogram...)
+
+    Construct a new histogram by merging `h` with `others`. All histograms
+    must have the same binning, shape of weights and properties (`closed` and
+    `isdensity`). The weights of all histograms are summed up for each bin,
+    the weights of the resulting histogram will have the same type as those
+    of `h`.
+"""
+
```
ararslan: No line break between docstring and function

andreasnoack:
Great. This functionality is really needed. The question is what to call it. The function `merge` in base is doing somethign related but is different.

Right now, `Histogrm` mimics the `Vector` interface by defining `push!` and `append!` so I'm wondering if that is  better to use for this functionality. I've actually been doing a similar excercise because I wanted to compute a hisogram of a distributed vector. That can be seen as a reduction operation but it requires that the same function is used within and accross nodes. Single node we'd have the left fold `(((Histogram() op x1) op x2) op x3) op x4` but distributed, it would be `((Histogram() op x1) op x2) op ((Histogram() op x3) op x4)`.

I think `similar` is fine but for arrays, it wouldn't initialize the memory and here we'd be using it under the assumption that the `Histogram` is initialized. An alternative could be to define `zero` but I don't think it is a big deal.

src/hist.jl
oschulz: Ooops! Sorry - fixed.

src/hist.jl
oschulz: Argh - I keep forgetting that one ... fixed.

oschulz:
> An alternative could be to define zero but I don't think it is a big deal.

You're right, similar normally doesn't initialize to zero, I wasn't entirely happy with it myself. I like the idea of using `zero` instead of `similar`, I changed the code accordingly.

oschulz:
> Right now, Histogrm mimics the Vector interface by defining push! and append! so I'm wondering if that is better to use for this functionality.

I don't like `append!` here for three reasons:

* We already defined `append!` for histograms as "add data points from a collection to this histogram". As a histogram is kind of a collection itself, using it for merge seems counter-intuitive (as we don't use the bin-values of the histograms as data points).
* There's no `append`, only `append!`, and having a version that returns a new histogram is very convenient (but a `!` other version is important to, esp. if memory allocation and GC have to be controlled, e.g. when using many threads).
* It doesn't apply symmetry - if you merge a bunch of histograms, the order shouldn't matter (in that respect, I should actually make `merge` use the promoted element type of all histograms).

I think using `merge` is well justified:

* People (at least in my experience) call it "merging" histograms, not "appending".
* It's not so different from what merge does for `Dict`: It combines it's arguments be combining theirs corresponding entries - only here the natural combination is addition, not replacement (as for `Dict`).

oschulz:
> in that respect, I should actually make merge use the promoted element type of all histograms

I tried, but I can't manage to make that type-stable. So I propose to leave it as it currently is (in any case, I think people will pretty much always merge histograms with the same element type).

oschulz:
I thought a bit more about `merge!` vs `append!` and their behavior, and I think `merge!` is indeed the right choice for merging histograms: Given my current definition for `merge!` for histograms, and given

* A filled `Dict` resp. `Histogram` `a`
* An unfilled `Dict`/`Histogram` `b`  of the same type as `a`
* A `Vector` `V` of an element type for which `push!(a, V[i])` is defined

Then the following holds true for both `Dict` and `Histogram`:

* `append!(a, b)` is undefined
* `merge!(a, b) leaves `a` unchanged (as `b` is "empty").
* `for x in V push!(a, x)` is equivalent to `for x in V push!(b, x); merge!(a, b)`

So even though a `Dict` and `Histogram` are obviously different things, I think the `merge!` behaves exactly the same way, in relation to `push!`. The difference lies only in what `push!` does: For a `Dict`, `push!` finds an "entry" and replaces it's value, while for a `Histogram!` it finds an "entry" and adds to it's value. But I feel that there's enough of a common underlying abstraction to justify using `merge!` for histograms.

oschulz:
> @andreasnoack: Right now, Histogrm mimics the Vector interface by defining push! and append!
> @oschulz: But I feel that there's enough of a common underlying abstraction to justify using merge! for histograms.

I actually think that our `push!` for `Histogram` is much closer to `push!` for `Dict` than for `Vector`: For `Dict` and `Histogram`, push modifies entries but doesn't change the size, while for `Vector` is leaves the previous contents unchanged, but changes the size.

Also, I believe the semantics for `append!` (for any type) are "repeated `push!` of compatible from a arbitrary collection", but *not* "combine two similar things". For example, you append to a `Vector` from an `Array` of completely different size and dimensionality.

So also from this perspective, I believe `merge!` is a much better fit for combining compatible histograms than `append!`.

andreasnoack:
...and the reduction argument? I think it would be nice if `((a op b) op c) op d` was the same as `(a op b) op (c op d)`?

oschulz:
@andreasnoack: Sorry up front for the lengthy answer.

As far as the reduction argument is concerned, I feel that the functions currently available in Julia cannot cover parallel execution for non-trivial reductions (i.e. histogramming) anyhow:

* `reduce` and `mapreduce` leave it "unspecified whether v0 is used for non-empty collections". So they are only usable for reduction operation that need no parameters (e.g. binning) and that can be expressed by a function that basically returns the same type as it's arguments.

* `foldl`, `mapfoldl`, etc. specify "guaranteed left associativity. v0 will be used exactly once." So they cannot be executed in parallel at all, without violating that statement.

* AFAIK, map/reduce-like frameworks (e.g. CouchDB) often allow a re-reduce function in addition to the reduce function, to address the problem that the reduction result may be of a different kind than the input values, and that in a multi-layer reduction scheme the secondary layers may therefore need to use a different reduction function. It would be nice if Julia would offer something like `rereduce(f_reduce, f_rereduce, data, ...)`.

I do not think the the function name used to merge histograms should be dictated by the fact that Julia currently doesn't give us a way to specify a secondary reduction function, especially as `reduce` and `foldl` are unsuitable for parallel execution anyhow (see above).

In any case, in a real-live scenario using multiple hosts/processes with multiple threads each, one will have to write more specific code anyhow: How the work is distributed on nodes/processes and on threads will in general be application specific. Here's an example that histograms data in a `DArray` using my fledgling new [MultiThreadingTools](https://github.com/oschulz/MultiThreadingTools.jl/tree/dev):

```julia
data = drandn(10^8)

result = Histogram(-4:0.1:4, Float64, :left)

hists = @everyworker begin
    A = localpart(data)
    tl_h = ThreadLocal(result)
    @everythread begin
        h = tl_h[]
        for i in threadpartition(eachindex(A))
            push!(h, A[i])
        end
    end
    merge(all_thread_values(tl_h)...)
end

merge!(result, hists...)
```

I'm currently playing with this with 4 workers on 2 machines (one worker per NUMA node on each) and 14 threads on each worker - it runs, though performance seems mem-I/O limited at the moment. In a real application there would be some expensive operations to be done on the data before histogramming, of course.

If we want to support `foldl` and friends (for simple non-parallel use cases) how about we do this:

* We keep `merge` and `merge!`, to have both an in-place and non-in-place variant with a semantically chosen name.
* We make `push!(h::Histogram, others::Histogram...)` a synonym for `merge!` to make `foldl` and friends happy.

It's not perfect, but I believe it's the best we can do. What do you think?

oschulz:
> We make push!(h::Histogram, others::Histogram...) a synonym for merge! to make foldl and friends happy.

It don't really like that workaround though, I have to admit - it's kind of hacky. Do we really need to support `foldl` for histograms? If people want to add the values in a vector (etc.) `V` to a histogram `h`, they can just do `append!(h, V)` instead of something like `foldl(push!, h, V)` - and if they want to transform it first, a simple `append!(h, (f(x) for x in V))` should do instead of using `mafoldl`. In any case, if `foldl(push!, h, V)` worked for a vector of real values, it would fail for a vector of a vector of real values - there, we'd need `foldl(append!, h, V)`. An vice versa.

Until Julia provides a reduction API that allows for a re-reduce function, I really don't think that `foldl` and `mafoldl` are a good fit for histogramming, without hacky comprimises  - and, like described above, I think it would only work for trivial use-cases. And anything done distributed resp. in parallel would violate the "use neutral element exactly once" contract of `foldl` anyhow (I guess that's why only `reduce` is implemented for `DArray`).

We need a more generic solution to reduce with complex types as the result, i.e. a different kind of (map)reduce functions - but StatsBase shouldn't be the place to solve it, IMHO it's a general problem. Maybe I can defining something that could do it in MultiThreadingTools (covering multiple workers with multiple threads).

oschulz:
> We need a more generic solution to reduce with complex types as the result, i.e. a different kind of (map)reduce functions

@andreasnoack : I just finished a proof-of-concept implementation of something like this in [MultiThreadingTools](https://github.com/oschulz/MultiThreadingTools.jl) (new package, nor registered yet). It now defines functions like `prefoldl` and `pmaprefoldl` for parallel map-reduce/fold operations that run on multiple processes and all their threads. The "refold" functions accept two functions `f_fold` and `f_refold` to support reductions where the reduction result is of a completely different type than the data elements, so that secondary stage reductions will often need a different function.

With that, one can simply do things like

```julia
prefoldl(push!, merge!, Histogram(-4:0.1:4, :left), drandn(10^7));
```

Here's a more involved example (needs the current master branch of Plots.jl). I 

```julia
data = drand(10^7)
h = Histogram(0:0.1:5, Float64, :left)

pmaprefoldl(x -> -log(x), push!, merge!, h, data);

normalize!(h)
using Plots
plot(h, st = :step, label = "h")
plot!(0:0.01:5, x -> exp(-x), label = "exp(-x)")
```

So it basically only needs a single `pmaprefoldl` to turn distributed uniform random numbers into a histogram of an exponential distibution. I ran this on my office PC using two compute servers for the workers (2 workers with 14 threads each on every server):

```julia
julia> info(workers())
INFO: [2,3,4,5]

julia> info(@everyworker nthreads())
INFO: [14,14,14,14]

julia> savefig("exp-dist.png")
```

exp-dist

Runs just fine, though performance could be better (definitely not a 4x14-fold speed up - though probably also not to be expected this example).

So, what do you think - if we can have things like this, can we go with `merge` and `merge!`?

andreasnoack:
@oschulz Sorry for the hesitation here. I've been arguing a lot with myself but, eventually, I decided to buy your argument. For `DArray`'s, I'd like the distributed aspect to be abstracted away so the parallel construct shouldn't require parallelism. However, I think you are right that it should be expressed as two separate operations.

Abstractly, I'd like to think of the operation as DnC, i.e. for an operatation `f(x1,...,xn)` we can supply a `g` such that `g(f(x1,...,xk),f(x{k+1},...,xn)) = f(x1,...,xn)`. In my case `f=t -> fit(Histogram, t)` and `g=merge!`. The higher order function is then free to repeat that pattern as many times as it wants to (including zero). Hence, the operation is not necessarily parallel. In the distributed case, the obvious number corresponds to the number of processes. I believe your `pmaprefoldl` can be considered a special case of that pattern.

Anyway, that was just my excuse for being slow here. Thanks again for the contribution.

oschulz:
> @andreasnoack Sorry for the hesitation here.

I understand - I know you care very much about the consistency and the quality of StatsBase, and I do appreciate it a lot. Thanks for merging!

> For DArray's, I'd like the distributed aspect to be abstracted away

I fully agree, for "simple" use cases, it should definitely be transparent (though maybe with an option to enable/disable multithreading or so). For complex cases, I think Julia needs better constructs to control explicit parallelism in more detail (beyond my modest beginning in MultiThreadingTools).

> Abstractly, I'd like to think of the operation as DnC [...] I believe your pmaprefoldl can be considered a special case of that pattern

Yes - I was actually thinking about offering a more general, resp. non-parallel version. Maybe one that allows to switch both multi-processing and multi-threading on/off via Bool kwargs or so? If you have any suggestions regarding `MultiThreadingTools` and `pmaprefoldl`, please let me know. I'd like to register it in the not too distant future, but I'd be glad to get a bit of advice from some long-time Julia experts on design and implementation.

PR #272, opened 2017-06-01
sample call fails with error

claireh93:
Since v0.15.0 was released sample doesn't seem to work any longer, e.g.:

```julia
sample([1.0,2.0], WeightVec([0.5,0.5]), (2, 1))
```

Fails with an error at line 722 of src/sampling.jl. There seemed to be a couple of problems and I have fixed this to match the lines above and it now works.


andreasnoack:
Thanks for fixing this. Could you add a test that exercises that function?

PR #274, opened 2017-06-21
add params

joshday:
From https://github.com/JuliaStats/Distributions.jl/issues/589, `params` is probably too general to live in Distributions, and StatsBase may be a better home for it.  A corresponding PR to Distributions will follow shortly.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -165,10 +165,14 @@ dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(ty
 """
     params(model)
 
-Return a tuple of model parameters.
+Return all parameters of a model.
 """
 params(obj) = error("params is not defined for $(typeof(obj))")
-params!(obj) = error("params! is not defined for $(typeof(obj))")
+
+"""
+Write the parameters of a model into preallocated storage
+"""
+function params! end
 
 ## coefficient tables with specialized show method
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -165,7 +165,7 @@ dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(ty
 """
     params(model)
 
-Return all parameters of a model.  For a linear model, this would return coefficients as well as variance components.
+Return all parameters of a model.  For a linear model, this would return coefficients as well as any other parameters.
 """
 params(obj) = error("params is not defined for $(typeof(obj))")
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -162,17 +162,8 @@ predict(obj::RegressionModel) = error("predict is not defined for $(typeof(obj))
 predict!(obj::RegressionModel) = error("predict! is not defined for $(typeof(obj)).")
 dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(typeof(obj)).")
 
-"""
-    params(model)
 
-Return all parameters of a model.  For a linear model, this would return coefficients as
-well as any other parameters.
-"""
 params(obj) = error("params is not defined for $(typeof(obj))")
-
-"""
-Write the parameters of a model into preallocated storage.
-"""
 function params! end
 
 ## coefficient tables with specialized show method
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -165,12 +165,12 @@ dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(ty
 """
     params(model)
 
-Return all parameters of a model.
+Return all parameters of a model.  For a linear model, this would return coefficients as well as variance components.
 """
 params(obj) = error("params is not defined for $(typeof(obj))")
 
 """
-Write the parameters of a model into preallocated storage
+Write the parameters of a model into preallocated storage.
 """
 function params! end
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -168,7 +168,7 @@ dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(ty
 Return a tuple of model parameters.
 """
 params(obj) = error("params is not defined for $(typeof(obj))")
-
+params!(obj) = error("params! is not defined for $(typeof(obj))")
 
 ## coefficient tables with specialized show method
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -162,6 +162,13 @@ predict(obj::RegressionModel) = error("predict is not defined for $(typeof(obj))
 predict!(obj::RegressionModel) = error("predict! is not defined for $(typeof(obj)).")
 dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(typeof(obj)).")
 
+"""
+    params(model)
+
+Return a tuple of model parameters.
+"""
+params(obj) = error("params is not defined for $(typeof(obj))")
+
 
 ## coefficient tables with specialized show method
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -162,7 +162,11 @@ predict(obj::RegressionModel) = error("predict is not defined for $(typeof(obj))
 predict!(obj::RegressionModel) = error("predict! is not defined for $(typeof(obj)).")
 dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(typeof(obj)).")
 
+"""
+    params(obj)
 
+Return all parameters of a model. 
+"""
 params(obj) = error("params is not defined for $(typeof(obj))")
 function params! end
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -165,7 +165,8 @@ dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(ty
 """
     params(model)
 
-Return all parameters of a model.  For a linear model, this would return coefficients as well as any other parameters.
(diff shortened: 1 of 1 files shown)


nalimilan:
I have two remarks:
- how different is this from `coef`? this should be mentioned in the docstring
- the plural `params` is inconsistent with the singular `coef`

joshday:
I think the suggested usage for `params` should be how Distributions already uses it, which is to return a tuple of parameter values.  For a linear model `y = x * β + σ²I`, `params(model)` would return `(β̂, σ̂²)`.  Since it returns a tuple, I think the plural name is appropriate.  Thoughts?

nalimilan:
The plural could be equally justified for `coef`, what matters is consistency IMHO. Then I wonder about the opportunity of having two identical functions, one returning an array and the other a tuple. Couldn't we use `coef` everywhere, and have it return a tuple where it makes sense? I.e. for distributions or model families where the number of parameters is fixed, returning a tuple can be more efficient.

andreasnoack:
If we conclude that they are too overlapping, I think that we should deprecate `coef` and just use `params` since I think it is more common to talk about the *parameters* of a statistical model than the coefficients.

nalimilan:
> If we conclude that they are too overlapping, I think that we should deprecate coef and just use params since I think it is more common to talk about the parameters of a statistical model than the coefficients.

Hmm, that's a tough call. I guess that really depends on your field; coefficients is pretty common in mine. R uses `coef`, but Python's StatsModels uses `params`. Though I have to admit a Google search for "model parameters" gives much more many results than "model coefficients".

Terminology aside, something I don't like about `params` or `coefs` is the pluralization of an abbreviation, as if `param` or `coef` were real words. Are there other examples of this in Julia APIs?

kleinschmidt:
I think `coef` and `params` are different enough to warrant two functions (we talk about model coefficients as a special subset of model parameters a lot in my field). I think this is a useful addition to the interface.

nalimilan:
@kleinschmidt How would you define these functions then? I guess parameters would include things like the dispersion parameter of the normal distribution in a linear regression model? And coefficients would only include parameters related to the effect of variables?

dmbates:
I think the distinction is that coefficients are the parameters that determine the linear predictor.  That is, they are the β in the linear predictor Xβ

kleinschmidt:
Parameters include additional things like the dispersion parameter and the random effects covariance matrices (in a mixed model)

nalimilan:
Makes sense. @joshday Could you mention this difference with `coef` somehow in the docstring? Maybe the function should only return a tuple for models/distributions whose number of parameters is fixed? An array would be better e.g. for linear regression, to avoid introducing a type instability.

Also, a docstring should be added for `params!`. Since `params!(obj)` isn't the actual signature that users will call, I'd attach the docstring to `function params! end` (I did that in https://github.com/JuliaStats/StatsBase.jl/pull/275).

Then there's the issue of singular `coef` vs. plural `params`. Would people be OK with `param`?

joshday:
I prefer `params` because it's well established in Distributions.  

> Maybe the function should only return a tuple for models/distributions whose number of parameters is fixed? An array would be better e.g. for linear regression, to avoid introducing a type instability.

I'm not sure I follow.  In Distributions, it always returns a tuple.  I imagine something like this:
```julia
params(linear_model) = (β̂, σ̂²)
```
similar to
```julia
julia> params(MvNormal(zeros(2), eye(2)))
([0.0, 0.0], PDMats.PDMat{Float64,Array{Float64,2}}(2, [1.0 0.0; 0.0 1.0], Base.LinAlg.Cholesky{Float64,Array{Float64,2}} with factor:
[1.0 0.0; 0.0 1.0]))
```

nalimilan:
> I prefer params because it's well established in Distributions.

Well, yes, but having two very similar functions with inconsistent naming rules is problematic.

> I'm not sure I follow. In Distributions, it always returns a tuple.

What I meant is that returning a tuple makes a lot of sense for distributions, where the number of parameters is fixed for a given type. But it doesn't sound like the best choice for models with a varying number of parameters, like a linear regression: the return type would be unstable since `N` cannot be determined at compile time. But that's OK now that you've removed the mention of tuple in the docstring.

The docstrings should still explain what's the difference with `coef` IMO.

src/statmodels.jl
```diff
@@ -165,12 +165,12 @@ dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(ty
 """
     params(model)
 
-Return all parameters of a model.
+Return all parameters of a model.  For a linear model, this would return coefficients as well as variance components.
```
nalimilan: Wrap lines at 92 chars.

Also, couldn't we find a more general description? @dmbates gave a possible one for `coef` (and "variance components" would only apply to a *mixed* model anyway).

src/statmodels.jl
```diff
@@ -162,6 +162,18 @@ predict(obj::RegressionModel) = error("predict is not defined for $(typeof(obj))
...
+"""
+    params(model)
+
+Return all parameters of a model.  For a linear model, this would return coefficients as
+well as any other parameters.
+"""
+params(obj) = error("params is not defined for $(typeof(obj))")
+
+"""
+Write the parameters of a model into preallocated storage.
```
ararslan: Since there are no methods defined and it isn't exported, perhaps not worth adding a docstring for now? Idk, I guess this is fine, I don't feel too strongly.

joshday: Perhaps removing the docstring is best for now.  I am fine either way.

ararslan: ¯\\\_(ツ)\_/¯ I'll leave it up to you

kleinschmidt:
I'm pro docstring. 

(please pardon my thumb-typing)

--
dave.f.kleinschmidt@gmail.com
http://davekleinschmidt.com
413-884-2741

src/statmodels.jl
nalimilan: Docstrings are required to build the Documenter manual in https://github.com/JuliaStats/StatsBase.jl/pull/275, at least if we want to include the function in the manual. Why wouldn't we export this function? Isn't that the whole point of this issue?

src/statmodels.jl
```diff
@@ -165,7 +165,8 @@ dof_residual(obj::RegressionModel) = error("dof_residual is not defined for $(ty
 """
     params(model)
 
-Return all parameters of a model.  For a linear model, this would return coefficients as well as any other parameters.
+Return all parameters of a model.  For a linear model, this would return coefficients as
```
nalimilan: I still don't like that explanation. How about something like this? To be improved if somebody else has suggestions.
```
Return all parameters of a model. Contrary to [`coef`](@ref), this includes
parameters other than those corresponding to the effect of independent variables
(e.g. in a linear model, those which determine the linear predictor).
```

Or we could add that mention to `coef` instead, pointing to `params` to get all parameters. In any case, it makes sense to have a link from each of the docstrings to the other.

src/statmodels.jl
ararslan: I dunno, I need to pay more attention

joshday: My reasoning for moving params from Distributions to StatsBase is so that LearnBase isn't importing Distributions for a single name.  I'm actually indifferent to whether StatsBase exports it or not.

andreasnoack:
I assume this was blocking downstream packages. We can adjust the doc string later and consider a general naming overhaul.

Evizero:
Any chance we can get this tagged?

ararslan:
https://github.com/JuliaLang/METADATA.jl/pull/10048

Evizero:
Thank you. This may seem like a little change but note that this is very appreciated.

Nosferican:
For linear models, could it return a dictionary with all model specifications? For example, à la [CovarianceMatrices#19](https://github.com/gragusa/CovarianceMatrices.jl/issues/19) for a requested variance covariance estimator?

nalimilan:
Could you be more specific? I can't find what you're referring to in that issue? I would have thought it should return a `NamedArray`/`AxisArray` rather than a dictionary.

Nosferican:
For various regression models a few relevant arguments could be the fit method / estimator (2SLS, MLE, etc.), solver (e.g., Cholesky, Gradient Descent, Conjugate Gradient, SVD), and variance covariance estimator (e.g., OLS, HC1, etc.) and cluster variables. 
I wanted to know if these methods should be implemented in their relevant packages or have a generic StatsBase method to get them... For example, `params` if it returns a collection of these. Don't know which design might be preferred.

nalimilan:
What do you mean by "a collection of these"? `params` is supposed to return the estimated parameters, not a list of the possible values of arguments.

Nosferican:
For example, if I am estimating random effects which estimates the error components (individual and idiosyncratic) would that be a candidate for `params` or is it meant only as:
```
params(model) = (coef(model), stderr(model))
```
if so, shouldn't this be the default behavior for `StatsBase.RegressionModel`?

nalimilan:
OK, I see. The definition of the function proposed above when it was added was to return estimates, not standard errors (which can be obtained via `stderr`). I guess for mixed models a two-element `NamedTuple` with fixed and random effects could be returned. But that's not something that StatsBase can do, it would have to be implemented by packages on a case by case basis.

Nosferican:
So `params` is not universally standardized or whatnot and is largely context dependent. Got it. Thanks.

PR #275, opened 2017-06-22
Port manual to Documenter

### LINKED ISSUE #260: Documenter.jl
It would be really nice if we could switch over to using Documenter.jl with automatic docs generation.

nalimilan:
Move all content to docstrings when they were missing or incomplete.
Fix many details in docstrings. Improve in particular docstrings
for sampling algorithms, analysis of signals, and stats models.

Fixes #260.

I would really appreciate if somebody could build the manual locally and look at the new manual a bit carefully, in particular in the areas mentioned above. Reviewing the .md files is hard since they have been completely rewritten, but paradoxically they contain little changes (except for the removal of RST docstrings). So reviewing the docstrings diff in the .jl files is much more likely to catch problems.


ararslan:
Really busy during JuliaCon but I've requested a review from myself to try not to forget to review this when I have time. Thanks so much for doing this!

rofinn:
@nalimilan I tried building the old and new docs locally, but the `docs/make.jl` file is currently missing.

nalimilan:
@rofinn Thanks for catching this, I must have forgotten to add that file. Should be fixed now.

.travis.yml
```diff
@@ -16,3 +16,5 @@ git:
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("StatsBase"); Pkg.test("StatsBase"; coverage=true)'
 after_success:
   - julia -e 'cd(Pkg.dir("StatsBase")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'Pkg.add("Documenter")'
+  - julia -e 'cd(Pkg.dir("StatsBase")); include(joinpath("docs", "make.jl"))'
```
ararslan: Might as well make this one line in keeping with the coverage line item and add the package in the same step as building the docs. Doesn't really matter though.

src/misc.jl
```diff
@@ -114,19 +123,25 @@ function levelsmap{T}(a::AbstractArray{T})
...
+# Examples
+
+```jldoctest
+julia> using StatsBase
+
+julia> indicatormat([1 2 2], 2)
+2×3 Array{Bool,2}:
+  true  false  false
+ false   true   true
+
```
ararslan: unnecessary blank line

src/weights.jl
```diff
@@ -458,6 +477,21 @@ function Base.median(v::AbstractArray, w::AbstractWeights)
...
 end
 
+"""
+    median(v::RealVector, w::AbstractWeights)
+
+Compute the weighted median of `x`, using weights given by a weight vector `w`
+(of type `AbstractWeights`). The weight and data vectors must have the same length.
+
+The weighted median ``x_k`` is the element of `x` that satisfies
+``\sum_{x_i < x_k} w_i \le \frac{1}{2} \sum_{j} w_j`` and
```
ararslan: Need more double backslashes for the TeX here

ararslan:
Built the documentation now and it looks beautiful

src/sampling.jl
```diff
@@ -109,17 +116,29 @@ end
...
+
+Pseudo-code:
+```
+n = length(a)
+k = length(x)
+
+create an array of index inds = [1:n]
+
+for i = 1:k
+    swap inds[i] with a random one in inds[i:n]
```
rofinn: Same here. It might make more sense to put the descriptive text in comments and have to code still look like julia code.
```
# swap element `i` for another random element
inds[i] = rand(inds)

# set element `i` in `x`
x[i] = a[inds[i]]
```

src/statmodels.jl
```diff
@@ -37,30 +67,44 @@ nobs(obj::StatisticalModel) = error("nobs is not defined for $(typeof(obj)).")
...
+"""
+    vcov(obj::StatisticalModel)
+
+Return the variance-covariance matrix for the coefficients of the model.
+"""
 vcov(obj::StatisticalModel) = error("vcov is not defined for $(typeof(obj)).")
-fit(obj::StatisticalModel, data...) = error("fit is not defined for $(typeof(obj)).")
-fit!(obj::StatisticalModel, data...) = error("fit! is not defined for $(typeof(obj)).")
+
+fit(obj::StatisticalModel, args...) = error("fit is not defined for $(typeof(obj)).")
```
rofinn: Missing docstring for `fit` and `fit!`?

docs/source/signalcorr.rst
```diff
@@ -1,86 +0,0 @@
...
-
-    Here, ``method`` is a keyword argument to specify the choice of algorithm, which can be either ``:regresion`` or ``:yulewalker``. The default value is ``:regression``. 
-
-
-.. function:: pacf!(r, x, lags[; method=:regression])
-
-    Write the computed partial auto-correlation to ``r``.
-
-
-.. function:: corspearman(x, y)
```
rofinn: `corspearman` and `corkendall` now appear in the  "Rank and Rank Correlations" sections. Was this intentional?

docs/src/statmodels.md
```diff
@@ -0,0 +1,33 @@
...
+```@docs
+adjr2
+aic
+aicc
+bic
+coef
+coeftable
+confint
+deviance
+dof
```
rofinn: Missing `df`, `fit` and `fit!`.

.travis.yml
nalimilan: Yes, I prefer that form, but for some reason the Documenter manual uses two separate lines.

docs/source/counts.rst
```diff
@@ -1,106 +0,0 @@
...
-
-.. function:: counts(x, a:b[, wv])
-
-    Count the number of times (or total weights if a weight vector ``wv`` is given) values in ``a:b`` appear in array ``x``. Here, the optional argument ``wv`` should be a weight vector of type ``AbstractWeights`` (see :ref:`weightvec`).
-
-    This function returns a vector ``r`` of length ``n``, with ``n = length(a:b) = b-a+1``. In particular, we have
-
-        ``r[i] == countnz(x .== a+(i-1))``
-
-    **Convenient forms:**
```
nalimilan: Makes sense, I hadn't noticed it. I've added the multiple forms to the signature for a general `counts` docstring, which is more standard than providing a table.

src/sampling.jl
```diff
@@ -109,17 +116,29 @@ end
...
+    fisher_yates_sample!([rng], a::AbstractArray, x::AbstractArray)
+
+Fisher-Yates shuffling (with early termination).
+
+Pseudo-code:
+```
+n = length(a)
+k = length(x)
+
+create an array of index inds = [1:n]
```
nalimilan: Indeed, this syntax does something different now.

src/sampling.jl
nalimilan: I tried that, but the correct code is actually more complex than what you posted, and in the end it was almost identical to the source, so I went back to pseudo-code. People can easily look at the source if they want details.

(BTW, that function probably doesn't work for a long time given that it calls `randi`, which doesn't exist.)

docs/source/signalcorr.rst
nalimilan: Yes, these are two rank correlation coefficients, they don't really have anything to do with signals AFAICT.

docs/src/statmodels.md
nalimilan: Good catch reading the diff, but `df` has been replaced with `dof`, and the RST files hadn't been updated correclty it seems.

nalimilan:
Thanks for the detailed reviews. I've added a commit which should address all comments.

Failure on nightlies is apparently due to https://github.com/JuliaLang/julia/pull/22296/. This PR shouldn't be merged until tests pass there, else the manual won't be deployed (and the README.md links will be broken).


BTW, I've noticed something which also applies to the existing manual (and therefore doesn't need to be fixed here): it's sometimes a bit weird that only methods added in this package are documented. It could be confusing for people who are not really familiar with Julia, and who might wonder e.g. why `median` and `quantile` are documented only for weighted versions. OTOH showing docs for Base methods could make the manual a bit long. I think we just need to switch a Documenter option, so it should be easy to test.

docs/make.jl
```diff
@@ -0,0 +1,32 @@
...
+             "signalcorr.md",
+             "misc.md",
+             "statmodels.md"]   
+)
+
+deploydocs(
+    # options
+    repo = "github.com/JuliaStats/StatsBase.jl.git",
+    target = "build",
+    julia  = "nightly",
```
KristofferC: Deploying on nightly is imo a bad idea because it can just randomly break. Just go with 0.6 and this can be merged?

nalimilan: The problem is that `0.6` will be outdated when 1.0 comes out, and we'll likely forget to update that file. We need a better solution (like `release`), but for now I'd rather use `nightly`.

KristofferC: So if Base breaks something in Documenter for some time you will just go without doc updates for that time? Seems worse than the potential fear that this file will be stale (which shouldn't even change the ouput of the docs assuming the documentation still goes through).

nalimilan: Yeah, neither option is ideal. I've filed https://github.com/JuliaDocs/Documenter.jl/issues/526. In the meantime, I've changed the config file to use `0.6`, let's hope we don't forget to update it.

nalimilan:
Let's merge?

rofinn:
I don't have visibility on this, but are the [ssh deploy keys setup](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#SSH-Deploy-Keys-1)?

ararslan:
The error on nightly is weird but most likely unrelated.

nalimilan:
> I don't have visibility on this, but are the ssh deploy keys setup?

I've just added the keys.

> The error on nightly is weird but most likely unrelated.

Yes, see my comment above, that's JuliaLang/julia#22296 (but there's PR open to fix it). Anyway we don't need it now that I've moved the deployment to 0.6.

ararslan:
Sorry didn't see the other comment about the error. This is good to go as far as I'm concerned.

nalimilan:
I had to push two more fixes, but now the manual is online!

ararslan:
Excellent work! Thanks so much for tackling this!

PR #277, opened 2017-06-27
Fix deploying Documenter manual

nalimilan:
The HTML backend needs these two options, else we need to provide a
docs/mkdocs.yml file.


nalimilan:
Merging since that's the only way to test deployment...

PR #279, opened 2017-06-28
Enable AppVeyor, remove randi()

nalimilan:


nalimilan:
@ararslan I've added a much more invasive commit. You may want to revise your approval...

For future reference, the `randi` vs `rand` performance issue is detailed at https://github.com/JuliaLang/julia/issues/22600.

ararslan:
> You may want to revise your approval

Nah, this is still fine by me. Any idea whether any packages that depend on StatsBase use that sampler directly?

nalimilan:
> Any idea whether any packages that depend on StatsBase use that sampler directly?

Looking on GitHub, looks like there are a few uses, mostly in old, unmaintained packages. One exception is Distributions. So I've added deprecations which fix the tests for that package until we port it.

The remaining failure on 32-bit 0.6 is unrelated, see https://github.com/JuliaLang/julia/issues/22613.

src/deprecates.jl
```diff
@@ -48,3 +48,15 @@ findat(a::AbstractArray, b::AbstractArray) = findat!(Array{Int}(size(b)), a, b)
 @deprecate df_residual(obj::StatisticalModel) dof_residual(obj)
 
 @deprecate_binding WeightVec Weights
+
+using Base.Random: RangeGenerator
+@deprecate_binding RandIntSampler RangeGenerator
+@deprecate RangeGenerator(K::Int) RangeGenerator(1:K)
```
ararslan: I'm confused by this. Did you mean for the former to be `RandIntSampler`?

nalimilan: That's a tricky situation. I first tried to do `@deprecate RandIntSampler(K::Int) RangeGenerator(1:K)`, but that triggers a deprecation warning when loading StatsBase. So that's the only solution I could find. That's a kind of type piracy, but hopefully we won't have to keep this for too long.

ararslan: What if you move the binding deprecation to the end?

nalimilan: Then I get a constant redefinition error (IIRC)...

ararslan: Hmmmmmmmmm, maybe
```julia
@deprecate (::Type{RandIntSampler})(K::Int) RangeGenerator(1:K)
```
?

nalimilan: That syntax doesn't work. I've also tried defining a standard function and calling `depwarn` inside it, but that still triggers the deprecation warning. Anyway, `@deprecate_binding` defines `const RandIntSampler = RangeGenerator`, so in the end the method is defined for `RangeGenerator` under the hood.

ararslan: Ah, okay. Well, not ideal, and we might get some pushback for the type piracy, but idk what else to suggest and I'm sure you've already carefully considered anything else that I could suggest.

tkelman: does it trigger the depwarn if you don't have deprecate_binding? what if RandIntSampler is defined as a function here for now where all of its methods trigger depwarn?

nalimilan: If I don't use `@deprecate_binding`, the depwarn is not triggered, but I need to use `const RandIntSampler = RangeGenerator`, and then the methods are added to `RangeGenerator`. So that's equivalent to the current situation of the PR. If I only define `RandIntSampler` as a function, the places where it's used as a type will break (there's one case in Distributions, see https://github.com/JuliaStats/Distributions.jl/issues/630). We can certainly do that, the question is whether it's better or worse than the type piracy.

tkelman: I guess another strategy would be to make a temporary type with a silly name like `UseRangeGeneratorInstead`, I think some of the linspace / float-range deprecations in base were done like that

nalimilan: I'm not so concerned about the absence of a deprecation warning: it's unlikely that a package would use the type without any of the associated functions, which do print warnings. But however the type would be named, how would ensure at the same type that the old code continues to work (both code using the type and code using the functions)? AFAICT that requires type piracy.

tkelman: You'd make it an entirely new type, only for the purposes of keeping the deprecated code working. The warnings could say the right thing for the function calls, but the deprecate_binding warnings when it's used in type signatures would have the funny name.

nalimilan: OK, actually the simplest solution seems to be moving the `:RandIntSampler` code to deprecates.jl, and printing deprecation warnings from the inner constructors. No type piracy involved. The use of the binding itself does not print a message, but you can't fix the warnings without removing the uses of the type anyway, and the only package which used this unexported type was Distributions. So should be good to go.

ararslan:
What happened to `covm` on nightly...?

nalimilan:
See https://github.com/JuliaStats/StatsBase.jl/pull/281.

ararslan:
Ah that's right, thanks for taking care of that!

PR #281, opened 2017-07-02
Fix failure on Julia nightlies due to covm change

nalimilan:
'corrected' is now a keyword argument (https://github.com/JuliaLang/julia/pull/21709).


ararslan:
Oh right, AppVeyor failure is because it's enabled for the repository but the actual AppVeyor YAML file is in a different PR.

PR #284, opened 2017-07-20
Convergence exception

alexmorley:
Ref discussion & pull request here. https://github.com/JuliaStats/MultivariateStats.jl/pull/37#issuecomment-304880116

Often useful to know how far away from convergence the algorithm was.


alexmorley:
Builds only failing on nightly. Not related to this PR.
`ERROR: LoadError: LoadError: syntax: invalid escape sequence`

src/statmodels.jl
```diff
@@ -336,12 +336,30 @@ function show(io::IO, ct::CoefTable)
...
 """
 type ConvergenceException <: Exception
     iters::Int
+    change::Real
+    tol::Real
+    function ConvergenceException(iters,change,tol)
+        if tol > change 
+            error("Change must be greater than tol.")
+        else
+            new(iters,change,tol)
```
nalimilan: Spaces after commas (here and elsewhere).

test/statmodels.jl
```diff
@@ -34,5 +34,13 @@ x3   0.453058  0.72525 0.999172 0.5567
 try
     throw(ConvergenceException(10))
 catch ex
-    @test sprint(showerror, ex) == "failure to converge after 10 iterations"
+    @test sprint(showerror, ex) == "failure to converge after 10 iterations."
 end
+
+try
```
nalimilan: `@test_throws` returns an object with a `value.msg` field, no need for a duplicate `try... catch` block.

src/statmodels.jl
```diff
@@ -336,12 +336,30 @@ function show(io::IO, ct::CoefTable)
 end
 
 """
-	ConvergenceException(iters::Int)
+	ConvergenceException(iters::Int, change::Real, tol::Real)
```
nalimilan: Use `change::Real=NaN, tol::Real=NaN`.

src/statmodels.jl
```diff
@@ -336,12 +336,30 @@ function show(io::IO, ct::CoefTable)
...
 
 The fitting procedure failed to converge in `iters` number of iterations.
+`change` on final iteration was greater than specified tolerance `tol`.
```
nalimilan: Could be more explicit about what `change` is.

src/statmodels.jl
```diff
@@ -336,12 +336,30 @@ function show(io::IO, ct::CoefTable)
...
+        if tol > change 
+            error("Change must be greater than tol.")
+        else
+            new(iters,change,tol)
+        end
+    end
 end
 
-Base.showerror(io::IO, ce::ConvergenceException) = print(io, "failure to converge after $(ce.iters) iterations")
+ConvergenceException(iters) = ConvergenceException(iters,NaN,NaN)
```
nalimilan: I think you should be able to drop this method in favor of default values for the corresponding arguments in the above constructor, right?

src/statmodels.jl
```diff
@@ -336,12 +336,30 @@ function show(io::IO, ct::CoefTable)
...
+
+function Base.showerror(io::IO, ce::ConvergenceException)
+    if isnan(ce.change)
+        print(io, "failure to converge after $(ce.iters) iterations.")
+    else
+        print(io, "failure to converge after $(ce.iters) iterations. Last change ($(ce.change)) > tol($(ce.tol)).")
```
nalimilan: I'd rather print something like `Last change ($(ce.change)) was greater than tolerance ($(ce.tol)).` You can also avoid repeating the part common to the two strings by printing it unconditionally, and print the second part separately inside an `if` block.

src/statmodels.jl
```diff
@@ -336,12 +336,30 @@ function show(io::IO, ct::CoefTable)
...
 """
-	ConvergenceException(iters::Int)
+	ConvergenceException(iters::Int, change::Real, tol::Real)
 
 The fitting procedure failed to converge in `iters` number of iterations.
+`change` on final iteration was greater than specified tolerance `tol`.
 """
 type ConvergenceException <: Exception
     iters::Int
+    change::Real
```
nalimilan: To avoid type instabilities, this should be a type parameter.

src/statmodels.jl
alexmorley: I didn't think you could? But maybe I'm wrong...

test/statmodels.jl
alexmorley: Are you referring to the @test_throws below (that is testing the `tol > change` condition)?

alexmorley:
There are now failures on 0.5. I can add @compat or just wait for #287 to be merged.

src/statmodels.jl
```diff
@@ -336,12 +336,31 @@ function show(io::IO, ct::CoefTable)
 end
 
 """
-	ConvergenceException(iters::Int)
+	ConvergenceException(iters::Int, change::Real=NaN, tol::Real=Nan)
```
nalimilan: `NaN`

src/statmodels.jl
```diff
@@ -336,12 +336,31 @@ function show(io::IO, ct::CoefTable)
...
 
 The fitting procedure failed to converge in `iters` number of iterations.
+i.e. where the `change` between the cost of the final & penultimate iteration was greater than
```
nalimilan: Comma rather than dot in the previous sentence. "and" rather than "&". Remove "where".

Also, calling the field/argument `lastchange` could be more explicit?

src/statmodels.jl
```diff
@@ -336,12 +336,31 @@ function show(io::IO, ct::CoefTable)
...
+specified tolerance `tol`.
 """
-type ConvergenceException <: Exception
+type ConvergenceException{T<:Real} <: Exception
```
nalimilan: Use `struct` when rebasing after #287 has been merged.

src/statmodels.jl
```diff
@@ -336,12 +336,31 @@ function show(io::IO, ct::CoefTable)
...
     iters::Int
+    change::T
+    tol::T
+    function ConvergenceException{T}(iters, change::T, tol::T) where T<:Real
+        if tol > change 
+            error("Change must be greater than tol.")
```
nalimilan: Use `throw(ArgumentError(...))`.

test/statmodels.jl
nalimilan: I mean that you can combine both tests and not use `try... catch` at all.

src/statmodels.jl
nalimilan: I've just tested and it works AFAICT.

test/statmodels.jl
alexmorley: But the latter is comparing and throwing the `ArgumentException`?

ararslan:
Looks like you have some coverage files in here (*.mem). It would be great if you could `git rm` those and push again.

src/statmodels.jl
```diff
@@ -336,12 +336,29 @@ function show(io::IO, ct::CoefTable)
 end
 
 """
-    ConvergenceException(iters::Int)
-
-The fitting procedure failed to converge in `iters` number of iterations.
+    ConvergenceException(iters::Int, lastchange::Real=NaN, tol::Real=NaN)
+The fitting procedure failed to converge in `iters` number of iterations,
```
nalimilan: Missing blank line.

src/statmodels.jl
```diff
@@ -336,12 +336,29 @@ function show(io::IO, ct::CoefTable)
...
+        if tol > lastchange 
+            throw(ArgumentError("Change must be greater than tol."))
+        else
+            new(iters, lastchange, tol)
+        end
+    end
 end
 
-Base.showerror(io::IO, ce::ConvergenceException) = print(io, "failure to converge after $(ce.iters) iterations")
+ConvergenceException(iters, lastchange::T=NaN, tol::T=NaN) where {T<:Real} = ConvergenceException{T}(iters, lastchange, tol)
```
nalimilan: Break lines at 92 chars.

test/statmodels.jl
nalimilan: Sorry, I misread. Though you could use `@test_throws` just like you do below for a slightly more compact code.

alexmorley: Then we would have to add a msg field to ConvergenceException?

nalimilan: Actually, you could just do `@test sprint(showerror, ConvergenceException(10, 0.2, 0.1)) == ...`. There's no need to use `try... catch` just for that. Same above.

nalimilan:
Thanks!

alexmorley:
Thanks for your help!

PR #287, opened 2017-07-24
Updates for Julia 0.6

ararslan:
Ref https://github.com/JuliaLang/METADATA.jl/pull/10398. Decided to go the whole nine yards:

* Remove Compat dependency
* Migrate to `where`
* Remove now-irrelevant `VERSION` checks
* Bump the Julia lower bound to 0.6

The diff is pretty big (I ended up touching pretty much every file) but the changes are pretty straightforward.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -10,25 +10,25 @@
 # covariant type notation, i.e. AbstractVector{<:Real}
 #
 
-@compat RealArray{T<:Real,N} = AbstractArray{T,N}
-@compat RealVector{T<:Real} = AbstractArray{T,1}
-@compat RealMatrix{T<:Real} = AbstractArray{T,2}
+const RealArray{T<:Real,N} = AbstractArray{T,N}
+const RealVector{T<:Real} = AbstractArray{T,1}
+const RealMatrix{T<:Real} = AbstractArray{T,2}
 
-@compat IntegerArray{T<:Integer,N} = AbstractArray{T,N}
-@compat IntegerVector{T<:Integer} = AbstractArray{T,1}
-@compat IntegerMatrix{T<:Integer} = AbstractArray{T,2}
+const IntegerArray{T<:Integer,N} = AbstractArray{T,N}
+const IntegerVector{T<:Integer} = AbstractArray{T,1}
+const IntegerMatrix{T<:Integer} = AbstractArray{T,2}
 
-@compat const RealFP = Union{Float32, Float64}
+const RealFP = Union{Float32, Float64}
 
 ## conversion from real to fp types
 
-@compat fptype{T<:Union{Float32,Bool,Int8,UInt8,Int16,UInt16}}(::Type{T}) = Float32
-@compat fptype{T<:Union{Float64,Int32,UInt32,Int64,UInt64,Int128,UInt128}}(::Type{T}) = Float64
+fptype{T<:Union{Float32,Bool,Int8,UInt8,Int16,UInt16}}(::Type{T}) = Float32
+fptype{T<:Union{Float64,Int32,UInt32,Int64,UInt64,Int128,UInt128}}(::Type{T}) = Float64
 fptype(::Type{Complex64}) = Complex64
 fptype(::Type{Complex128}) = Complex128
 
 # A convenient typealias for deprecating default corrected Bool
-@compat const DepBool = Union{Bool, Void}
+const DepBool = Union{Bool, Void}
 
 function depcheck(fname::Symbol, b::DepBool)
     if b == nothing
--- a/src/deviation.jl
+++ b/src/deviation.jl
@@ -46,7 +46,7 @@ end
 Compute the squared L2 distance between two arrays: ``\\sum_{i=1}^n |a_i - b_i|^2``.
 Efficient equivalent of `sumabs2(a - b)`.
 """
-function sqL2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
+function sqL2dist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
@@ -64,7 +64,7 @@ end
 Compute the L2 distance between two arrays: ``\\sqrt{\\sum_{i=1}^n |a_i - b_i|^2}``.
 Efficient equivalent of `sqrt(sumabs2(a - b))`.
 """
-L2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = sqrt(sqL2dist(a, b))
+L2dist(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:Number} = sqrt(sqL2dist(a, b))
 
 
 # L1 distance
@@ -74,7 +74,7 @@ L2dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = sqrt(sqL2dist(a, b
 Compute the L1 distance between two arrays: ``\\sum_{i=1}^n |a_i - b_i|``.
 Efficient equivalent of `sum(abs, a - b)`.
 """
-function L1dist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
+function L1dist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
@@ -93,7 +93,7 @@ Compute the L∞ distance, also called the Chebyshev distance, between
 two arrays: ``\\max_{i\\in1:n} |a_i - b_i|``.
 Efficient equivalent of `maxabs(a - b)`.
 """
-function Linfdist{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T})
+function Linfdist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
@@ -115,7 +115,7 @@ Compute the generalized Kullback-Leibler divergence between two arrays:
 ``\\sum_{i=1}^n (a_i \\log(a_i/b_i) - a_i + b_i)``.
 Efficient equivalent of `sum(a*log(a/b)-a+b)`.
 """
-function gkldiv{T<:AbstractFloat}(a::AbstractArray{T}, b::AbstractArray{T})
+function gkldiv(a::AbstractArray{T}, b::AbstractArray{T}) where T<:AbstractFloat
     n = length(a)
     r = 0.0
     for i = 1:n
@@ -137,7 +137,7 @@ end
 
 Return the mean absolute deviation between two arrays: `mean(abs(a - b))`.
 """
-meanad{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = L1dist(a, b) / length(a)
+meanad(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:Number} = L1dist(a, b) / length(a)
 
 
 # MaxAD: maximum absolute deviation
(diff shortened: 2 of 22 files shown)


ararslan:
32-bit Windows failure on 0.6 is unrelated, and the 0.5 failures are weird because I removed 0.5 from the AppVeyor config...

src/common.jl
```diff
@@ -10,25 +10,25 @@
 # covariant type notation, i.e. AbstractVector{<:Real}
 #
 
-@compat RealArray{T<:Real,N} = AbstractArray{T,N}
-@compat RealVector{T<:Real} = AbstractArray{T,1}
-@compat RealMatrix{T<:Real} = AbstractArray{T,2}
+const RealArray{T<:Real,N} = AbstractArray{T,N}
```
nalimilan: FWIW, `const` isn't needed when there are type parameters -- but using it is fine.

src/deviation.jl
```diff
@@ -137,7 +137,7 @@ end
 
 Return the mean absolute deviation between two arrays: `mean(abs(a - b))`.
 """
-meanad{T<:Number}(a::AbstractArray{T}, b::AbstractArray{T}) = L1dist(a, b) / length(a)
+meanad(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:Number} = L1dist(a, b) / length(a)
```
nalimilan: Split line at 92 chars (same below).

nalimilan: Split line at 92 chars (same below).

src/scalarstats.jl
```diff
@@ -477,9 +475,8 @@ function crossentropy{T<:Real}(p::AbstractArray{T}, q::AbstractArray{T})
     return -s
 end
 
-function crossentropy{T<:Real}(p::AbstractArray{T}, q::AbstractArray{T}, b::Real)
-    return crossentropy(p,q) / log(b)
-end
+crossentropy(p::AbstractArray{T}, q::AbstractArray{T}, b::Real) where {T<:Real} =
```
nalimilan: Is there a rule to decide when to use `{}` and when to omit it? Looks like kind of inconsistent.

src/signalcorr.jl
```diff
@@ -453,23 +454,23 @@ If both `x` and `y` are vectors, return a vector of the same length as
...
     crosscor!(Matrix{fptype(T)}(length(lags), size(y,2)), float(x), float(y), lags; demean=demean)
 end
 
-function crosscor{T<:Real}(x::AbstractMatrix{T}, y::AbstractMatrix{T}, lags::IntegerVector; demean::Bool=true)
+function crosscor(x::AbstractMatrix{T}, y::AbstractMatrix{T}, lags::IntegerVector; demean::Bool=true) where T<:Real
     crosscor!(Array{fptype(T),3}(length(lags), size(x,2), size(y,2)), float(x), float(y), lags; demean=demean)
 end
 
-crosscor{T<:Real}(x::AbstractVecOrMat{T}, y::AbstractVecOrMat{T}; demean::Bool=true) = crosscor(x, y, default_crosslags(size(x,1)); demean=demean)
+crosscor(x::AbstractVecOrMat{T}, y::AbstractVecOrMat{T}; demean::Bool=true) where {T<:Real} = crosscor(x, y, default_crosslags(size(x,1)); demean=demean)
```
nalimilan: `T` isn't needed here.

src/deviation.jl
nalimilan: Split line at 92 chars.

nalimilan: Split line at 92 chars (same below).

src/hist.jl
```diff
@@ -186,56 +186,56 @@ binindex{T,N,E}(h::Histogram{T,N,E}, xs::NTuple{N,Real}) =
...
 
-@inline _edge_binvolume{V}(::Type{V}, edge::AbstractVector, i::Integer) = V(edge[i+1]) - V(edge[i])
-@inline _edge_binvolume{V}(::Type{V}, edge::Range, i::Integer) = V(step(edge))
+@inline _edge_binvolume(::Type{V}, edge::AbstractVector, i::Integer) where {V} = V(edge[i+1]) - V(edge[i])
+@inline _edge_binvolume(::Type{V}, edge::Range, i::Integer) where {V} = V(step(edge))
 @inline _edge_binvolume(edge::AbstractVector, i::Integer) = _edge_binvolume(eltype(edge), edge, i)
 
 
-@inline _edges_nbins{N}(edges::NTuple{N,AbstractVector}) = map(_edge_nbins, edges)
+@inline _edges_nbins(edges::NTuple{N,AbstractVector}) where {N} = map(_edge_nbins, edges)
```
nalimilan: While we're at it, I wonder whether this was intended to be `::Tuple{Vararg{AbsractVector}}`. IIRC since 0.6 `NTuple` implies homogeneous element types, but maybe our tests don't cover this. (No need to fix it in this PR though.)

src/hist.jl
```diff
@@ -186,56 +186,56 @@ binindex{T,N,E}(h::Histogram{T,N,E}, xs::NTuple{N,Real}) =
...
 
 Histogram(edge::AbstractVector, closed::Symbol=:default_left, isdensity::Bool=false) =
     Histogram((edge,), closed, isdensity)
 
 
-push!{T,E}(h::AbstractHistogram{T,1,E}, x::Real, w::Real) = push!(h, (x,), w)
-push!{T,E}(h::AbstractHistogram{T,1,E}, x::Real) = push!(h,x,one(T))
-append!{T}(h::AbstractHistogram{T,1}, v::AbstractVector) = append!(h, (v,))
-append!{T}(h::AbstractHistogram{T,1}, v::AbstractVector, wv::Union{AbstractVector,AbstractWeights}) = append!(h, (v,), wv)
+push!(h::AbstractHistogram{T,1,E}, x::Real, w::Real) where {T,E} = push!(h, (x,), w)
```
nalimilan: `E` isn't used here and on the next line.

src/hist.jl
```diff
@@ -186,56 +186,56 @@ binindex{T,N,E}(h::Histogram{T,N,E}, xs::NTuple{N,Real}) =
...
+fit(::Type{Histogram{T}},v::AbstractVector, wv::AbstractWeights; closed::Symbol=:default_left, nbins=sturges(length(v))) where {T} =
     fit(Histogram{T}, (v,), wv; closed=closed, nbins=nbins)
 
-fit{W}(::Type{Histogram}, v::AbstractVector, wv::AbstractWeights{W}, args...; kwargs...) = fit(Histogram{W}, v, wv, args...; kwargs...)
+fit(::Type{Histogram}, v::AbstractVector, wv::AbstractWeights{W}, args...; kwargs...) where {W} = fit(Histogram{W}, v, wv, args...; kwargs...)
 
 # N-dimensional
 
-function push!{T,N}(h::Histogram{T,N},xs::NTuple{N,Real},w::Real)
+function push!(h::Histogram{T,N},xs::NTuple{N,Real},w::Real) where {T,N}
```
nalimilan: Same question about `NTuple`.

src/hist.jl
```diff
@@ -350,23 +338,23 @@ h = fit(Histogram, (rand(100),rand(100)),nbins=10)
 ```
 """
 fit(::Type{Histogram}, args...; kwargs...) = fit(Histogram{Int}, args...; kwargs...)
-fit{N,W}(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::AbstractWeights{W}, args...; kwargs...) = fit(Histogram{W}, vs, wv, args...; kwargs...)
+fit(::Type{Histogram}, vs::NTuple{N,AbstractVector}, wv::AbstractWeights{W}, args...; kwargs...) where {N,W} = fit(Histogram{W}, vs, wv, args...; kwargs...)
 
 
 # Get a suitable high-precision type for the norm of a histogram.
-norm_type{T, N, E}(h::Histogram{T, N, E}) =
+norm_type(h::Histogram{T,N,E}) where {T,N,E} =
```
nalimilan: `N` and `E` aren't used. Same in some places below (including docstrings).

src/hist.jl
```diff
@@ -391,19 +379,19 @@ Calculate the norm of histogram `h` as the absolute value of its integral.
...
+float(h::Histogram{T,N,E}) where {T<:AbstractFloat,N,E} = h
 
-float{T, N, E}(h::Histogram{T, N, E}) = Histogram(h.edges, float(h.weights), h.closed, h.isdensity)
+float(h::Histogram{T,N,E}) where {T,N,E} = Histogram(h.edges, float(h.weights), h.closed, h.isdensity)
 
 
 
 """
-    normalize!{T<:AbstractFloat, N, E}(h::Histogram{T, N, E}, aux_weights::Array{T,N}...; mode::Symbol = :pdf)
+    normalize!(h::Histogram{T,N,E}, aux_weights::Array{T,N}...; mode::Symbol=:pdf) where {T<:AbstractFloat,N,E}
```
nalimilan: You can omit `E` in the docstring at least.

src/scalarstats.jl
ararslan: In Base the convention is `{}` for inline definitions and no braces for `function` definitions.

src/signalcorr.jl
ararslan: It is, it's constraining `x` and `y` to have the same element type.

src/common.jl
ararslan: Yeah, I just personally prefer it since it adds a nice visual distinction from randomly assigning some variable in the middle of package code. 😛

src/deviation.jl
ararslan: That's actually a larger issue with this package: the lines are looooooooong. I just did the quick and dirty syntax rewrite in this PR but we should consider going through and doing a stylistic cleanup in a separate PR.

ararslan:
> I wonder whether it's still justified to drop spaces after commas for type parameters: that's a really subtle exception.

I agree that it's a subtle exception but Base tends not to use spaces after commas in type parameters (and personally I prefer it with no spaces). As for documenting the style, it's not too bad to say "space after a comma except for the trailing comma in a 1-tuple and after commas in type parameters."

src/deviation.jl
nalimilan: Yeah, I noticed that too. I mentioned it because the new syntax is slightly longer in that case, which makes this line go beyond the 92 char threshold.

src/hist.jl
ararslan: Actually I think this works okay as-is:
```julia
julia> ([1], [1.0]) isa NTuple{2,AbstractVector}
true
```

src/deviation.jl
ararslan: This one got bumped to 93. So close!

ararslan:
Why are Codecov comments so difficult to disable properly...

ararslan:
I was overzealous in cleaning up the use of typevars. I'm just going to remove that commit and do it over.

src/hist.jl
```diff
@@ -132,30 +132,30 @@ function sturges(n)  # Sturges' formula
...
+                              closed::Symbol, isdensity::Bool=false) where {T,N,E}
         closed == :right || closed == :left || error("closed must :left or :right")
         isdensity && !(T <: AbstractFloat) && error("Density histogram must have float-type weights")
         _edges_nbins(edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
         new{T,N,E}(edges,weights,closed,isdensity)
     end
 end
 
-Histogram{T,N}(edges::NTuple{N,AbstractVector},weights::AbstractArray{T,N},closed::Symbol=:default_left, isdensity::Bool=false) =
+Histogram(edges::NTuple{N,AbstractVector},weights::AbstractArray{T,N},closed::Symbol=:default_left, isdensity::Bool=false) where {T,N} =
```
ararslan: Okay, pushed a commit to shorten those. We should do a separate pass over the code here for stylistic updates in another PR.

src/hist.jl
nalimilan: Got it, the form that requires homogeneous eltypes is `f(edges::NTuple{N,T}) where {N, T}`.

PR #291, opened 2017-08-04
Support empirical PDF via Histograms

anriseth:
I couldn't find a function in the Julia-verse that creates a pdf from data/a histogram. Hope this is the place to PR...

### DIFF
--- a/src/empirical.jl
+++ b/src/empirical.jl
@@ -45,3 +45,34 @@ function ecdf(X::RealVector{T}) where T<:Real
 end
 
 
+"""
+    epdf(X)
+
+Return an empirical probability distribution function (EPDF) based on a vector of samples
+given in `X`.
+
+Note: this is a higher-level function that returns a function, which can then be applied
+to evaluate PDF values on other samples.
+"""
+function epdf(X::AbstractVector; nbins::Int = sturges(length(X)), closed::Symbol=:left)
+    # TODO: Support edges
+    hg = normalize(fit(Histogram, X; closed=closed, nbins=nbins); mode=:pdf)
+    if hg.closed == :right
+        le = <=
+        ge = >
+    else
+        le = <
+        ge = >=
+    end
+
+    function ef(x)
+        if le(x,hg.edges[1][1]) || ge(x,hg.edges[1][end])
+            return zero(eltype(hg.weights))
+        end
+
+        idx = findfirst(le.(x,hg.edges[1]))
+        return hg.weights[idx-1]
+    end
+
+    return ef
+end
--- a/test/empirical.jl
+++ b/test/empirical.jl
@@ -1,7 +1,8 @@
 using StatsBase
 using Base.Test
 
-fnecdf = ecdf(randn(10000000))
+X = randn(10000000)
+fnecdf = ecdf(X)
 @test isapprox(fnecdf([-1.96, -1.644854, -1.281552, -0.6744898, 0, 0.6744898, 1.281552, 1.644854, 1.96]),
     [0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 0.9, 0.95, 0.975], atol=1e-3)
 @test isapprox(fnecdf(1.96), 0.975, atol=1e-3)
@@ -10,3 +11,9 @@ fnecdf = ecdf(randn(10000000))
 
 fnecdf = ecdf([0.5])
 @test fnecdf([zeros(5000); ones(5000)]) == [zeros(5000); ones(5000)]
+
+fnepdf = epdf(X)
+@test isapprox(fnepdf.([-10.0,0.0]), [0.0,0.383]; atol = 2e-3)
+
+fnepdf = epdf(X; nbins = 100)
+@test isapprox(fnepdf.([-10.0,0.0]), [0.0,0.398]; atol = 2e-3)
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -119,13 +119,14 @@ module StatsBase
 
     ## empirical
     ecdf,           # empirical cumulative distribution function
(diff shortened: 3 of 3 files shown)


ararslan:
Nice, thanks for the contribution! I wonder whether this package or [KernelDensity](https://github.com/JuliaStats/KernelDensity.jl) would be the best home for this. Thoughts @simonbyrne, master of KDEs, and @nalimilan?

anriseth:
Yeah, I wasn't sure exactly where this belonged. I chose StatsBase because
it already had an empirical CDF function.

src/empirical.jl
```diff
@@ -45,3 +45,34 @@ function ecdf(X::RealVector{T}) where T<:Real
...
+    if hg.closed == :right
+        le = <=
+        ge = >
+    else
+        le = <
+        ge = >=
+    end
+
+    function ef(x)
+        if le(x,hg.edges[1][1]) || ge(x,hg.edges[1][end])
```
nalimilan: Spaces after commas (here and elsewhere).

src/empirical.jl
```diff
@@ -45,3 +45,34 @@ function ecdf(X::RealVector{T}) where T<:Real
...
+            return zero(eltype(hg.weights))
+        end
+
+        idx = findfirst(le.(x,hg.edges[1]))
```
nalimilan: Rather do `findfirst(y -> le(y, hg.edges[1]), x)` to avoid an allocation. Could also use `searchsortedfirst`.

test/empirical.jl
```diff
@@ -10,3 +11,9 @@ fnecdf = ecdf(randn(10000000))
 
 fnecdf = ecdf([0.5])
 @test fnecdf([zeros(5000); ones(5000)]) == [zeros(5000); ones(5000)]
+
+fnepdf = epdf(X)
+@test isapprox(fnepdf.([-10.0,0.0]), [0.0,0.383]; atol = 2e-3)
+
+fnepdf = epdf(X; nbins = 100)
+@test isapprox(fnepdf.([-10.0,0.0]), [0.0,0.398]; atol = 2e-3)
```
nalimilan: Should also test `closed=:right`.

Also, why use `isapprox`? AFAICT we can hardcode the exact value and use `==`.

anriseth: I just copied the ECDF tests, where they use `isapprox`. The values depend on the sample `X`, so this would require that `X` is the same on all platforms etc. Does Julia generate the same pseudo-random numbers across different architectures?

src/empirical.jl
anriseth: Thanks, I wasn't aware of `searchsortedfirst`

anriseth:
> Regarding the implementation, I'm not sure defining le and ge is the best approach: have you checked whether the compiler was able to resolve these statically? The code would also look slightly clearer and wouldn't be much longer if you wrote one version for :right and one for :left.

I've not checked if the compiler resolves this. I rewrote the `epdf` to create different functions depending on `closed`. I'm not sure if this code is more clear, however :p

test/empirical.jl
nalimilan: Yes, normally the random numbers are the same as long as you call `srand`. But if the existing tests use that pattern, let's keep the same approach.

nalimilan:
> I've not checked if the compiler resolves this. I rewrote the epdf to create different functions depending on closed. I'm not sure if this code is more clear, however :p

Yeah, it's a bit verbose too, but that's not terrible.

anriseth:
Are there any other changes I should make to this PR?

andreasnoack:
Sorry for joining late here. Wouldn't it make sense to overload `pdf` for `Histogram`s instead introducing a new function? It would also avoid the possible confusion that the two `e` functions `epdf` and `ecdf` wouldn't reflect the same probability measure.

anriseth:
> Wouldn't it make sense to overload pdf for Histograms instead of introducing a new function?

That was my initial plan before creating `epdf`. I can PR Distributions.jl instead if the consensus is that this is a better option?

nalimilan:
Makes sense. But what's the discrepancy between `epdf` and `ecdf` you mention?

andreasnoack:
The `ecdf(x)` is always just the relative frequency of observations below `x`. However, the `pdf` corresponding to the `ecdf` would be a weird rough guy. The function would be stepwise constant between each of the sorted observations but the value of the pdf would jump up and down. It would be a histogram with very small and unevenly sized bins. Hence, you'd like to smooth it by using larger bins but the `cdf` corresponding to such a histogram with larger bins would be different from the `ecdf`.

anriseth:
Okay, adding this to `Distributions.jl` would probably require that I create a `ContinuousUnivariateDistribution` for `Histogram`. I don't need this functionality anymore myself, so that is more effort than I'm willing to put into that. A pdf function by itself may not be that useful, however, so I'm not sure if we want to have something like that for Histograms at all.

I propose three alternatives
- Leave it as `epdf` and document that it is different from `ecdf`
- Create a `pdf(hg::Histogram,x)` function in StatsBase (would there be a way to do this without checking `hg.closed` each time?)
- Close this PR and let someone else take care of implementing a `HistogramDistribution` in `Distributions.jl`

nalimilan:
I'd say option 2 is the best one. I'm not sure about option 3, it sounds like we would have to move all the histogram functionality to Distributions.jl if we did that.

andreasnoack:
I've thought a bit more about this after the latest changes to `Histogram`. I think we might have to do this as part of a proper distribution in `Distributions` where we can assume certain properties about the `Histogram` after the construction of the distribution. Here in `StatsBase` we allow a `Histrogram` to be so many things, instead of just a density estimate, that overloading `pdf` might not be practical because I think it would have to check the normalization at each call to pdf.

PR #293, opened 2017-08-16
Add proportion normalization to histograms

mkborregaard:
This adds the option of a `:proportion` normalization to histograms. This was suggested here: https://discourse.julialang.org/t/plots-how-to-create-a-histogram-such-that-sum-of-bar-heights-1/5392/16

The proportion normalization is the normalization where the sum of all weights (the height of the histogram bars in a plot) is = 1. It is not a density-controlled normalization. It can be more intuitive to read from a diagram than a pdf normalization.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -442,8 +442,8 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
             for A in aux_weights
                 A ./= s
             end
-        else mode != :pdf && mode != :density
-            throw(ArgumentError("Normalization mode must be :pdf, :density, :proportion or :none"))
+        else 
+            throw(ArgumentError("Normalization mode must be :pdf, :density, :fraction or :none"))
         end
         h
     end
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -425,7 +425,6 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
                 end
             end
             h.isdensity = true
-        elseif mode == :proportion
             if h.isdensity #if it already is a density, reverse that operation
                 SumT = norm_type(h)
                 vs_0 = one(SumT)
@@ -437,6 +436,7 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
                 end
                 h.isdensity = false
             end
+        elseif mode == :fraction
             s = sum(weights)
             weights ./= s
             for A in aux_weights
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -425,8 +425,25 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
                 end
             end
             h.isdensity = true
+        elseif mode == :proportion
+            if h.isdensity #if it already is a density, reverse that operation
+                SumT = norm_type(h)
+                vs_0 = one(SumT)
+                @inbounds @nloops $N i weights d->(vs_{$N-d+1} = vs_{$N-d} * _edge_binvolume(SumT, edges[d], i_d)) begin
+                    (@nref $N weights i) *= $(Symbol("vs_$N"))
+                    for A in aux_weights
+                        (@nref $N A i) *= $(Symbol("vs_$N"))
+                    end
+                end
+                h.isdensity = false
+            end
+            s = sum(weights)
+            weights ./= s
+            for A in aux_weights
+                A ./= s
+            end
         else mode != :pdf && mode != :density
-            throw(ArgumentError("Normalization mode must be :pdf, :density or :none"))
+            throw(ArgumentError("Normalization mode must be :pdf, :density, :proportion or :none"))
         end
         h
     end
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -399,6 +399,10 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
             (size(A) != size(weights)) && throw(DimensionMismatch("aux_weights must have same size as histogram weights"))
         end
 
+        if mode == :fraction && h.isdensity
+            mode = :pdf
+        end
+
         if mode == :none
             # nothing to do
         elseif mode == :pdf || mode == :density
@@ -425,17 +429,6 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
                 end
             end
             h.isdensity = true
-            if h.isdensity #if it already is a density, reverse that operation
-                SumT = norm_type(h)
-                vs_0 = one(SumT)
-                @inbounds @nloops $N i weights d->(vs_{$N-d+1} = vs_{$N-d} * _edge_binvolume(SumT, edges[d], i_d)) begin
-                    (@nref $N weights i) *= $(Symbol("vs_$N"))
-                    for A in aux_weights
-                        (@nref $N A i) *= $(Symbol("vs_$N"))
-                    end
-                end
-                h.isdensity = false
-            end
         elseif mode == :fraction
             s = sum(weights)
             weights ./= s
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -401,9 +401,9 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
 
         if mode == :none
             # nothing to do
-        elseif mode == :pdf || mode == :density || mode == :fraction
+        elseif mode == :pdf || mode == :density || mode == :probability
             if h.isdensity
-                if mode == :pdf || mode == :fraction
+                if mode == :pdf || mode == :probability
                     # histogram already represents a density, just divide weights by norm
                     s = 1/norm(h)
                     weights .*= s
@@ -426,7 +426,7 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
                     end
                     h.isdensity = true
                 else
-                    # :fraction - divide weights by sum of weights
+                    # :probability - divide weights by sum of weights
                     nf = inv(sum(weights))
                     weights .*= nf
                     for A in aux_weights
@@ -435,7 +435,7 @@ arrays appropriately. See description of `normalize` for details. Returns `h`.
                 end
             end
         else
-            throw(ArgumentError("Normalization mode must be :pdf, :density, :fraction or :none"))
+            throw(ArgumentError("Normalization mode must be :pdf, :density, :probability or :none"))
         end
         h
     end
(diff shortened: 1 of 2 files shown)


mkborregaard:
Ping @oschulz for comments.

oschulz:
Personally, I'm not against it, in principle - I find it hard to describe the mathematical result of this normalization, though. A few points:

* Maybe better call it `:weights` or so, instead of  `:proportion`?

* I'd go for a simpler implementation that just fails if the histogram is density-normalized. Silently reversing a density-normalization seems a bit dangerous to me, and I can't see typical use-cases for it. IMHO we should only allow this kind of normalization  for `h.isdensity == false`.

oschulz:
If we had this, then a `:weights` normalization followed by a `:density` normalization would be the equivalent of a `:pdf` normalization. But what would we get from a `:density` normalization followed by a `:weights` normalization? Hm, I'm not sure I like the implications of this.

mkborregaard:
I'm fine with both of those suggestions.

oschulz:
Strike my last - if only allow `:weights`-normalization for `h.isdensity == false`, then the problematic case of a `:density` normalization followed by a `:weights` normalization would be disallowed, and there wouldn't be a problem.

mkborregaard:
But note that in the current implementation
```julia
a, b = randn(1000), randn(1000)
h = fit(Histogram, (a,b))
h_prop = normalize(h, mode = :proportion)
h_density = normalize(h, mode = :density)
h_propafterdensity = normalize(h_density, mode = :proportion)
h_prop == h_propafterdensity
# true
``

mkborregaard:
Travis error on nightly is unrelated - Histogram tests pass.

oschulz:
> But note that in the current implementation density-then-weights should not be problematic at all

It would result in something different then weights-then-density, though. We'd have density then weights equivalent to weights only, but weights then density equivalent to pdf. This seems very ugly/asymmetic to me, and dangerous, because users won't expect it. Plus, I really don't think a normalization should revert another one.

mkborregaard:
I've

1. changed the name from :proportion to :fraction
2. changed the behaviour of the normalization when the histogram is already a density. Now applying both :density and :fraction to the same histogram will always result in a :pdf, regardless of the order.
Everybody happy with that?
The changes should also address the codecov report above.

oschulz:
I was a bit sceptical about this, initially, but I like the final outcome: The new `:fraction`-normalization fits in very naturally now, and IMHO in a mathematically clean way:

* `:fraction` normalizes by sum of weights
* `:density` normalizes by bin sizes
* `:pdf` normalizes by sum of weights and bin sizes

Therefore, a `:fraction` normalization followed by a `:density` normalization (or the other way round) is equivalent to a `:pdf` normalization. All modes are idempotent.

oschulz:
Note: The above will hold true only after @mkborregaard merges my PR mkborregaard/StatsBase.jl#1, so please don't merge this yet.

mkborregaard:
Merged, thanks @oschulz that's a lot nicer.

oschulz:
Thanks, Michael!
IMHO this is ready for review (and hopefully merging) now.

mkborregaard:
Bump? :-)

oschulz:
I haven't forgotten - just a bit overloaded, atm. :-)

mkborregaard:
I didn't mean you, you already gave feedback :-)

sethchandler:
It was suggested that I add my comments in this forum. So here they are. I apologize if this is not the right format or forum; I am not really a software developer (more of a user) and have never commented about a PR before.

Two more plugs for more flexible normalization options within histogram. (And I know these are not the world's strongest arguments, but ...)

1. [Matlab does it](https://www.mathworks.com/help/matlab/ref/histogram.html?requestedDomain=www.mathworks.com). 'Normalization' with a valid option ('count', 'probability', 'countdensity', 'pdf', 'cumcount', or 'cdf') 
2. [Wikipedia](https://en.wikipedia.org/wiki/Histogram) says it's fine. "A histogram may also be normalized to display "relative" frequencies. It then shows the proportion of cases that fall into each of several categories, with the sum of the heights equaling 1.
However, bins need not be of equal width; in that case, the erected rectangle is defined to have its area proportional to the frequency of cases in the bin. The vertical axis is then not the frequency but frequency density — the number of cases per unit of the variable on the horizontal axis."

oschulz:
> However, bins need not be of equal width; in that case, the erected rectangle is defined to have its area proportional to the frequency of cases in the bin.

We don't do this currently, in this PR, but I like it, it seems cleaner than just ignoring unequal bin widths. We'll have to change the code though, to make ensure fraction ∘ denstiy is equivalent to pdf. I'll wait for @simonbyrne's initial review, though, to see if this new normalization is acceptable in principle.

oschulz:
> However, bins need not be of equal width; in that case, the erected rectangle is defined to have its area proportional to the frequency of cases in the bin.

We don't do this currently, in this PR - but I like it, it seems cleaner than just ignoring unequal bin widths.
It would mean that the new normalization is simply PDF multiplied by the total volume of all bins. I'll wait for @simonbyrne's initial review, though, to see if this new normalization is acceptable in principle.

mkborregaard:
"the erected rectangle is defined to have its area proportional to the frequency of cases in the bin" is just the current :density transform.

andreasnoack:
Just catching up here. I think this is fine to have. It is true that many people are not comfortable with densities. The result of this scaling is a funny pmf where the outcomes are contiguous intervals so I can see why the label "probability" makes sense and since it is the choice in both Matlab and Mathematica I'm wondering if we shouldn't just follow along.

The test failure should be fixed by https://github.com/JuliaStats/StatsBase.jl/pull/296 so rebasing should hopefully give all green lights.

mkborregaard:
In the PR the label is currently `:fraction` though, but happy to go with whichever name you feel makes the best sense.

andreasnoack:
> In the PR the label is currently `:fraction` though

I'm aware. That was the "wondering" part of my comment but I should probably have been more clear.

> but happy to go with whichever name you feel makes the best sense.

Since the two others use that name and that I think the name makes sense, I'd prefer to use `:probability` unless you'd argue strongly against it (which it appears that you don't 😄 )

oschulz:
@mkborregaard,  I'll do the changes regarding bin-width treatment and make a PR against your PR, Ok?

mkborregaard:
What are those changes? Try to look at the Wikipedia page itself. It introduces everything exactly as it is / will be with this PR (my headings, but the text is unchanged):

"
**Normalization = none**
If the bins are of equal size, a rectangle is erected over the bin with height proportional to the frequency — the number of cases in each bin. 

**Normalization = probability**
A histogram may also be normalized to display "relative" frequencies. It then shows the proportion of cases that fall into each of several categories, with the sum of the heights equaling 1.

**Normalization = density**
However, bins need not be of equal width; in that case, the erected rectangle is defined to have its area proportional to the frequency of cases in the bin.[3] The vertical axis is then not the frequency but frequency density — the number of cases per unit of the variable on the horizontal axis. Examples of variable bin width are displayed on Census bureau data below.

As the adjacent bins leave no gaps, the rectangles of a histogram touch each other to indicate that the original variable is continuous.[4]

**Normalization = pdf**
Histograms give a rough sense of the density of the underlying distribution of the data, and often for density estimation: estimating the probability density function of the underlying variable. The total area of a histogram used for probability density is always normalized to 1."

oschulz:
> What are those changes?

Currently, we don't take unequal bin widths into account. The Wikipedia page reads

> If the bins are of equal size, a rectangle is erected over the bin with height proportional to the frequency — the number of cases in each bin. A histogram may also be normalized to display "relative" frequencies. It then shows the proportion of cases that fall into each of several categories, with the sum of the heights equaling 1.
>
> However, bins need not be of equal width; in that case, the erected rectangle is defined to have its area proportional to the frequency of cases in the bin.[3] The vertical axis is then not the frequency but frequency density [...]

The question whether the second paragraph applies to said normalization by "relative" frequencies. If yes, we have to code this up a bit differently, and we'd have `isdensity == true` after the normalization. It would still be different from `:density` because the latter doesn't set the sum of weights to 1. So it would change the meaning: `:probability` would actually be more similar to `:pdf`, just with the sum of weights as one, instead of the integral, so the ratio between them would the sum of all bin volumes.

mkborregaard:
The second paragraph clearly doesn't pertain to relative frequencies. You have two paragraphs, one starting "If bins are of equal size...", the other starting "However, bins need not be of equal width...". The "probability" histogram is in the first paragraph.

oschulz:
> The second paragraph clearly doesn't pertain to relative frequencies.

It's not absolutely obvious, to me - but I would actually prefer that interpretation, because otherwise our current neat "probability-norm plus density-norm equals pdf-norm" would no longer hold true. So if the others agree, I'm fine with merging it as it is.

mkborregaard:
Thanks @andreasnoack !

There is one consideration you may want to think about with histograms. In the current implementation, the normalization is an _operation_ on the histogram. So if you have a :density transformed histogram and :probability-transform it, it will become a :pdf-histogram. Any normalization on a :pdf-transformed histogram will leave it unchanged. This is the neat correspondence between normalizations  that @oschulz refers to. (The :none transformation does nothing.)

It also means that it is not possible to revert the effect of a normalization. If the histogram had an `N::Int` field, an alternative implementation would instead allow calling `normalize(hist, mode = :count)` and regenerate the original count histogram, `normalize(hist, mode = :density)` and always regenerate the :density-histogram (i.e. area of boxes in units of the x axis equal the count in each box). 

These two methods are not compatible, and you simply have to choose one of them. Just flagging this for your awareness.

oschulz:
> If the histogram had an N::Int field, an alternative implementation would instead allow calling normalize(hist, mode = :count) and regenerate the original count histogram [...]

Here's my take on this: This would be a breaking change without a strong need. I'm also not comfortable with a normalization trying to "undo" a previous one - especially since the result may not be correct: It's not unusual that one has to deal with pre-histogrammed, pre-normalized data, so the value of `N` is not available. So I'm really not in favour of adding an `N` field. I fear that in actual use, it may often end up set to the wrong (or no) value.

mkborregaard:
Yes, those are also very valid points. In fact I think there are pros and cons to both. I just wanted to highlight this concern to the StatsBase crew.

PR #295, opened 2017-08-19
added normalization in docstrings

### LINKED ISSUE #294: Please state in docstrings the normalization of autocor and crosscor
One can show [analytically](https://math.stackexchange.com/questions/1518206/finding-the-autocorrelation-of-a-sine-wave) that the autocorrelation of a sine or cosine signal `cos(t)` is `0.5cos(τ)` with `τ` the delay, aka lag.

This MWE:
```julia
using StatsBase
t = 0:0.01:200π
x = cos.(t)
a = autocor(x)
``` 
produces clearly `a = cos(τ)` instead of `a = 0.5cos(τ)`. Why? (because you normalize with std^2)

What is the normalization you are using? Looking at the source code I can see that it is the `dot(x,x)`. Since there are so many different digital processing and timeseries analysis books and each scientist may use a different approach (I have seen some normalizing by the std^2 of the signal (so that the autocor is 1 at τ=0) and some others only by the length), could you consider stating the normalization in the docstrings?

EDIT: Sorry if it wasn't obvious from my question: It is clear after looking at the source that you normalize for the std squared so that the correlation is always 1 at 0. But since there are also other approaches, you could add *only one* sentence to the doc strings to make people avoid looking at source (or just assuming you do this normalization).

Datseris:
Please be sure to check if everything I've written is actually correct!

I also tried to not mess around with the absence or not of `!` .

Fixes  #294.

### DIFF
--- a/src/signalcorr.jl
+++ b/src/signalcorr.jl
@@ -56,6 +56,8 @@ before computing the autocovariance.
 If `x` is a vector, `r` must be a vector of the same length as `x`.
 If `x` is a matrix, `r` must be a matrix of size `(length(lags), size(x,2))`, and
 where each column in the result will correspond to a column in `x`.
+
+The output is normalized by `size(x, 1)`. See `autocor!` for normalization including the variance.
 """
 function autocov!(r::RealVector, x::AbstractVector{T}, lags::IntegerVector; demean::Bool=true) where T<:RealFP
     lx = length(x)
@@ -101,6 +103,9 @@ where each column in the result corresponds to a column in `x`.
 
 When left unspecified, the lags used are the integers from 0 to
 `min(size(x,1)-1, 10*log10(size(x,1)))`.
+
+
+The output is normalized by `size(x, 1)`. See `autocor` for normalization including the variance.
 """
 function autocov(x::AbstractVector{T}, lags::IntegerVector; demean::Bool=true) where T<:Real
     autocov!(Vector{fptype(T)}(length(lags)), float(x), lags; demean=demean)
@@ -124,6 +129,9 @@ be subtracted from `x` before computing the ACF.
 If `x` is a vector, `r` must be a vector of the same length as `x`.
 If `x` is a matrix, `r` must be a matrix of size `(length(lags), size(x,2))`, and
 where each column in the result will correspond to a column in `x`.
+
+The output is normalized by `size(x, 1)` as well as the variance of `x`. See `autocov!` for normalization
+without the variance.
 """
 function autocor!(r::RealVector, x::AbstractVector{T}, lags::IntegerVector; demean::Bool=true) where T<:RealFP
     lx = length(x)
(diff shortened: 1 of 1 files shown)


src/signalcorr.jl
```diff
@@ -56,6 +56,8 @@ before computing the autocovariance.
 If `x` is a vector, `r` must be a vector of the same length as `x`.
 If `x` is a matrix, `r` must be a matrix of size `(length(lags), size(x,2))`, and
 where each column in the result will correspond to a column in `x`.
+
+The output is normalized by `size(x, 1)`. See `autocor!` for normalization including the variance.
```
andreasnoack: I don't think the first sentence should be there. It would correspond to saying that the sample mean is scaled by the sample size.  Maybe something like "See `autocor!` for a version normalized with the variance." Similar comments apply to the other docstrings.

Datseris: To be honest I did found it a bit meaningless to talk about normalizing with the size. I will remove it and simply say that this version does not normalize with the variance., etc. etc.

Datseris:
So, is this okay?

simonbyrne:
cross ref #294.

src/signalcorr.jl
```diff
@@ -343,6 +355,8 @@ If both `x` and `y` are vectors, `r` must be a vector of the same length as
 `(length(lags), size(x, 2))`; if `x` is a vector and `y` is a matrix, `r` must be a matrix
 of size `(length(lags), size(y, 2))`. If both `x` and `y` are matrices, `r` must be a
 three-dimensional array of size `(length(lags), size(x, 2), size(y, 2))`.
+
+The output is normalized by the `sqrt(var(x)*var(y))`. See `crosscov!` if you do not want this normalization.
```
nalimilan: Just "by `sqrt`...".

src/signalcorr.jl
```diff
@@ -56,6 +56,8 @@ before computing the autocovariance.
 If `x` is a vector, `r` must be a vector of the same length as `x`.
 If `x` is a matrix, `r` must be a matrix of size `(length(lags), size(x,2))`, and
 where each column in the result will correspond to a column in `x`.
+
+The output is not normalized by the variance. See `autocor!` for normalization including the variance.
```
nalimilan: Could just say "is not normalized." (either by the variance of by anything). Also, it would be good to normalize (pun intended) the second sentence so that the phrasing is the same across docstrings.

Datseris:
I think it is best if you simply merge this and make the other minor changes as you see fit. In the end of the day, you will decide what will pass not me, plus these changes take at maximum 2 minutes to actually make them.

The way we are doing it now is simply extremely inefficient.

src/signalcorr.jl
```diff
@@ -124,6 +128,8 @@ be subtracted from `x` before computing the ACF.
 If `x` is a vector, `r` must be a vector of the same length as `x`.
 If `x` is a matrix, `r` must be a matrix of size `(length(lags), size(x,2))`, and
 where each column in the result will correspond to a column in `x`.
+
+The output is normalized by the variance of `x`. See `autocov!` if you do not want this normalization.
```
nalimilan: The second sentence still isn't consistent with the one above. Sorry, I'm not on my computer right now so I can't make edits directly. (Anyway, what takes time and can sound inefficient is to discuss the best formulation. Who makes the final edit after everybody agrees is secondary.)

Issue #280, opened 2017-06-30
Improve consistency of model API


nalimilan:
Our current `StatisticalModel` API is heavily inspired from R, adapted to match Julia rules (plus some improvements), but it is not really consistent (since R isn't either). Below is a summary of the situation. Names are considered OK if they are either a standard abbreviation, or are spelled in full (and the name is not too long).

Names consistent with R and OK: `coef`, `nobs`, `deviance`, `confint`, `fitted`, `residuals`, `predict`, `aic`, `bic`, `aicc`, `r2`/`r²`, `adjr2`, `adjr²`, `stderr`

Names that do not exist in R and are OK: `aicc`, `r2`/`r²`, `adjr2`, `adjr²`, `stderr`, `nulldeviance`

Names different from R for good reasons: `dof`, `dof_residual` (to avoid two-letter objects, limiting confusion with `df = DataFrame(...)`)

Names not consistent with R for bad reasons, and proposed change:
- `loglikelihood` -> `loglik` (shorter, relatively standard abbreviation, and consistent with R)
- `nullloglikelihood` -> `null_loglik` (shorter and more readable; if changed, `nulldeviance` should be called `null_deviance`)

Names consistent with R which should be changed:
- `vcov` -> `cov`
- `model_response` -> `response` ("model" is redundant with multiple dispatch)

Name to add and which does not exist in R: `params`, `param` or `parameters`. The choice depends on whether we accept abbreviations, or prefer full names everywhere. Names which should be changed accordingly include potentially `coef`, `loglikelihood`/`nullloglikelihood`, `nobs`, `adjr2`/`adjr²`, `confint` and `dof`.

rofinn:
I like all of this except that I'm not a fan of the `lik` convention in R (maybe `like`?). Also, seems a little weird to have `null_loglik` and `null_deviance` when everything else seems to skip the `_`. Maybe we could make `null` a keyword which defaults to `false`?

rofinn:
This is probably a separate issue, but would it make sense to merge the `StatisticalModel` API into [StatsModels.jl](https://github.com/JuliaStats/StatsModels.jl/blob/master/REQUIRE)?

nalimilan:
> I like all of this except that I'm not a fan of the lik convention in R (maybe like?). Also, seems a little weird to have null_loglik and null_deviance when everything else seems to skip the _. Maybe we could make null a keyword which defaults to false?

[...]

PR #301, opened 2017-09-11
Added coefnames and model_matrix

Nosferican:
The functions `coefnames` and `model_matrix` are implemented to expand the
RegressionModel API.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -14,7 +14,7 @@ coef(obj::StatisticalModel) = error("coef is not defined for $(typeof(obj)).")
 
 Return the names of the coefficients.
 """
-coefnames(obj::RegressionModel) = error("coefnames is not defined for $(typeof(obj)).")
+coefnames(obj::StatisticalModel) = error("coefnames is not defined for $(typeof(obj)).")
 
 """
     coeftable(obj::StatisticalModel)
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -200,6 +200,13 @@ const adjr² = adjr2
 abstract type RegressionModel <: StatisticalModel end
 
 """
+    coefnames(obj::RegressionModel)
+
+Return the names of the coefficients as a `Vector{Symbol}`.
+"""
+coefnames(obj::RegressionModel) = error("coefnames is not defined for $(typeof(obj)).")
+
+"""
     fitted(obj::RegressionModel)
 
 Return the fitted values of the model.
@@ -213,6 +220,12 @@ Return the model response (a.k.a. the dependent variable).
 """
 model_response(obj::RegressionModel) = error("model_response is not defined for $(typeof(obj)).")
 
+"""
+    model_matrix(obj::RegressionModel)
+
+Return the model matrix (a.k.a. the design matrix).
+"""
+model_matrix(obj::RegressionModel) = error("model_matrix is not defined for $(typeof(obj)).")
 
 """
     residuals(obj::RegressionModel)
@@ -347,7 +360,7 @@ struct ConvergenceException{T<:Real} <: Exception
     lastchange::T
     tol::T
     function ConvergenceException{T}(iters, lastchange::T, tol::T) where T<:Real
-        if tol > lastchange 
+        if tol > lastchange
             throw(ArgumentError("Change must be greater than tol."))
         else
             new(iters, lastchange, tol)
@@ -355,7 +368,7 @@ struct ConvergenceException{T<:Real} <: Exception
     end
 end
 
-ConvergenceException(iters, lastchange::T=NaN, tol::T=NaN) where {T<:Real} = 
+ConvergenceException(iters, lastchange::T=NaN, tol::T=NaN) where {T<:Real} =
     ConvergenceException{T}(iters, lastchange, tol)
 
 function Base.showerror(io::IO, ce::ConvergenceException)
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -10,6 +10,13 @@ Return the coefficients of the model.
 coef(obj::StatisticalModel) = error("coef is not defined for $(typeof(obj)).")
 
 """
+    coefnames(obj::StatisticalModel)
+
+Return the names of the coefficients.
+"""
+coefnames(obj::RegressionModel) = error("coefnames is not defined for $(typeof(obj)).")
+
+"""
     coeftable(obj::StatisticalModel)
 
 Return a table of class `CoefTable` with coefficients and related statistics.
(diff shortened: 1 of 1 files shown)


src/statmodels.jl
```diff
@@ -200,6 +200,13 @@ const adjr² = adjr2
 abstract type RegressionModel <: StatisticalModel end
 
 """
+    coefnames(obj::RegressionModel)
+
+Return the names of the coefficients as a `Vector{Symbol}`.
```
nalimilan: I wouldn't specify that it's a `Vector{Symbol}`. As noted on the issue, it could also be a `NamedTuple` for types with fixed parameters.

src/statmodels.jl
```diff
@@ -213,6 +220,12 @@ Return the model response (a.k.a. the dependent variable).
 """
 model_response(obj::RegressionModel) = error("model_response is not defined for $(typeof(obj)).")
 
+"""
+    model_matrix(obj::RegressionModel)
+
+Return the model matrix (a.k.a. the design matrix).
+"""
+model_matrix(obj::RegressionModel) = error("model_matrix is not defined for $(typeof(obj)).")
```
nalimilan: @andreasnoack suggested getting rid of the underscore (`modelmatrix`). I guess we can do that, and deprecate `model_response` later.

src/statmodels.jl
```diff
@@ -200,6 +200,13 @@ const adjr² = adjr2
 abstract type RegressionModel <: StatisticalModel end
 
 """
+    coefnames(obj::RegressionModel)
```
nalimilan: Should allow any `StatisticalModel`, just like `coef`.

src/statmodels.jl
Nosferican: Went ahead and changed it to `modelmatrix`. Is `model_response` being depreciated in favor of `modelresponse` for the next release? I think the decision was taken some time ago, but don't know the timeline.

src/statmodels.jl
Nosferican: Made it inherit from `StatisticalModel`. Makes sense since `coef` is at the `StatisticalModel` level.

src/statmodels.jl
Nosferican: I was thinking having the same expected behavior since it has the same function name as DataFrames / StatsModels, but I guess for some packages the flexibility of specifying the return type overweights this.

src/statmodels.jl
nalimilan: We're still discussing the best naming policy for these methods (#280), so I'm not sure it's worth deprecating the function right now, but indeed it's likely to happen at some point.

nalimilan: Actually, I suggested renaming it to just `response` in that issue. OTOH, calling `modelmatrix` just `matrix` would be weird. `designmatrix` could be an interesting alternative.

Nosferican: I come from an R background with `model.matrix` and `model.response` so I might be a bit biased on the issue. I know the Julian naming conventions recommend not using `_` if it can be understood without (e.g., `completecases`). I like the model prefix for consistency, but I think `response` and `designmatrix` are a good alternative. However, I am not complete sure the design matrix concept is shared among all fields that might use it... I know econometrics uses it as the standard and in machine learning it is used even if it is not at the same level (e.g., model matrix is also commonly used in ML as well, but design matrix is understood). If design matrix doesn't exclude any practitioners to the best of your knowledge I would be in favor of `designmatrix`.

ararslan: I like both `modelmatrix` and `designmatrix`. In my previous field (clinical biostats) I think "model matrix" was the dominant term, though "design matrix" was equally well understood; no one would be confused by that term. As another data point, one of my classes in grad school used the term "design matrix" exclusively.

dmbates: Please don't confound a model matrix with a design matrix.  They are different and there is a very good reason that the term "model matrix" is used in R.

If you know the sequence of (formula + data) -> ModelFrame -> ModelMatrix the "design" is reflected in the ModelFrame.  If you have a categorical factor, it is a single categorical variable at the level of the design.  At the level of the model matrix categorical factors are expanded to multiple columns of contrasts, interactions are expanded, polynomial terms are expanded, ...

This is why it is useful to retain not only the numerical values in the model matrix but also information on how you got them.  The ModelFrame with its terms information was based on the R implementation as a way to do this.  It may not be the best way to do this in a language as flexible as Julia but the thing to avoid is ending up with nothing but a matrix of numbers and no idea where these numbers came from.

If you consider designing experiments in the sense of choosing the levels for each of the primary covariates at each run you can see that a design matrix doesn't involve interactions (if you know what the levels of factors A and B are you know the level of the interaction A&B) or contrasts or polynomial terms.

nalimilan: Thanks @dmbates. Let's go with `modelmatrix` then. It's probably easier to find/understand anyway. (Though [Wikipedia](https://en.wikipedia.org/wiki/Design_matrix) doesn't appear to make a distinction between "model matrix" and "design matrix".)

Nosferican: Sounds good. `modelmatrix` is suppose to return the actual matrix used when fitting the model which includes the matrix that has already been through complete cases, contrasts, expansions, interactions, weighted, transformed (e.g., error component models), and made full rank when applicable. Should the updated documentation after merging the pull request explicitly indicate the above?

dmbates: @nalimilan I agree that model matrix and design matrix are often used interchangeably including in many textbooks where "X" is described as the design matrix.  Some of us who are old enough to have had courses on experimental design when it was taught separately from regression prefer to make the distinction.  For example, a constant column (a column of 1's) makes perfect sense as part of the model but not as part of the design.

nalimilan: > Should the updated documentation after merging the pull request explicitly indicate the above?

That surely wouldn't hurt, but better do that directly in this PR.

src/statmodels.jl
```diff
@@ -10,6 +10,13 @@ Return the coefficients of the model.
 coef(obj::StatisticalModel) = error("coef is not defined for $(typeof(obj)).")
 
 """
+    coefnames(obj::StatisticalModel)
+
+Return the names of the coefficients.
+"""
+coefnames(obj::RegressionModel) = error("coefnames is not defined for $(typeof(obj)).")
```
andreasnoack: Should be `StatisticalModel`

Nosferican:
Feel free to edit it. We lost power in PR since yesterday (hurricane María) and not sure when will it be restored.

andreasnoack:
I've made the change. Take care.

PR #306, opened 2017-09-21
Clean up docstring

### LINKED ISSUE #304: Doc issue(?): maximum absolute deviation/median absolute deviation
(Continued from Slack)

I think this is just a docstring error, since the non-mutating function uses the mutating internally (`mad()` vs `mad!()`)

https://github.com/JuliaStats/StatsBase.jl/blob/9d851b704cc135ee4ca9bf68f9b0541d64d5814f/src/scalarstats.jl#L239-L269

If it's median, I can quickly submit a PR (or someone else who can confirm can just change it real quick)

randyzwitch:
Fix miswritten "maximum" to read "median" for consistency. Fixes #304

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -254,7 +254,7 @@ end
 """
     StatsBase.mad!(v, center=median!(v); constant=k)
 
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks!

Issue #102, opened 2014-11-25
pacf() gives strange values


skeppet:
When I use pacf() with method=:regression I get that it's almost zero for time lag 0. This should always be one, as far as I know.This does on the other hand not happen with method=:yulewalker. Am I the only one experiencing that?

If I compare the output of pacf(method=:yulewalker) with aryule() in MATLAB i get the same values but negative. Is there a good reason for that, or is it a bug?

Nosferican:
Status of this?

andreasnoack:
Looks like it was fixed in #309

PR #309, opened 2017-10-15
Ensure that pacf at lag 0 is 1.0

BenjaminBorn:
This change makes sure that the partial acf at lag zero is always one, which hasn't been the case so far (see issue #102).

There are two open issues:

1. We could just throw an error if lag zero is requested but this would be inconsistent with the `autocor` function where lag zero is a valid input. Therefore I would like to allow it for `pacf` as well.

2. My fix ensures that the pacf at lag zero is one but it is inefficient in the sense that the pacf at that lag is still looped through. This could potentially be relevant if someone wants to compute pacfs for a large number of dataseries. So if you like, I could also make a larger change so that the pacf at lag zero is taken out of the loop.

### DIFF
--- a/src/signalcorr.jl
+++ b/src/signalcorr.jl
@@ -513,7 +513,7 @@ function pacf_regress!(r::RealMatrix, X::AbstractMatrix{T}, lags::IntegerVector,
         for i = 1 : length(lags)
             l = lags[i]
             sX = view(tmpX, 1+l:lx, 1:l+1)
-            r[i,j] = (cholfact!(sX'sX)\(sX'view(X, 1+l:lx, j)))[end]
+            r[i,j] = l == 0 ? 1 : (cholfact!(sX'sX)\(sX'view(X, 1+l:lx, j)))[end]
         end
     end
     r
@@ -580,4 +580,3 @@ end
 function pacf(x::AbstractVector{T}, lags::IntegerVector; method::Symbol=:regression) where T<:Real
     vec(pacf(reshape(x, length(x), 1), lags, method=method))
 end
-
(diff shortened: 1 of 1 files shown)


src/signalcorr.jl
```diff
@@ -513,7 +513,7 @@ function pacf_regress!(r::RealMatrix, X::AbstractMatrix{T}, lags::IntegerVector,
         for i = 1 : length(lags)
             l = lags[i]
             sX = view(tmpX, 1+l:lx, 1:l+1)
-            r[i,j] = (cholfact!(sX'sX)\(sX'view(X, 1+l:lx, j)))[end]
+            r[i,j] = l == 0 ? one(T) : (cholfact!(sX'sX)\(sX'view(X, 1+l:lx, j)))[end]
```
ararslan: Will the result of `(cholfact!(sX'sX)\(sX'view(X, 1+l:lx, j)))[end]` definitely have type `T`?

BenjaminBorn: The way I understand it, `T <:RealFP` (where `RealFP = Union{Float32, Float64}` - defined in common.jl, line 21) means that the matrix `X` and therefore also `sX` is a real floating point matrix. That would also imply that the correlation coefficient is a floating point number (Float32 if `X` is Float32 and Float64 if `X` is Float64).

This is also how it is handled in `pacf_yulewalker!()` (same file, line 528). Do you have a better way? I'm happy to change it.

nalimilan: AFAICT `one(T)` isn't better than `1` anyway since implicit conversion will happen.

BenjaminBorn: Ok. I replaced `one(T)` by `1`. Should we then also change it in `pacf_yulewalker!()` to be consistent?

nalimilan: No idea, I don't know that code.

BenjaminBorn: Ok, thanks. Maybe @andreasnoack can help here.

andreasnoack: Yes, I think it would be good to change it there for consistency.

BenjaminBorn: Done! Thanks for the quick feedback.

PR #312, opened 2017-10-31
Added  `getfrequent` function to get  most frequent elements from a vector

jayvn:
I'm pretty new to this.
Felt giving a PR was the best way to say it.
Please tell me the right place to put this (if it's needed anywhere at all).
`
julia> v1 = rand(1:5, 10)
10-element Array{Int64,1}:
 5
 5
 4
 1
 1
 3
 5
 4
 4
 1

julia> StatsBase.getfrequent(v1,10)
5-element Array{Int64,1}:
 1
 4
 5
 3
 2
`


nalimilan:
Thanks. Unfortunately, I'm not sure whether we should add such a function or not. `counts` is currently quite restrictive as it only accepts integer vectors, and there's no reason why `getfrequent` should be so limited too. Overall I think we need to redesign our counting API first before extending it. For example, `countmap` could gain an argument to return only the `n` most frequent elements, but that doesn't work for `counts`. I think that function should be renamed to something more specific (e.g. `counts(Vector{Int}, ...)`) as it's quite unexpected that it only works for integers.

jayvn:
Thank you for the response. I think the redesign makes sense. Although I don't get why we'd rename it ;). So I'll decline this for now ?

nalimilan:
If you can suggest a consistent API for counting which includes this feature, you're welcome. Else I'm afraid this will have to wait until somebody works on it.

nalimilan:
Here's a possible plan: 
- deprecate `counts(x)` in favor of `counts(x, levels)`
- make `counts(x)` be equivalent to `countmap(x)`  (i.e. returning a `Dict`, and supporting any type), and deprecate the latter
- as a possible extension, add support for non-integer arrays in `counts(x, levels)`, still returning a `Vector` of counts in the order of `levels`
- add a `n` keyword argument to optionally return only the most frequent elements

Issue #313, opened 2017-11-01
Quantiles don't respect weights of 0


joshday:
I would expect these two to be the same, but maybe I'm missing something?

```julia
julia> quantile([1,2,3,4,5], fweights([0,1,2,1,0]), [0, .25, .5, .75, 1])
5-element Array{Float64,1}:
 1.0
 2.42857
 3.0
 3.57143
 5.0

julia> quantile([2,3,3,4])
5-element Array{Float64,1}:
 2.0
 2.75
 3.0
 3.25
 4.0
```

nalimilan:
@matthieugomez Remarks?

joshday:
The [Implementation in StatsBase](https://github.com/JuliaStats/StatsBase.jl/blob/master/src/weights.jl#L540) points to http://stats.stackexchange.com/questions/13169/defining-quantiles-over-a-weighted-sample, which doesn't provide a reference or theoretical guarantees.  

My understanding is that any subtype of `AbstractWeights{<:Integer, <:Integer}` is essentially run length encoding, and it may be that StatsBase only need a new `quantile` method to handle this.  

For context, my use case is creating approximate quantiles from a histogram (where bins may contain zero observations).

If anyone has a good reference for weighted quantiles (a few google searches returns there's no well-established method), I would be happy to take a look.

nalimilan:
These issues are really tricky. As [Wikipedia](https://en.wikipedia.org/wiki/Quantile) notes, there are 9 different ways of computing sample quantiles. AFAIK we use the same definition as R (R-7), and indeed we get the same result for unweighted quantiles. For weighted quantiles, the implementation by @matthieugomez from https://github.com/JuliaStats/StatsBase.jl/pull/117 gives the same result as the unweighted quantiles when all weights are 1.

But then it doesn't appear to follow the (a priori natural) property that it's equivalent to replicating each entry the number of times corresponding to its integer weight:
```julia
julia> quantile([2,3,3,4], [0, .25, .5, .75, 1])
5-element Array{Float64,1}:
 2.0 
 2.75
 3.0 
 3.25
 4.0 

julia> quantile([2,3,3,4], fweights([1,1,1,1]), [0, .25, .5, .75, 1])
5-element Array{Float64,1}:
 2.0 
 2.75
 3.0 
 3.25
 4.0 

julia> quantile([2,3,4], fweights([1,2,1]), [0, .25, .5, .75, 1])
5-element Array{Float64,1}:
 2.0
[...]

PR #314, opened 2017-11-01
WIP: expand data by FrequencyWeights

joshday:
This PR adds a `repeat(x::RealVector, w::FrequencyWeight{<:Integer})` method to create a new vector where each element in `x` gets repeated by its associated weight in `w`.  Would this be useful, if only for better testing?  It would help catch cases like https://github.com/JuliaStats/StatsBase.jl/issues/313

```julia
x = randn(10)
w = rand(1:5, 10)
p = [0, .25, .5, .75, 1]
@test quantile(x, w, p) == quantile(repeat(x, w), p)
```

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -137,6 +137,18 @@ See the documentation for [`FrequencyWeights`](@ref) for more details.
 fweights(vs::RealVector) = FrequencyWeights(vs)
 fweights(vs::RealArray) = FrequencyWeights(vec(vs))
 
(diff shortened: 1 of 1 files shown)


joshday:
Oops, this already exists.

```julia
StatsBase.inverse_rle([1,2,3], fweights([1,2,3]))
```

nalimilan:
It would make sense to rename that function as it's pretty hard to discover. I even had filed an issue against Julia to add this kind of function without noticing it existed here: https://github.com/JuliaLang/julia/issues/16443.

PR #316, opened 2017-11-09
modify weighted quantile (aweights + fweights)

matthieugomez:
Solves #313
The quantile method for frequency weights is now equivalent to the unweighted method with a vector of repeated values
The quantile method for non frequency weight now removes zeros.


src/weights.jl
```diff
@@ -549,15 +545,15 @@ Compute the weighted quantiles of a vector `x` at a specified set of probability
...
 
-The quantile for `p` is defined as follows. Denoting
-``S_k = (k-1)w_k + (n-1) \\sum_{i<k}w_i``, define ``x_{k+1}`` the smallest element of `x`
-such that ``S_{k+1}/S_{n}`` is strictly superior to `p`. The function returns
-``(1-\\gamma) x_k + \\gamma x_{k+1}`` with  ``\\gamma = (pS_n- S_k)/(S_{k+1}-S_k)``.
-
-This corresponds to  R-7, Excel, SciPy-(1,1) and Maple-6 when `w` contains only ones
-(see [Wikipedia](https://en.wikipedia.org/wiki/Quantile)).
+With frequency weights, the function returns the same result as `quantile` for a vector with repeated values.
+With non frequency weights,  denote N the length of the vector, w the vector of weights normalized to sum to 1, `h = p (N - 1) + 1`  and ``S_k = 1 + (k-1) * wk + (N-1) \\sum_{i<=k}w_i/\\sum_{i<=N}w_i``, define ``x_{k+1}`` the smallest element of `x` such that ``S_{k+1}`` is strictly superior to `h`. The function returns
```
nalimilan: Could you say what `S` and `h` represent?

Remove the double spaces, use double backticks everywhere (including around variable names) and break lines at 92 chars. Also better have ``[frequency weights](@ref FrequencyWeights)``.

src/weights.jl
```diff
@@ -549,15 +545,15 @@ Compute the weighted quantiles of a vector `x` at a specified set of probability
...
+``x_k + \\gamma (x_{k+1} -x_k)`` with  ``\\gamma = (h - S_k)/(S_{k+1}-S_k)``
+In particular, when `w` is a vector of one, the function returns the same result as `quantile`.
 """
-function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where {V,W<:Real}
+
```
nalimilan: No line breaks before nor after the signature. I think the style of this package is not to have spaces after commas for type parameters.

src/weights.jl
```diff
@@ -569,47 +565,62 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
         x < 0 && error("weight vector cannot contain negative entries")
     end
 
-    # full sort
-    vw = sort!(collect(zip(v, w.values)))
+    wvalues = w.values
+    nz = find(w.values)
+    #normalize if non frequencyweight
```
nalimilan: Space after # (also below) and between "frequency" and "weights". Better also say that the sum will be 1.

src/weights.jl
```diff
@@ -569,47 +565,62 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
         x < 0 && error("weight vector cannot contain negative entries")
     end
 
-    # full sort
-    vw = sort!(collect(zip(v, w.values)))
+    wvalues = w.values
+    nz = find(w.values)
```
nalimilan: Use `find(!iszero, w.values)` since the previous form is deprecated on 0.7. Or probably even better, just do `nz = .!iszero.(w.values)`, since a boolean vector will be more efficient (you know the size in advance and it's smaller).

Also move this below to the place where it's used for clarity.

Finally, why is that operation needed for frequency weights? Shouldn't the algorithm be able to skip entries with zero weights on its own?

src/weights.jl
```diff
@@ -569,47 +565,62 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
...
     # prepare out vector
-    N = length(vw)
     out = Vector{typeof(zero(V)/1)}(length(p))
     fill!(out, vw[end][1])
 
     # start looping on quantiles
-    cumulative_weight, Sk, Skold =  zero(W), zero(W), zero(W)
-    vk, vkold = zero(V), zero(V)
-    k = 1
+    Sk, Skold =  zero(W), zero(W)
```
nalimilan: Double space.

src/weights.jl
```diff
@@ -569,47 +565,62 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
...
+    vk, vkold, cumwk, wk = zero(V), zero(V), zero(V), zero(V)
+    k = 0
+
```
nalimilan: A single line break is enough.

src/weights.jl
```diff
@@ -569,47 +565,62 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
...
+
     for i in 1:length(p)
-        h = p[i] * (N - 1) * wsum
-        if h == 0
-            # happens when N or p or wsum equal zero
-            out[ppermute[i]] = vw[1][1]
-        else
+        if isa(w, FrequencyWeights)
```
nalimilan: It feels weird to use a completely different path for frequency weights. Couldn't a common path be defined, moving some type-specific computations out of the loop like you did for normalization? For example, for `h = p[i] * (wsum - 1) + 1`, `wsum` just needs to be replaced with `N` for non frequency weights, so you could as well define the variable before the loop.

BTW, wouldn't it make more sense to normalize the weights to sum to `N` (rather than 1), since it plays a role equivalent to `wsum`?

test/weights.jl
```diff
@@ -261,7 +261,48 @@ end
...
+        [-10, 1, 1, -10, -10],
+    )
+    p = [0.0, 0.25, 0.5, 0.75, 1.0]
+    for x in data
+        @test quantile(x, fweights(ones(Int64, length(x))), p) ≈ quantile(x, p)
+    end
+    # zero don't count
+    x = [1, 2, 3, 4, 5]
+    @test quantile(x, fweights([0,1,1,1,0]), p) ≈ quantile([2, 3, 4], p)
+    # repetitions dont count
```
nalimilan: "don't"

test/weights.jl
```diff
@@ -284,81 +325,84 @@ end
...
-    @test wquantile(data[1], w, 0.5)            ≈  answer
+    answer = 6.1818
+    @test quantile(data[1], aweights(w), 0.5)    ≈  answer atol = 1e-4
+    @test wquantile(data[1], aweights(w), [0.5]) ≈ [answer] atol = 1e-4
+    @test wquantile(data[1], aweights(w), 0.5)   ≈  answer atol = 1e-4
+    @test wquantile(data[1], w, [0.5])          ≈ [answer] atol = 1e-4
+    @test wquantile(data[1], w, 0.5)            ≈  answer atol = 1e-4
 end
 
+
```
nalimilan: Not needed.

test/weights.jl
```diff
@@ -284,81 +325,84 @@ end
...
+    [1.0, 1.625, 2.33333, 3.25, 5.0],
+    [-2.0, -0.538462, 1.53846, 2.7, 6.0],
+    [-10.0, -10.0, -10.0, 1.0, 1.0],
     )
     p = [0.0, 0.25, 0.5, 0.75, 1.0]
 
     srand(10)
     for i = 1:length(data)
-        @test quantile(data[i], wt[i], p) ≈ quantile_answers[i]
+        @test quantile(data[i], aweights(wt[i]), p) ≈ quantile_answers[i] atol = 1e-3
```
nalimilan: 1e-3 sounds quite high, why not keep more precision?

test/weights.jl
```diff
@@ -261,7 +261,48 @@ end
...
+        [2.3],
+        [-2, -3, 1, 2, -10],
+        [1, 2, 3, 4, 5],
+        [5, 4, 3, 2, 1],
+        [-2, 2, -1, 3, 6],
+        [-10, 1, 1, -10, -10],
+    )
+    p = [0.0, 0.25, 0.5, 0.75, 1.0]
+    for x in data
+        @test quantile(x, fweights(ones(Int64, length(x))), p) ≈ quantile(x, p)
```
nalimilan: It would be nice to test more thoroughly the results by combining various inputs with various weights, as done below with non frequency weights. You could use a `rep` helper function (https://github.com/JuliaLang/julia/issues/16443) to generate vectors with repeated entries to call the unweighted `quantile` method on.

Also, there aren't any zeros no negative non-integer values. Wouldn't hurt to add some.

src/weights.jl
matthieugomez: No, the algorithm does not work with zero weights in its current form. There is an issue for instance if the highest value has zero weight. The algorithm also only keeps track of the last visited value, instead of the last visiting value with non zero weight.

src/weights.jl
matthieugomez: I have thought about this but I think it is better to do two different paths. Joining the two looks more confusing than enlightening.

matthieugomez:
Thanks for the detailed comments. I commented on the ones I disagreed with.
About the normalization of weights to 1 or N, the actual normalizing factor is irrelevant — the important thing is that the algorithm does not depend on the sum of weights. It turns out that the formula with weights normalized to 1 are simpler to digest, so that's what I end up doing.

nalimilan:
But why is it more problematic to give different results depending on the sum of weights than to normalize them arbitrarily to 1? Is there a rationale or a precedent in other software? That combined with the fact that we don't even use the same algorithm for frequency and other kinds of weights makes it sound like a totally arbitrary choice. For example, Hmisc::wtd.quantile normalizes to `N` when `normwt=TRUE`, and does not provide an argument to normalize to `1`. Is that better? Is that worse? I wouldn't want to make a choice without solid arguments.

matthieugomez:
Take an algorithm that does not depend on the sum of weight. You can always
rewrite it as an algorithm that takes a vector of weight that sum to one.
Rewriting it like that allows to replace all the expressions in sum wi by
1.  That’s all there is. This does not mean that the normalization is
arbitrary.

matthieugomez:
Ok I have updated the definition of aweights. There are now only two differences between fweights and aweights in the algorithm.  The intuition for why they need to be different is that fweight gives you more datapoint on the empirical CDF while aweight does not.

For instant, I think it is intuitive that 
```
quantile([1, 2], weights([1, 2])) = 1.5
``` while
```
quantile([1, 2], fweights([1, 2])) = 2
```

File: test/weights.jl [hunk 12]
```diff
@@ -284,81 +286,155 @@ end
...
-        f([5, 5, 4, 1]),
-        f([30, 56, 144, 24, 55, 43, 67]),
-        f([0.1, 0.2, 0.3, 0.4, 0.5, 0.6]),
-        f([12]),
-        f([7, 1, 1, 1, 6]),
-        f([1, 0, 0, 0, 2]),
-        f([1, 2, 3, 4, 5]),
-        f([0.1, 0.2, 0.3, 0.2, 0.1]),
-        f([1, 1, 1, 1, 1]),
+        Int[3, 1, 1, 1, 3],
```
nalimilan: Why specify `Int`?

src/weights.jl
```diff
@@ -537,25 +537,22 @@ wmedian(v::RealVector, w::AbstractWeights{<:Real}) = median(v, w)
...
-# In the non weighted version, we compute a vector of index h(N, p)
-# and take interpolation between floor and ceil of this index
-# Here there is a supplementary function from index to weighted index k -> Sk
 
 """
     quantile(v, w::AbstractWeights, p)
 
-Compute the weighted quantiles of a vector `x` at a specified set of probability
-values `p`, using weights given by a weight vector `w` (of type `AbstractWeights`).
+Compute the weighted quantiles of a vector ``x`` at a specified set of probability
```
nalimilan: Actually, references to Julia objects should use single backticks... :-) And `v` should be changed to `x` in the signature above to match descriptions below (or the other way around).

src/weights.jl
```diff
@@ -537,25 +537,22 @@ wmedian(v::RealVector, w::AbstractWeights{<:Real}) = median(v, w)
...
 Weights must not be negative. The weights and data vectors must have the same length.
 
-The quantile for `p` is defined as follows. Denoting
-``S_k = (k-1)w_k + (n-1) \\sum_{i<k}w_i``, define ``x_{k+1}`` the smallest element of `x`
-such that ``S_{k+1}/S_{n}`` is strictly superior to `p`. The function returns
-``(1-\\gamma) x_k + \\gamma x_{k+1}`` with  ``\\gamma = (pS_n- S_k)/(S_{k+1}-S_k)``.
-
-This corresponds to  R-7, Excel, SciPy-(1,1) and Maple-6 when `w` contains only ones
-(see [Wikipedia](https://en.wikipedia.org/wiki/Quantile)).
+With [FrequencyWeights](@ref FrequencyWeights), the function returns the same result as
```
nalimilan: Missing backticks here and below.

src/weights.jl
```diff
@@ -537,25 +537,22 @@ wmedian(v::RealVector, w::AbstractWeights{<:Real}) = median(v, w)
...
-The quantile for `p` is defined as follows. Denoting
-``S_k = (k-1)w_k + (n-1) \\sum_{i<k}w_i``, define ``x_{k+1}`` the smallest element of `x`
-such that ``S_{k+1}/S_{n}`` is strictly superior to `p`. The function returns
-``(1-\\gamma) x_k + \\gamma x_{k+1}`` with  ``\\gamma = (pS_n- S_k)/(S_{k+1}-S_k)``.
-
-This corresponds to  R-7, Excel, SciPy-(1,1) and Maple-6 when `w` contains only ones
-(see [Wikipedia](https://en.wikipedia.org/wiki/Quantile)).
+With [FrequencyWeights](@ref FrequencyWeights), the function returns the same result as 
+`quantile` for a vector with repeated values.
+With non FrequencyWeights,  denote N the length of the vector, w the vector of weights, ``h = p (\\sum_{i<= N}w_i - w_1) + w_1`` the cumulative weight corresponding to the probability `p` and
```
nalimilan: Cut line at 92 chars. Still missing backticks.

src/weights.jl
```diff
@@ -537,25 +537,22 @@ wmedian(v::RealVector, w::AbstractWeights{<:Real}) = median(v, w)
...
-This corresponds to  R-7, Excel, SciPy-(1,1) and Maple-6 when `w` contains only ones
-(see [Wikipedia](https://en.wikipedia.org/wiki/Quantile)).
+With [FrequencyWeights](@ref FrequencyWeights), the function returns the same result as 
+`quantile` for a vector with repeated values.
+With non FrequencyWeights,  denote N the length of the vector, w the vector of weights, ``h = p (\\sum_{i<= N}w_i - w_1) + w_1`` the cumulative weight corresponding to the probability `p` and 
+ ``S_k = \\sum_{i<=k}w_i`` the cumulative weight for to each observation,
+  define ``x_{k+1}`` the smallest element of ``x`` such that ``S_{k+1}`` is strictly 
+  superior to ``h``. The function returns``x_k + \\gamma (x_{k+1} -x_k)`` 
+  with  ``\\gamma = (h - S_k)/(S_{k+1}-S_k)``. In particular, when ``w`` is a vector
+   of one, the function returns the same result as `quantile`.
```
nalimilan: "of ones". Also remove indentation on these lines.

src/weights.jl
```diff
@@ -537,25 +537,22 @@ wmedian(v::RealVector, w::AbstractWeights{<:Real}) = median(v, w)
...
-``S_k = (k-1)w_k + (n-1) \\sum_{i<k}w_i``, define ``x_{k+1}`` the smallest element of `x`
-such that ``S_{k+1}/S_{n}`` is strictly superior to `p`. The function returns
-``(1-\\gamma) x_k + \\gamma x_{k+1}`` with  ``\\gamma = (pS_n- S_k)/(S_{k+1}-S_k)``.
-
-This corresponds to  R-7, Excel, SciPy-(1,1) and Maple-6 when `w` contains only ones
-(see [Wikipedia](https://en.wikipedia.org/wiki/Quantile)).
+With [FrequencyWeights](@ref FrequencyWeights), the function returns the same result as 
+`quantile` for a vector with repeated values.
+With non FrequencyWeights,  denote N the length of the vector, w the vector of weights, ``h = p (\\sum_{i<= N}w_i - w_1) + w_1`` the cumulative weight corresponding to the probability `p` and 
+ ``S_k = \\sum_{i<=k}w_i`` the cumulative weight for to each observation,
```
nalimilan: "for to"

src/weights.jl
```diff
@@ -569,47 +566,47 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
         x < 0 && error("weight vector cannot contain negative entries")
     end
 
-    # full sort
-    vw = sort!(collect(zip(v, w.values)))
 
-    wsum = w.sum
+    #remove zeros weights and sort
```
nalimilan: Space after `#`.

src/weights.jl
```diff
@@ -569,47 +566,47 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
...
+    wsum = sum(w.value)
```
nalimilan: Use `sum(w)`, which is equivalent to ` w.sum` thanks to a special method. I don't think you need to access the `values` field below either: `AbstractWeight` implements the `AbstractVector` interface (but does not guarantee that the `values` field exists).

src/weights.jl
```diff
@@ -569,47 +566,47 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
...
-    N = length(vw)
     out = Vector{typeof(zero(V)/1)}(length(p))
     fill!(out, vw[end][1])
 
     # start looping on quantiles
-    cumulative_weight, Sk, Skold =  zero(W), zero(W), zero(W)
-    vk, vkold = zero(V), zero(V)
-    k = 1
+    Sk, Skold = zero(W), zero(W)
+    vk, vkold= zero(V), zero(V)
```
nalimilan: Space before `=`.

test/weights.jl
nalimilan: This comment hasn't been addressed.

BTW, it would be nice to duplicate each `@test` line to test `pweights` too.

nalimilan:
OK, I trust you when you say it's more intuitive, that's not completely obvious to me. ;-)

Is there any way to check our results against another implementation? We don't return the same results as `Hmisc::wtd.quantile` for non-frequency weights, which is worrying since it claims to implement the same R-7 version. Of course there may well be problems in that implementation rather than in ours, 

I've also found a MIT-licensed [implementation](https://github.com/IoSR-Surrey/MatlabToolbox/blob/4bff1bb2da7c95de0ce2713e7c710a0afa70c705/%2Biosr/%2Bstatistics/quantile.m#L23) of weighted quantiles for MATLAB (they use the R-8 definition by default, but support all variants). Unfortunately, even if they say that equal weights should give the same results as the unweighted version, that doesn't appear to be the case:
```MATLAB
>> iosr.statistics.quantile([7, 1, 2, 4, 10], [0, .25, .5, .75, 1], [], ['R-7'])

ans =

     1     2     4     7    10

>> iosr.statistics.quantile([7, 1, 2, 4, 10], [0, .25, .5, .75, 1], [], ['R-7'], [1, 1, 1, 1, 1])

ans =

    1.0000    1.5000    3.0000    5.5000   10.0000
```

It's incredible how hard it is to find completely reliable implementations, and the existence of so many variants makes it difficult to compare them.

nalimilan:
BTW, should `ProbabilityWeights` behave like `FrequencyWeights`, like `AnalyticWeights`, or implement a third behavior?

nalimilan:
Bump. Do you want to finish this?

matthieugomez:
Ok. I've updated it to incorporate your comments. I think `AnalyticalWeights` should be similar to `ProbabilityWeights`.

matthieugomez:
Thanks a lot! Really appreciate your help

tbeason:
What is the status on this PR / branch? It seems like the functionality is there. Are the tests not passing?

nalimilan:
It's just that I had left the PR open in case somebody wanted to comment, but nobody objected nor merged it.

PR #321, opened 2017-11-26
Export Methods and Updates Documentation

Nosferican:
Exports `coefnames` and `modelmatrix`. Updated the documentation as well.


docs/src/statmodels.md
```diff
@@ -10,8 +10,9 @@ aic
 aicc
 bic
 coef
-coeftable
 confint
+coefnames
```
nalimilan: `coefnames` and `coeftable` should come before `confint` AFAIK.

src/statmodels.jl
```diff
@@ -224,6 +224,11 @@ model_response(obj::RegressionModel) = error("model_response is not defined for
     modelmatrix(obj::RegressionModel)
 
 Return the model matrix (a.k.a. the design matrix).
+
+The model matrix may be used for diagnostic tests and variance covariance estimation.
```
nalimilan: I don't think this is the place to say what the model matrix may be used for. There are just too many different kinds of models. Also remember this docstring is going to be read by users, it's not mainly intended at package developers. As an end user, reading that models "should" do something isn't really useful.

Nosferican: I thought it was similar to the docstring of `r2` or `adjr2` which specify what the accessor is returning when it can be ambiguous. I would prefer to give the bit extra information, but if you think is not worthwhile I could remove the extra bit for information.

docs/src/statmodels.md
Nosferican: Oops. You are right.

src/statmodels.jl
Nosferican: Let me know what you think and I will address the comments in the same commit.

nalimilan: R² is kind of a special case since the classic linear model is really the reference there. And even for that case I'm not really happy about mentioning it in StatsBase rather than in GLM. Better give details in package-specific docstrings, which among other things has the advantage of only being visible to people who use the corresponding kind of model.

Nosferican: The advantage of including it in the base package is that it will be universal for all users (either package developers or end-users). Package specific usage would be ambiguous (i.e., might lead to fallback specific accessors). We can try it to leave it at the discretion of the packages and re-assess later on if major issues arise.

docs/src/statmodels.md
nalimilan: Still not fixed?

Nosferican:
Fixed now. Should be ready to merge.

PR #322, opened 2017-11-26
faster sampling without replacement using  shrinking array algorithm

xiaodaigh:
See benchmark

```julia
using StatsBase
x = rand(1:1_000_000, 1_000_000)

@time StatsBase.self_avoid_sample!(x, similar(x)) # 5 seconds
@time StatsBase.shrinking_array_sample!(x, similar(x)) # 0.6 second
```

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -207,6 +207,30 @@ self_avoid_sample!(a::AbstractArray, x::AbstractArray) =
     self_avoid_sample!(Base.GLOBAL_RNG, a, x)
 
 """
+    shrinking_array_sample!([rng], a::AbstractArray, x::AbstractArray)
+
+Shrinking-array sampling: samples from an array by replacing each selected
+element with the last element in the array and decrementing the array length
+counter by 1 each time - hence shrinking-array.
+"""
+function shrinking_array_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
+    n = length(a)
+    k = length(x)
+    k <= n || error("length(x) should not exceed length(a)")
+
+    ca = copy(a)
+    for (i, m) in zip(1:k, n:-1:n-k+1)
+        rgen = RangeGenerator(1:m)
+        idx = rand(rng, rgen)
+        x[i], ca[idx] = ca[idx], ca[m]
+    end
+
+    return x
+end
+shrinking_array_sample!(a::AbstractArray, x::AbstractArray) =
+    shrinking_array_sample!(Base.GLOBAL_RNG, a, x)
+
+"""
     seqsample_a!([rng], a::AbstractArray, x::AbstractArray)
 
 Random subsequence sampling using algorithm A described in the following paper (page 714):
@@ -346,7 +370,7 @@ function sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray;
             elseif n < k * 24
                 fisher_yates_sample!(rng, a, x)
             else
-                self_avoid_sample!(rng, a, x)
+                shrinking_array_sample!(rng, a, x)
             end
         end
     end
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -221,7 +221,7 @@ function shrinking_array_sample!(rng::AbstractRNG, a::AbstractArray, x::Abstract
     for (i, m) in zip(1:k, n:-1:n-k+1)
         rgen = RangeGenerator(1:m)
         idx = rand(rng, rgen)
-        x[i], a[idx] = a[idx], a[m]
+        @inbounds x[i], a[idx] = a[idx], a[m]
     end
 
     return x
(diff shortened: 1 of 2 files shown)


src/sampling.jl
```diff
@@ -206,6 +206,30 @@ end
...
+    shrinking_array_sample!([rng], a::AbstractArray, x::AbstractArray)
+
+Shrinking-array sampling: samples from an array by replacing each selected
+element with the last element in the array and decrementing the array length
+counter by 1 each time - hence shrinking-array.
+"""
+function shrinking_array_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
+    n = length(a)
+    k = length(x)
+    k <= n || error("length(x) should not exceed length(a)")
```
ararslan: Rather than a generic error this would be a good situation to throw a typed error, e.g. `DimensionMismatch` or `ArgumentError`.

src/sampling.jl
```diff
@@ -206,6 +206,30 @@ end
...
+
+    ca = copy(a)
```
ararslan: Since this function mutates `x` we'd might as well avoid copying `a` and just let it mutate `a` as well. For downstream functions that use it, we can always explicitly `copy` an array used as an argument to it.

xiaodaigh: Yeah. Just didn't want to break too many things. This is a good point.

src/sampling.jl
```diff
@@ -206,6 +206,29 @@ end
...
+    shrinking_array_sample!([rng], a::AbstractArray, x::AbstractArray)
+
+Shrinking-array sampling: samples from an array by replacing each selected
+element with the last element in the array and decrementing the array length
+counter by 1 each time - hence shrinking-array.
+"""
+function shrinking_array_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
+    n = length(a)
+    k = length(x)
+    k <= n || error(DimensionMismatch("length(x) should not exceed length(a)"))
```
ararslan: For typed errors, you need to use `throw` instead of `error`

src/sampling.jl
```diff
@@ -206,6 +206,32 @@ end
...
+Shrinking-array sampling: samples from an array by replacing each selected
+element with the last element in the array and decrementing the array length
+counter by 1 each time - hence shrinking-array.
+"""
+function shrinking_array_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
+    n = length(a)
+    k = length(x)
+    k <= n || throw(DimensionMismatch("length(x) should not exceed length(a)"))
+
+    # need to use an index instead as in the case where a = 1:j then a[jj] = val
```
nalimilan: What is `jj`?

xiaodaigh: `jj` is just an arbitrary integer. Should make more clear. The point is `a[jj]` for arbitrary integer `jj` will fail if `a` is a `UnitRange` . Basically, when `a` doesn't have a `setindex!` but is still `typeof(a) <: AbstractArray` then the original code will fail, hence this fix.

xiaodaigh:
@nalimilan In regards to review, I would like to point out the following two points which will hopefully help increase comfort level and aid in the review process

- `shrinking_array_sample!` is a straight replacement of `self_avoid_sample!` and you should see that it passes all tests that was intended for `self_avoid_sample!`, see line 159-166 for test/sampling.jl
- I can't seem to find direct references in other literature, but the algorithm is quite short and straight-forward so perhaps is part of a more complicated theory. But I found [this in the r-devel](https://stat.ethz.ch/pipermail/r-devel/2017-October/075012.html) which references R's implementation of sample without replacement. 

You can see R's C implementation (for sampling integers) is basically implementing the same idea as shrinking_array_sample!, e.g. see the shrinking array code `x[j] = x[--n]`,

```r
  for (i = 0; i < n; i++)
            x[i] = i;
        for (i = 0; i < k; i++) {
            j = n * unif_rand();
            y[i] = x[j] + 1;
            x[j] = x[--n];
        }
```

I remember this algorithm as it was a question in an interview and at the end the interviewer showed me this clever solution.

Please let me know anything else that might help speed up the process

xiaodaigh:
Thanks for the efforts in reviewing. After going through the code again I realized that `fisher_yates_sample!` is a more generalized algorithm than my `shrinking_array_sample!` and that `self_avoid_sample!` is not meant to be used when the number of elements being sampled, `k`, is large relative to `n` (large being `24k > n`) - which is one of the main reasons for why I thought my implementation would be faster.

Apologies for taking up your valuable time; all is not lost as I think I can repurpose what I learned here to improve `fisher_yates_sample!` in another PR.

This is one of my first efforts at contributing to open source so I've made some mistakes regarding the thoroughness of my research.

nalimilan:
No worries. Though, note that we should avoid posting code which uses a different license on a package's issues, as anybody who reads it could be accused of taking inspiration from it. Since R is GPL and StatsBase is MIT, it's not possible to read R code and adapt it to write StatsBase code. I've removed the snippet to be on the safe side.

PR #325, opened 2017-12-06
Extend Base.isequal to AbstractWeights

lbittarello:


ararslan:
This seems reasonable to me. Could you add some tests?

src/weights.jl
```diff
@@ -481,7 +481,7 @@ The weighted median ``x_k`` is the element of `x` that satisfies
 
 If a weight has value zero, then its associated data point is ignored.
 If none of the weights are positive, an error is thrown.
-`NaN` is returned if `x` contains any `NaN` values.
```
lbittarello: Not sure what happened here

lbittarello:
Done!

src/weights.jl
ararslan: Not a big deal. Some editors eat trailing whitespace when a file is opened.

ararslan:
Cool, thanks! Out of curiosity, did the examples in the test work before this change? I'm not sure what the behavior of `isequal` was prior to this.

lbittarello:
I was just getting `true` every time... So two of the three examples failed.

ararslan:
Okay great, thanks for the explanation.

src/weights.jl
```diff
@@ -27,6 +27,12 @@ isempty(wv::AbstractWeights) = isempty(wv.values)
 Base.getindex(wv::AbstractWeights, i) = getindex(wv.values, i)
 Base.size(wv::AbstractWeights) = size(wv.values)
 
+function Base.isequal(w₁::W, w₂::W) where {W <: AbstractWeights}
```
rofinn: Most of the existing `isequal` methods I've seen use `x` & `y` or `a` & `b` instead of  subscripts.

src/weights.jl
```diff
@@ -27,6 +27,12 @@ isempty(wv::AbstractWeights) = isempty(wv.values)
...
+    isequal(sum(w₁), sum(w₂)) && isequal(values(w₁), values(w₂))
+end
+
+Base.isequal(w₁::AbstractWeights, w₂::AbstractWeights) = false
```
rofinn: I'm guessing this is the fall back condition where the `typeof(w₁) != typeof(w₂)`, but it might we worth including a comment.

nalimilan: Is it really needed though? The fallback to `===` should be enough.

test/weights.jl
```diff
@@ -36,6 +36,22 @@ weight_funcs = (weights, aweights, fweights, pweights)
     @test sum(sa, wv) === 7.0
 end
 
+# isequal
+
+@testset "isequal" begin
+
```
nalimilan: Remove empty line. It would make sense to test with `NaN` or `-0.0`, since that's what makes `isequal` different from `==`.

test/weights.jl
```diff
@@ -36,6 +36,22 @@ weight_funcs = (weights, aweights, fweights, pweights)
...
+    x = FrequencyWeights([1; 2; 3], 6)
+
+    y = FrequencyWeights([1; 2; 3], 6) # same values, type and parameters
+    @test isequal(x, y) == true
```
nalimilan: FWIW, `== true` is actually redundant here since `@test` will only pass if the returned value is `=== true` (which is even stricter).

src/weights.jl
lbittarello: I've just realized that this two-method implementation is equivalent to `x === y`. It returns `false` if `x = fweights([1; 2; 3])` and `y = fweights([1.0; 2.0; 3.0])`, whereas `isequal([1; 2; 3], [1.0; 2.0; 3.0])` returns true. It is stricter than I intended – namely, distinguishing between the different weight types (`FrequencyWeights`, `ProbabilityWeights`). I don't know if we want to return `false` because a difference in parameters. I'm trying to figure out a way around it (other than brute force).

nalimilan: Using `isequal` as you do is fine here, indeed we want `1` and `1.0` to be considered as equal. What I meant is that you should be able to get rid of the `= false` method.

lbittarello: But then `isequal(x, y) = true` if `x = fweights([1; 2; 3])` and `y = pweights([1; 2; 3])`

nalimilan: Not if you define only `Base.isequal(w₁::W, w₂::W) where {W <: AbstractWeights}` as you do above.

lbittarello: It does. As a matter of fact, `isequal(x, y) = true` even if `x = FrequencyWeights([1; 2; 3], 6)` and `y = ProbabilityWeights([1; 2; 3], 0)` when we only define the first method.

Here's my guess. Suppose that we only define the first method. When `x` and `y` have the same type and the same parameters, the new method is called to check the values and the sum. When `x ` and `y` have different types or different parameters, the default method is called. Because `AbstractWeights` are subtypes of `AbstractVector`, the fallback is `isequal(x::Array, y::Array)`, which only cares about the `values` field.

nalimilan: Ah, right, then we indeed need the method. Probably worth adding a comment to explain why it's there.

lbittarello: Here's a proposition:

    for w in [AnalyticWeights; FrequencyWeights; ProbabilityWeights; Weights]
        eval(quote
            Base.isequal(x::$w, y::$w) = isequal(x.sum, y.sum) && isequal(x.values, y.values)
            Base.:(==)(x::$w, y::$w)   = (x.sum == y.sum) && (x.values == y.values)
        end)
    end

    Base.isequal(x::AbstractWeights, y::AbstractWeights) = false
    Base.:(==)(x::AbstractWeights, y::AbstractWeights)   = false

Under this implementation, parameters aren't a problem: `isequal(x, y) = true` if `x = fweights([1])` and `y = fweights([1.0])`, but `isequal(x, y) = false` if `y = pweights[1.0]`.

Does it seem reasonable?

nalimilan: Yeah, but I don't think you need the loop. Isn't `Base.isequal(w₁::W, w₂::W) where {W <: AbstractWeights}` enough, rather than defining a separate function for each type?

lbittarello: Under that implementation, `isequal(x, y) = false` if `x = weights([1])` and `y = weights([1.0])`.* I thought we didn't want that behavior.

* Reason being: `x` is `Weights{Int64, Int64, Vector{Int64}}` and `y` is `Weights{Float64, Float64, Vector{Float64}}`, so `isequal(x, y)` calls the fallback.

nalimilan: Damn, you're right. I can't believe I repeatedly got this wrong so many times...

test/weights.jl
```diff
@@ -36,6 +36,34 @@ weight_funcs = (weights, aweights, fweights, pweights)
     @test sum(sa, wv) === 7.0
 end
 
+# isequal and ==
+
+@testset "isequal" begin
```
nalimilan: Should be "isequal and ==", and you can drop the above comment.

It would be good to test this for all weights types. You can do that with a `for` on the testset line, iterating over the type. To test that different types do not compare equal, you can do `if T !== FrequencyWeights`.

test/weights.jl
```diff
@@ -36,6 +36,34 @@ weight_funcs = (weights, aweights, fweights, pweights)
...
+    @test isequal(x, y)
+    @test x == y
+
+    y = FrequencyWeights([1; 2; 3], 6.0) # same values and type, different parameters
+    @test isequal(x, y)
+    @test x == y
+
+    y = ProbabilityWeights([1; 2; 3], 6) # same values and parameters, different types
+    @test !isequal(x, y)
+    @test !(x == y)
```
nalimilan: You can use `!=`.

test/weights.jl
```diff
@@ -36,6 +36,34 @@ weight_funcs = (weights, aweights, fweights, pweights)
...
+# isequal and ==
+
+@testset "isequal" begin
+    x = FrequencyWeights([1; 2; 3], 6)
+
+    y = FrequencyWeights([1; 2; 3], 6) # same values, type and parameters
+    @test isequal(x, y)
+    @test x == y
+
+    y = FrequencyWeights([1; 2; 3], 6.0) # same values and type, different parameters
```
nalimilan: Would be good to use `Float64` for the weights themselves too, just in case.

src/weights.jl
```diff
@@ -194,6 +194,18 @@ pweights(vs::RealArray) = ProbabilityWeights(vec(vs))
     end
 end
 
+##### Equality tests #####
+
+for w in [AnalyticWeights; FrequencyWeights; ProbabilityWeights; Weights]
```
nalimilan: Better use a tuple.

Below, I think it's more standard to do `@eval begin`. Not sure.

test/weights.jl
```diff
@@ -36,6 +36,34 @@ weight_funcs = (weights, aweights, fweights, pweights)
     @test sum(sa, wv) === 7.0
 end
 
+# isequal and ==
+
+@testset "isequal" begin
+    x = FrequencyWeights([1; 2; 3], 6)
```
nalimilan: It would be more natural to use commas, semicolons are only needed for concatenation of vectors.

src/weights.jl
rofinn: This seems like something we should be able to express with an appropriately parameterized method.

nalimilan: Maybe... But which one? :-p

omus: How about:
```julia
function Base.isequal(x::W, y::W) where {W <: AbstractWeights}
    isequal(sum(x), sum(y)) && isequal(values(x), values(y))
end

function Base.isequal(x::AbstractWeights, y::AbstractWeights)
    Base.datatype_name(typeof(x)) == Base.datatype_name(typeof(y)) || throw(MethodError(Base.isequal, (x, y)))
    isequal(sum(x), sum(y)) && isequal(values(x), values(y))
end
```

rofinn: That would also work if someone wants to define their own custom weights type... for some reason :)

nalimilan: Yeah. But using the unexported `Base.datatype` name isn't terribly clean. And the generic definition could even be incorrect if somebody defines an `AbstractWeights` type with a type parameter specifying the kind of weights (i.e. types with different parameters should be considered as different).

lbittarello: Both implementations work. I leave the decision to smarter people than me. :)

nalimilan:
Thanks @lbittarello. Please continue to port features you need to StatsBase when that makes sense.

PR #331, opened 2017-12-27
Fix a couple more 0.7 deprecations

ararslan:


src/cov.jl
```diff
@@ -1,5 +1,9 @@
 ## extended methods for computing covariance and scatter matrix
 
+if !isdefined(Base, :CartesianIndices)
```
nalimilan: No Compat definition for this?

src/rankcorr.jl
```diff
@@ -110,7 +110,7 @@ corkendall(X::RealMatrix, Y::RealMatrix) = Float64[corkendall!(float(X[:,i]), fl
 
 function corkendall(X::RealMatrix)
     n = size(X, 2)
-    C = eye(n)
+    C = Matrix(Diagonal(ones(n)))
```
nalimilan: Isn't there a more efficient way to do this using `I`? IIUC this will allocate the `ones(n)` vector and discard it immediately.

ararslan: There is in 0.7 but it's purposefully not in Compat due to the change in default element type for `I`. Not sure what else would be good. I guess I could do something like
```julia
C = zeros(Float64, n, n)
C[diagind(C)] .= 1.0
```
if that would be better, or just live with the `eye` deprecation...

src/cov.jl
ararslan: Nope

src/rankcorr.jl
nalimilan: OK. Maybe just add a version check? That will allow Femtocleaner to retain only the new efficient and concise approach automatically.

quinnj:
Alright, I picked up where @ararslan left off and got everything working/tests passing on 0.7/0.6 for me locally. Let's see how CI does; nightly will probably fail due to the weird "can't find the module" error, but oh well, we'll see how long it takes to fix that in Base.

ararslan:
Thanks Jacob! I completely forgot about this PR.

src/moments.jl
```diff
@@ -137,6 +137,8 @@ weights used:
 """
 Base.std(v::RealArray, w::AbstractWeights; mean=nothing, corrected::DepBool=nothing) =
     sqrt.(var(v, w; mean=mean, corrected=depcheck(:std, corrected)))
+Base.std(v::AbstractArray{<:AbstractFloat}, w::AbstractWeights; mean=nothing, corrected::DepBool=nothing) =
```
nalimilan: Why is this needed? To fix an ambiguity?

test/ranking.jl
```diff
@@ -2,6 +2,7 @@ using StatsBase
 using Compat
 using Compat.Test
 
+@testset "ranking" begin
```
nalimilan: Missing line break after this line.

test/sampling.jl
```diff
@@ -1,10 +1,8 @@
 using StatsBase
-using Compat
-using Compat.Test
-import Base: maxabs
+using Compat, Compat.Test, Compat.Random
 import StatsBase: norepeat
 
-srand(1234)
+Random.srand(1234)
```
nalimilan: No need to use the prefix.

src/moments.jl
quinnj: Yes, see my comments [here](https://github.com/JuliaLang/julia/commit/2cc82d29e1063aad0acc680bb546a5bd475743ee)

ExpandingMan:
Anything more needed for this PR?  Just fixing @nalimilan s to comments above?

PR #336, opened 2018-01-02
Loosened `countmap` and `addcount` signatures.

### LINKED ISSUE #335: Allow numeric Vectors as weights
When I first started using `countmap` I was confused about how to do a count with weights. I thought `countmap(x, wgt)` would do where `wgt` is a numeric vector. But actually we need `countmap(x, weights(wgt))`. 

When I looked into the documentation there are 3 types of weights: they are Analytical, Frequency and Probability; these seem to be useful. But before reading the source code I only thought of weights as a numeric weight for each value and I suspect a lot of people would think like that too.

So do you think it's ok to add this signature?
`countmap(Vector{T}, Vectr{S}) where {T, S<:Number}`

rofinn:
They can now accept any `AbstractVector{<:Real}` as weights w/o introducing
method ambiguities. Also, `AbstractWeights` is a subtype of `AbstractVector`, so the existing behaviour is preserved.

Closes #335

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -317,7 +317,7 @@ function addcounts_radixsort!(cm::Dict{T}, x::AbstractArray{T}) where T
     return cm
 end
 
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}, wv::AbstractWeights{W}) where {T,W}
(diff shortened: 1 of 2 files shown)


rofinn:
Looks like appveyor is failing on the histogram code.

nalimilan:
I wonder whether we should make weights a keyword argument now that they are fast. That would be more explicit, especially when passing a standard (non-`AbstractWeights`) vector.

PR #337, opened 2018-01-07
generalize ranking

piever:
This PR allows to pass any `AbstractArray` to the ranking functions (they only need to be sortable, which is true much more generally than `Array{<:Real}` which is the current restriction). Also, one can now pass a custom "less than" function or reverse the order of sorting. `sortperm` also provides a `by` keyword (to give a function to be applied before sorting), but that wouldn't work here as it would screw up the equality comparison, the user will have to do `competerank(by.(x))`

### DIFF
--- a/src/ranking.jl
+++ b/src/ranking.jl
@@ -16,7 +16,7 @@ end
 
 
 # Ordinal ranking ("1234 ranking") -- use the literal order resulted from sort
-function ordinalrank!(rks::RealArray, x::RealArray, p::IntegerArray)
+function ordinalrank!(rks::RealArray, x::AbstractArray, p::IntegerArray)
     n = _check_randparams(rks, x, p)
 
     if n > 0
@@ -32,18 +32,20 @@ end
 
 
 """
-    ordinalrank(x)
+    ordinalrank(x; lt = isless, rev::Bool = false)
 
 Return the [ordinal ranking](https://en.wikipedia.org/wiki/Ranking#Ordinal_ranking_.28.221234.22_ranking.29)
-("1234" ranking) of a real-valued array.
+("1234" ranking) of an array. The `lt` keyword allows providing a custom "less
+than" function; use `rev=true` to reverse the sorting order.
 All items in `x` are given distinct, successive ranks based on their
-position in `sort(x)`.
+position in `sort(x; lt = lt, rev = rev)`.
 """
-ordinalrank(x::RealArray) = ordinalrank!(Array{Int}(uninitialized, size(x)), x, sortperm(x))
+ordinalrank(x::AbstractArray; lt = isless, rev::Bool = false) =
+    ordinalrank!(Array{Int}(uninitialized, size(x)), x, sortperm(x; lt = lt, rev = rev))
 
 
 # Competition ranking ("1224" ranking) -- resolve tied ranks using min
-function competerank!(rks::RealArray, x::RealArray, p::IntegerArray)
+function competerank!(rks::RealArray, x::AbstractArray, p::IntegerArray)
     n = _check_randparams(rks, x, p)
 
     if n > 0
@@ -70,18 +72,20 @@ end
 
 
 """
-    competerank(x)
+    competerank(x; lt = isless, rev::Bool = false)
 
 Return the [standard competition ranking](http://en.wikipedia.org/wiki/Ranking#Standard_competition_ranking_.28.221224.22_ranking.29)
-("1224" ranking) of a real-valued
-array. Items that compare equal are given the same rank, then a gap is left
+("1224" ranking) of an array. The `lt` keyword allows providing a custom "less
+than" function; use `rev=true` to reverse the sorting order.
+Items that compare equal are given the same rank, then a gap is left
 in the rankings the size of the number of tied items - 1.
 """
-competerank(x::RealArray) = competerank!(Array{Int}(uninitialized, size(x)), x, sortperm(x))
+competerank(x::AbstractArray; lt = isless, rev::Bool = false) =
+    competerank!(Array{Int}(uninitialized, size(x)), x, sortperm(x; lt = lt, rev = rev))
 
 
 # Dense ranking ("1223" ranking) -- resolve tied ranks using min
-function denserank!(rks::RealArray, x::RealArray, p::IntegerArray)
+function denserank!(rks::RealArray, x::AbstractArray, p::IntegerArray)
     n = _check_randparams(rks, x, p)
 
     if n > 0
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks! Looks good.

I wonder what to do about missing values though. With the current state of the PR, they will be sorted last. That's consistent with R's default, but not with what we do elsewhere (propagate). So I'd rather either throw an error in the presence of missing values for now, or use a `missing` rank for missing values. Else it will be hard to change this behavior later.

piever:
I was unsure whether the missing data handling should happen here or elsewhere. Here actually giving `missing` would error on some methods (as they have `if x == y` where `x` could be missing) but not others, so it's really not ideal. I'll update this PR to allow `missing` and give `missing` rank to `missing` values (I also think it's consistent with our handling of missing data).

src/ranking.jl
```diff
@@ -168,7 +171,23 @@ of an array. The `lt` keyword allows providing a custom "less
...
+Missing values are assigned rank `missing`.
 """
 tiedrank(x::AbstractArray; lt = isless, rev::Bool = false) =
     tiedrank!(Array{Float64}(uninitialized, size(x)), x, sortperm(x; lt = lt, rev = rev))
 
+for (f, f!, S) in zip([:ordinalrank, :competerank, :denserank, :tiedrank],
+                      [:ordinalrank!, :competerank!, :denserank!, :tiedrank!],
+                      [Int64, Int64, Int64, Float64])
+    @eval begin
+        function $f(x::AbstractArray{Union{T, Missing}}; lt = isless, rev::Bool = false) where {T}
```
nalimilan: Better use `AbstractArray{>: Missing}`.

src/ranking.jl
```diff
@@ -168,7 +171,23 @@ of an array. The `lt` keyword allows providing a custom "less
...
+            xv = collect(skipmissing(x))
+            sp = sortperm(xv; lt = lt, rev = rev)
+            rks = missings($S, length(x))
+            $(f!)(view(rks, .!ismissing.(x)), xv, sp)
+            rks
+        end
+
+        $f(x::AbstractArray{Missing}; lt = isless, rev::Bool = false) = missings($S, length(x))
```
nalimilan: Why is this needed? Because of ambiguities?

test/ranking.jl
```diff
@@ -1,31 +1,45 @@
...
 
 a = [1.0, 2.0, 2.0, 3.0, 4.0, 4.0, 4.0, 5.0]
 x = [3.0, 1.0, 2.0, 4.0, 4.0, 2.0, 5.0, 4.0]  # x is a permutated version of a
+xm = [3.0, 1.0, 2.0, 4.0, 4.0, 2.0, 5.0, 4.0, missing]
 s = ["c", "a", "b", "d", "d", "b", "e", "d"] # s is a vector of strings ordered like x
 
 @test ordinalrank(a) == [1, 2, 3, 4, 5, 6, 7, 8]
 @test ordinalrank(x) == [4, 1, 2, 5, 6, 3, 8, 7]
+@test collect(skipmissing(ordinalrank(xm))) == [4, 1, 2, 5, 6, 3, 8, 7]
+@test ismissing(ordinalrank(xm)[end])
```
nalimilan: It would be good to check that the returned value has the expected values apart from the last one. Use `isequal` to compare it with the expected array. Same below.

src/ranking.jl
```diff
@@ -168,7 +171,23 @@ of an array. The `lt` keyword allows providing a custom "less
...
 
+for (f, f!, S) in zip([:ordinalrank, :competerank, :denserank, :tiedrank],
+                      [:ordinalrank!, :competerank!, :denserank!, :tiedrank!],
+                      [Int64, Int64, Int64, Float64])
+    @eval begin
+        function $f(x::AbstractArray{Union{T, Missing}}; lt = isless, rev::Bool = false) where {T}
+            xv = collect(skipmissing(x))
+            sp = sortperm(xv; lt = lt, rev = rev)
+            rks = missings($S, length(x))
+            $(f!)(view(rks, .!ismissing.(x)), xv, sp)
```
nalimilan: Since you need to compute `.!ismissing.(x)`, better do something like `mask = .!ismissing.(x); xv = x[mask]` above, that could be more efficient, and preserve the array type. Or maybe you can use a view there too?

src/ranking.jl
```diff
@@ -168,7 +171,23 @@ of an array. The `lt` keyword allows providing a custom "less
...
 Items that compare equal receive the mean of the
 rankings they would have been assigned under ordinal ranking.
+Missing values are assigned rank `missing`.
 """
 tiedrank(x::AbstractArray; lt = isless, rev::Bool = false) =
     tiedrank!(Array{Float64}(uninitialized, size(x)), x, sortperm(x; lt = lt, rev = rev))
 
+for (f, f!, S) in zip([:ordinalrank, :competerank, :denserank, :tiedrank],
+                      [:ordinalrank!, :competerank!, :denserank!, :tiedrank!],
+                      [Int64, Int64, Int64, Float64])
```
nalimilan: Better use `Int`?

piever:
Should be good now. We can always add a keyword argument later if we want to allow specifying a different handling of missing values.

src/ranking.jl
piever: In that case don't I get a performance penalty as `x[mask]` has eltype `Union{T, Missing}` ? I thought that was to be avoided as `sortperm` is probably the most costly operation.

piever: Maybe `disallowmissing(view(x, mask))` is the best solution.

src/ranking.jl
piever: My bad, it was actually not needed.

src/ranking.jl
nalimilan: Ah, right, and `disallowmissing` will have to check that all entries are non-missing, so that's not more efficient either. Better keep the current code then.

src/ranking.jl
```diff
@@ -158,8 +167,26 @@ end
...
+tiedrank(x::AbstractArray; lt = isless, rev::Bool = false) =
+    tiedrank!(Array{Float64}(uninitialized, size(x)), x, sortperm(x; lt = lt, rev = rev))
+
+for (f, f!, S) in zip([:ordinalrank, :competerank, :denserank, :tiedrank],
+                      [:ordinalrank!, :competerank!, :denserank!, :tiedrank!],
+                      [Int, Int, Int, Float64])
+    @eval begin
+        function $f(x::AbstractArray{>: Missing}; lt = isless, rev::Bool = false)
+            mask = .!ismissing.(x)
+            xv = disallowmissing(view(x, mask))
```
nalimilan: Ah, I hadn't seen you had updated the code. Then keep it that way, but instead of `mask = ...` do `inds = find(!ismissing, x)`. That should be more efficient since that's what  `view` does under the hood, and you can avoid doing it twice.

test/ranking.jl
```diff
@@ -1,18 +1,42 @@
...
+@test isequal(ordinalrank(xm), [4, 1, 2, 5, 6, 3, 8, 7, missing])
+@test all(ismissing, ordinalrank([missing, missing]))
+@test ordinalrank(s) == ordinalrank(x)
+@test ordinalrank(x, rev = true) == ordinalrank(-x)
+@test ordinalrank(x, lt = (x, y) -> isless(y, x)) == ordinalrank(-x)
 
 @test competerank(a) == [1, 2, 2, 4, 5, 5, 5, 8]
 @test competerank(x) == [4, 1, 2, 5, 5, 2, 8, 5]
+@test isequal(competerank(xm), [4, 1, 2, 5, 5, 2, 8, 5, missing])
+@test all(ismissing, competerank([missing, missing]))
```
nalimilan: Here too, better check the result using `isequal`, so that you also check the dimensions. Same below.

src/ranking.jl
piever: Good point, I was wondering how Julia could get a view from a Boolean indexing...

piever:
Bump? Is there anything else that needs to be done for this?

PR #339, opened 2018-01-13
Faster countmap for small bits type

xiaodaigh:
See below benchmarking code which shows that the new algorithms for small bits types significantly outperforms the existing algorithms, e.g. for `Int16` the timing goes from 56.4s to 2.3s

timings vs new algorithm for small bits types vs old

```julia
import StatsBase: addcounts!, addcounts_dict!, addcounts_radixsort!

using Plots
bool = rand(Bool, 1_000_000_000);
bool_new = @elapsed addcounts!(Dict{Bool,Int}(), bool); # 0.7 
bool_old = @elapsed addcounts_dict!(Dict{Bool,Int}(), bool); # 13

uint8 = rand(UInt8, 1_000_000_000);
uint8_new = @elapsed addcounts!(Dict{UInt8,Int}(), uint8); # 1.1
uint8_old = @elapsed addcounts_radixsort!(Dict{UInt8,Int}(), uint8); # 8.78

int8 = rand(Int8, 1_000_000_000);
int8_new = @elapsed addcounts!(Dict{Int8,Int}(), int8); # 1.1
int8_old = @elapsed addcounts_radixsort!(Dict{Int8,Int}(), int8); # 8.5


uint16 = rand(UInt16, 1_000_000_000);
uint16_new = @elapsed addcounts!(Dict{UInt16,Int}(), uint16); # 2.3
uint16_old = @elapsed addcounts_radixsort!(Dict{UInt16,Int}(), uint16); # 59.73

int16 = rand(Int16, 1_000_000_000);
int16_new = @elapsed addcounts!(Dict{Int16,Int}(), int16) # 2.3
int16_old = @elapsed addcounts_radixsort!(Dict{Int16,Int}(), int16) # 56.4


using DataFrames

resdf = DataFrame(
    datatype = repeat(["Bool","UInt8","Int8","UInt16","Int16"], inner=2),
    new_old = repeat(["new","old"], outer=5),
    timings = [bool_new, bool_old, uint8_new, uint8_old, int8_new, int8_old, uint16_new, uint16_old, int16_new, int16_old])

resdf[:xlabel] = resdf[:datatype].*resdf[:new_old]
using Plots, StatPlots
groupedbar(resdf[:datatype], resdf[:timings], group=resdf[:new_old], bar_position = :dodge, ylabel="time (s)")
savefig("Timings vs new algorithm for small bits types vs old")
```

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -283,6 +283,40 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
     return cm
 end
 
+"""Specialist addcounts methods for small bits types"""
+function addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool})
+    sumx = sum(x)
+    cm[true] = sumx
+    cm[false] =length(x) = sumx
+    cm
+end
+
+"""toindex functions to convert an integer to an array index between 1 to typemax{T}"""
+function toindex(x::Int16)
+  Int(x) + 32769
+end
+
+function toindex(x::Int8)
+  Int(x) + 129
+end
+
+function toindex(x::T) where T <: Union{UInt8, UInt16}
+  Int(x) + 1
+end
+
+function addcounts!(cm::Dict{T}, x::Vector{T}) where T <: Union{UInt8, UInt16, Int8, Int16}
+  arr = zeros(Int, 2^(sizeof(T)*8))
+
+  for xi in x
+    @inbounds arr[toindex(xi)] += 1
+  end
+
+  for (i,arr1) in zip(typemin(T):typemax(T),arr)
+    arr1 == 0 || @inbounds cm[i] = arr1
+  end
+  cm
+end
+
 const BaseRadixSortSafeTypes = Union{Int8, Int16, Int32, Int64, Int128,
                                      UInt8, UInt16, UInt32, UInt64, UInt128,
                                      Float32, Float64}
@@ -351,6 +385,9 @@ of occurrences.
                      RAM and is safe for any data type.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
+# specialist function for Bool
+countmap(x::AbstractArray{T}) where T<:Union{Bool, UInt8, UInt16, Int8, Int16} = addcounts!(Dict{T,Int}(), x) 
+    
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
 
 
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -258,9 +258,13 @@ raw counts.
                      RAM and is safe for any data type.
 """
 function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :auto) where T
+    # If T <: Union{UInt8, UInt16, Int8, Int16, Bool} then another `addcounts!` function is called.
+    # If the bits type is of small size i.e. it can have up to 65536 distinct values then it is
+    # always better to apply a counting-sort like reduce algorithm for faster results and less memory usage
+
     # if it's safe to be sorted using radixsort then it should be faster
     # albeit using more RAM
-    if radixsort_safe(T) && (alg == :auto || alg == :radixsort)
+   if radixsort_safe(T) && (alg == :auto || alg == :radixsort)
         addcounts_radixsort!(cm, x)
     elseif alg == :radixsort
         throw(ArgumentError("`alg = :radixsort` is chosen but type `radixsort_safe($T)` did not return `true`; use `alg = :auto` or `alg = :dict` instead"))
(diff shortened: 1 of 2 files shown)


src/counts.jl
```diff
@@ -283,6 +283,40 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
     return cm
 end
 
+"""Specialist addcounts methods for small bits types"""
```
nalimilan: This docstring doesn't really add anything given that the signature already gives this information. Also, docstrings are attached to a given method, so you cannot use the plural here. Better make this a standard comment, with a line break after it so that it's clear it describes several functions and not just this one.

src/counts.jl
```diff
@@ -283,6 +283,40 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
...
+function addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool})
+    sumx = sum(x)
+    cm[true] = sumx
```
nalimilan: `addcounts!` is supposed to add counts, not overwrite them, right?

src/counts.jl
```diff
@@ -283,6 +283,40 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
...
+    cm[false] = length(x) - sumx
+    cm
+end
+
+"""toindex functions to convert an integer to an array index between 1 to typemax{T}"""
```
nalimilan: Same remark here. And "between 1 and typemax(T)".

Also better use the short syntax `toindex(x::Int16) = ...`, and use `typemax` for clarity.

src/counts.jl
```diff
@@ -283,6 +283,40 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
...
+function toindex(x::Int16)
+  Int(x) + 32769
+end
+
+function toindex(x::Int8)
+  Int(x) + 129
+end
+
+function toindex(x::T) where T <: Union{UInt8, UInt16}
```
nalimilan: No need to define `T`, just use the union in the signature.

src/counts.jl
```diff
@@ -283,6 +283,40 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
...
+  Int(x) + 1
+end
+
+function addcounts!(cm::Dict{T}, x::Vector{T}) where T <: Union{UInt8, UInt16, Int8, Int16}
+  arr = zeros(Int, 2^(sizeof(T)*8))
```
nalimilan: Maybe better call this `counts`? Below, `arr1` would better be e.g. `c`, as it's not an array.

src/counts.jl
```diff
@@ -283,6 +283,40 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
...
+
+function toindex(x::Int8)
+  Int(x) + 129
+end
+
+function toindex(x::T) where T <: Union{UInt8, UInt16}
+  Int(x) + 1
+end
+
+function addcounts!(cm::Dict{T}, x::Vector{T}) where T <: Union{UInt8, UInt16, Int8, Int16}
```
nalimilan: This method has to accept the `alg` keyword argument, or it won't be possible to write generic code. It's fine to ignore it if we consider that this implementation is always better than the others (but add a comment to explain this).

src/counts.jl
```diff
@@ -351,6 +385,9 @@ of occurrences.
                      RAM and is safe for any data type.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
+# specialist function for Bool
+countmap(x::AbstractArray{T}) where T<:Union{Bool, UInt8, UInt16, Int8, Int16} = addcounts!(Dict{T,Int}(), x)
```
nalimilan: You can remove this function after adding support for the `alg` keyword argument to `addcounts!` above.

test/counts.jl
```diff
@@ -114,3 +114,19 @@ pm = proportionmap(x, weights(w))
 @test pm["a"] ≈ (5.5 / 13.5)
 @test pm["b"] ≈ (4.5 / 13.5)
 @test pm["c"] ≈ (3.5 / 13.5)
+
+# testing small bits type
+bx = [true, false, true, true, false]
+@test countmap(bx) == Dict(true => 3, false => 2)
+
+uint8x = UInt8[0, 10, 255, 34, 10]
```
nalimilan: Better write a loop to avoid repeating values.

src/counts.jl
nalimilan: Actually, wouldn't it be enough to subtract `typemin(T)` and add `1` for all types? Then you wouldn't need these functions at all.

nalimilan:
Oh, and while you're at it, why not support `Char` too?

src/counts.jl
xiaodaigh: would that slow things down? actually compiler might be small enough

src/counts.jl
xiaodaigh: I think `T` is needed for `typemin(T)`

test/counts.jl
xiaodaigh: what do you mean? I though the tests as they are are quite straight-forward and easy to understand and change.

xiaodaigh:
For `Char` will do. But want to close this piece of work first and then do that in a separate PR, if that's alright.

nalimilan:
There are still a few indentation issues.

> For Char will do. But want to close this piece of work first and then do that in a separate PR, if that's alright.

Unless that really requires a lot of work, I'd prefer if you did this directly here. AFAICT it should be quite trivial to add.

src/counts.jl
```diff
@@ -283,6 +283,31 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
...
+# to write generic algorithms, therefore the two methods below still accept the 
+# `alg` argument but it is ignored.
+function addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
+    sumx = sum(x)
+    cm[true] = get(cm, true, 0) + sumx
+    cm[false] = get(cm, false, 0) + length(x) - sumx
+    cm
+end
+
+function addcounts!(cm::Dict{T}, x::Vector{T}; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
```
nalimilan: Should use `::AbstractArray{T}`.

src/counts.jl
```diff
@@ -283,6 +283,31 @@ function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
...
+    counts = zeros(Int, 2^(8sizeof(T)))
+
+    for xi in x
+        @inbounds counts[Int(xi) - typemin(T) + 1] += 1
+    end
+
+    for (i, c) in zip(typemin(T):typemax(T), counts)
+        c == 0 || @inbounds cm[i] = get(cm, i, 0) + c
```
nalimilan: I don't think `@inbounds` achieves anything here. FWIW, I think in this loop and loop moving `@inbounds` before the `for` can make the code faster, because (currently) `for xi in x` still includes some bounds checks. Sometimes it prevents the activate of SIMD.

Also, if you really want the maximal performance you could use the same strategy as in `addcounts_dict!` to avoid doing the dict lookup twice. For 16-bit types it could make a small difference.

xiaodaigh: For `Char` I checked on my system `sizeof(Char)` it's actually 4, so 32 bit. So radixsort is the best bet for this.

nalimilan: Ah, of course, it's funny how traditional C definitions pollute my mind...

test/counts.jl
```diff
@@ -114,3 +114,19 @@ pm = proportionmap(x, weights(w))
...
 @test pm["c"] ≈ (3.5 / 13.5)
+
+# testing small bits type
+bx = [true, false, true, true, false]
+@test countmap(bx) == Dict(true => 3, false => 2)
+
+tx = UInt8[0, 10, 255, 34, 10]
+@test countmap(tx) == Dict(0 => 1, 10 => 2, 255 => 1, 34 => 1)
+
+tx = UInt16[0, 8, 64000, 27, 8]
```
nalimilan: Why `64000`? It would make more sense to test `65535`, to ensure no overflow happens. Same for other types (better write this as `typemax(T)`, and also include `typemin(T)` while you're at it). As I suggested before, put these inside a `for T in (UInt8, ...` loop (using the same values for all except for typemin/typemax).

PR #345, opened 2018-02-09
Add midpoints, which was deprecated in Base.

tpapp:
Currently not exported, to avoid symbol conflict.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -15,7 +15,7 @@ module StatsBase
         using Printf
     end
 
-    if VERSION < v"0.7-"
+    if VERSION < v"0.7.0-DEV.3335"
         import Base: midpoints
     end
 
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -130,7 +130,7 @@ module StatsBase
     Histogram,
     hist,
     # histrange,
-    midpoints,
+    # midpoints,    # TODO export when depending on Julia v0.7
 
     ## robust
     trim,           # trimmed set
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -15,6 +15,10 @@ module StatsBase
         using Printf
     end
 
+    if VERSION < v"0.7-"
+        import Base: midpoints
+    end
+
     ## tackle compatibility issues
 
     export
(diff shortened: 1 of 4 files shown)


src/StatsBase.jl
```diff
@@ -130,7 +130,7 @@ module StatsBase
     Histogram,
     hist,
     # histrange,
-    midpoints,
+    # midpoints,    # TODO export when depending on Julia v0.7
```
nalimilan: Better put this under a `VERSION` check, so that Femtocleaner fixes it automatically. I think the best approach is to always export the symbol (what is currently done), and conditionally import if from Base on 0.6. Then you can drop the module prefix in the docs.

nalimilan: The new commit doesn't correspond to what I suggested. Have you tried that?

src/hist.jl
```diff
@@ -524,3 +524,12 @@ have the same binning, shape of weights and properties (`closed` and
...
 Base.merge(h::Histogram, others::Histogram...) = merge!(zero(h), h, others...)
+
+"""
+    midpoints(v)
+
+Calculate the midpoints (pairwise mean of consecutive elements).
+"""
+midpoints(v::AbstractVector) = [middle(v[i - 1], v[i]) for i in 2:length(v)]
+
+midpoints(r::Range) = r[1:(end - 1)] + 0.5 * step(r)
```
andreasnoack: Probably better to divide with two here to avoid promotion to `Float64`.

tpapp: Thanks, fixed.

tpapp:
Friendly ping: is there anything else I can do to get this merged?

nalimilan:
@andreasnoack Fine with you?

andreasnoack:
It'll be really slow though since it goes through the deprecation machinery. Is there a way to overwrite without a warning?

nalimilan:
Ah, right, I forgot that it was deprecated on 0.6.2 (usually it's the other way around). We could live with the warning, but it's too bad to annoy all users with a function most don't use. It makes it look like something is broken.

Another solution is to avoid importing and exporting `midpoints` on 0.6, so that people would need to do `StatsBase.midpoints` there. Probably the best solution.

Issue #347, opened 2018-02-16
Naming for mad()?


mikkling:
I'm just learning Julia, and it took a while to find out what was wrong with the code for my simple outlier detection algorithm test -- turns out that mad() doesn't yield the expected median of the absolute deviations from the median... 

I looked at earlier issues in StatsBase about MAD and found this (https://github.com/JuliaStats/StatsBase.jl/issues/197): 

"Just to be clear, constant is what make this function produce a consistent estimator of the standard deviation for normally distributed data. So if you intend to remove it, you need to either decide that (a) your estimator is inconsistent or (b) you don't provide any mechanism for producing the number some people believe MAD should generate."

This is certainly a good function, but perhaps it could be named something that indicates that it estimates standard deviation? I, for one, was expecting a function named mad to yield by-the-book MAD. 

[This is also my very first post on GitHub. Sorry if I offend against proper posting conventions!]

|m|
Mikael

nalimilan:
Yes, I think at the very minimum the docs should mention this.

@ararslan Wasn't it a bit radical to stop documenting what `mad` actually returns while removing the argument in https://github.com/JuliaStats/StatsBase.jl/pull/198? I'm also not convinced it was a good idea to remove that argument: what problem does it create? At least it seems the OP is one counter-example to @andreasnoack's statement that nobody uses it.

FWIW, [MATLAB](https://fr.mathworks.com/help/stats/mad.html) doesn't apply any multiplicative factor. OTOH R and Numpy's StatsModels apply it by default, but provide an argument to change it. I really think we should do the same.

ararslan:
> Wasn't it a bit radical to stop documenting what mad actually returns while removing the argument in #198?

Yeah, probably. I guess that didn't occur to me nearly two years ago. 😛 Definitely a good idea to improve the documentation for it.

nalimilan:
OK. Thoughts about reintroducing that argument while we're at it, given that other software provide it?

mikkling:
Thank you for prompt feedback!

So far, i'm loving Julia. Thank you for making it even better!


|m|
Mikael

nalimilan:
See https://github.com/JuliaStats/StatsBase.jl/pull/349.

PR #349, opened 2018-02-18
mad() and mad!() improvements

nalimilan:
Add `center` and `constant` arguments to `mad`, as for `mad!`. Make `center` a keyword argument for `mad!`. Improve docstrings.

Fixes https://github.com/JuliaStats/StatsBase.jl/issues/347.

### DIFF
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -82,3 +82,6 @@ rand(s::RandIntSampler) = rand(Base.GLOBAL_RNG, s)
 @deprecate randi(rng::AbstractRNG, a::Int, b::Int) rand(rng, a:b)
 @deprecate randi(a::Int, b::Int) rand(a:b)
 
+@deprecate(mad!(v::AbstractArray{T}, center;
+                constant::Real = 1 / (-sqrt(2 * one(T)) * erfcinv(3 * one(T) / 2))) where T<:Real,
+           mad!(v, center=center, constant=constant))
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -238,29 +238,40 @@ sem(a::AbstractArray{<:Real}) = sqrt(var(a) / length(a))
 
 # Median absolute deviation
 """
-    mad(v)
+    mad(v; center=median(v), constant=1/quantile(Normal(), 3/4))
 
-Compute the median absolute deviation of `v`.
+Compute the median absolute deviation (MAD) of `v` around `center`
+(by default, around the median).
+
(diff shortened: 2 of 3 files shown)


andreasnoack:
Great with the improved documentation but there [was a reason why](https://github.com/JuliaStats/StatsBase.jl/issues/197#issuecomment-235775085) we decided to remove the options so when adding them again it would be useful to address that, i.e. this version makes an extra copy of the data in the common case where the center isn't known. In addition, this version isn't inferrable.
```julia
julia> @inferred mad([1,2,3])
ERROR: return type Float64 does not match inferred return type Any
Stacktrace:
 [1] error(::String) at ./error.jl:21
```
I'm not totally against providing more options but I think we should think about why we want to add these arguments. @mikkling asked that the `mad` function returned the unscaled version and this PR doesn't accommodate that. Instead, it provides a general scaling argument which I find a very odd way of multiplying the result by a number. In R, which I assume is the inspiration, this might be excused by the fact they avoid some boxing but that doesn't apply to Julia.

If we want to provide the option of not scaling the result then I think a Boolean argument makes more sense but it would be great if @mikkling could provide a little more context. If the issue is just that "the name `mad` suggest that it computed `median(abs(x) - median(x))`" then this PR doesn't fix the issue since the default is unchanged. If the issue is "I *need* the unscaled version" then I'd be interested in knowing why. I can think of two reasons to use this function: as an objective function and as an estimator. For the former, a scaling factor shouldn't matter and for the latter, the factor is appropriate. You could then again argue that this choice depends on normality but that argument applies equally to `var` where we scale with `n-1`.

nalimilan:
Agreed regarding using a boolean argument instead. People can easily do the multiplication by hand if they want. How should we call it? `corrected` or something more specific mentioning the SD?

About inferrability, that's fixed on 0.7 so I don't think it's a problem.

nalimilan:
I've pushed an updated version using `normalize=true` instead of `constant`, and `center=nothing` to avoid allocating a copy just to compute the median. I also agree with @mikkling that it's weird to have `mad` not return the MAD. Though we can't change this without a deprecation period, so I've added a deprecation warning telling the caller to specify `normalize` explicitly, before we change its default.

mikkling:
@andreasnoack, I don’t have anything against mad() as is, except for the scaling factor being undocumented. MAD is a very clearly defined statistical algorithm. I wasted several hours error-checking my code because it didn’t occur to me that mad() wouldn’t return MAD. Had it been documented, I wouldn’t have raised this issue. 

|m|
Mikael

PR #352, opened 2018-02-23
Add weighted IQR methods

tbeason:
Now that weighted quantile methods work, this adds methods for weighted IQR.

I did some testing and it seemed to work fine for me, but would definitely appreciate somebody else testing this. Also, now the REPL help for `iqr` has the new `iqr(v,w)` listed first rather than second, but I don't know how to fix that.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -35,6 +35,7 @@ module StatsBase
     wmean!,             # weighted mean across dimensions with provided storage
     wmedian,            # weighted median
     wquantile,          # weighted quantile
+	wiqr,				# weighted IQR
 
     ## moments
     skewness,       # (standardized) skewness
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -650,3 +650,24 @@ wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::RealVector) = quantile(v
 wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::Number) = quantile(v, w, [p])[1]
 wquantile(v::RealVector, w::RealVector, p::RealVector) = quantile(v, weights(w), p)
 wquantile(v::RealVector, w::RealVector, p::Number) = quantile(v, weights(w), [p])[1]
+
+
+"""
+    wiqr(v, w)
+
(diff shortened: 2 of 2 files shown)


src/StatsBase.jl
```diff
@@ -35,6 +35,7 @@ module StatsBase
     wmean!,             # weighted mean across dimensions with provided storage
     wmedian,            # weighted median
     wquantile,          # weighted quantile
+	wiqr,				# weighted IQR
```
tbeason: Hmm... The indentation looks fine on mine computer. I wonder why it looks funky here?

nalimilan: That's because the character is a tab. Change it to normal spaces.

src/weights.jl
```diff
@@ -650,3 +650,24 @@ wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::RealVector) = quantile(v
...
+or an `AbstractWeights` vector.
+"""
+wiqr{W <: Real}(v::RealVector,w::AbstractWeights{W}) = (q = quantile(v, w, [.25, .75]); q[2] - q[1])
+wiqr(v::RealVector,w::RealVector) = (q = quantile(v, weights(w), [.25, .75]); q[2] - q[1])
+
+
+"""
+    iqr(v, w)
+
+Compute the interquartile range (IQR) of an array with weights `w`, given as either a vector
```
nalimilan: I'd rather require `w` to be `AbstractWeights`, as we do for other methods. Then instead of mentioning it in the text, add it to the signature.

BTW, could you also give the definition of the IQR?

tbeason: `quantile` and `wquantile` both have methods to accept `w` as a `RealVector` (which gets converted to general `Weights`) in addition to the methods where `w` is passed as `AbstractWeights`. I just followed that convention. Would you still like me to change it?

nalimilan: Are you sure? AFAICT only `wquantile` accepts a `RealVector` (which sounds fine and `wiqr` can do the same thing).

tbeason: Ah, subtle differences. I just glanced over that. Ok I'll make these changes on Monday.

PR #355, opened 2018-03-12
Additional abstractions for statsmodels

Nosferican:
Added: `adjr2(::StatisticalModel)` `informationmatrix`, `isfitted`, `islinear`, `leverage`, `mss`, `meanresponse`, `r2(::StatisticalModel)`,  `rss`, `score`, `weights(::StatisticalModel)`.

Depreciated `model_response` to `response`.

Fixes https://github.com/JuliaStats/StatsBase.jl/issues/356.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -305,7 +305,7 @@ modelmatrix(obj::RegressionModel) = error("modelmatrix is not defined for $(type
 """
     leverage(obj::RegressionModel)
 
-Returns the diagonal of the projection matrix.
+Return the diagonal of the projection matrix.
 """
 leverage(obj::RegressionModel) = error("leverage is not defined for $(typeof(obj)).")
 


Nosferican:
Error is unrelated (issue is with ambiguities)

src/statmodels.jl
```diff
@@ -79,20 +108,48 @@ when applicable the intercept and the distribution's dispersion parameter.
...
     stderr(obj::StatisticalModel)
 
 Return the standard errors for the coefficients of the model.
 """
 stderr(obj::StatisticalModel) = sqrt.(diag(vcov(obj)))
 
+"""
+    tss(obj::StatisticalModel)
+
+Return the standard errors for the coefficients of the model.
```
andreasnoack: Should this be the total sum of squares?

Nosferican: I figured a good amount of methods will be dependent on their types, but leaving those with the error methods has no drawbacks that I can think of.
Thanks for catching that.

src/statmodels.jl
```diff
@@ -227,6 +302,13 @@ Return the model matrix (a.k.a. the design matrix).
 """
 modelmatrix(obj::RegressionModel) = error("modelmatrix is not defined for $(typeof(obj)).")
 
+"""
+    modelweights(obj::RegressionModel)
```
nalimilan: Maybe we could just override the `weights` function here? You renamed `model_response` to `response`, so better not introduce a new function starting with "model". Also why limit this to `RegressionModel`?

src/statmodels.jl
```diff
@@ -137,13 +194,15 @@ the likelihood of the model,  ``k`` its number of consumed degrees of freedom
...
 """
+    r2(obj::StatisticalModel)
+    r²(obj::StatisticalModel)
     r2(obj::StatisticalModel, variant::Symbol)
     r²(obj::StatisticalModel, variant::Symbol)
 
 Coefficient of determination (R-squared).
 
 For a linear model, the R² is defined as ``ESS/TSS``, with ``ESS`` the explained sum of squares
-and ``TSS`` the total sum of squares, and `variant` can be omitted.
```
nalimilan: Split the docstring in two, attaching each part to the corresponding method.

src/statmodels.jl
```diff
@@ -79,20 +108,48 @@ when applicable the intercept and the distribution's dispersion parameter.
 """
 dof(obj::StatisticalModel) = error("dof is not defined for $(typeof(obj)).")
 
+"""
+    ess(obj::StatisticalModel)
+
+Return the the explained sum of squares.
+"""
+ess(obj::StatisticalModel) = sum(abs2, fitted(obj) - meanresponse(obj))
```
nalimilan: Better use a generator to avoid allocating a copy. Same below.

src/statmodels.jl
```diff
@@ -214,11 +282,18 @@ Return the fitted values of the model.
...
 """
-model_response(obj::RegressionModel) = error("model_response is not defined for $(typeof(obj)).")
+response(obj::RegressionModel) = error("response is not defined for $(typeof(obj)).")
+
+"""
+    meanresponse(obj::RegressionModel)
+
+Return the mean of the response.
+"""
+meanresponse(obj::RegressionModel) = error("meanresponse is not defined for $(typeof(obj)).")
```
nalimilan: Could you explain why this is useful? Also I guess we could have a `mean(response(obj))` fallback?

src/statmodels.jl
simonbyrne: ess also means "effective sample size" in MCMC

src/statmodels.jl
```diff
@@ -39,6 +39,35 @@ the likelihood of the model.
 """
 deviance(obj::StatisticalModel) = error("deviance is not defined for $(typeof(obj)).")
 
+"""
+    hessian(obj::StatisticalModel)
+
+Return the Hessian of the statistical model.
+"""
+hessian(obj::StatisticalModel) = error("hessian is not defined for $(typeof(obj)).")
```
simonbyrne: [Hessian](https://en.wikipedia.org/wiki/Hessian_matrix) is a much more general term. In fact, several other libraries already define `hessian` functions (e.g. Calculus and ForwardDiff)

src/statmodels.jl
```diff
@@ -39,6 +39,35 @@ the likelihood of the model.
...
+Returns the diagonal of the projection matrix.
+"""
+leverage(obj::StatisticalModel) = error("leverage is not defined for $(typeof(obj)).")
+
+"""
+    linear(obj::StatisticalModel)
+
+Return Boolean indicator whether model is linear.
+"""
+linear(obj::StatisticalModel) = error("linear is not defined for $(typeof(obj)).")
```
simonbyrne: `islinear`?

src/statmodels.jl
```diff
@@ -79,20 +108,48 @@ when applicable the intercept and the distribution's dispersion parameter.
...
 Return the standard errors for the coefficients of the model.
 """
 stderr(obj::StatisticalModel) = sqrt.(diag(vcov(obj)))
 
+"""
+    tss(obj::StatisticalModel)
+
+Return the total sum of squares.
+"""
+tss(obj::StatisticalModel) = error("tss is not defined for $(typeof(obj)).")
```
simonbyrne: Isn't this equivalent to `deviance`?

src/statmodels.jl
Nosferican: I originally had `islinear`, but don't have a preference between the two. We can go with `islinear`.

src/statmodels.jl
Nosferican: The deviance of some models use `tss` in the calculation. For example,
```
struct OLS <: RegressionModel end
deviance(model::OLS) = tss(model) / dof_residual(model)
```

src/statmodels.jl
Nosferican:
```
weights(obj::StatisticalModel)
```
seem a better choice.

src/statmodels.jl
Nosferican: @nalimilan, could you give me a quick example to adopt it to using generators in this case?

Nosferican: @simonbyrne Do you have a recommendation for a name with less ambiguity just in case?

src/statmodels.jl
Nosferican: I was thinking of GLM frameworks where the `link` might require the mean response. I also didn't want to use the fallback due to potentially using a wrong value for models with weights.

src/statmodels.jl
Nosferican: This is true. This might be a case where it is justified to use `model_hessian`? A big part would be to provide a nice way for statistical models to play well with optimization packages.

nalimilan: Is that really a problem? As you noted, several packages already use the same name and that's not the end of the world. It's a case where automatic merging of methods would make sense.

src/statmodels.jl
nalimilan: Good point about weights. What do you mean exactly about the link?

src/statmodels.jl
simonbyrne: Perhaps don't export it then?

src/statmodels.jl
nalimilan: Something like `m = meanresponse(obj); sum(abs2(x - m) for x in fitted(obj))`.

Maybe we could use `mss` for "[Model sum of squares](https://en.wikipedia.org/wiki/Explained_sum_of_squares)"?

src/statmodels.jl
nalimilan: Mmm, that's not what GLM does:
https://github.com/JuliaStats/GLM.jl/blob/287c668acd592c05ec90140dfa70a2887045b4f3/src/lm.jl#L46-L61

src/statmodels.jl
nalimilan: It's not a problem per se to export functions with conflicting names, the error only happens if somebody tries to call it while conflicting packages are loaded. If we don't export it from any of the potentially conflicting packages, we're end up in a strictly worse situation where you always need to fully qualify the function.

src/statmodels.jl
Nosferican: I stand corrected. The `deviance` for non-linear models is still computed the same way (by the generalization). Since they are equivalent as far as I believe now, we can just keep `deviance`.

src/statmodels.jl
Nosferican: Calculus exports `hessian` and ForwardDiff doesn't. I think it would be best at least for now to go with `model_hessian` to avoid any potential conflicting namespace with other optimization packages.

src/statmodels.jl
Nosferican: For example, `GLM.linkfun(L::Link, μ)`.

src/statmodels.jl
nalimilan: If we're trying to get rid of these `model_*` functions, better not introduce new ones now.

src/statmodels.jl
nalimilan: But in what context do you need to call it? Couldn't we provide an API to call the link on the mean response instead?

src/statmodels.jl
Nosferican: I think the idea was to remove it from function that do not need it (e.g., `model_response` to `response` or `model_weights` to `weights`). For functions that justify it, it should be fine. For example, `modelmatrix` is justified as `matrix` would be bad. I would have liked `hessian`, but sadly it is expected to be used in optimization packages and these not all require `StatsBase`.

As an aside, should `weights(obj::AbstractVector{T}) where T <: Real` be depreciated to just `Weights`? It seems like the Julian thing would be to keep constructors with uppercase. `pweights`, `aweights`, and the others seem just like short-hand for long to type structs.

nalimilan: Yeah, but "model matrix" is a standard term in the field, probably because "matrix" is indeed too broad. "Model hessian" doesn't exist AFAIK.

Regarding `weights`, that pattern of lower case convenience constructors is used elsewhere in Julia, for example in Base.Iterators. Anyway, better discuss that elsewhere.

src/statmodels.jl
Nosferican: There was a request to include a `link` argument to `response` from the people at GLM. However, I think that requires that the `link` code be ported to a small package or StatsBase. Other instances of when one might want the `meanresponse` include estimation for intercept term, and `mss`.

src/statmodels.jl
Nosferican: What about skipping the `hessian` and implementing `eim` `oim` for the Expected Information Matrix and Observed Information Matrix? These are specific to statistical models (components for the `vcov`).

src/statmodels.jl
```diff
@@ -79,6 +101,37 @@ when applicable the intercept and the distribution's dispersion parameter.
 """
 dof(obj::StatisticalModel) = error("dof is not defined for $(typeof(obj)).")
 
+"""
+    mss(obj::StatisticalModel)
+
+Return the model sum of squares.
+"""
+function mss(obj::StatisticalModel)
+    μ = meanresponse(obj)
```
nalimilan: Should be `m`. (We really need to add tests in StatsBase for these methods at some point...)

Also, use the same approach for other `sum` calls.

src/statmodels.jl
nalimilan: Not sure. I kind of like the idea of using the same name `hessian` across packages given that it has a consistent meaning. We can avoid exporting it if people are concerned about that.

What do other software (languages, packages...) do?

Cc: @dmbates

src/statmodels.jl
```diff
@@ -137,15 +204,25 @@ the likelihood of the model,  ``k`` its number of consumed degrees of freedom
...
+"""
+function r2(obj::StatisticalModel)
+    mss(obj) / deviance(obj)
+end
+
+"""
+    r2(obj::StatisticalModel, variant::Symbol)
+    r²(obj::StatisticalModel, variant::Symbol)
+
+Coefficient of determination (R-squared).
```
nalimilan: Would better be "Pseudo-coefficient of determination".

src/statmodels.jl
nalimilan: Though if we add `mss` and `rss`, it would make sense to have  `tss` too, and let linear models implement it by calling `deviance` (or vice-versa).

src/statmodels.jl
```diff
@@ -93,6 +146,20 @@ Return the variance-covariance matrix for the coefficients of the model.
...
+    weights(obj::StatisticalModel)
+
+Return the weights used in the model.
+"""
+weights(obj::StatisticalModel) = error("weights is not defined for $(typeof(obj)).")
+
+"""
+    isfitted(obj::StatisticalModel)
+
+Return Boolean indicator of whether the model has been fitted.
```
nalimilan: Maybe just "Indicate whether the model has been fitted."

Please check the other docstrings, as some say "the"/"a", and some have no determinant. Some also repeat "the" twice.

src/statmodels.jl
```diff
@@ -174,16 +245,27 @@ end
...
+    adjr²(obj::StatisticalModel)
 
 Adjusted coefficient of determination (adjusted R-squared).
 
 For linear models, the adjusted R² is defined as ``1 - (1 - (1-R^2)(n-1)/(n-p))``, with ``R^2``
 the coefficient of determination, ``n`` the number of observations, and ``p`` the number of
 coefficients (including the intercept). This definition is generally known as the Wherry Formula I.
+"""
+function adjr2(obj::StatisticalModel)
+    n, p = nobs(obj), length(coef(obj)) # Default behavior (p could potentially be incorrect)
```
nalimilan: Mmm, indeed it could be incorrect if some coefficients could not be estimated (and would be `NaN`, `0` or anything depending on the implementation). Probably better use `dof` here? Though there's a problem since it includes the dispersion parameter and we have no way to know that here. I see two possible solutions:
- do not define a fallback for now
- add `dof(obj::StatisticalModel, coef_only::Bool = false)` which custom types have to implement, counting only coefficients per se when `coef_only=true`

src/StatsBase.jl
```diff
@@ -163,22 +163,32 @@ module StatsBase
...
+    score,
     stderr,
     vcov,
     predict,
     predict!,
     residuals,
     r2,
     r²,
-    model_response,
+    # weights is also used in statsmodels with a different method implemented
```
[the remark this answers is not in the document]
Nosferican: Sorry about the confusion. Alright.

src/statmodels.jl
```diff
@@ -39,6 +39,20 @@ the likelihood of the model.
 """
 deviance(obj::StatisticalModel) = error("deviance is not defined for $(typeof(obj)).")
 
+"""
+    leverage(obj::StatisticalModel)
+
+Returns the diagonal of the projection matrix.
```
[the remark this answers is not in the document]
Nosferican: I am not sure, but since `modelmatrix` is defined for `RegressionModel ` I would think that leverage should be defined at that level too.

src/statmodels.jl
simonbyrne: I prefer slightly longer names where possible, so how about `fisherinf` and `observedinf`?

docs/src/statmodels.md
```diff
@@ -15,22 +15,32 @@ coeftable
 confint
 deviance
 dof
+fisherinf
```
ararslan: The suffix "inf" here is confusing. I assume you mean Fisher information, but "inf" makes it sound like a reference to `Inf`.

Nosferican: What about `fisherim` as in information matrix or `fisherinformation`? I believe these methods will mostly be used by other packages and not user so we could have longer than usual names.

nalimilan: Is there any precedent in other software? I could only find mentions of fields or functions called `hessian`.

Nosferican: Stata has uses it as options for variance covariance estimations. It uses `eim` for expected and `oim` for the observed. Typically the `score` and `fisherinf` are used to compute the variance covariance which is what is typically returned. I would be fine not exporting those too to encourage their use through other packages that compute the variance covariance.

nalimilan: OK. The situation is a bit different in Stata since it's an option, so short names do not polluted the namespace.

`fisherinformation` isn't too bad, but `observedfisherinformation` would be really too long. Yet having `fisherinformation` and `observedinformation` would be weird too, since both refer to Fisher. Maybe `informationmatrix(obj, kind::Symbol)` with `kind` being either `:observed` or `:expected`?

Nosferican: I believe Fisher is the expected of the observed information matrix. However, I think having these and selecting on kind is preferred. How about,
```
informationmatrix(model::StatisticalModel; expected::Bool=true)
```

nalimilan: Yeah, that's another possibility. Not sure the argument should be called `expected` or `observed` (disregarding the default, which I agree should be expected).

Nosferican: How about `fisher::Bool=true` as Fisher indicates the expected rather than the observed?

ararslan: I think `expected` is a better name for the argument. We can document that `expected=true` indicates that the result is the Fisher information.

Nosferican: Latest version reflects those changes.

Nosferican:
Thanks to all the reviewers!

Nosferican:
Could this be merged? It will make it easier to start porting some code.

PR #362, opened 2018-03-26
Compatibility with v0.7.0

dmbates:
- Use `Compat.sum`, `Compat.mean`, etc. with named `dims` argument
- Replace `uninitialized` with `undef` (which is exported from `Compat`)
- Add `using Compat.LinearAlgebra`, `Compat.Random`, etc. where needed
- **Comment out 4 tests in `test/ranking.jl`**. I don't know how to fix these.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -24,11 +24,11 @@ const RealFP = Union{Float32, Float64}
 
 fptype(::Type{T}) where {T<:Union{Float32,Bool,Int8,UInt8,Int16,UInt16}} = Float32
 fptype(::Type{T}) where {T<:Union{Float64,Int32,UInt32,Int64,UInt64,Int128,UInt128}} = Float64
-fptype(::Type{Complex64}) = Complex64
-fptype(::Type{Complex128}) = Complex128
+fptype(::Type{Compat.ComplexF32}) = Compat.ComplexF32
+fptype(::Type{Compat.ComplexF64}) = Compat.ComplexF64
 
 # A convenient typealias for deprecating default corrected Bool
-const DepBool = Union{Bool, Void}
+const DepBool = Union{Bool, Compat.Nothing}
 
 function depcheck(fname::Symbol, b::DepBool)
     if b == nothing
--- a/src/common.jl
+++ b/src/common.jl
@@ -24,11 +24,11 @@ const RealFP = Union{Float32, Float64}
 
 fptype(::Type{T}) where {T<:Union{Float32,Bool,Int8,UInt8,Int16,UInt16}} = Float32
 fptype(::Type{T}) where {T<:Union{Float64,Int32,UInt32,Int64,UInt64,Int128,UInt128}} = Float64
-fptype(::Type{Compat.ComplexF32}) = Compat.ComplexF32
-fptype(::Type{Compat.ComplexF64}) = Compat.ComplexF64
+fptype(::Type{Complex{Float32}}) = Complex{Float32}
+fptype(::Type{Complex{Float64}}) = Complex{Float64}
 
 # A convenient typealias for deprecating default corrected Bool
-const DepBool = Union{Bool, Compat.Nothing}
+const DepBool = Union{Bool, Nothing}
 
 function depcheck(fname::Symbol, b::DepBool)
     if b == nothing
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -533,4 +533,8 @@ Calculate the midpoints (pairwise mean of consecutive elements).
 """
 midpoints(v::AbstractVector) = [middle(v[i - 1], v[i]) for i in 2:length(v)]
 
-midpoints(r::AbstractRange) = broadcast(+, r[1:(end - 1)], step(r) / 2)
+if VERSION < v"0.7.0-DEV.4713"
+    midpoints(r::AbstractRange) = r[1:(end - 1)] + step(r) / 2
+else
+    midpoints(r::AbstractRange) = broadcast(+, r[1:(end - 1)], step(r) / 2)
+end
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -533,4 +533,4 @@ Calculate the midpoints (pairwise mean of consecutive elements).
 """
 midpoints(v::AbstractVector) = [middle(v[i - 1], v[i]) for i in 2:length(v)]
 
-midpoints(r::AbstractRange) = r[1:(end - 1)] .+ step(r) / 2
+midpoints(r::AbstractRange) = broadcast(+, r[1:(end - 1)], step(r) / 2)
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -1,6 +1,7 @@
 using Base.Cartesian
 
-import Base: show, ==, push!, append!, float, norm, normalize, normalize!
+import Base: show, ==, push!, append!, float
+import Compat.LinearAlgebra: norm, normalize, normalize!
 
 # Mechanism for temporary deprecation of default for "closed" (because default
 # value has changed). After deprecation is lifed, remove "_check_closed_arg"
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -533,4 +533,4 @@ Calculate the midpoints (pairwise mean of consecutive elements).
 """
 midpoints(v::AbstractVector) = [middle(v[i - 1], v[i]) for i in 2:length(v)]
 
-midpoints(r::Compat.AbstractRange) = r[1:(end - 1)] .+ step(r) / 2
+midpoints(r::AbstractRange) = r[1:(end - 1)] .+ step(r) / 2
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -533,4 +533,4 @@ Calculate the midpoints (pairwise mean of consecutive elements).
 """
 midpoints(v::AbstractVector) = [middle(v[i - 1], v[i]) for i in 2:length(v)]
 
-midpoints(r::Range) = r[1:(end - 1)] + step(r) / 2
+midpoints(r::Compat.AbstractRange) = r[1:(end - 1)] .+ step(r) / 2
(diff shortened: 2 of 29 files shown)


src/common.jl
```diff
@@ -24,11 +24,11 @@ const RealFP = Union{Float32, Float64}
 
 fptype(::Type{T}) where {T<:Union{Float32,Bool,Int8,UInt8,Int16,UInt16}} = Float32
 fptype(::Type{T}) where {T<:Union{Float64,Int32,UInt32,Int64,UInt64,Int128,UInt128}} = Float64
-fptype(::Type{Complex64}) = Complex64
-fptype(::Type{Complex128}) = Complex128
+fptype(::Type{Compat.ComplexF32}) = Compat.ComplexF32
```
ararslan: The `Compat.` prefix isn't needed for these (or `Nothing` below), since they're exported from Compat. Also, just personal preference, but I think it might be clearer to spell out the `Complex` type explicitly, e.g. `::Type{Complex{Float32}}`, which will work across versions.

src/misc.jl
```diff
@@ -1,16 +1,14 @@
-# Miscelleneous stuff
+# Miscellaneous stuff
 
 # test whether a contains no repeated elements
 function norepeat(a::AbstractArray)
```
ararslan: We should just deprecate this to `allunique`, which is in Base as of Julia 0.6.

src/robust.jl
```diff
@@ -44,8 +44,8 @@ function trim!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
         0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
     end
 
-    select!(x, (n-count+1):n)
-    select!(x, 1:count)
+    Compat.partialsort!(x, (n-count+1):n)
```
ararslan: `Compat.` prefix also isn't needed here

test/hist.jl
```diff
@@ -126,10 +121,10 @@ end
 @testset "Histogram show" begin
     # hist show
     show_h = sprint(show, fit(Histogram,[0,1,2], closed=:left))  # FIXME: closed
-    @test contains(show_h, "edges:\n  0.0:1.0:3.0")
-    @test contains(show_h, "weights: $([1,1,1])")
-    @test contains(show_h, "closed: left")
-    @test contains(show_h, "isdensity: false")
+    @test Compat.occursin("edges:\n  0.0:1.0:3.0", show_h)
```
ararslan: Can drop the `Compat.` here too

test/ranking.jl
```diff
@@ -11,31 +11,31 @@ s = ["c", "a", "b", "d", "d", "b", "e", "d"] # s is a vector of strings ordered
 @test ordinalrank(a) == [1, 2, 3, 4, 5, 6, 7, 8]
 @test ordinalrank(x) == [4, 1, 2, 5, 6, 3, 8, 7]
 @test isequal(ordinalrank(xm), [4, 1, 2, 5, 6, 3, 8, 7, missing])
-@test isequal(ordinalrank([missing, missing]), [missing, missing])
+#@test isequal(ordinalrank([missing, missing]), [missing, missing])
```
ararslan: What's the issue with this test?

test/hist.jl
```diff
@@ -228,7 +223,7 @@ end
 
 @testset "midpoints" begin
     @test StatsBase.midpoints([1, 2, 4]) == [1.5, 3.0]
-    @test StatsBase.midpoints(linspace(0, 1, 5)) == 0.125:0.25:0.875
+    @test StatsBase.midpoints(Compat.range(0, stop = 1, length = 5)) == collect(0.125:0.25:0.875)
```
nalimilan: Does this means we no longer return a range? Sounds like a regression.

test/hist.jl
nalimilan: There are a few other places where the prefix isn't needed either. And sometimes `using Compat: ...` would make sense instead of repeating `Compat.`.

src/common.jl
dmbates: It might be simpler to define 
```
fptype(::Type{T}) where T = float(T)
fptype(::Type{T}) where {T<:Union{Bool,Int8,UInt8,Int16,UInt16}} = Float32
```

test/hist.jl
dmbates: Thanks for catching that.  Reverted in https://github.com/JuliaStats/StatsBase.jl/pull/362/commits/d5d5436e92d31c3f6344b7be1218bbb472a10c19

src/hist.jl
```diff
@@ -532,4 +533,4 @@ Calculate the midpoints (pairwise mean of consecutive elements).
 """
 midpoints(v::AbstractVector) = [middle(v[i - 1], v[i]) for i in 2:length(v)]
 
-midpoints(r::Range) = r[1:(end - 1)] + step(r) / 2
+midpoints(r::AbstractRange) = broadcast(+, r[1:(end - 1)], step(r) / 2)
```
nalimilan: Doesn't `.+` give the same result? Or is there a precedence issue?

More importantly, it looks like a bug on 0.6 makes `broadcast` return a vector rather than a range. The simplest solution is to use `if VERSION`, keeping `+` on 0.6. Femtocleaner will remove it automatically when we require 0.7.

test/weights.jl
```diff
@@ -85,8 +84,8 @@ w2 = rand(8)
 @test size(wsum(x, w1, 1)) == (1, 8)
 @test size(wsum(x, w2, 2)) == (6, 1)
 
-@test wsum(x, w1, 1) ≈ sum(x .* w1, 1)
-@test wsum(x, w2, 2) ≈ sum(x .* w2', 2)
+@test wsum(x, w1, 1) ≈ Compat.sum(x .* w1, dims = 1)
```
nalimilan: Would `using Compat: sum` allow not repeating the prefix?

ararslan: It would have to be `import Compat: sum` in order to shadow the symbol from Base.

nalimilan: Are you sure? `using` seems to be enough in my tests.

ararslan: Nope, I'm not sure. :) The behavior probably changed in 0.7 when Jeff made packages no longer load into Main by default.

src/hist.jl
dmbates: Okay.  Any hints on how I would find the minimum version number in the 0.7.0-DEV series on which to base the conditional?  That is, what change should I be looking for?

nalimilan: To be safe, you can just take the commit you are using. That way on older 0.7 versions the deprecation warning will be printed but tests will pass.

dmbates:
As I see it, the tasks that still need to be done are to

- [ ] Fix the tests that have been commented out in `test/ranking.jl`. E.g.
```
#@test isequal(ordinalrank([missing, missing]), [missing, missing])
```

- [ ] Replace calls to `scale!` by `rmul!` for v0.7 only.  I don't think this can be done with `Compat`.  I think it will be necessary to write a placeholder that calls `scale!` for v0.6 and `rmul!` for v0.7.  I would be happy to be told that I am wrong, however.

- [ ] Settle the issue of whether it is possible to have `Compat.sum` be the generic for all `sum` methods.

- [ ] In testing I have found that there are occasional failures from attempting to evaluate the square root of a negative number in one particular test

```
Test Summary:             | Pass  Error  Total
StatsBase.Covariance      |  292      1    293
  StatsBase.weights       |   56            56
  StatsBase.aweights      |   84            84
  StatsBase.fweights      |   68      1     69
    Scattermat            |   16            16
    Uncorrected           |   28            28
    Corrected             |   20      1     21
      Weighted Covariance |    8             8
      Mean and covariance |   12            12
      Conversions         |           1      1
    Correlation           |    4             4
  StatsBase.pweights      |   84            84
```
It looks to be in tests of `cov2cor` but, as a Heisenbug, it is difficult to check it.

By the way, does anyone understand why `cov2cor` needs two arguments?  It should only need the covariance matrix.

src/hist.jl
```diff
@@ -533,4 +533,8 @@ Calculate the midpoints (pairwise mean of consecutive elements).
 """
 midpoints(v::AbstractVector) = [middle(v[i - 1], v[i]) for i in 2:length(v)]
 
-midpoints(r::AbstractRange) = broadcast(+, r[1:(end - 1)], step(r) / 2)
+if VERSION < v"0.7.0-DEV.4713"
+    midpoints(r::AbstractRange) = r[1:(end - 1)] + step(r) / 2
+else
+    midpoints(r::AbstractRange) = broadcast(+, r[1:(end - 1)], step(r) / 2)
```
nalimilan: Better simply use `.+` here.

dmbates:
@ararslan and @nalimilan : Regarding replacing the calls to `scale!` with `rmul!`, the only way I can think of doing this is to define a `myscale!` function with an `if VERSION < v"0.7.0-DEV.xxx"` wrapper.  Is there a better way?  I don't think that `Compat` provides any facilities for this.

nalimilan:
I don't really know, but `if VERSION ` isn't the end of the world. But better do that inline rather than creating a wrapper function, so that the code is completely clean once Femtocleaner removes the check.

src/moments.jl
```diff
@@ -54,7 +54,7 @@ end
 function Base.varm!(R::AbstractArray, A::RealArray, w::AbstractWeights, M::RealArray,
                     dim::Int; corrected::DepBool=nothing)
     corrected = depcheck(:varm!, corrected)
-    scale!(_wsum_centralize!(R, abs2, A, values(w), M, dim, true),
+    myscale!(_wsum_centralize!(R, abs2, A, values(w), M, dim, true),
            varcorrection(w, corrected))
```
nalimilan: Need to adjust indentation.

nalimilan:
Should we merge this and investigate the random failure afterwards?

dmbates:
@nalimilan Yes, please do merge.

nalimilan:
I have no idea why the random error only appears on 0.7, but I've found the culprit. See https://github.com/JuliaStats/StatsBase.jl/pull/365.

PR #364, opened 2018-04-05
Fix countmap for views

### LINKED ISSUE #363: Default `countmap` broken for `SubArray`
looks like #319 broke `countmap` for array views (unless `alg = :dict` is explicitly specified).

```julia
julia> using StatsBase

julia> X = view([1,1,1,2,2], 1:5)
5-element SubArray{Int64,1,Array{Int64,1},Tuple{UnitRange{Int64}},true}:
 1
 1
 1
 2
 2

julia> StatsBase.countmap(X)
ERROR: TypeError: addcounts_radixsort!: in typeassert, expected SubArray{Int64,1,Array{Int64,1},Tuple{UnitRange{Int64}},true}, got Array{Int64,1}
```

andreasnoack:
Fixes #363

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -323,16 +323,13 @@ const BaseRadixSortSafeTypes = Union{Int8, Int16, Int32, Int64, Int128,
 radixsort_safe(::Type{T}) where {T<:BaseRadixSortSafeTypes} = true
 radixsort_safe(::Type) = false
 
-function addcounts_radixsort!(cm::Dict{T}, x::AbstractArray{T}) where T
-    # sort the x using radixsort
-    sx = sort(x, alg = RadixSort)::typeof(x)
-
+function _addcounts_radix_sort_loop!(cm::Dict{T}, sx::AbstractArray{T}) where T
     tmpcount = 1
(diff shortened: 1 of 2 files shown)


src/counts.jl
```diff
@@ -323,16 +323,13 @@ const BaseRadixSortSafeTypes = Union{Int8, Int16, Int32, Int64, Int128,
 radixsort_safe(::Type{T}) where {T<:BaseRadixSortSafeTypes} = true
 radixsort_safe(::Type) = false
 
-function addcounts_radixsort!(cm::Dict{T}, x::AbstractArray{T}) where T
-    # sort the x using radixsort
-    sx = sort(x, alg = RadixSort)::typeof(x)
-
+function _addcounts_radix_sort_loop!(cm, sx)
```
nalimilan: Probably better keep the type signatures, it makes it easier to read the code.

It would also be useful to have a comment to explain that the separate function is needed because `sort` isn't inferred (is it still on 0.7?).

andreasnoack: Done. It looks like `sort` is inferred in Julia 0.7.

Issue #366, opened 2018-04-06
Issues with mad!


Nosferican:
I took a look at the code for `mad!` since it is the only piece of code that requires `SpecialFunctions` which introduces `BinDeps` as a dependency. The code used from SpecialFunctions is one line
```
1 / (-sqrt(2 * one(T)) * erfcinv(3 * one(T) / 2))
```
where `T <: Real`. However, `erfcinv` only supports `erfcinv(y::Float64)`, `erfcinv(y::Float32)`, and `erfcinv(x::Integer) = erfcinv(float(x))`. Using `T <: Real` can result in error if `T <: Union{Rational, Irrational, BigFloat, Float16}`. This suggest that we can just hardcode the value as `1.4826022185056018::Float64` and drop those two dependencies.

In regards to the `mad!` function, accepting `T<:Integer` also has issues due to the in-place nature,
```
mad!([1,2,3,4,5,6,8,9])
```
will result in `InexactError()`.

Possible solutions,

1. Strict `!` replaces
 
```
mad!(v::AbstractArray{T};
     center::Real=median!(v),
     normalize::Union{Bool,Nothing}=true,
     constant=nothing) where T<:Union{Float64,Float32}`
```
and make the adjustment either
```
     1.4826022185056018 * median!(v)
```
or
```
    convert(T, 1.4826022185056018) * median!(v)
```

2. May `!` replace

```
mad!(v::AbstractArray{T};
     center::Real=median!(v),
     normalize::Union{Bool,Nothing}=true,
     constant=nothing) where T<:Real`
```
and use `float.(v)` for a special dispatch on `T<:Integer`. Same considerations for hardcoded value apply, but using the updated type.

nalimilan:
Good catch. I think the simplest solution would be to call ` m = median!(v)` first, and then return `oftype(m, 1.4826022185056018) * m`. If we want to be extra careful, we could check that `precision(m)` isn't higher than `precision(Float64)`, in which case the result wouldn't be as precise as it should and better throw an error (what happens currently anyway).

We could also hardcode more digits so that it also works with `BigFloat` (at least with the default precision), but that's strictly an improvement compared to the current code.

Regarding the in-place failure with `Integer`, it's not completely absurd given that there's no way to compute that in-place. I'm not sure whether it's a good idea to allocate a new vector when needed. On the one hand it would allow writing generic code which is efficient when possible, but on the other it would make a potentially large allocation silently.

[...]

PR #367, opened 2018-04-07
Special functions (`mad!`)

Nosferican:
This addresses #366 
The changes seem to work just fine when I checked them by itself, but when I tried calling them from the package I kept getting segmentation error 11. Might be a local issue so gonna see what Travis says.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -5,8 +5,6 @@ module StatsBase
     import Base.Cartesian: @nloops, @nref, @nextract
     import DataStructures: heapify!, heappop!, percolate_down!
 
-    import SpecialFunctions: erfcinv
-
     using Compat, SortingAlgorithms, Missings
     using Compat.LinearAlgebra
     using Compat.Random
@@ -14,6 +12,10 @@ module StatsBase
     using Compat.SparseArrays
     import Compat.Random: rand, rand!
     import Compat.LinearAlgebra: BlasReal, BlasFloat
+
+    ## Constants
+    ## inv(sqrt(2) * erfinv(.5)) From https://www.wolframalpha.com/input/?i=1+%2F+(sqrt(2)+*+erfinv(.5))
+    const k = BigFloat("1.482602218505601860547076529360423431326703202590312896536266275245674447622701")
 
     ## tackle compatibility issues
 
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -2,18 +2,18 @@ __precompile__()
 
 module StatsBase
     import Base: length, isempty, eltype, values, sum, mean, mean!, show, quantile
-    import Base: rand, rand!
-    import Base.LinAlg: BlasReal, BlasFloat
     import Base.Cartesian: @nloops, @nref, @nextract
     import DataStructures: heapify!, heappop!, percolate_down!
 
     import SpecialFunctions: erfcinv
 
     using Compat, SortingAlgorithms, Missings
-
-    if VERSION >= v"0.7.0-DEV.3052"
-        using Printf
-    end
+    using Compat.LinearAlgebra
+    using Compat.Random
+    using Compat.Printf
+    using Compat.SparseArrays
+    import Compat.Random: rand, rand!
+    import Compat.LinearAlgebra: BlasReal, BlasFloat
 
     ## tackle compatibility issues
 
@@ -186,6 +186,11 @@ module StatsBase
     export midpoints
 end
 
+if VERSION < v"0.7.0-DEV.3665"
+    myscale!(A::AbstractArray, b::Number) = scale!(A, b)
+else
+    myscale!(A::AbstractArray, b::Number) = rmul!(A, b)
+end
     # source files
 
     include("common.jl")
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -3,6 +3,7 @@ __precompile__()
 module StatsBase
     import Base: length, isempty, eltype, values, sum, mean, mean!, show, quantile
     import Base.Cartesian: @nloops, @nref, @nextract
+    using Base: @irrational
     import DataStructures: heapify!, heappop!, percolate_down!
 
     using Compat, SortingAlgorithms, Missings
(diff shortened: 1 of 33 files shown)


src/StatsBase.jl
```diff
@@ -15,6 +13,10 @@ module StatsBase
     import Compat.Random: rand, rand!
     import Compat.LinearAlgebra: BlasReal, BlasFloat
 
+    ## Constants
+    ## inv(sqrt(2) * erfinv(.5)) From https://www.wolframalpha.com/input/?i=1+%2F+(sqrt(2)+*+erfinv(.5))
+    const k = parse(BigFloat,"1.482602218505601860547076529360423431326703202590312896536266275245674447622701")
```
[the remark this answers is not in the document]
Nosferican: Apparently for some people `parse(BigFloat, x)` wasn't giving them the segmentation error so I try it, but didn't seem to work now.

src/scalarstats.jl
```diff
@@ -249,7 +249,7 @@ of the standard deviation under the assumption that the data is normally distrib
 """
 function mad(v::AbstractArray{T};
              center::Union{Real,Nothing}=nothing,
-             normalize::Union{Bool, Nothing}=nothing) where T<:Real
+             normalize::Union{Bool, Nothing}=nothing) where T<:Union{AbstractFloat,Integer}
```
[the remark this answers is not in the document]
Nosferican: `AbstractIrrational` and `Rational` just error out. I can I try to see if we can manage `Rational`.

src/scalarstats.jl
```diff
@@ -277,22 +277,23 @@ of the standard deviation under the assumption that the data is normally distrib
 function mad!(v::AbstractArray{T};
               center::Real=median!(v),
               normalize::Union{Bool,Nothing}=true,
-              constant=nothing) where T<:Real
-    for i in 1:length(v)
+              constant=nothing) where T<:Union{AbstractFloat,Integer}
+    isempty(v) && throw(ArgumentError("mad is not defined for empty arrays"))
+    for i ∈ eachindex(v)
```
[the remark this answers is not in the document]
Nosferican: Will do.

src/scalarstats.jl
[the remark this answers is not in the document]
Nosferican: It will error out with `Rational` due to
```
oftype(m, BigFloat("1.482602218505601860547076529360423431326703202590312896536266275245674447622701")
```
throwing `InexactError`... We can either put a catch and ask them to not use `normalize=true` with `T>:Rational` or use an approximation for `k`, and give a warning that they are losing precision in the process.

Should I verify the precision of `BigFloat` and give a warning if the precision exceeds the 256 hardcoded value?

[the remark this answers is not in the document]
Nosferican: The `Rational` value for would be `6677046798800499//4503599627370496`. 
For `Irrational`, this works `mad!([ℯ, π])`, but `mad!([π])` or `mad!([ℯ])` won't as the single element `Vector` are `Array{Irrational{:π},1}` rather than `Vector{Float64}`. Could make it work by doing the `oftype` when `T <: Irrational` go to `Float64`.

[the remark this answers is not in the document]
nalimilan: `Rational{T}` can use any precision depending on `T`, so there's no absolutely correct value. I'd just leave it alone (and `Irrational` too), but still use `Real`.

src/scalarstats.jl
```diff
@@ -274,25 +274,25 @@ If `normalize` is set to `true`, the MAD is multiplied by
...
     end
-    if constant !== nothing
+    if !isa(constant, Nothing)
         Base.depwarn("keyword argument `constant` is deprecated, use `normalize` instead or apply the multiplication directly", :mad)
-        constant * median!(v)
+        m * constant
     elseif normalize
-        k * median!(v)
+        ## inv(sqrt(2) * erfinv(.5)) From https://www.wolframalpha.com/input/?i=1+%2F+(sqrt(2)+*+erfinv(.5))
+        m * oftype(m, BigFloat("1.482602218505601860547076529360423431326703202590312896536266275245674447622701"))
```
nalimilan: Actually for performance we should probably define a custom `Irrational` with special `convert` methods for `Float64` and `Float32`. Or maybe just define a function taking `m` and returning the ajusted value, with a general `BigFloat` method and two optimized methods for `Float64` and `Float32`.

nalimilan: No, I meant we should define an `Irrational` for the constant `k`.

nalimilan: I am *not* speaking about `Vector{<:Irrational}`. I'm just saying that `oftype(m, BigFloat("1.48260..."))` isn't very efficient, and that using a custom `Irrational` with specialized `convert` methods could be a better approach. That's how this kind of situation is handled in Base.

Nosferican: Got it. I couldn't find in Base how exactly to make an irrational number...
I did a few attempts at `@irrational(sym, val, def)`, but could use some guidance there.

nalimilan: I guess you'd need to define methods for an `Irrational` with a particular symbol as type parameter. But given that it's only used internally it's probably simpler not to use `Irrational` and just define a function taking a type and returning the value converted to that type, with specialized methods.

Nosferican: The `Irrational` way actually works quite well. I learned how to do implement it from the StatsFuns code. If it compiles and all we can go this way and see if we can use it for the depreciated code as well.

nalimilan: Cool! I don't think we should use it for the deprecation, since people should just omit the argument if they want to use the constant.

It would be clearer to use a longer name than `k` though.

Nosferican: How about `MAD_Normal_Scale_Factor`?

nalimilan: Usually Julia objects use lowercase names; also no need to be too long either. So maybe `mad_constant`?

Nosferican: The value would be a `DataType` which usually has upper case letters except that for `T<:Irrational` it seems the convention is to use lowercase as values. Sounds good.

PR #368, opened 2018-04-08
Change identifer of stderr to se

yukota:
Identifer confilict between meaning of IO and statical.
https://github.com/JuliaLang/julia/pull/26636

### DIFF
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -20,6 +20,15 @@ import Base.varm, Base.stdm
 @deprecate AICc(obj::StatisticalModel) aicc(obj)
 @deprecate BIC(obj::StatisticalModel) bic(obj)
 
+if !isdefined(Base, :stderr)
+    @deprecate stderr(obj::StatisticalModel) stderror(obj)
+else
+    function (io::IO)(obj::StatisticalModel)
+        Base.depwarn("stderr(obj::StatisticalModel) is deprecated, use stderror(obj) instead", :stderr)
+            io === stderr ? stderror(obj) : throw(MethodErrror(io, (obj,)))
+    end
+end
+
 @deprecate R2(obj::StatisticalModel, variant::Symbol) r2(obj, variant)
 @deprecate R²(obj::StatisticalModel, variant::Symbol) r²(obj, variant)
 @deprecate adjR2(obj::StatisticalModel, variant::Symbol) adjr2(obj, variant)
@@ -37,7 +46,6 @@ function findat!(r::IntegerArray, a::AbstractArray{T}, b::AbstractArray{T}) wher
     end
     return r
 end
-
 
 """
     findat(a, b)
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -23,9 +23,9 @@ import Base.varm, Base.stdm
 if !isdefined(Base, :stderr)
     @deprecate stderr(obj::StatisticalModel) stderror(obj)
 else
-    function (io::IO)(obj::StatisticalModel)
+    function (io::typeof(stderr))(obj::StatisticalModel)
         Base.depwarn("stderr(obj::StatisticalModel) is deprecated, use stderror(obj) instead", :stderr)
-            io === stderr ? stderror(obj) : throw(MethodErrror(io, (obj,)))
+        io === stderr ? stderror(obj) : throw(MethodErrror(io, (obj,)))
     end
 end
 
--- a/docs/src/statmodels.md
+++ b/docs/src/statmodels.md
@@ -27,7 +27,7 @@ nulldeviance
 r2
 rss
 score
-stderr
+stderror
 vcov
 weights(::StatisticalModel)
 ```
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -180,7 +180,6 @@ module StatsBase
     nullloglikelihood,
     rss,
     score,
-    stderr,
     stderror,
     vcov,
     predict,
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -181,6 +181,7 @@ module StatsBase
     rss,
     score,
     stderr,
+    stderror,
     vcov,
     predict,
     predict!,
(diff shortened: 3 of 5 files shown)


Nosferican:
Seems good.

kmsquire:
As suggested [here](https://github.com/JuliaLang/julia/pull/26636#comment-376861507), how about `stderror` instead, to make it easier for new users (and old) to understand the meaning at a glance?

Nosferican:
The Slack channel favored `se` over `stderror` as it was deemed weird to abbreviate the first part and not the second one. `se` was also believed to be quite common and understood in the Statistical Model domain. The reason for not going too short (e.g., two letters symbol) was mostly to avoid another `df` for `dof` which conflicted with a common variable name for `DataFrame` structs. However, that does not seem to be the case with `se`.

ararslan:
FWIW I don't think we should use `se` here, especially since our standard deviation function isn't `sd`.

Nosferican:
I totally agree about Slack not being the best for these kinds of decisions. The best place probably (would have/is) an issue on StatsBase.jl This is an excerpt form the Slack discussion on the `sd` argument.
```
Andreas Noack [4:14 PM]
Actually I think that `se` would be fine if it wasn’t for `std` and `ste` looks really
weird to me. We have time for a renaming of `std` to `sd`, though, and…
Milan Bouchet-Valat [4:15 PM]
I've never liked `std`. It's an inconsistent abbreviation (STandard Deviation?),
and AFAIK the standard one is SD.
```
which led to the conversation about whether to lift the ban on using two-letter functions.

ararslan:
I'd actually prefer we go in the opposite direction and use a longer name for standard deviation, e.g. `stddev`.

Nosferican:
`stddev` would then be inconsistent with `stderror` as error would not be abbreviated as deviation. It doesn't seem there is a consensus on how to proceed. Gonna open an issue about `std` and `stderr` and we revisit the PR with the decision there.

src/deprecates.jl
```diff
@@ -20,7 +20,7 @@ import Base.varm, Base.stdm
 @deprecate AICc(obj::StatisticalModel) aicc(obj)
 @deprecate BIC(obj::StatisticalModel) bic(obj)
 
-@deprecate stderr(obj::StatisticalModel) se(obj)
+@deprecate stderr(obj::StatisticalModel) stderror(obj)
```
stevengj: Maybe something like
```jl
if !isdefined(Base, :stderr)
    @deprecate stderr(obj::StatisticalModel) stderror(obj)
else
    @deprecate (io::Base.TTY)(obj::StatisticalModel) stderror(obj)
end
```
so that it still mostly works in 0.7.   (Technically, the second deprecation message should be `io == stderr ? stderror(obj) : error(...)`, but that would be a rather confusing deprecation message, and in practice I don't think people are likely to pass `StatisticalModel` objects to `stdin` or `stdout`.)

yukota: Thank you for reviewing.
The second deprecate message cause syntax error.
```
ERROR: LoadError: LoadError: invalid usage of @deprecate
```
I'm tring to fix it, but it is difficult for me.
 Could you tell me how to write?

nalimilan: You should be able to work around this limitation by doing something like:
```julia
function (io::IO)(obj::StatisticalModel)
    Base.depwarn("stderr(obj::StatisticalModel) is deprecated, use stderror(obj) instead", :stderr)
    io === stderr ? stderror(obj) : throw(MethodErrror(io, (obj,)))
end
```

Nosferican:
Would be nice to get this tagged before [#StatsModels PR#53](https://github.com/JuliaStats/StatsModels.jl/pull/53/).

src/deprecates.jl
```diff
@@ -23,9 +23,9 @@ import Base.varm, Base.stdm
 if !isdefined(Base, :stderr)
     @deprecate stderr(obj::StatisticalModel) stderror(obj)
 else
-    function (io::IO)(obj::StatisticalModel)
+    function (io::typeof(stderr))(obj::StatisticalModel)
```
stevengj: The problem with this is that `typeof(stderr)` can change at runtime if `redirect_stderr` is called.

So this should just be `::IO` to catch all usages.   I don't think it is a problem to have a fairly general method here as a temporary deprecation since it continues to throw a `MethodError` for non-`stderr` objects. (People are unlikely to pass `StatisticalModel` objects to random `IO` objects by accident anyway.)

src/deprecates.jl
```diff
@@ -20,6 +20,15 @@ import Base.varm, Base.stdm
 @deprecate AICc(obj::StatisticalModel) aicc(obj)
 @deprecate BIC(obj::StatisticalModel) bic(obj)
 
+if !isdefined(Base, :stderr)
+    @deprecate stderr(obj::StatisticalModel) stderror(obj)
+else
+    function (io::typeof(stderr))(obj::StatisticalModel)
```
stevengj: The problem with this is that `typeof(stderr)` can change at runtime if `redirect_stderr` is called.

So this should just be `::IO` to catch all usages. I don't think it is a problem to have a fairly general method here as a temporary deprecation since it continues to throw a MethodError for non-`stderr` objects. (People are unlikely to pass `StatisticalModel` objects to random IO objects by accident anyway.)

stevengj: Why doesn't `IO` work?  On 0.7 (unless something has changed very recently), `typeof(stderr)` is `TTY`, which is a subtype of `IO`.

stevengj: > I'm not sure we should care about supporting that deprecation after calling `redirect_stderr`

This would mean that the deprecation wouldn't work in `IJulia`, for example.

nalimilan: Because of https://github.com/JuliaLang/julia/issues/14919:
```julia
julia> (io::IO)() = 1
ERROR: cannot add methods to an abstract type
```

stevengj: I see… yes, I can't think of a way to make it work for IJulia (without adding an IJulia dependency), since IJulia redirects to [its own `IO` type](https://github.com/JuliaLang/IJulia.jl/blob/9cba26457abe25058e1331b76f7b117406d72507/src/stdio.jl#L7-L9).

PR #375, opened 2018-04-28
Added `setindex!` support to weights (made weights mutable)

rofinn:
In order to write several performant weighted statistics algorithms I've needed to mutate a weight vector in-place. I've been manually extracting the weight values for my specific functions, but this may be a better (more general purpose) solution moving forward.

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -9,7 +9,7 @@ and stores the `values` (`V<:RealVector`) and `sum` (`S<:Real`).
 """
 macro weights(name)
     return quote
-        struct $name{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractWeights{S, T, V}
+        mutable struct $name{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractWeights{S, T, V}
             values::V
             sum::S
         end
@@ -24,6 +24,15 @@ isempty(wv::AbstractWeights) = isempty(wv.values)
 
 Base.getindex(wv::AbstractWeights, i) = getindex(wv.values, i)
 Base.size(wv::AbstractWeights) = size(wv.values)
+
+function Base.setindex!(wv::AbstractWeights{S, T}, v::T, i::Int) where {S, T}
+    s = sum(wv)
+    s -= wv[i]
+    s += v
+
+    wv.values[i] = v
+    wv.sum = s
+end
 
 """
     varcorrection(n::Integer, corrected=false)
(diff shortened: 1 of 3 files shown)


rofinn:
NOTE: I realize this needs tests, but I didn't want to spend a lot of time on it if folks were vehemently opposed to mutable weights.

src/weights.jl
```diff
@@ -25,6 +25,15 @@ isempty(wv::AbstractWeights) = isempty(wv.values)
 Base.getindex(wv::AbstractWeights, i) = getindex(wv.values, i)
 Base.size(wv::AbstractWeights) = size(wv.values)
 
+function Base.setindex!(wv::AbstractWeights{S, T}, v::T, i::Int) where {S, T}
```
nalimilan: `v` could be any `Real`, no need to restrict it. But then to avoid type instabilities better do `wv.sum = wv.sum - wv[i] + v`.

Would also be good to use `@propagate_inbounds`.

test/weights.jl
```diff
@@ -36,6 +36,23 @@ weight_funcs = (weights, aweights, fweights, pweights)
...
+
+    wv[1] = 4.
+    @test wv[1] == 4.
+    @test sum(wv) == 9.
+
+    wv[1:2] = [3., 5.]
+    @test wv[1] == 3.
+    @test wv[2] == 5.
+    @test sum(wv) == 11.
+end
```
nalimilan: It would be nice to check 1) that conversion from e.g. `Int` to `Float64` works, 2) that `setindex!` returns the original value (*before* the conversion), and 3) that if assignment fails (e.g. because of conversion error) the sum isn't corrupted.

Otherwise, looks good!

test/weights.jl
```diff
@@ -36,6 +36,23 @@ weight_funcs = (weights, aweights, fweights, pweights)
     @test sum(sa, wv) === 7.0
 end
 
+@testset "$f, setindex!" for f in weight_funcs
+    w = [1., 2., 3.]
+    wv = f(w)
+
+    @test wv[1] == 1.
+    sum(wv) == 6.
```
nalimilan: Missing `@test`. Also if you use trailing dots, use `===`, else it has no effect and it's misleading.

src/weights.jl
```diff
@@ -25,13 +25,9 @@ isempty(wv::AbstractWeights) = isempty(wv.values)
...
 Base.size(wv::AbstractWeights) = size(wv.values)
 
-function Base.setindex!(wv::AbstractWeights{S, T}, v::T, i::Int) where {S, T}
-    s = sum(wv)
-    s -= wv[i]
-    s += v
-
+@propagate_inbounds function Base.setindex!(wv::AbstractWeights, v::Real, i::Int)
+    wv.sum += v - wv[i]
     wv.values[i] = v
```
nalimilan: Actually I realize the assignment should be done first, so that if it fails the sum isn't modified. But you still need to return `v` explicitly (cf. comment below).

test/weights.jl
```diff
@@ -36,6 +36,23 @@ weight_funcs = (weights, aweights, fweights, pweights)
...
+
+    wv[1] = 4.
+    @test wv[1] == 4.
```
nalimilan: While you're at it, better test the full contents of the vector. Same below.

rofinn:
Bump?

PR #377, opened 2018-05-14
Fix bug in addcounts_radixsort! when passing existing counts

### LINKED ISSUE #376: addcounts! using radix sort doesn't add counts
Instead it gives the counts of any elements in the new array as well the previous counts of any elements that do not appear in the new array.

I believe this is because of the initialization`tmpcount = 1` in `_addcounts_radix_sort_loop!` rather than using the previous value + 1 [if it exists].

Example:

```
using StatsBase
dict = Dict{Int64, Int64}()
vals1 = [1 ; 2 ; 3]
vals2 = [1 ; 2 ; 4]
StatsBase.addcounts!(dict, vals1)
println(dict)
StatsBase.addcounts!(dict, vals2)
println(dict)
```

gives as output:

```
Dict(2=>1,3=>1,1=>1)
Dict(4=>1,2=>1,3=>1,1=>1)
```

xiaodaigh:
The addcounts_radixsort! wasn't adding the counts! It is now fixed with tests to prevent it from happening again. See #376

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -324,9 +324,9 @@ radixsort_safe(::Type{T}) where {T<:BaseRadixSortSafeTypes} = true
 radixsort_safe(::Type) = false
 
 function _addcounts_radix_sort_loop!(cm::Dict{T}, sx::AbstractArray{T}) where T
-    tmpcount = 1
     last_sx = sx[1]
-
+    tmpcount = get(cm, last_sx, 0) + 1
+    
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks, but tests all failed. :-/

xiaodaigh:
Seems to be failing to install GLM.jl. Tests pass on my computer

andreasnoack:
I've restarted the tests. Hopefully, GLM should work correctly now.

PR #382, opened 2018-05-29
add a BaseCompat submodule to replace Compat.cor etc

fredrikekre:


martinholters:
You could add the tests about to be removed from Compat in https://github.com/JuliaLang/Compat.jl/pull/559 here with `Compat` being replaced with `BaseCompat`:

https://github.com/JuliaLang/Compat.jl/blob/0f612e10cf62d6b1a15da9276df92a5dd2e3605c/test/runtests.jl#L1572-L1601

fredrikekre:
Will do. Is `BaseCompat` a good name for this?

martinholters:
LGTM. I've also verified locally that it works (on 0.6) with a Compat stripped of these definitions.

src/StatsBase.jl
```diff
@@ -204,13 +204,40 @@ end
...
 else
     import Base: cov, var, varm, std, stdm, sqrt!,
         unscaled_covzm, cor, varm!, covm, corm, cov2cor!
-    const Compatvarm = Compat.varm
 end
 
+module StatsCompat
+    if VERSION < v"0.7.0-DEV.4064"
+        var(a::AbstractArray; dims=nothing, kwargs...) =
+                    dims===nothing ? Base.var(a; kwargs...) : Base.var(a, dims; kwargs...)
```
nalimilan: Weird alignment.

test/cov.jl
```diff
@@ -118,14 +118,14 @@ weight_funcs = (weights, aweights, fweights, pweights)
             cor2 = cor(X, wv2, 2)
 
             @testset "cov2cor" begin
-                @test cov2cor(Compatcov(X, dims = 1), Compatstd(X, dims = 1)) ≈ Compatcor(X, dims = 1)
-                @test cov2cor(Compatcov(X, dims = 2), Compatstd(X, dims = 2)) ≈ Compatcor(X, dims = 2)
+                @test cov2cor(StatsBase.StatsCompat.cov(X, dims = 1), StatsBase.StatsCompat.std(X, dims = 1)) ≈ StatsBase.StatsCompat.cor(X, dims = 1)
```
nalimilan: `StatsBase.` isn't needed, right?

martinholters:
> That's not very pretty, but hopefully we'll drop 0.6 support soon.

However, even after dropping 0.6 support, it would be nice to carry the `StatsCompat` module for a bit so that packages that still use it to still support 0.6 don't have to put a version upper bound on StatsBase. The module definition could of course be trivial then.

fredrikekre:
I can't merge to this repo if that is what any of you are waiting for.

martinholters:
Neither can I.

PR #384, opened 2018-05-31
Use `something` instead of `coalesce`

martinholters:


martinholters:
OSX failure is unrelated (and should go away with Julia > 0.7.0-DEV.5268).
AV 32bit failure also looks unrelated to me.

src/base.jl
```diff
@@ -1,5 +1,9 @@
 # This file contains code formerly part of Julia. License is MIT: https://julialang.org/license
 
+if !isdefined(Base, :something)
```
nalimilan: Better check for `VERSION`, which will allow Femtocleaner to remove this block automatically. Or add this directly to Compat, which will be useful anyway.

martinholters: > Better check for VERSION, which will allow Femtocleaner to remove this block automatically. 

Good point, update incoming.

> Or add this directly to Compat, which will be useful anyway.

True. But do we want to wait for that to be included in Compat and the release published?

nalimilan:
The 32-bit failure on Julia 0.6.3 is https://github.com/JuliaLang/julia/issues/27345.

PR #385, opened 2018-06-01
Deploy docs from Julia 0.7

nalimilan:
The functions which have been moved from Base don't appear on the manual when deploying from 0.6.

Currently doesn't work due to https://github.com/JuliaLang/julia/issues/27358.

Cc: @fredrikekre @martinholters


nalimilan:
Should be good to go now.

PR #390, opened 2018-06-18
WIP: Require Julia 0.7-alpha to run Femtocleaner

nalimilan:


REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.7.0-alpha
+julia 0.7-alpha
```
ararslan: Why is this necessary? Seems like 0.7.0-alpha should be sufficient, as it's the same as 0.7-alpha.

nalimilan: Looks like neither works... :-/

ararslan: ...welp.

nalimilan: CC: @Keno

Keno: 0.7.0-alpha would have been fine. Looks like femtocleaner got itself stuck somewhere. I'll restart it for now.

Keno: Try again (you can close and re-open the PR to re-trigger the check).

Keno: Hmm, it's possible that this package itself causes it to hang. I'll take a look.

Keno: Nope, there it is. It just takes some time.

nalimilan: Thanks, it worked! It even finished in -1m :-)

Keno: It tried to do the following:
```
  ## histograms ##
  function sturges(n)  # Sturges' formula
      n==0 && return one(n)
-     ceil(Integer, digits=log2(n))+1
+     ceil(Integer, digits=digits=log2(n))+1
  end
I think you can see where that'd go. I'll deploy a fix and report back.
```

Keno: I have deployed https://github.com/JuliaComputing/Deprecations.jl/commit/42ca003101a57341046acd8f2bc3528d1e064cc3 which fixes that particular issue. If you hit the "Fix it" button again, you should be able to trigger femtocleaner to get you a fixed commit.

ararslan:
I have things working in #394 (in retrospect I should have picked it up from here), so I'll close this in favor of that PR.

PR #394, opened 2018-07-02
Updates for 0.7, drop 0.6

ararslan:
Let's see what happens

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -1,18 +1,34 @@
 __precompile__()
 
 module StatsBase
-    import Base: length, isempty, eltype, values, sum, mean, mean!, show, quantile
+    import Base: length, isempty, eltype, values, sum, show
     import Base.Cartesian: @nloops, @nref, @nextract
     using Base: @irrational, @propagate_inbounds
     import DataStructures: heapify!, heappop!, percolate_down!
+    using SortingAlgorithms
+    using Missings
 
-    using Compat, SortingAlgorithms, Missings
-    using Compat.LinearAlgebra
-    using Compat.Random
-    using Compat.Printf
-    using Compat.SparseArrays
-    import Compat.Random: rand, rand!
-    import Compat.LinearAlgebra: BlasReal, BlasFloat
+    if VERSION >= v"0.7.0-beta.85"
+        using Statistics
+        import Statistics: mean, mean!, var, varm, varm!, std, stdm, cov, covm,
+                           cor, corm, cov2cor!, unscaled_covzm, quantile, sqrt!,
+                           median, middle
+    else
+        import Base: mean, mean!, var, varm, varm!, std, stdm, cov, covm,
+                     cor, corm, cov2cor!, unscaled_covzm, quantile, sqrt!,
+                     median, middle
+    end
+
+    # Location of the `mean` function, used for calling the function inside of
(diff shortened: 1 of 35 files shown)


ararslan:
Should be squashed on merge.

src/StatsBase.jl
```diff
@@ -1,18 +1,34 @@
 __precompile__()
 
 module StatsBase
-    import Base: length, isempty, eltype, values, sum, mean, mean!, show, quantile
+    import Base: length, isempty, eltype, values, sum, show
     import Base.Cartesian: @nloops, @nref, @nextract
     using Base: @irrational, @propagate_inbounds
     import DataStructures: heapify!, heappop!, percolate_down!
+    using SortingAlgorithms
+    using Missings
```
nalimilan: What do we need Missings for?

src/StatsBase.jl
```diff
@@ -1,18 +1,34 @@
...
+                           median, middle
+    else
+        import Base: mean, mean!, var, varm, varm!, std, stdm, cov, covm,
+                     cor, corm, cov2cor!, unscaled_covzm, quantile, sqrt!,
+                     median, middle
+    end
+
+    # Location of the `mean` function, used for calling the function inside of
+    # functions definitions that have `mean` as a keyword argument
+    const MEANHOME = (VERSION >= v"0.7.0-beta.85" ? Statistics : Base)::Module
```
nalimilan: I don't really understand why it's still needed. It's used below when calling `mean` when it doesn't appear to be necessary.

src/StatsBase.jl
```diff
@@ -189,53 +206,16 @@ module StatsBase
...
-end
-
-const BASESTATS_IN_STATSBASE = VERSION >= v"0.7.0-DEV.5238"
-
-if VERSION < v"0.7.0-DEV.3665"
-    myscale!(A::AbstractArray, b::Number) = scale!(A, b)
-else
-    myscale!(A::AbstractArray, b::Number) = rmul!(A, b)
-end
+const BASESTATS_IN_STATSBASE = v"0.7.0-DEV.5238" <= VERSION < v"0.7.0-beta.85"
```
nalimilan: Since that variable is only used here, better put it directly with the `if` so that Femtocleaner can remove it.

src/StatsBase.jl
ararslan: `disallowmissing`

src/StatsBase.jl
ararslan: It's used in a few other places as well:
```
$ sift -e BASESTATS
test/runtests.jl:28:if StatsBase.BASESTATS_IN_STATSBASE
src/StatsBase.jl:209:const BASESTATS_IN_STATSBASE = v"0.7.0-DEV.5238" <= VERSION < v"0.7.0-beta.85"
src/StatsBase.jl:211:@static if BASESTATS_IN_STATSBASE
```

src/StatsBase.jl
ararslan: There are cases where we're calling the `mean` function when the keyword argument `mean` is `nothing`. In these cases we need to qualify the call to `mean` so that it doesn't pick up the keyword argument instead for the call. But what we qualify it with, Base or Statistics, depends on the version. Thus I stuck in this quick and dirty workaround to avoid version checks in every case of this.

PR #395, opened 2018-07-09
Require Julia v0.7.0-beta.85 (Statistics stdlib)

ararslan:
This permits sizable code deletion, but means that we can only test on nightly, at least until the next 0.7 prerelease after the current beta.


ararslan:
Should definitely be squashed on merge.

Issue #135, opened 2015-09-30
A Ones type


matthieugomez:
A Ones type would be something like

``` julia
immutable Ones{N} <: AbstractArray{Int,N}
    dims::NTuple{N, Int}
end
Base.size(O::Ones) = O.dims
Base.getindex(O::Ones, I::Int...) = (checkbounds(O, I...); 1)
```

(see [this stackoverflow answer](http://stackoverflow.com/a/30968709/3662288) for the original proposal by @mbauman).

It's helpful because it allows to write efficient, generic code:
- Its default behavior matches a vector of ones, without the memory allocation
- Frequently,  a non weighted operation can be written more efficiently than the weighted case with a vector of ones. Defining a neutral weight type such as Ones allows to branch elegantly between the weighted and the non weighted versions.

It's a good application of multiple dispatch. It would be great to define something like this in StatsBase.

matthieugomez:
Why should such a type be defined in `StatsBase`? 
Suppose I need to compute the median within a function that may or may not be passed weights. For efficiency, I need to call `median(x)` in the no weight case vs `median(x, w)` in the weight case. I'm not sure how to do that in a elegant way. In contrast, if  `median(x, w::Ones) = x` is defined in `StatsBase` defined, I just need to write `median(x, w)` to handle either case. In effect, dispatch on the number of argument is replaced by a type dispatch. 
Even better, suppose I actually did not know that `median` was more efficient than the unweighted case. I would still write `median(x, Ones(length(x)))` to avoid the allocation of a weight vector (back to the first point). But now, the definition in `StatsBase` automatically dispatches to the efficient method when `w` is of type `Ones`

johnmyleswhite:
I think this is pretty reasonable and a potentially good change, but it adds more complexity.

I would like us to improve the design of our weighting functionality at some point to handle weights with different interpretations (right now everything is a frequency weight); maybe we can address this concern then?

We probably need a different name than `Ones`, which is too vague. Something like `UniformWeights` would be better.

nalimilan:
`UnitWeights` might be even better. In the literature, uniform weights sometimes mean `1/N`, `N` being the number of cases. Unit weights doesn't have that ambiguity.

matthieugomez:
Great! I'm also a supporter of different kinds of weights.

sbromberger:
Hi,

[...]

PR #400, opened 2018-07-22
Correct R² formulas and add Efron's R²

### LINKED ISSUE #396: Methods for StatsModel
I have some questions:

* Could we add a function to recover the label of the response (similar to `coefnames`)? `respname`, maybe?
* The method `weights` was recently added. It has the same name as the constructor of `Weights`. Isn't it confusing?
* Why is `r2` based on deviances rather than likelihoods? The original formulae use likelihoods and so do programs like Stata.
* `r2` gained a new method: `r2(obj::StatisticalModel) = mss(obj) / deviance(obj)`. Wouldn't it make more sense to add the Efron R² to the standard r² instead? Something like:

```
function r2(obj::StatisticalModel, variant::Symbol)

    if variant == :Efron

        y = model_response(obj)
        ŷ = fitted(obj)
        μ = meanresponse(obj)
        1.0 - sum(abs2, y .- ŷ) / sum(abs2, y .- μ)

    else

        ll = loglikelihood(obj)
        ll0 = nullloglikelihood(obj)

        if variant == :McFadden
            1 - ll/ll0
        elseif variant == :CoxSnell
            1 - exp(2 * (ll0 - ll) / nobs(obj))
        elseif variant == :Nagelkerke
            (1 - exp(2 * (ll0 - ll) / obs(obj))) / (1 - exp(2 * ll0/nobs(obj)))
        else
            error("variant must be one of :McFadden, :CoxSnell, :Nagelkerke or :Efron")
        end
    end
end
```

lbittarello:
This PR fixed the R² formulas, which were previously based on the deviance instead of the likelihood function. It also adds Efron's R².
The following variants were benchmarked against Stata's fitstat: McFadden, Cox and Snell, Nagelkerke and Efron.
Cox and Snell's and Efron's R² should match the classical R² for linear models.

Closes #396.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -198,6 +198,10 @@ For a linear model, the R² is defined as ``ESS/TSS``, with ``ESS`` the explaine
 and ``TSS`` the total sum of squares.
 """
 function r2(obj::StatisticalModel)
+
+    Base.depwarn("The default r² method for linear models is deprecated. " *
+                 "Packages should define their own methods.", :r2)
+
     μ = meanresponse(obj)
     mss(obj) / sum(x -> abs2(x - μ), response(obj))
 end
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -197,7 +197,7 @@ Coefficient of determination (R-squared).
 For a linear model, the R² is defined as ``ESS/TSS``, with ``ESS`` the explained sum of squares
 and ``TSS`` the total sum of squares.
 """
-r2(obj::StatisticalModel) = mss(obj) / deviance(obj)
+r2(obj::StatisticalModel) = mss(obj) / sum(abs2, response(obj) .- meanresponse(obj))
 
 """
     r2(obj::StatisticalModel, variant::Symbol)
@@ -207,27 +207,39 @@ Pseudo-coefficient of determination (pseudo R-squared).
 
 For nonlinear models, one of several pseudo R² definitions must be chosen via `variant`.
 Supported variants are:
-- `:MacFadden` (a.k.a. likelihood ratio index), defined as ``1 - \\log L/\\log L0``.
-- `:CoxSnell`, defined as ``1 - (L0/L)^{2/n}``
-- `:Nagelkerke`, defined as ``(1 - (L0/L)^{2/n})/(1 - L0^{2/n})``, with ``n`` the number
-of observations (as returned by [`nobs`](@ref)).
+- `:MacFadden` (a.k.a. likelihood ratio index), defined as ``1 - \\log (L)/\\log (L_0)``;
+- `:CoxSnell`, defined as ``1 - (L_0/L)^{2/n}``;
+- `:Nagelkerke`, defined as ``(1 - (L_0/L)^{2/n})/(1 - L_0^{2/n})``;
+- `:Efron`, defined as ``1 - \\sum_i (y_i - \\hat{y})^2 / \\sum_i (y_i - \\bar{y})^2``.
 
-In the above formulas, ``L`` is the likelihood of the model, ``L0`` that of the null model
-(the model including only the intercept). These two quantities are taken to be minus half
-`deviance` of the corresponding models.
+In the above formulas, ``L`` is the likelihood of the model,
+``L_0`` is the likelihood of the null model (the model with only an intercept),
+``n`` is the number of observations, ``y_i`` is the response, 
+``\\hat{y}_i`` are fitted values and ``\\bar{y}`` is the average response.
 """
-function r2(obj::StatisticalModel, variant::Symbol)
-    ll = -deviance(obj)/2
-    ll0 = -nulldeviance(obj)/2
 
-    if variant == :McFadden
-        1 - ll/ll0
-    elseif variant == :CoxSnell
-        1 - exp(2/nobs(obj) * (ll0 - ll))
-    elseif variant == :Nagelkerke
-        (1 - exp(2/nobs(obj) * (ll0 - ll)))/(1 - exp(2/nobs(obj) * ll0))
+# The following variants were benchmarked against Stata's fitstat:
+# McFadden, Cox and Snell, Nagelkerke and Efron.
+# Cox and Snell's and Efron's R² should match the classical R² for linear models.
+
+function r2(obj::StatisticalModel, variant::Symbol)
+    if variant == :Efron
+        y = response(obj)
+        ŷ = fitted(obj)
+        μ = meanresponse(obj)
+        1 - sum(abs2, y .- ŷ) / sum(abs2, y .- μ)
     else
-        error("variant must be one of :McFadden, :CoxSnell or :Nagelkerke")
+        ll = loglikelihood(obj)
+        ll0 = nullloglikelihood(obj)
+        if variant == :McFadden
+            1 - ll/ll0
+        elseif variant == :CoxSnell
+            1 - exp(2 * (ll0 - ll) / nobs(obj))
+        elseif variant == :Nagelkerke
+            (1 - exp(2 * (ll0 - ll) / obs(obj))) / (1 - exp(2 * ll0 / nobs(obj)))
+        else
+            error("variant must be one of :McFadden, :CoxSnell, :Nagelkerke or :Efron")
+        end
     end
 end
 
@@ -259,10 +271,9 @@ In this formula, ``L`` is the likelihood of the model, ``L0`` that of the null m
 of the model (as returned by [`dof`](@ref)).
 """
 function adjr2(obj::StatisticalModel, variant::Symbol)
-    ll = -deviance(obj)/2
-    ll0 = -nulldeviance(obj)/2
+    ll = likelihood(obj)
+    ll0 = nulllikelihood(obj)
     k = dof(obj)
-
     if variant == :McFadden
         1 - (ll - k)/ll0
     else
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -228,7 +228,7 @@ function r2(obj::StatisticalModel, variant::Symbol)
         y = response(obj)
         ŷ = fitted(obj)
         μ = meanresponse(obj)
-        1 - sum(abs2, y .- ŷ) / sum(x -> abs2(x - μ), response(obj))
+        1 - sum(x -> abs2(x[1] - x[2]), zip(y, ŷ)) / sum(x -> abs2(x - μ), response(obj))
     else
         ll = loglikelihood(obj)
         ll0 = nullloglikelihood(obj)
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -201,8 +201,7 @@ function r2(obj::StatisticalModel)
     Base.depwarn("The default r² method for linear models is deprecated. " *
                  "Packages should define their own methods.", :r2)
 
-    μ = meanresponse(obj)
-    mss(obj) / sum(x -> abs2(x - μ), response(obj))
+    mss(obj) / deviance(obj)
 end
 
 """
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -197,7 +197,10 @@ Coefficient of determination (R-squared).
 For a linear model, the R² is defined as ``ESS/TSS``, with ``ESS`` the explained sum of squares
 and ``TSS`` the total sum of squares.
 """
-r2(obj::StatisticalModel) = mss(obj) / sum(abs2, response(obj) .- meanresponse(obj))
+function r2(obj::StatisticalModel)
+    μ = meanresponse(obj)
+    mss(obj) / sum(x -> abs2(x - μ), response(obj))
+end
 
 """
     r2(obj::StatisticalModel, variant::Symbol)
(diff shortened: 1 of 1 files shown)


src/statmodels.jl
```diff
@@ -197,7 +197,7 @@ Coefficient of determination (R-squared).
 For a linear model, the R² is defined as ``ESS/TSS``, with ``ESS`` the explained sum of squares
 and ``TSS`` the total sum of squares.
 """
-r2(obj::StatisticalModel) = mss(obj) / deviance(obj)
+r2(obj::StatisticalModel) = mss(obj) / sum(abs2, response(obj) .- meanresponse(obj))
```
nalimilan: Better do `μ = meanresponse(obj); sum(x -> abs2(x - μ), response(obj))` to avoid allocating a copy. Maybe turn this into a long-form function for readability.

@Nosferican Is this formula OK for your package?

src/statmodels.jl
```diff
@@ -207,27 +207,39 @@ Pseudo-coefficient of determination (pseudo R-squared).
...
+- `:Nagelkerke`, defined as ``(1 - (L_0/L)^{2/n})/(1 - L_0^{2/n})``;
+- `:Efron`, defined as ``1 - \\sum_i (y_i - \\hat{y})^2 / \\sum_i (y_i - \\bar{y})^2``.
 
-In the above formulas, ``L`` is the likelihood of the model, ``L0`` that of the null model
-(the model including only the intercept). These two quantities are taken to be minus half
-`deviance` of the corresponding models.
+In the above formulas, ``L`` is the likelihood of the model,
+``L_0`` is the likelihood of the null model (the model with only an intercept),
+``n`` is the number of observations, ``y_i`` is the response, 
+``\\hat{y}_i`` are fitted values and ``\\bar{y}`` is the average response.
```
nalimilan: If you use plural "are" here, also use it for the response.

src/statmodels.jl
```diff
@@ -207,27 +207,39 @@ Pseudo-coefficient of determination (pseudo R-squared).
...
+# The following variants were benchmarked against Stata's fitstat:
+# McFadden, Cox and Snell, Nagelkerke and Efron.
+# Cox and Snell's and Efron's R² should match the classical R² for linear models.
+
+function r2(obj::StatisticalModel, variant::Symbol)
+    if variant == :Efron
+        y = response(obj)
+        ŷ = fitted(obj)
+        μ = meanresponse(obj)
+        1 - sum(abs2, y .- ŷ) / sum(abs2, y .- μ)
```
nalimilan: Same remark as above about avoiding allocations.

src/statmodels.jl
```diff
@@ -207,27 +207,39 @@ Pseudo-coefficient of determination (pseudo R-squared).
...
 For nonlinear models, one of several pseudo R² definitions must be chosen via `variant`.
 Supported variants are:
-- `:MacFadden` (a.k.a. likelihood ratio index), defined as ``1 - \\log L/\\log L0``.
-- `:CoxSnell`, defined as ``1 - (L0/L)^{2/n}``
-- `:Nagelkerke`, defined as ``(1 - (L0/L)^{2/n})/(1 - L0^{2/n})``, with ``n`` the number
-of observations (as returned by [`nobs`](@ref)).
+- `:MacFadden` (a.k.a. likelihood ratio index), defined as ``1 - \\log (L)/\\log (L_0)``;
+- `:CoxSnell`, defined as ``1 - (L_0/L)^{2/n}``;
+- `:Nagelkerke`, defined as ``(1 - (L_0/L)^{2/n})/(1 - L_0^{2/n})``;
+- `:Efron`, defined as ``1 - \\sum_i (y_i - \\hat{y})^2 / \\sum_i (y_i - \\bar{y})^2``.
```
nalimilan: Should be `\\har{y}_i`.

src/statmodels.jl
```diff
@@ -207,27 +207,39 @@ Pseudo-coefficient of determination (pseudo R-squared).
...
-    ll = -deviance(obj)/2
-    ll0 = -nulldeviance(obj)/2
 
-    if variant == :McFadden
-        1 - ll/ll0
-    elseif variant == :CoxSnell
-        1 - exp(2/nobs(obj) * (ll0 - ll))
-    elseif variant == :Nagelkerke
-        (1 - exp(2/nobs(obj) * (ll0 - ll)))/(1 - exp(2/nobs(obj) * ll0))
+# The following variants were benchmarked against Stata's fitstat:
```
nalimilan: I'd rather put this mention in the tests next to the expected values. The explanation about linear models can go to the docstring.

src/statmodels.jl
```diff
@@ -259,10 +271,9 @@ In this formula, ``L`` is the likelihood of the model, ``L0`` that of the null m
 of the model (as returned by [`dof`](@ref)).
 """
 function adjr2(obj::StatisticalModel, variant::Symbol)
-    ll = -deviance(obj)/2
-    ll0 = -nulldeviance(obj)/2
+    ll = likelihood(obj)
```
nalimilan: AFAICT you want `loglikelihood` here. The fact it wasn't caught by testing is a bit worrying: isn't this covered by GLM's tests?

src/statmodels.jl
```diff
@@ -259,10 +271,9 @@ In this formula, ``L`` is the likelihood of the model, ``L0`` that of the null m
 of the model (as returned by [`dof`](@ref)).
 """
```
nalimilan: Also add parentheses after `log` in this docstring for consistency.

lbittarello: Done!

src/statmodels.jl
lbittarello: > avoid allocating a copy

I thought that `sum` didn't allocate transform scalars into vectors with the dot syntax.

> we'd better deprecate this function and leave packages to implement it for linear models

Do you mean removing the function and leaving a warning or leaving the function and adding a warning?

src/statmodels.jl
lbittarello: check

src/statmodels.jl
lbittarello: check

src/statmodels.jl
lbittarello: check

src/statmodels.jl
lbittarello: Tests in GLM.jl?

src/statmodels.jl
lbittarello: Sorry. You're right. I hadn't tested the adjusted R².

src/statmodels.jl
Nosferican: Consider this script,
```
using BenchmarkTools
a(x, y) = sum(abs2, x - y)
b(x, y) = sum(abs2, x .- y)
c(x, y) = sum(xy -> abs2(xy[1] - xy[2]), zip(x, y))
d(x, y) = sum(abs2(a - b) for (a, b) ∈ zip(x, y))
using Random: seed!
seed!(0)
const x = rand(1000);
const y = rand(1000);
@btime a($x, $y)
@btime b($x, $y)
@btime c($x, $y)
@btime d($x, $y)
  703.416 ns (1 allocation: 7.94 KiB)
  659.205 ns (1 allocation: 7.94 KiB)
  823.250 ns (1 allocation: 32 bytes)
  969.684 ns (2 allocations: 48 bytes)
```
It seems to be trivially faster, but it certainly allocates as opposed to `c` or `d`.

If the `R²` matches Stata, it will match mine. Some models will still need to define the method for some panel estimators, but this seems good.

If the statsmodels functions are being moved to StatsModels it might be best to make the decision there (better to trim there, than to have lost useful code then),

src/statmodels.jl
Nosferican: Some tests here are run using GLM which uses this package... Might be best test these eventually without another package for it.

src/statmodels.jl
nalimilan: @lbittarello:
> I thought that sum didn't allocate transform scalars into vectors with the dot syntax.

As @Nosferican shows, it allocates, and it will probably stay that way in the future. Maybe another syntax will be introduced to do that.

> Do you mean removing the function and leaving a warning or leaving the function and adding a warning?

I mean print a warning using `Base.depwarn(..., :r2)`, saying that packages should define their custom method. Then we'll turn the warning into an error just like for other function stubs in the file.

@Nosferican:
> If the R² matches Stata, it will match mine. Some models will still need to define the method for some panel estimators, but this seems good.

OK, cool.

> If the statsmodels functions are being moved to StatsModels it might be best to make the decision there (better to trim there, than to have lost useful code then),

It's not clear yet what will happen and when, and given how small this function is that's really not an issue.

src/statmodels.jl
nalimilan: Yeah, in GLM.jl (since the check against Stata applies to GLMs AFAICT).

It's not ideal to have this split in two packages, but testing things directly here implies writing pseudo-model types. That's not so hard, but it will take some work.

src/statmodels.jl
nalimilan: OK. Do you confirm it works know?

src/statmodels.jl
```diff
@@ -207,27 +210,37 @@ Pseudo-coefficient of determination (pseudo R-squared).
...
-    elseif variant == :CoxSnell
-        1 - exp(2/nobs(obj) * (ll0 - ll))
-    elseif variant == :Nagelkerke
-        (1 - exp(2/nobs(obj) * (ll0 - ll)))/(1 - exp(2/nobs(obj) * ll0))
+function r2(obj::StatisticalModel, variant::Symbol)
+    if variant == :Efron
+        y = response(obj)
+        ŷ = fitted(obj)
+        μ = meanresponse(obj)
+        1 - sum(abs2, y .- ŷ) / sum(x -> abs2(x - μ), response(obj))
```
nalimilan: `sum(abs2, y .- ŷ)` also needs to be changed.

lbittarello: check

src/statmodels.jl
lbittarello: Yes, sir!

src/statmodels.jl
lbittarello: I've added a deprecation warning. Let me know if it's okay.

src/statmodels.jl
```diff
@@ -197,7 +197,13 @@ Coefficient of determination (R-squared).
 For a linear model, the R² is defined as ``ESS/TSS``, with ``ESS`` the explained sum of squares
 and ``TSS`` the total sum of squares.
 """
-r2(obj::StatisticalModel) = mss(obj) / deviance(obj)
+function r2(obj::StatisticalModel)
+    Base.depwarn("The default r² method for linear models is deprecated. " *
+                 "Packages should define their own methods.", :r2)
+
+    μ = meanresponse(obj)
+    mss(obj) / sum(x -> abs2(x - μ), response(obj))
```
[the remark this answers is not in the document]
nalimilan: Good catch. That also applies to the Efron R². For the method above, since it's deprecated, we'd better keep using `deviance` for now. For the Efron R², I guess we need to multiply observations by their weights, but will it work for all kinds of models?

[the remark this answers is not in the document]
lbittarello: Precisely: packages can handle weights through `mss(obj)` and `loglikelihood(obj)`. There's no need to provide for weights here.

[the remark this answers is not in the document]
Nosferican: Not sure for `ProbabilityWeights`, but should work fine for the other weights supported if not mistaken. The tests should cover the weight cases for sure.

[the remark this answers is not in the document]
lbittarello: Efron's R² does need special treatment, since it uses the sum of residuals.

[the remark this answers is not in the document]
Nosferican: Shouldn't that be
```
mss(obj) / sum(w * abs2(x - μ) for (w, x) in zip(weights(obj), response(obj)))
```
It might also be best to keep the `deviance` since models might cache the value in the struct rather than having to compute it in the fly every time it is called.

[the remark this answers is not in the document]
lbittarello: Good idea. It will automatically handle the weights too.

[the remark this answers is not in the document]
nalimilan: > Not sure for ProbabilityWeights, but should work fine for the other weights supported if not mistaken. The tests should cover the weight cases for sure.

IIRC it should be OK even for probability weights since weights appear both at the numerator and the denominator (therefore multiplying weights doesn't change the result, which is often an issue with probability weights). But I'm afraid we don't have tests for that in GLM (since it only supports frequency weights), maybe you have some in your package?

> Precisely: packages can handle weights through mss(obj) and loglikelihood(obj). There's no need to provide for weights here.

The point is that the new code uses `sum(x -> abs2(x - μ), response(obj))` instead of `deviance`, so it no longer takes weights into account.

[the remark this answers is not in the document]
Nosferican: I am not sure if I ended up supporting `ProbabilityWeights` yet. I know `FrequencyWeights` and `AnalyticalWeights` were supported (basically these worked the same way except in the robust variance covariance estimator which had a subtle difference). If I have time, might help out with developing a simple test suite for the statistical models inheritance. I had a bunch of tests and the computation so one should be able to make a dummy `RegressionModel` for it. In reality one only needs a struct to hold certain values and implement the API for it. No need to actually fit the model so I find the dependency on other packages for fitting it not required.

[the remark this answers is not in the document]
lbittarello: > The point is that the new code uses `sum(x -> abs2(x - μ), response(obj))` instead of `deviance`, so it no longer takes weights into account

I know. I'd removed `deviance` in case a package hadn't defined it. But I reverted this change in the interest of weights in the newest commit.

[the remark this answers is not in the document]
nalimilan: OK. Do you know the formula for weighted Efron R²? I guess we just need to multiply both the numerator and the denominator?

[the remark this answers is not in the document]
lbittarello: > Do you know the formula for weighted Efron R²? I guess we just need to multiply both the numerator and the denominator?

I think so. How are we going to implement it? We can retrieve weights via `weights`, but how do we check if weights were defined in the first place?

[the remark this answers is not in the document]
Nosferican: I think the default is just a one array. Might be nice to follow [FillArrays.jl](https://github.com/JuliaArrays/FillArrays.jl) for it to implement the UnitWeight or something.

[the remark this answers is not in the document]
nalimilan: Let's call `weights` and assume it returns a valid `AbstractVector`. Then it would make sense to define `UnitWeights` in this package, to complement other kinds of weights (https://github.com/JuliaStats/StatsBase.jl/issues/135).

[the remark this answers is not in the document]
nalimilan: Bump. Can you drop the Efron R² for now so that we can merge the PR? We can discuss improvements later.

nalimilan:
Thanks! Can you file a PR against GLM to update R² tests?

lbittarello:
As far as I can tell, the only tests involve linear models. This PR shouldn't affect them.

nalimilan:
OK. Then it would be nice to add tests to ensure results remain correct in the future. ;-)

src/statmodels.jl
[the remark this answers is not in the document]
nalimilan: Since https://github.com/JuliaStats/StatsBase.jl/pull/515 we have `UnitWeights`, so we can require models to return this when they are unweighted.

PR #406, opened 2018-08-16
Add level argument to confint

nalimilan:
Fixes https://github.com/JuliaStats/GLM.jl/issues/241.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -17,18 +17,19 @@ Return the names of the coefficients.
 coefnames(obj::StatisticalModel) = error("coefnames is not defined for $(typeof(obj)).")
 
 """
-    coeftable(obj::StatisticalModel)
+    coeftable(obj::StatisticalModel; level::Real=0.95)
 
 Return a table of class `CoefTable` with coefficients and related statistics.
+`level` determines the level for confidence intervals (by default, 95%).
 """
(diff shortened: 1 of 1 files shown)


nalimilan:
OK to merge?

lbittarello:
Why not add a similar argument to `coeftable`?

nalimilan:
Good idea. I see you implemented a `level` keyword argument in Microeconometrics. Two questions:
- Wouldn't it be clearer to call it `confint`, since in the context of `coeftable` nothing indicates that it applies to a confidence interval?
- Should `confint` take a keyword argument too rather than a positional argument (as is currently the case in GLM, but undocumented)?

lbittarello:
> Wouldn't it be clearer to call it `confint`, since in the context of `coeftable` nothing indicates that it applies to a confidence interval?

I don't have a strong preference. On the one hand, `confint` is marginally clearer. On the other hand, `level` can't refer to anything else in this context. I used `level` because Stata does so. Another option is `cilevel`.

> Should `confint` take a keyword argument too rather than a positional argument (as is currently the case in GLM, but undocumented)?

I like keywords. `r2` uses a positional argument though, so it might be better to be consistent.

nalimilan:
OK, I've switched to keyword arguments, and added it to `coeftable`. Since keyword arguments don't participate in dispatch, the default cannot be set in the generic method, and needs to be defined by each implementation. So this is now purely a doc PR.

Good to merge?

PR #407, opened 2018-08-18
fix documentation of vardim

jamesonquinn:
Vardim is misplaced in the function signatures in the documentation, this pull fixes that.


matbesancon:
Thanks!

PR #412, opened 2018-09-12
Remove deprecation warnings for histogram closed-argument default

oschulz:
It's been a long time since we switched the default for histogram `closed` arguments from `:right` to `:left`. Past time to remove the deprecation warnings.

@andreasnoack, sorry it took me a while to do the PR, this is what we discussed at JuliaCon.

### DIFF
--- a/test/hist.jl
+++ b/test/hist.jl
@@ -1,5 +1,3 @@
-# See src/hist.jl for meaning of "FIXME: closed" comments.
-
 using StatsBase
 using LinearAlgebra, Random, Test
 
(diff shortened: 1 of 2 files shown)


test/hist.jl
```diff
@@ -26,7 +27,6 @@ end
 
 
 @testset "Histogram append" begin
-    # FIXME: closed (all lines in this block):
```
nalimilan: Do you have an idea why this block is marked, while the `closed` argument must have been present in the first place?

oschulz: I probably put it in by mistake, back then. I had lots of these `FIXME: closed` comments in the test file already, but then I restructured the test file into testsets when I added normalization. I guess that's how it snuck into that block, even though here `:left` can't be left out (to test with different values for `isdensity`).

oschulz:
Is the AppVeyor failure on 32-bit to be expected?

nalimilan:
AFAICT that's a download problem which isn't specific to StatsBase.

oschulz:
Thanks!

PR #415, opened 2018-09-21
Fixed typo in comment

PGS62:
Corrected attribution of algorithm for Kendall tau calculation - See https://www.jstor.org/stable/2282833?seq=1#page_scan_tab_contents


andreasnoack:
Thanks

PR #416, opened 2018-09-21
Fix appveyor.yml

andreasnoack:


appveyor.yml
```diff
@@ -1,17 +1,18 @@
 environment:
   matrix:
   - julia_version: 0.7
-  - julia_version: latest
+  - julia_version: 1
```
nalimilan: I'd rather use `1.0` to ensure we test that that version for as long as we claim to support it.

appveyor.yml
```diff
@@ -1,17 +1,18 @@
...
-  - x64
+  - x86 # 32-bit
+  - x64 # 64-bit
 
-## uncomment the following lines to allow failures on nightly julia
-## (tests will run but not make your overall status red)
-#matrix:
-#  allow_failures:
-#  - julia_version: latest
+# # Uncomment the following lines to allow failures on nightly julia
```
nalimilan: How about uncommenting this while you're at it?

PR #427, opened 2018-11-05
Use Wikipedia's language in the docstring to kldivergence.

mschauer:
I find "KL of `q` from `p`" confusing if not wrong.

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -509,7 +509,8 @@ crossentropy(p::AbstractArray{T}, q::AbstractArray{T}, b::Real) where {T<:Real} 
 """
     kldivergence(p, q, [b])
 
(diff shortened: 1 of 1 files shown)


src/scalarstats.jl
```diff
@@ -509,7 +509,8 @@ crossentropy(p::AbstractArray{T}, q::AbstractArray{T}, b::Real) where {T<:Real}
 """
     kldivergence(p, q, [b])
 
-Compute the Kullback-Leibler divergence of `q` from `p`, optionally specifying
+Compute the Kullback-Leibler divergence from `q` to `p`,
+or the sum `pi * log(pi / qi)`, optionally specifying
```
nalimilan:
```suggestion
i.e. the sum ``p_i * log(p_i / q_i)``, optionally specifying
```

While you're at it, maybe mention that it's also called relative entropy?

PR #429, opened 2018-11-20
Return a callable type from ecdf instead of a closure.

tpapp:
Also define minimum, maximum, and extrema.

**Rationale:** this should allow other packages to dispatch on the results of `ecdf`, eg for plotting. Accessor functions are defined for this purpose.

**Implementation:** code was moved mostly unchanged. Tests added for new functionality.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -1,7 +1,7 @@
 __precompile__()
 
 module StatsBase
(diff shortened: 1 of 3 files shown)


andreasnoack:
Seems like a good idea to me.

tpapp:
Anything I can improve on this to get it merged?

andreasnoack:
I don't think so

Issue #435, opened 2018-12-07
Wrong results for weighted quantile


andreasnoack:
From https://discourse.julialang.org/t/median-vs-50th-quantile-giving-different-answers/18414/4
```julia
julia> x = [1; 4; 3; 2; 2.5; 7];

julia> w = [0.1;0.3;0.05;0.05;0.2;0.3];

julia> quantile(x, weights(w), 0.5)
3.5
```
but
```julia
julia> sum(w[x .<= 3.5]) >= 0.5
false
```
so `0.5` is not a median.

cc @matthieugomez

matthieugomez:
There are multiple ways to extend the base function `quantile` to the case of weights, and I am very happy to discuss how to do it properly.

But it is not obvious to me that the existing function has the wrong output. You reject that 3.5 is the median because you say: if there is only one value `v` in the vector that satisfies `P(x <=v) >= p` and `P(x >= v ) >= p`, then `v` should be the `p` quantile.

But this is not true with the definition of `quantile` in Julia. For instance:
```julia
x = [1, 2, 3, 4]
w_n = [0.25, 0.25, 0.25, 0.25]
df= DataFrame(plow = cumsum(w_n), phigh = reverse(cumsum(reverse(w_n))), x = x)
#>│ Row │ plow    │ phigh   │ x     │
#>│     │ Float64 │ Float64 │ Int64 │
#>├─────┼─────────┼─────────┼───────┤
#>│ 1   │ 0.25    │ 1.0     │ 1     │
#>│ 2   │ 0.5     │ 0.75    │ 2     │
#>│ 3   │ 0.75    │ 0.5     │ 3     │
#>│ 4   │ 1.0     │ 0.25    │ 4     │
```
Julia says that the 0.49 quantile is 2.47 but, according to your definition, it should be 2.

That being said, I think the thread shows two problems with the existing implementation:
- Whatever the definition of weighted quantile we end up choosing, `wmedian` should return the same output as `wquantile(., 0.5)`. For now they return different things because they defined weighted median to be different things.
- with frequency weigths, `wquantile` fails. This is because I wrote this function assuming that `fweights` were restricted to integer. This is not the case?

matthieugomez:
At the end of the day, quantile type 7 is just not straighforward to extend to weights. I would be happier to have quantile defined as type 1 or 3 in base. It would make it simpler to extend it to weight. It would also make it possible to use quantile with types that are only ordinal.

[...]

PR #436, opened 2018-12-09
Change the definition of `wmedian` to `wquantile(., 0.5)`

matthieugomez:
and returns an error for non integer FrequencyWeights. 
See https://github.com/JuliaStats/StatsBase.jl/issues/435
I wish there was a NEWS section to document the change, because `wmedian` now gives a different result.


nalimilan:
We should revive this PR. Sorry for the delay.

For future reference, can you explain why `median` differs from `quantile` currently? What definition does it implement? @andreasnoack thinks there aren't so many different definitions of the weighted median as there are for quantiles, but I'm not sure that's the case.

matthieugomez:
median currently is: 
- take the maximum value such that sum of weight of observations below the value is lower or equal than  1/2
- take the minimum value such that sum of weight of observations above is lower or equal than 1/2. 
If same value, it is fine. If these are different values, take the average.

It corresponds to a nice extension of the median definition from unweighted to  weighted vectors. The issue is this generalization does not help when generalizing other quantiles, so I think is better to use a generalization that is defined for any quantile.

The other issue with the current implementation is that, with frequency weights, it does not give the same thing as the unweighted median of a repeated variable, but it is a different issue.

src/weights.jl
```diff
@@ -640,9 +567,50 @@ quantile(v::RealVector, w::AbstractWeights{<:Real}, p::Number) = quantile(v, w,
...
 
 Compute the `p`th quantile(s) of `v` with weights `w`, given as either a vector
-or an `AbstractWeights` vector.
+or an `AbstractWeights` vector. See the documentation for [`quantile`](@ref) for more details.
 """
-wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::RealVector) = quantile(v, w, p)
-wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::Number) = quantile(v, w, [p])[1]
-wquantile(v::RealVector, w::RealVector, p::RealVector) = quantile(v, weights(w), p)
-wquantile(v::RealVector, w::RealVector, p::Number) = quantile(v, weights(w), [p])[1]
+function wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::RealVector)
```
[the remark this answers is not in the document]
matthieugomez: Done.

src/weights.jl
```diff
@@ -579,6 +500,12 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
     end
 
 
+    isa(w, FrequencyWeights) && !isa(w, AbstractVector{<:Integer}) && error("the vector of `FrequencyWeights` must be of type `AbstractVector{<:Integer}`. Use `ProbabilityWeights` or `AnalyticWeights` if applicable.")
```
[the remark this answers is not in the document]
matthieugomez: Done.

src/deprecates.jl
```diff
@@ -26,3 +26,10 @@ end
 @deprecate scattermat_zm(x::DenseMatrix, wv::AbstractWeights, dims::Int) scattermat_zm(x::DenseMatrix, wv::AbstractWeights, dims=dims)
 @deprecate mean!(R::AbstractArray, A::AbstractArray, w::AbstractWeights, dims::Int) mean!(R, A, w, dims=dims)
 @deprecate mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Int) where {T<:Number,W<:Real} mean(A, w, dims=dims)
+
+@deprecate wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::RealVector) quantile(v, w, p)
+@deprecate wquantile(v::RealVector, w::AbstractWeights{<:Real}, p::Number) quantile(v, w, [p])[1]
+@deprecate wquantile(v::RealVector, w::RealVector, p::RealVector) quantile(v, weights(w), p)
+@deprecate wquantile(v::RealVector, w::RealVector, p::Number) quantile(v, weights(w), [p])[1]
+@deprecate wmedian(v::RealVector, w::AbstractWeights{<:Real}) median(v, w)
+@deprecate wmedian(v::RealVector, w::RealVector) median(v, weights(w))
```
nalimilan: This makes me realize that we probably shouldn't accept `Weights` objects in `quantile` and `median` since their meaning is ambiguous: they could be frequency weights, or another type of weights. So here we'd better recommend using `pweights`or `fweights` (whichever is closer to the current behavior of `wmedian`).

src/weights.jl
```diff
@@ -473,116 +473,50 @@ _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Nothing) where {T<:Numbe
...
-`quantile` for a vector with repeated values.
-With non `FrequencyWeights`,  denote ``N`` the length of the vector, ``w`` the vector of weights,
-``h = p (\\sum_{i<= N}w_i - w_1) + w_1`` the cumulative weight corresponding to the
-probability ``p`` and ``S_k = \\sum_{i<=k}w_i`` the cumulative weight for each
+`NaN` is returned if `x` contains any `NaN` values. An error is raised if `w` contains 
+any `NaN` values.
+
+With [`FrequencyWeights`](@ref), the function returns the same result as the unweighted 
+`quantile` applied to a vector with repeated values. The function returns an error if the 
+weights are not of type `AbstractVector{<:Integer}`.
```
```suggestion
weights are not integers.
```

src/weights.jl
```diff
@@ -473,116 +473,50 @@ _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Nothing) where {T<:Numbe
...
+With non `FrequencyWeights`, denote ``N`` the length of the vector, ``w`` the vector of
```
nalimilan: This paragraph shouldn't be changed (rebasing mistake?).

src/weights.jl
```diff
@@ -473,116 +473,50 @@ _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Nothing) where {T<:Numbe
...
+weights, ``h = p (\\sum_{i<= N}w_i - w_1) + w_1`` the cumulative weight corresponding to 
+the probability ``p`` and ``S_k = \\sum_{i<=k}w_i`` the cumulative weight for each
 observation, define ``v_{k+1}`` the smallest element of `v` such that ``S_{k+1}``
-is strictly superior to ``h``. The weighted ``p`` quantile is given by ``v_k + \\gamma (v_{k+1} -v_k)``
-with  ``\\gamma = (h - S_k)/(S_{k+1}-S_k)``. In particular, when `w` is a vector
-of ones, the function returns the same result as `quantile`.
+is strictly superior to ``h``. The weighted ``p`` quantile is given by 
+``v_k + \\gamma (v_{k+1} -v_k)`` with  ``\\gamma = (h - S_k)/(S_{k+1}-S_k)``. 
 """
+
```
nalimilan: Remove this line.

src/weights.jl
```diff
@@ -473,116 +473,50 @@ _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Nothing) where {T<:Numbe
...
 function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where {V,W<:Real}
     # checks
     isempty(v) && error("quantile of an empty array is undefined")
     isempty(p) && throw(ArgumentError("empty quantile array"))
 
     w.sum == 0 && error("weight vector cannot sum to zero")
-    length(v) == length(w) || error("data and weight vectors must be the same size, got $(length(v)) and $(length(w))")
+    length(v) == length(w) || error("data and weight vectors must be the same size, 
+        got $(length(v)) and $(length(w))")
```
nalimilan: Line breaks in strings are transformed to `\n`. Use two strings concatenated with `*` if you want to break them. Same below.

Also better use `throw(ArgumentError(...))` here and below (and actually everywhere, but...).

src/weights.jl
```diff
@@ -473,116 +473,50 @@ _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Nothing) where {T<:Numbe
...
         isnan(x) && error("weight vector cannot contain NaN entries")
         x < 0 && error("weight vector cannot contain negative entries")
     end
 
 
+    isa(w, FrequencyWeights) && !(eltype(w) <: Integer) && any(!isinteger, w) && 
+        error("The values of the vector of `FrequencyWeights` must be numerically equal to 
+        integers. Use `ProbabilityWeights` or `AnalyticWeights` instead.")
+
+    @inbounds for x in v
```
nalimilan: To be 100% certain the return type is type-stable, better call `fill!(x, out)` below, after allocating `out`.

src/weights.jl
```diff
@@ -628,25 +562,23 @@ function quantile(v::RealVector{V}, w::AbstractWeights{W}, p::RealVector) where
     return out
 end
 
-# similarly to statistics.jl in Base
+# similar function in Base statistics.jl
 function bound_quantiles(qs::AbstractVector{T}) where T<:Real
```
nalimilan: Somewhat unrelated, but this function has been removed from Base. We should probably remove it too.

src/weights.jl
```diff
@@ -473,116 +473,50 @@ _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Nothing) where {T<:Numbe
...
     isempty(p) && throw(ArgumentError("empty quantile array"))
 
     w.sum == 0 && error("weight vector cannot sum to zero")
-    length(v) == length(w) || error("data and weight vectors must be the same size, got $(length(v)) and $(length(w))")
+    length(v) == length(w) || error("data and weight vectors must be the same size, 
+        got $(length(v)) and $(length(w))")
     for x in w.values
         isnan(x) && error("weight vector cannot contain NaN entries")
         x < 0 && error("weight vector cannot contain negative entries")
     end
```
nalimilan: A single empty line is enough.

test/weights.jl
```diff
@@ -491,10 +407,30 @@ end
...
-    @test wquantile(data[1], w, [0.5])    ≈ [answer] atol = 1e-5
-    @test wquantile(data[1], w, 0.5)      ≈  answer atol = 1e-5
+end
+
+
+@testset "Median $f" for f in weight_funcs
+    data = [4, 3, 2, 1]
+    wt = [0, 0, 0, 0]
+    @test_throws ErrorException median(data, f(wt))
+    @test_throws ErrorException median((Float64)[], f((Float64)[]))
```
nalimilan:
```suggestion
    @test_throws ErrorException median(Float64[], f(Float64[]))
```

test/weights.jl
```diff
@@ -491,10 +407,30 @@ end
...
+    wt = [1, 2, 3, 4, 5]
+    @test_throws ErrorException median(data, f(wt))
+    if VERSION >= v"1.0"
+        @test_throws MethodError median([4 3 2 1 0], f(wt))
+        @test_throws MethodError median([[1 2];[4 5];[7 8];[10 11];[13 14]], f(wt))
```
nalimilan:
```suggestion
        @test_throws MethodError median([1 2; 4 5; 7 8; 10 11; 13 14], f(wt))
```

test/weights.jl
```diff
@@ -491,10 +407,30 @@ end
...
+    @test isnan(median(data, f(wt)))
+    wt = [1, 2, NaN, 4, 5]
+    @test_throws ErrorException median(data, f(wt))
+    data = [1, 3, 2, 1, 2]
+    @test_throws ErrorException median(data, f(wt))
+    wt = [-1, -1, -1, -1, -1]
+    @test_throws ErrorException median(data, f(wt))
+    wt = [-1, -1, -1, 0, 0]
+    @test_throws ErrorException median(data, f(wt))
 end
```
nalimilan: Also test a valid case?

src/deprecates.jl
matthieugomez: Isn't it ok to assume that default weights are probability weights, as it is now? Honestly, most of the time these differences don't matter, and I don't want users starting to feel overwhelmed by the exact weight type they should use.

In Stata:
> Each command has its own idea of the "natural" kind of weight.  The command will tell you what kind of weight it is assuming and perform the request as if you specified that kind of weight.

src/weights.jl
matthieugomez: No, it's correct. I added some docs that was in `median` about the behavior with NaN + error about frequency weights that are not integer

test/weights.jl
matthieugomez: done.

matthieugomez:
Incorporated your suggestion. Sorry I answered to your remarks as new comments — you can see them in "Files Changed".

src/weights.jl
nalimilan: The changes you mention are above. If you change the wrapping here and remove one sentence that my PR https://github.com/JuliaStats/StatsBase.jl/pull/458 also removes, it won't apply cleanly anymore.

matthieugomez: I tried to undo some of the change. Does that work for you?

matthieugomez: I don't understand why it gives conflicts.

matthieugomez: Ok I went over some git commands I barely understand and it seems to work now.

src/deprecates.jl
matthieugomez: I'd like that too. I don't think it's too late.

nalimilan:
> Honestly the difference does not matter for 99.99% of vectors. For me it'd be a bit like having to specify the type of quantile (1, 2... 7) etc every time I use the `quantile` function: while it would be more explicit, it would be much more cumbersome.

That's not really comparable IMHO, as these quantile types all aim to estimate the same quantity (and they converge with large sample sizes). OTC if you have frequency weights and you pass them without noticing they are treated as analytical weights, the interpretation is clearly incorrect. What's the point in spending so much work in finding the right definitions if in the end we don't think it matters? Also, people could be confused if they pass `weights(w)` for frequency weights and don't get the same result as replicating observations with the unweighted `quantile`.

>  In any case, if you still want to do it, maybe it'd be better to do it in a separate pull request?

Adding an error can wait for another PR, but the deprecation should suggest using `pweights` immediately, so that people don't use something which will throw an error soon.

src/deprecates.jl
nalimilan: I don't really like commands that talk to the user. Either a command succeeds and it should just do what is requested, or it doesn't and it should indicate a way to make it succeed. Printing warnings during normal operation is just annoying, we'd better explain how to do things properly. Adding `f`, `p` or `a` in front of `weights` isn't that costly, and anyway if users doesn't know what letter to use they won't understand the warning (and likely do something incorrect).

Also I don't think is common for Julia functions to print messages like that.

matthieugomez: I was talking about the second point ;)

matthieugomez:
Thanks!

PR #442, opened 2019-01-09
Use keyword argument for dims in scatter and weighted mean

### LINKED ISSUE #219: Scattermat documentation incorrect
```
help?> scattermat
search: scattermat

  scattermat(X, [wv::WeightVec]; mean=nothing, vardim=1)

  Compute the scatter matrix, which is an unnormalized covariance matrix. A weighting vector wv can be specified to weight the estimate.

     Arguments
    ≡≡≡≡≡≡≡≡≡≡≡

    •    mean=nothing: a known mean value. nothing indicates that the mean is


  unknown, and the function will compute the mean. Specifying mean=0 indicates that the data are centered and hence there's no need to subtract the mean.

    •    vardim=1: the dimension along which the variables are organized.


  When vardim = 1, the variables are considered columns with observations in rows; when vardim = 2, variables are in rows with observations in columns.
```

```
julia> scattermat(randn(3,3), mean=zeros(3))
ERROR: function scattermat does not accept keyword arguments
 in kwfunc(::Any) at ./boot.jl:236
```

```
julia --version
julia version 0.5.0
```

Looking at cov.jl, the scattermat function definitions are different depending on the julia version being run.

andreasnoack:
Fixes #219 

We also need to make these changes to weighted `var`, `std`, `mean_and_var`, `mean_and_std`, `cov`, and `cor`. Annoyingly, we didn't manage to make things consistent in `Statistics` so `dims` is `dimvar` in unweighted `cor` and `cov` so we might want to make that change to `Statistics` before changing them here.

### DIFF
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -85,7 +85,7 @@ _scattermatm(x::DenseMatrix, mean, dims::Int=1) =
 scattermat(x::DenseMatrix, wv::AbstractWeights; mean=nothing, dims::Int=1) =
     _scattermatm(x, wv, mean, dims)
 _scattermatm(x::DenseMatrix, wv::AbstractWeights, ::Nothing, dims::Int) =
-    _unscaled_covzm(x .- mean(x, wv, dims), wv, dims)
+    _unscaled_covzm(x .- mean(x, wv, dims=dims), wv, dims)
 _scattermatm(x::DenseMatrix, wv::AbstractWeights, mean, dims::Int) =
     _unscaled_covzm(x .- mean, wv, dims)
 
@@ -103,8 +103,8 @@ covm(x::DenseMatrix, mean, w::AbstractWeights, dims::Int=1;
     rmul!(scattermat(x, w, mean=mean, dims=dims), varcorrection(w, depcheck(:covm, corrected)))
 
 
-cov(x::DenseMatrix, w::AbstractWeights, vardim::Int=1; corrected::DepBool=nothing) =
-    covm(x, mean(x, w, vardim), w, vardim; corrected=depcheck(:cov, corrected))
+cov(x::DenseMatrix, w::AbstractWeights, dims::Int=1; corrected::DepBool=nothing) =
+    covm(x, mean(x, w, dims=dims), w, dims; corrected=depcheck(:cov, corrected))
 
 function corm(x::DenseMatrix, mean, w::AbstractWeights, vardim::Int=1)
     c = covm(x, mean, w, vardim; corrected=false)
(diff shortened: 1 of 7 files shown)


src/cov.jl
```diff
@@ -74,27 +75,36 @@ Finally, bias correction is applied to the covariance calculation if
...
+    _unscaled_covzm(x .- mean, dims)
+
+scattermat(x::DenseMatrix, wv::AbstractWeights; mean=nothing, dims::Int=1) =
+    _scattermatm(x, wv, mean, dims)
+_scattermatm(x::DenseMatrix, wv::AbstractWeights, ::Nothing, dims::Int) =
+    _unscaled_covzm(x .- mean(x, wv, dims=dims), wv, dims)
+_scattermatm(x::DenseMatrix, wv::AbstractWeights, mean, dims::Int) =
+    _unscaled_covzm(x .- mean, wv, dims)
+
+### Deprecated January 2019
```
nalimilan: Move these to deprecated.jl.

src/weights.jl
```diff
@@ -439,35 +439,45 @@ function wmean(v::AbstractArray{<:Number}, w::AbstractVector)
...
-"""
-mean!(R::AbstractArray, A::AbstractArray, w::AbstractWeights, dim::Int) =
-    rmul!(Base.sum!(R, A, w, dim), inv(sum(w)))
-
-wmeantype(::Type{T}, ::Type{W}) where {T,W} = typeof((zero(T)*zero(W) + zero(T)*zero(W)) / one(W))
-wmeantype(::Type{T}, ::Type{T}) where {T<:BlasReal} = T
-
-mean(A::AbstractArray{T}, w::AbstractWeights{W}, dim::Int) where {T<:Number,W<:Real} =
-    mean!(similar(A, wmeantype(T, W), Base.reduced_indices(axes(A), dim)), A, w, dim)
+mean(A::AbstractArray{T}, w::AbstractWeights{W}; dims=nothing) where {T<:Number,W<:Real} =
```
```suggestion
mean(A::AbstractArray{T}, w::AbstractWeights{W}; dims::Union{Int,Nothing}=nothing) where {T<:Number,W<:Real} =
```
Same elsewhere.
(We could accept any `Integer` BTW, that's what Base does AFAICT.)

src/weights.jl
```diff
@@ -469,16 +466,13 @@ w = rand(n)
 mean(x, weights(w))
 ```
 """
-mean(A::AbstractArray{T}, w::AbstractWeights{W}; dims=nothing) where {T<:Number,W<:Real} =
-    _mean(A, w, dims)
+mean(A::AbstractArray{T}, w::AbstractWeights{W};
+    dims::Union{Nothing,Int}=nothing) where {T<:Number,W<:Real} = _mean(A, w, dims)
```
nalimilan: Missing space.

PR #448, opened 2019-01-18
Improve entropy

oxinabox:
There is no reason to only allow entropy to be calculated on arrays
so this generalizes it to be for any iterable.

It also drops the requirement that the elements all be `Real`.
If somone wants to make a nonreal type, that has addition, multiplication and log,
and then wants to use `entropy` on that, then good for them.
(Realistic causes of this is something like symbolic math types).

Finally, It removes the thing where negative probabilities were just dropped from the calculation of entropy.
Because that is insane AFAIK.

If anything, we should be throwing an error in that case (which this will do, since `log` will throw).

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -450,22 +450,11 @@ zscore(X::AbstractArray{<:Real}, dim::Int) = ((μ, σ) = mean_and_std(X, dim); z
 """
     entropy(p, [b])
 
-Compute the entropy of an array `p`, optionally specifying a real number
+Compute the entropy of a collection of probabilities `p`, optionally specifying a real number
 `b` such that the entropy is scaled by `1/log(b)`.
 """
-function entropy(p::AbstractArray{T}) where T<:Real
-    s = zero(T)
-    z = zero(T)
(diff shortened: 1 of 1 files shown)


rofinn:
LGTM, but I'm not sure I agree about changing the negative probability behaviour. I feel like the old behaviour was more robust, but should thrown a warning.

oxinabox:
> LGTM, but I'm not sure I agree about changing the negative probability behaviour. I feel like the old behaviour was more robust, but should have thrown a warning.

How was it more robust?
I do not understand why under any circumstances anyone would want to do this ever.
This is even the wrong behavour as to what to do if numerical issues turn a `10^-16` into a `-1*10^-16`.
Since it is effectively changing every negitive "probability" into 1.

rofinn:
> Since it is effectively changing every negitive "probability" into 1 -- very distant from 10^-16

Good point. Since I was mostly thinking of handling numeric issues maybe we could include a `tol` keyword to bump zero and slightly negative values by?

rofinn:
Yeah, a probability number type would be nice.

Issue #449, opened 2019-01-18
Be less restrictive about things being AbstractArrays


oxinabox:
The vast majority of the time when we say `AbstractArray{<:Real}`
we only actually require that it be iterable.

Consider this pain:
```
julia> crossentropy(data.plant_id_eia, data.plant_id_pudl)
ERROR: MethodError: no method matching crossentropy(::Array{Union{Missing, Int64},1}, ::Array{Union{Missing, Int64},1})
Stacktrace:
 [1] top-level scope at none:0

julia> crossentropy(skipmissing(data.plant_id_eia), skipmissing(data.plant_id_pudl))
ERROR: MethodError: no method matching crossentropy(::Base.SkipMissing{Array{Union{Missing, Int64},1}}, ::Base.SkipMissing{Array{Union{Missing, Int64},1}})
Stacktrace:
 [1] top-level scope at none:0
```

PR #451, opened 2019-01-19
relax type constraints on means

oxinabox:
This contributtes towards #449 

Should be a little faster too, with any luck, not that I've benchmarked it

As an aside it is incredibly cute that  all 3 fancy means now depend on `mean`.

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -11,31 +11,17 @@
 """
     geomean(a)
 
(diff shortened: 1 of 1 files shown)


oxinabox:
I just removed the premature optimisation to avoid `float`ing unnescarily.
As A.) I had it backwards, and B.) the optimiser will do that for me during specialisation

oxinabox:
Merge?

PR #455, opened 2019-01-26
Generalise distance functions to work on all iterators

oxinabox:
As a rule all these changes are:

 - Slightly more allocating but much faster for acting on `AbstractArray`s
 - Work for all iterators.

Benchmarking done with
```
a = rand(1:10, 1000)
b = shuffle(a)
```

## sqL2dist
Also respects the type of the input

```
julia> @btime sum(abs2.($a.-$b))
  445.434 ns (1 allocation: 7.94 KiB)
16044

julia> @btime sqL2dist($a,$b)
  884.460 ns (0 allocations: 0 bytes)
16044.0
```

## L2dist

```
julia> @btime sqrt(sum(abs2, ($a .- $b)))
  534.909 ns (1 allocation: 7.94 KiB)
126.6649122685521

julia> @btime  L2dist($a,$b)
  884.920 ns (0 allocations: 0 bytes)
126.6649122685521
```

## L1dist

```
julia> @btime sum(abs.($a .- $b))
  425.117 ns (1 allocation: 7.94 KiB)
3250

julia> @btime L1dist($a,$b)
  887.367 ns (0 allocations: 0 bytes)
3250.0
```

## Linfdist

```
julia> @btime maximum(abs.($a .- $b))
  848.592 ns (1 allocation: 7.94 KiB)
9

julia> @btime Linfdist($a,$b)
  1.234 μs (0 allocations: 0 bytes)
9
```

## meanad

```
julia> @btime mean(abs.($a .- $b))
  422.730 ns (1 allocation: 7.94 KiB)
3.25

julia> @btime meanad($a, $b)
  887.521 ns (0 allocations: 0 bytes)
3.25
```

## msd
```
julia> @btime mean(abs2.($a .- $b))
  463.201 ns (1 allocation: 7.94 KiB)
16.044

julia> @btime msd($a, $b)
  884.840 ns (0 allocations: 0 bytes)
16.044
```

### DIFF
--- a/src/deviation.jl
+++ b/src/deviation.jl
@@ -5,10 +5,10 @@
 """
     counteq(a, b)
 
-Count the number of indices at which the elements of the arrays
+Count the number of indices at which the elements of the collectin
 `a` and `b` are equal.
 """
-function counteq(a::AbstractArray, b::AbstractArray)
+counteq(a::AbstractArray, b::AbstractArray) = sum(map(==, a,b))
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
     c = 0
@@ -43,28 +43,17 @@ end
 """
     sqL2dist(a, b)
 
-Compute the squared L2 distance between two arrays: ``\\sum_{i=1}^n |a_i - b_i|^2``.
-Efficient equivalent of `sumabs2(a - b)`.
+Compute the squared L2 distance between `a` and `b`: ``\\sum_{i=1}^n |a_i - b_i|^2``.
 """
-function sqL2dist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
-    r = 0.0
-    for i = 1:n
-        @inbounds r += abs2(a[i] - b[i])
-    end
-    return r
-end
-
+sqL2dist(a, b) = sum(abs2.(a .- b))
 
 # L2 distance
 """
     L2dist(a, b)
 
-Compute the L2 distance between two arrays: ``\\sqrt{\\sum_{i=1}^n |a_i - b_i|^2}``.
-Efficient equivalent of `sqrt(sumabs2(a - b))`.
+Compute the L2 distance between `a` and b`: ``\\sqrt{\\sum_{i=1}^n |a_i - b_i|^2}``.
 """
-L2dist(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:Number} = sqrt(sqL2dist(a, b))
+L2dist(a, b) = sqrt(sqL2dist(a, b))
 
 
 # L1 distance
@@ -72,17 +61,8 @@ L2dist(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:Number} = sqrt(sqL2di
     L1dist(a, b)
 
 Compute the L1 distance between two arrays: ``\\sum_{i=1}^n |a_i - b_i|``.
-Efficient equivalent of `sum(abs, a - b)`.
 """
-function L1dist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
-    r = 0.0
-    for i = 1:n
-        @inbounds r += abs(a[i] - b[i])
-    end
-    return r
-end
+L1dist(a, b) = sum(abs.(a .- b))
 
 
 # Linf distance
@@ -93,19 +73,7 @@ Compute the L∞ distance, also called the Chebyshev distance, between
 two arrays: ``\\max_{i\\in1:n} |a_i - b_i|``.
 Efficient equivalent of `maxabs(a - b)`.
 """
-function Linfdist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
-    r = 0.0
-    for i = 1:n
-        @inbounds v = abs(a[i] - b[i])
-        if r < v
-            r = v
-        end
-    end
-    return r
-end
-
+function Linfdist(a, b) = maximum(abs.(a .- b))
 
 # Generalized KL-divergence
 """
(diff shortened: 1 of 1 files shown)


src/deviation.jl
```diff
@@ -5,7 +5,7 @@
 """
     counteq(a, b)
 
-Count the number of indices at which the elements of the arrays
+Count the number of indices at which the elements of the arrays
```
nalimilan:
```suggestion
Count the number of indices at which the elements of the arrays
```

src/deviation.jl
```diff
@@ -43,46 +43,26 @@ end
...
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
-    r = 0.0
-    for i = 1:n
-        @inbounds r += abs2(a[i] - b[i])
-    end
-    return r
-end
-
+sqL2dist(a, b) = sum(abs2.(a .- b))
```
nalimilan: Better do `sum(abs2(x - y) for (x, y) in zip(a, b)))`. Same elsewhere.

oxinabox: I  benchmarked a number of options around this.
That one is slower.
Constructing a `zip` generally needs to be avoided to get best performance


```
julia> @btime sum(abs2(x-y) for (x,y) in zip($a,$b));
  796.989 ns (3 allocations: 64 bytes)
```

vs the current code I have (copy pasted from above)

```
julia> @btime sum(abs2.($a.-$b))
  445.434 ns (1 allocation: 7.94 KiB)
16044
```

nalimilan: Hmm, I don't get the same result here:
```julia
julia> @btime sum(abs2(x-y) for (x,y) in zip($a,$b));
  133.841 μs (3 allocations: 64 bytes)

julia> @btime sum(abs2.($a.-$b));
  158.430 μs (2 allocations: 781.33 KiB)
```

Also, `.-` only works for arrays, so that defeats the point of the PR (even if we could use it in a specialized method if it was faster).

oxinabox: I am testing on julia 1.1 that might be the source of differences.
(Or it could be system dependent, I feel like cache might be involved)

It was my understanding that as of julia 1.0 broadcast defaults to iterating.
Thus this is general, no?

```
julia> Iterators.drop([1,2,3], 1) .- [10,20]
2-element Array{Int64,1}:
  -8
 -17
```

nalimilan: Funny. The second variant is slower here both with 1.1 and master. I have a Skylake CPU. Do you get the same result if you wrap the code in functions?

> It was my understanding that as of julia 1.0 broadcast defaults to iterating.
> Thus this is general, no?

No, that's only the case for `HasShape` iterators:
```julia
julia> (i for i in [1,2,3]) .- [10,20]
ERROR: DimensionMismatch("arrays could not be broadcast to a common size")
```

Anyway I think we should use the generator approach as it avoids an intermediate allocation. Then it would be interesting to file an issue in Julia to investigate why it's faster on your machine.

oxinabox: Avoiding the intermidate allocation may not be worth it if it hurt cache locality.

II also have a Skylake CPU.
```
julia> versioninfo()
Julia Version 1.1.0
Commit 80516ca202 (2019-01-21 21:24 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

nalimilan: Looking at the code, what happens is that the second version is able to use SIMD, while the first one isn't. I guess depending on the CPU the gain due to SIMD can offset the overhead of making a copy, or not. It's hard to decide which approach should be the fastest in general without testing more CPUs. But in theory I imagine LLVM could improve and be able to use SIMD without making a copy with the first version.

nalimilan: Can you also try the original implementation after adding `@inbounds @simd` in front of the loop? In my tests, it's even faster. Using a single index for both arrays allows adding `@simd`, which isn't possible with `zip` (we could use `eachindex`, though, which would be more general).

Then we could keep both a specialized method for arrays and a more general one.

oxinabox: Predicting the future is hard.
 - LLVM could improve to use SIMD in the first version
 - Julia could improve to avoid the copy in the second one
 - Computers with better SIMD could dominate the market as time moves forward.

oxinabox: > Can you also try the original implementation after adding @inbounds @simd in front of the loop? 

At some point maybe. Tonight no.

more broadly I wonder if we should just deprecate all these into Distances.jl ?

nalimilan: Yes, prediction is hard. That's why I'd favor a simpler criterion: avoid allocating a copy. If you pass a very large vector, you don't necessarily have the space to allocate a copy (it could also be mmapped), and computing a simple sum shouldn't require that. And indeed the docs mention this as a rule: http://juliastats.github.io/StatsBase.jl/latest/deviation.html#Computing-Deviations-1 (at the bottom)

But I agree it could make sense to deprecate all of these. Let's hear what others think.

nalimilan: Any news on this?

oxinabox: I was convinced by your arguments to deprecate them here.
And then blocked by my desire to create a better way to do deprecations.

nalimilan: What better way? :-/

oxinabox: I have this plan for a package
https://github.com/invenia/Deprecations.jl/pull/1/files

nalimilan: Hmm, I'm not sure we want to take a new dependency just for deprecations. The current mechanism sounds good enough to me.

nalimilan: Bump.

oxinabox: I don't think I want to deprecate things in this PR.
That should be a different PR

nalimilan: But changing the implementation of something we want to deprecate immediately sounds like unnecessary churn. Does this PR touch functions that shouldn't be deprecated?

oxinabox: yes, `psnr` at least

nalimilan: OK, as you like. But you're going to spend much more time changing functions that are to be deprecated than on  `psnr`.

src/deviation.jl
```diff
@@ -182,10 +147,8 @@ end
 """
     psnr(a, b, maxv)
 
-Compute the peak signal-to-noise ratio between two arrays `a` and `b`.
+Compute the peak signal-to-noise ratio between `a` and `b`.
 `maxv` is the maximum possible value either array can take. The PSNR
 is computed as `10 * log10(maxv^2 / msd(a, b))`.
```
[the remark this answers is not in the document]
oxinabox: Errg good catch.
I will look into it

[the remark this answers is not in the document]
oxinabox: Wait, isn't it that same?
Under log identities?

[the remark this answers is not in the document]
nalimilan: Yeah, I just wonder what's the best way to represent it, taking into account simplicity and accuracy.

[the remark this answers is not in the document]
oxinabox: The way it is in the code right now is the best way, IMO.
Not run tests but it hits my instincts right.

PR #458, opened 2019-02-01
Make weighted quantile tests stricter and documentation more precise

nalimilan:
For `ProbabilityWeights` and `AnalyticWeights`, the magnitude of weights does not matter when they are equal. Also clean a few things.

CC: @matthieugomez

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -558,13 +558,14 @@ Weights must not be negative. The weights and data vectors must have the same le
 
 With [`FrequencyWeights`](@ref), the function returns the same result as
 `quantile` for a vector with repeated values.
+
 With non `FrequencyWeights`,  denote ``N`` the length of the vector, ``w`` the vector of weights,
-``h = p (\\sum_{i<= N}w_i - w_1) + w_1`` the cumulative weight corresponding to the
-probability ``p`` and ``S_k = \\sum_{i<=k}w_i`` the cumulative weight for each
+``h = p (\\sum_{i<= N} w_i - w_1) + w_1`` the cumulative weight corresponding to the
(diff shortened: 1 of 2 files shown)


matthieugomez:
Yes, very good.

nalimilan:
Thanks. BTW, is it necessary to sort the vector on both values *and weights* (see [this line](https://github.com/JuliaStats/StatsBase.jl/blob/1480c70a25eb86e479934173b70cfb30d98d889f/src/weights.jl#L589))? Otherwise I guess it could be slightly faster to sort only on values.

matthieugomez:
For future reference, note that the result of the weighted quantile function is sensitive to whether or not weights for observations with the same values are summed.

The reason is that it is impossible to have a weighted quantile definition that (i) is independent of the sum of weights (ii) gives the same result if weights are collapsed over observation with the same values (iii) gives the same result as unweighted quantile for weights equal to one. For instance:
```julia
1 = quantile([1, 1, 2, 2], 0.25)
= quantile([1, 1, 2, 2], Weights([1, 1, 1, 1]), 0.25) (by iii)
= quantile([1, 2], Weights([2, 2]), 0.25)  (by ii)
= quantile([1, 2], Weights([1, 1]), 0.25) (by i)
= 1.25
```julia


For now, the implementation satisfies (i) and (iii), but not (ii). Maybe it would be better to satisfy (i) and (ii) but not (iii) for `pweights'.

matthieugomez:
To answer your question, yes, it is necessary to sort by both since the order of weights matters within each value (this is related to the fact the definition does not satisfy (ii)). In any case, since `sortperm` is slower than `sort!`, I don't think this would be much more efficient (but I have not checked).

matthieugomez:
As you prefer! In any case, these are trade-offs to keep in mind for the future.

Issue #439, opened 2018-12-29
Interface for custom covariance estimation algorithms


mateuszbaran:
Here: https://github.com/JuliaStats/MultivariateStats.jl/pull/84#discussion_r244261509 an issue was raised that it would be a good idea to have an abstract interface for custom covariance estimation methods either in StatsBase or Statistics stdlib. I've recently been working on https://github.com/mateuszbaran/CovarianceEstimation.jl that provides such interface. I could make a PR here but I'm not sure what related methods of `cov` (if any) should be added to StatsBase.

nalimilan:
Sure. What would be needed concretely? Just `abstract type AbstractCovarianceEstimator end`?

Cc: @gragusa

nalimilan:
We can add an empty method for `::AbstractCovarianceEstimator`, with docstrings showing the API, like we do for statistical models. But defining a `SimpleCovarianceEstimator` in StatsBase doesn't sound very useful if we don't provide other types.

mateuszbaran:
What do you mean by an empty method for `::AbstractCovarianceEstimator`? API description could just be in the docstring for `AbstractCovarianceEstimator`.

As for `SimpleCovarianceEstimator`, I was thinking about it and I agree that it should remain in CovarianceEstimation.jl. API description in a docstring is sufficient.

mateuszbaran:
OK, so adding something like this:
```julia

abstract type AbstractCovarianceEstimator end

"""
    cov(x::AbstractVector, ce::AbstractCovarianceEstimator)

Return the sample variance of the vector `x` using estimator `ce`.
"""
cov(x::AbstractVector, ce::AbstractCovarianceEstimator) =
    error("cov is not defined for $(typeof(x)) and $(typeof(ce))")


"""
    cov(x::AbstractVector, y::AbstractVector, ce::AbstractCovarianceEstimator)

Return the covariance of the vectors `x` and `y` using estimator `ce`.
"""
cov(x::AbstractVector, y::AbstractVector, ce::AbstractCovarianceEstimator) =
    error("cov is not defined for $(typeof(x)), $(typeof(y)) and $(typeof(ce))")


"""
    cov(X::AbstractMatrix, ce::AbstractCovarianceEstimator, vardim::Int=1)

Return the covariance matrix associated with `X` along dimension `vardim`
using estimator `ce`.
"""
cov(X::AbstractMatrix, ce::AbstractCovarianceEstimator, vardim::Int=1) =
[...]

PR #460, opened 2019-02-03
Basic abstract covariance estimation interface

mateuszbaran:
I have prepared basic abstract covariance estimation support based on discussion in #439 . Let me know if there is anything I should change.

### DIFF
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -182,21 +182,20 @@ cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector) =
 
 Compute the covariance matrix of the matrix `X` along dimension `dims`
 using estimator `ce`. A weighting vector `w` can be specified.
-The keyword `mean` can be
+The keyword argument `mean` can be:
 
-    * `nothing` (default) in which case the mean is estimated then subtracted
+    * `nothing` (default) in which case the mean is estimated and subtracted
     from the data `X`,
-    * a given number in which case that provided mean is subtracted from the
-    data `X`.
+    * a precalculated mean in which case it is subtracted from the data `X`.
 """
-cov(ce::CovarianceEstimator, X::AbstractMatrix; dims::Int=1) =
+cov(ce::CovarianceEstimator, X::AbstractMatrix; mean=nothing, dims::Int=1) =
     error("cov is not defined for $(typeof(ce)) and $(typeof(X))")
 
-cov(ce::CovarianceEstimator, X::AbstractMatrix, w::AbstractWeights; dims::Int=1) =
+cov(ce::CovarianceEstimator, X::AbstractMatrix, w::AbstractWeights; mean=nothing, dims::Int=1) =
     error("cov is not defined for $(typeof(ce)), $(typeof(X)) and $(typeof(w))")
 
 """
-    SimpleCovariance(corrected::Bool)
+    SimpleCovariance(;corrected::Bool=false)
 
 Simple covariance estimator. Estimation calls `cov(x; corrected=corrected)`,
 `cov(x, y; corrected=corrected)` or `cov(X, w, dims; corrected=corrected)`
@@ -204,8 +203,8 @@ where `x`, `y` are vectors, `X` is a matrix and `w` is a weighting vector.
 """
 struct SimpleCovariance <: CovarianceEstimator
     corrected::Bool
+    SimpleCovariance(;corrected::Bool=false) = new(corrected)
 end
-SimpleCovariance(;corrected::Bool = false) = SimpleCovariance(corrected)
 
 cov(sc::SimpleCovariance, x::AbstractVector) =
     cov(x; corrected=sc.corrected)
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -184,9 +184,13 @@ Compute the covariance matrix of the matrix `X` along dimension `dims`
 using estimator `ce`. A weighting vector `w` can be specified.
 The keyword argument `mean` can be:
 
-    * `nothing` (default) in which case the mean is estimated and subtracted
-    from the data `X`,
-    * a precalculated mean in which case it is subtracted from the data `X`.
+* `nothing` (default) in which case the mean is estimated and subtracted
+  from the data `X`,
+* a precalculated mean in which case it is subtracted from the data `X`.
+  Assuming `size(X)` is `(N,M)`, `mean` can either be:
+  * when `dims=1`, an `AbstractMatrix` of size `(1,M)`,
+  * when `dims=2`, an `AbstractVector` of length `N` or an `AbstractMatrix`
+    of size `(N,1)`.
 """
 cov(ce::CovarianceEstimator, X::AbstractMatrix; mean=nothing, dims::Int=1) =
     error("cov is not defined for $(typeof(ce)) and $(typeof(X))")
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -153,3 +153,37 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
     end
     return C
 end
+
+"""
+    CovarianceEstimator
+
+Abstract type for covariance estimators.
+"""
+abstract type CovarianceEstimator end
+
+"""
+    cov(ce::CovarianceEstimator, x::AbstractVector)
+
+Compute the sample variance of the vector `x` using estimator `ce`.
+"""
+cov(ce::CovarianceEstimator, x::AbstractVector) =
+    error("cov is not defined for $(typeof(ce)) and $(typeof(x))")
+
+
+"""
+    cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector)
+
+Compute the covariance of the vectors `x` and `y` using estimator `ce`.
+"""
+cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector) =
+    error("cov is not defined for $(typeof(ce)), $(typeof(x)) and $(typeof(y))")
+
+
+"""
+    cov(ce::CovarianceEstimator, X::AbstractMatrix, vardim::Int=1)
+
+Compute the covariance matrix associated with `X` along dimension `vardim`
+using estimator `ce`.
+"""
+cov(ce::CovarianceEstimator, X::AbstractMatrix, vardim::Int=1) =
+    error("cov is not defined for $(typeof(ce)) and $(typeof(X))")
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -214,15 +214,19 @@ cov(sc::SimpleCovariance, x::AbstractVector, y::AbstractVector) =
     cov(x, y; corrected=sc.corrected)
 
 function cov(sc::SimpleCovariance, X::AbstractMatrix; dims::Int=1, mean=nothing)
-    @assert dims ∈ [1, 2] "Argument dims can only be 1 or 2 (given: $dims)"
+    dims ∈ (1, 2) || throw(ArgumentError("Argument dims can only be 1 or 2 (given: $dims)"))
     if mean === nothing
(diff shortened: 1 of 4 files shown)


src/cov.jl
```diff
@@ -153,3 +153,37 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+    cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector)
+
+Compute the covariance of the vectors `x` and `y` using estimator `ce`.
+"""
+cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector) =
+    error("cov is not defined for $(typeof(ce)), $(typeof(x)) and $(typeof(y))")
+
+
+"""
+    cov(ce::CovarianceEstimator, X::AbstractMatrix, vardim::Int=1)
```
nalimilan: While we're at it, `vardim` would better be a keyword argument (like `dims` in Base now). We should also change other methods (but this can be done in another PR).

src/cov.jl
```diff
@@ -153,3 +153,37 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+"""
+abstract type CovarianceEstimator end
+
+"""
+    cov(ce::CovarianceEstimator, x::AbstractVector)
+
+Compute the sample variance of the vector `x` using estimator `ce`.
+"""
+cov(ce::CovarianceEstimator, x::AbstractVector) =
+    error("cov is not defined for $(typeof(ce)) and $(typeof(x))")
```
nalimilan: How about providing a default estimator which would just call the standard `cov` function?

It would also make sense to add weights as an optional argument (in brackets), for consistency with `cov`.

src/cov.jl
mateuszbaran: OK, I'll change it then and prepare a separate PR for other methods.

src/cov.jl
mateuszbaran: Sure, I've got a default estimator here: https://github.com/mateuszbaran/CovarianceEstimation.jl/blob/master/src/simplecov.jl , so I can just move it to StatsBase.

Weights are a bit more problematic. I don't know how to make a generic implementation that works for all estimators, so most of them would not support it anyway. In CovarianceEstimation.jl I've put it in as a part of estimator but I can change it if you think it's a good idea. But estimator objects are not (and can't) be always data-independent anyway.

nalimilan: Regarding the `Simple` estimator, that's more code that I expected. Why can't it just call `cov`? Also the docs probably don't need to be repeated and can just point at the default method.

It's fine if only some estimators support weights. We just need to define a standard interface here. I'd rather keep weights out of the estimator object, passing them as an argument: they depend on the data, so you generally can't reuse an estimator for several operations if it stores weights.

src/cov.jl
nalimilan: Thanks. Actually I'm not sure whether it should be called `vardim` or `dims`. The former was probably more explicit, and we don't expect supporting multiple dimensions I guess.

mateuszbaran: `dims` is used in Statistics stdlib, so making it consistent makes sense. For me it works either way, so I'll do whatever you decide.

src/cov.jl
mateuszbaran: OK, makes sense. I'll trim the code and documentation.

I'll extract weights to an additional positional argument of `cov`. BTW, would you prefer to have `corrected` as a keyword argument? It somewhat depends on the data as well.

nalimilan: Having `corrected` in the estimator sounds fine, I'd say it depends more on the setting than on the actual data. It could almost be considered as a separate estimator.

src/cov.jl
nalimilan: OK since `cov` in `Statistics` uses `dims` I guess we don't really have a choice.

src/cov.jl
```diff
@@ -153,3 +153,37 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+    CovarianceEstimator
+
+Abstract type for covariance estimators.
+"""
+abstract type CovarianceEstimator end
+
+"""
+    cov(ce::CovarianceEstimator, x::AbstractVector)
+
+Compute the sample variance of the vector `x` using estimator `ce`.
```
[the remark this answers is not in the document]
mateuszbaran: Right, that was a mistake. `x` is a sample and the population variance is estimated (typically).

src/cov.jl
mateuszbaran: There is also some confusion because `cov` from `Statistics` has `dims` as a keyword argument and `StatsBase` has `vardim` that is a positional argument. I actually prefer the approach of `Statistics` as it's a bit more explicit.

src/cov.jl
```diff
@@ -178,12 +177,52 @@ Compute the covariance of the vectors `x` and `y` using estimator `ce`.
...
+SimpleCovariance(;corrected::Bool = false) = SimpleCovariance(corrected)
+
+cov(sc::SimpleCovariance, x::AbstractVector) =
+    cov(x; corrected=sc.corrected)
+
+cov(sc::SimpleCovariance, x::AbstractVector, y::AbstractVector) =
+    cov(x, y; corrected=sc.corrected)
+
+function cov(sc::SimpleCovariance, X::AbstractMatrix; dims::Int=1, mean=nothing)
+    @assert dims ∈ [1, 2] "Argument dims can only be 1 or 2 (given: $dims)"
```
nalimilan: Should be `throw(ArgumentError(...))`. Also better use `(1, 2)` to avoid an allocation.

src/cov.jl
```diff
@@ -178,12 +177,52 @@ Compute the covariance of the vectors `x` and `y` using estimator `ce`.
...
+    if mean === nothing
+        return cov(X; dims=dims, corrected=sc.corrected)
+    else
+        return cov(X .- mean; dims=dims, corrected=sc.corrected)
```
nalimilan: I think you can call `Statistics.covm`.

test/cov.jl
```diff
@@ -250,6 +250,26 @@ weight_funcs = (weights, aweights, fweights, pweights)
...
+        scc = SimpleCovariance(corrected=true)
+        @test cov(scc, X) ≈ cov(X, corrected=true)
+        @test cov(scc, X, mean=Xm) ≈ cov(X .- Xm, corrected=true)
+        @test cov(scc, X, mean=Xm) ≈ cov(X .- Xm, corrected=true)
+        if f !== weights
+            @test cov(scc, X, wv1, dims=1) ≈ cov(X, wv1, 1, corrected=true)
+            @test cov(scc, X, wv2, dims=2) ≈ cov(X, wv2, 2, corrected=true)
+        end
+
+        scu = SimpleCovariance(corrected=false)
```
nalimilan: Could use a loop over `corrected` to avoid duplication.

src/cov.jl
```diff
@@ -153,3 +153,76 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+    data `X`.
+"""
+cov(ce::CovarianceEstimator, X::AbstractMatrix; dims::Int=1) =
+    error("cov is not defined for $(typeof(ce)) and $(typeof(X))")
+
+cov(ce::CovarianceEstimator, X::AbstractMatrix, w::AbstractWeights; dims::Int=1) =
+    error("cov is not defined for $(typeof(ce)), $(typeof(X)) and $(typeof(w))")
+
+"""
+    SimpleCovariance(corrected::Bool)
```
nalimilan: Is "simple" the best term?

mateuszbaran: It's the best term I could think of but I'm open to suggestions.

src/cov.jl
```diff
@@ -153,3 +153,80 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+
+Simple covariance estimator. Estimation calls `cov(x; corrected=corrected)`,
+`cov(x, y; corrected=corrected)` or `cov(X, w, dims; corrected=corrected)`
+where `x`, `y` are vectors, `X` is a matrix and `w` is a weighting vector.
+"""
+struct SimpleCovariance <: CovarianceEstimator
+    corrected::Bool
+end
+SimpleCovariance(;corrected::Bool = false) = SimpleCovariance(corrected)
```
nalimilan: Maybe make this an inner constructor if `SimpleCovariance(::Bool)` isn't supposed to be used (and adjust docstring).

src/cov.jl
```diff
@@ -153,3 +153,80 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+"""
+    cov(ce::CovarianceEstimator, X::AbstractMatrix, [w::AbstractWeights]; mean=nothing, dims::Int=1)
+
+Compute the covariance matrix of the matrix `X` along dimension `dims`
+using estimator `ce`. A weighting vector `w` can be specified.
+The keyword `mean` can be
+
+    * `nothing` (default) in which case the mean is estimated then subtracted
+    from the data `X`,
+    * a given number in which case that provided mean is subtracted from the
```
nalimilan: I thought I had noted this, but I can't find it: `mean` should be a vector, not a number right? Signatures should be adjusted everywhere. Also, the fallback methods below should accept it.

Above, say "keyword *argument*", and add `:`.

src/cov.jl
mateuszbaran: OK, good idea.

src/cov.jl
mateuszbaran: Yes, `mean` is not a number but technically it doesn't need to be a `Vector` so maybe it's better to be a bit more vague here.

mateuszbaran:
I'm looking at this interface again and I'm wondering if `cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector)` should accept means of `x` and `y` as keyword arguments?

src/cov.jl
nalimilan: I guess it needs to be an `AbstractVector` at least, right? It can also be useful to mention that it must have a length matching the dimension not specified by `dims`.

nalimilan:
> I'm looking at this interface again and I'm wondering if `cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector)` should accept means of `x` and `y` as keyword arguments?

That makes sense, but unless we have an implementation which takes these arguments, maybe better leave them out for now.

src/cov.jl
```diff
@@ -153,3 +153,37 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+Compute the covariance of the vectors `x` and `y` using estimator `ce`.
+"""
+cov(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector) =
+    error("cov is not defined for $(typeof(ce)), $(typeof(x)) and $(typeof(y))")
+
+
+"""
+    cov(ce::CovarianceEstimator, X::AbstractMatrix; dims::Int=1)
+
+Compute the covariance matrix associated with `X` along dimension `dims`
```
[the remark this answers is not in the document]
mschauer: without "associated"

src/cov.jl
```diff
@@ -153,3 +153,83 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
+    CovarianceEstimator
+
+Abstract type for covariance estimators.
+"""
+abstract type CovarianceEstimator end
+
+"""
+    cov(ce::CovarianceEstimator, x::AbstractVector; mean=nothing)
+
+Compute the variance of the vector `x` using estimator `ce`.
```
mschauer:
```suggestion
Compute a variance estimate from the observation vector `x` using the  estimator `ce`.
```

src/cov.jl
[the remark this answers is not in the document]
mateuszbaran: Yes, I've already removed the word "associated".

PR #463, opened 2019-02-10
Remove bound_quantiles

nalimilan:
`Statistics.quantile` no longer applies this transformation.


matthieugomez:
In this case,  I guess you would need to add an error if the quantile is not in [0, 1] similar to [Base](https://github.com/JuliaLang/julia/blob/e5e1253b5ec4ed3fb5b6e12f72dda1dd4348e667/stdlib/Statistics/src/Statistics.jl#L857), and maybe add some `trunc` too

nalimilan:
Good catch, I've added a check. I'm not sure I understand where the truncation should be added, though. I guess it would apply only to frequency weights? Maybe better you make a PR, otherwise I'm just going to ask you to post the code. :-)

matthieugomez:
Sorry, you're right, I don't think trunc is really needed.

nalimilan:
Good to go?

Issue #430, opened 2018-11-23
describe with missing values


nalimilan:
`describe` should treat `AbstractArray{Union{Missing, <:Real}}` like `AbstractArray{<:Real}`, skipping missing values and giving their number. Currently, the output is not very useful:
```julia
julia> describe([1,missing])
Summary Stats:
Length:         2
Type:           Union{Missing, Int64}
Number Unique:  2

julia> describe([1])
Summary Stats:
Mean:           1.000000
Minimum:        1.000000
1st Quartile:   1.000000
Median:         1.000000
3rd Quartile:   1.000000
Maximum:        1.000000
Length:         1
Type:           Int64
```

PR #465, opened 2019-02-22
 Handle missing values and empty inputs in SummaryStats

ararslan:
Currently, `SummaryStats` will error when computing quantiles if the input is empty or contains any missing values. Instead, we can simply inform the user that the input is empty or that there were missing
values that caused the statistics to be missing.

I've also added a Project.toml file as a separate commit, since they're indispensable for local development and no longer confuse METADATA CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 docs/build
 docs/site
 perf/*.csv
+Manifest.toml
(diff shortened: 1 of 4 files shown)


matbesancon:
ok let's ship this :boat:

nalimilan:
We should probably call `skipmissing` for vectors that allow for missing values. `describe(::DataFrame)` does that and that's much more useful than bailing out.

ararslan:
That's a good idea. Could you open an issue to track that?

nalimilan:
Actually that's https://github.com/JuliaStats/StatsBase.jl/issues/430.

PR #468, opened 2019-02-27
fix entropy

matbesancon:
There was a bug in the definition of entropy, we should have entropy of a Dirac peak = 0.0

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -438,16 +438,7 @@ zscore(X::AbstractArray{<:Real}, dim::Int) = ((μ, σ) = mean_and_std(X, dim); z
 Compute the entropy of a collection of probabilities `p`, optionally specifying a real number
 `b` such that the entropy is scaled by `1/log(b)`.
 """
-function entropy(p)
-    s = zero(eltype(p))
-    z = s
-    for pi in p
-        if pi > z
-            s -= pi * log(pi)
-        end
-	end
-  return s
-end
+entropy(p) = -sum(pᵢ -> iszero(pᵢ) ? zero(pᵢ) : pᵢ * log(pᵢ), p)
 
 entropy(p, b::Real) = entropy(p) / log(b)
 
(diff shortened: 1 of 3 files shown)


src/scalarstats.jl
```diff
@@ -438,7 +438,17 @@ zscore(X::AbstractArray{<:Real}, dim::Int) = ((μ, σ) = mean_and_std(X, dim); z
 Compute the entropy of a collection of probabilities `p`, optionally specifying a real number
 `b` such that the entropy is scaled by `1/log(b)`.
 """
-entropy(p) = -sum(pi -> pi * log(pi), p)
+function entropy(p)
```
andreasnoack: I'd forgotten that `xlogx` isn't available here. We should probably move it here togeher with things like `logistic`. 

Probably better to just mimic `xlogx` with something like `-sum(pᵢ -> iszero(pᵢ) ? zero(pᵢ) : pᵢ * log(pᵢ), p)` since it will throw for negative probabilities.

nalimilan: `xlogx` sounds so general that it could live in Base, no? We're trying to move things to Statistics so it could go there too. Anyway it doesn't make a lot of sense in StatsFuns.

simonbyrne: Why can't you just call `StatsFuns.xlogx`?

matbesancon: Importing StatsFuns (which depends on R) just for xlogx seemed overkill

simonbyrne: Ah I thought it was already imported. Yes, we could definitely add it here.

oxinabox:
You really should have added to the docstring about this behaviour.

test/scalarstats.jl
```diff
@@ -119,6 +119,8 @@ it = (xᵢ for xᵢ in x)
 
 @test entropy([0.5, 0.5],2)       ≈ 1.0
 @test entropy([0.2, 0.3, 0.5], 2) ≈ 1.4854752972273344
+@test entropy([0.2, 0.3, 0.5], 2) ≈ 1.4854752972273344
```
nstiurca: FYI, it looks like you duplicated one of the tests here.

Issue #342, opened 2018-01-19
`sem` does not accept iterables


piever:
The typing of `sem` may be too strict:

```julia
julia> itr = (i for i in 1:3)
Base.Generator{UnitRange{Int64},##3#4}(#3, 1:3)

julia> mean(itr)
2.0

julia> var(itr)
1.0

julia> std(itr)
1.0

julia> sem(itr)
ERROR: MethodError: no method matching sem(::Base.Generator{UnitRange{Int64},##3#4})
Closest candidates are:
  sem(::AbstractArray{#s26,N} where N where #s26<:Real) at /home/pietro/.julia/v0.6/StatsBase/src/scalarstats.jl:237
```

Could we simply relax the definition [here](https://github.com/JuliaStats/StatsBase.jl/blob/558d8cfb4be0ebea755ffd92ecdbfd46e4446d2c/src/scalarstats.jl#L237) and allow it to accept anything?

nalimilan:
I think this problem is quite general, we often require arrays of `Real` when we actually only need iterables. It would make sense to go over functions to check which one could use a broader signature. This is particularly important for `Array{Union{<:Real, Missing}}` and for passing `skipmissing(x)`.

piever:
I hadn't thought about it, but after [#280](https://github.com/JuliaStats/DataArrays.jl/pull/280) things don't work even with `DataArrays`:

```julia
julia> sem(@data [1,2,3])
ERROR: MethodError: no method matching sem(::DataArrays.DataArray{Int64,1})
Closest candidates are:
  sem(::AbstractArray{#s26,N} where N where #s26<:Real) at /home/pietro/.julia/v0.6/StatsBase/src/scalarstats.jl:237
```

piever:
A possible plan could be the following, for all of these functions:

```julia
# actual definition for iterables
sem(a) = sqrt(var(a) / length(a))

# automatic missing handling
sem(a::AbstractArray{>: Missing}) = sem(skipmissing(a))
```

I'm also wondering whether `sem` of an empty iterator should be missing.

piever:
I've tried going through scalarstats.jl to see how many functions could widen their signature (to work with `skipmissing` but it's actually a bit tricky). There are at least 3 cases:

1. It would simply work by widening the signature (`sem`, `variation`)
[...]

PR #472, opened 2019-02-28
Accept any iterable in scalar stats

nalimilan:
Also improve a few docstrings.

Part of https://github.com/JuliaStats/StatsBase.jl/issues/449. Closes https://github.com/JuliaStats/StatsBase.jl/issues/342. Cc: @oxinabox


src/scalarstats.jl
```diff
@@ -105,17 +105,18 @@ function modes(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
     return ms
 end
 
-# compute mode over arbitrary array
-function mode(a::AbstractArray{T}) where T
-    isempty(a) && error("mode: input array cannot be empty.")
-    cnts = Dict{T,Int}()
+# compute mode over arbitrary iterable
+function mode(a)
+    isempty(a) && error("mode: input collection cannot be empty.")
```
oxinabox: Should be an `ArgumentException`

src/scalarstats.jl
```diff
@@ -126,19 +127,22 @@ function mode(a::AbstractArray{T}) where T
...
+        y = iterate(a, st)
     end
     return mv
 end
 
-function modes(a::AbstractArray{T}) where T
-    isempty(a) && error("modes: input array cannot be empty.")
-    cnts = Dict{T,Int}()
+function modes(a)
+    isempty(a) && error("modes: input collection cannot be empty.")
```
oxinabox: should throw and `ArgumentException`

src/scalarstats.jl
```diff
@@ -197,75 +194,117 @@ nquantile(v::AbstractArray{<:Real}, n::Integer) = quantile(v, (0:n)/n)
...
+function sem(x)
+    y = iterate(x)
+    if y === nothing
+        T = eltype(x)
+        return oftype(sqrt((abs2(zero(T)) + abs2(zero(T)))/2), NaN)
+    end
+    count = 1
+    value, state = y
+    y = iterate(x, state)
+    # Use Welford algorithm as seen in (among other places)
```
nalimilan: It's too bad I needed to copy the code from `var`, but if we move this to Statistics it will be easy to share the code in a common internal method.

src/scalarstats.jl
```diff
@@ -197,75 +194,117 @@ nquantile(v::AbstractArray{<:Real}, n::Integer) = quantile(v, (0:n)/n)
...
 
-Return the coefficient of variation of an array `x`, optionally specifying
+Return the coefficient of variation of collection `x`, optionally specifying
 a precomputed mean `m`. The coefficient of variation is the ratio of the
 standard deviation to the mean.
 """
-variation(x::AbstractArray{<:Real}, m::Real) = stdm(x, m) / m
-variation(x::AbstractArray{<:Real}) = variation(x, mean(x))
+variation(x, m) = stdm(x, m) / m
+variation(x) = ((m, s) = mean_and_std(x); s/m)
```
oxinabox:
```suggestion
variation(x) = \(mean_and_std(x)...)
```

This is is a bit gross, but I still think less gross than multistatement single-liners

src/scalarstats.jl
```diff
@@ -197,75 +194,117 @@ nquantile(v::AbstractArray{<:Real}, n::Integer) = quantile(v, (0:n)/n)
...
-    S = promote_type(T, typeof(middle(first(v))))
-    v2 = LinearAlgebra.copy_oftype(v, S)
+function mad(x; center=nothing, normalize::Union{Bool, Nothing}=nothing)
+    isempty(x) && throw(ArgumentError("mad is not defined for empty collections"))
+
+    if x isa AbstractArray
+        T = eltype(x)
+        S = promote_type(T, typeof(middle(zero(T))))
+        x2 = LinearAlgebra.copy_oftype(x, S)
+    else # Use type of first element
```
nalimilan: Using the first element sounds a bit arbitrary. A possibly better alternative would be to progressively widen the element type as we encounter new types, but that potentially requires allocating several copies.

src/scalarstats.jl
nalimilan: Why? :-/

src/scalarstats.jl
```diff
@@ -197,75 +194,117 @@ nquantile(v::AbstractArray{<:Real}, n::Integer) = quantile(v, (0:n)/n)
...
-Return the standard error of the mean of `a`, i.e. `sqrt(var(a) / length(a))`.
+Return the standard error of the mean of collection `x`,
+i.e. `sqrt(var(x, corrected=true) / length(x))`.
 """
-sem(a::AbstractArray{<:Real}) = sqrt(var(a) / length(a))
+function sem(x)
+    y = iterate(x)
+    if y === nothing
+        T = eltype(x)
+        return oftype(sqrt((abs2(zero(T)) + abs2(zero(T)))/2), NaN)
```
oxinabox: This is mildly insane,
but I see why you have to do it.

Probably should comment on it.

Noting that

```
julia> T = Complex
Complex

julia> convert(T, NaN)
NaN + 0.0im

julia> oftype(sqrt((abs2(zero(T)) + abs2(zero(T)))/2), NaN)
NaN
```

nalimilan: Note that this is a slightly adaption from `Base.var`, so I'd rather not change it.

src/scalarstats.jl
```diff
@@ -197,75 +194,117 @@ nquantile(v::AbstractArray{<:Real}, n::Integer) = quantile(v, (0:n)/n)
...
-             center::Union{Real,Nothing}=nothing,
-             normalize::Union{Bool, Nothing}=nothing) where T<:Real
-    isempty(v) && throw(ArgumentError("mad is not defined for empty arrays"))
-
-    S = promote_type(T, typeof(middle(first(v))))
-    v2 = LinearAlgebra.copy_oftype(v, S)
+function mad(x; center=nothing, normalize::Union{Bool, Nothing}=nothing)
+    isempty(x) && throw(ArgumentError("mad is not defined for empty collections"))
+
+    if x isa AbstractArray
```
oxinabox: break these two into two helper methods?

src/scalarstats.jl
oxinabox: too many equals  signs.
Though feel free to disagree

src/scalarstats.jl
nalimilan: That would force duplicating code...

nalimilan: Or I guess you mean calling a method from here? I don't think that's clearer (unless we use it elsewhere, but that's not the case currently).

oxinabox:
A solid improvement on the state of the world

src/scalarstats.jl
oxinabox: Fair.

I wonder if there is a further abstraction we need to help with working out types.
A lot of that code is pretty hairy

src/scalarstats.jl
oxinabox: I am not suggesting changing it.
But explaining it might be good.
Even if it is just 
`Return the Nan of the type that we would get, had this collection contained any elements`

src/scalarstats.jl
oxinabox: This would be fine to progressively widen, since it is very rare because most containers are homogenious.

src/scalarstats.jl
ararslan: FWIW I prefer what's written currently.

src/scalarstats.jl
nalimilan: Actually it sounds simpler to make a copy when the element type is not concrete. That way we choose the correct element type. Anyway performance doesn't matter a lot in that case and a copy is cheap compared with computing the median.

piever:
Nice! Is this good to go? I think it's a really important improvement now that  `mean(skipmissing(v))` is the recommended way to compute summary statistics ignoring missing values.

src/moments.jl
```diff
@@ -153,66 +153,68 @@ std(v::RealArray, w::AbstractWeights, dim::Int; mean=nothing,
...
+function mean_and_std(x; corrected::Bool=true)
+    m = mean(x)
+    s = stdm(x, m; corrected=corrected)
     m, s
 end
 
-function mean_and_var(A::RealArray, w::AbstractWeights; corrected::DepBool=nothing)
-    m = mean(A, w)
-    v = varm(A, w, m; corrected=depcheck(:mean_and_var, corrected))
+function mean_and_var(x::RealArray, w::AbstractWeights; corrected::DepBool=nothing)
```
bkamins: Why do we still expect `RealArray` here? I thought this should also work with other arrays. E.g. `Vector{Union{Missing, Float64}}` as long as it does not contain `missing` (which will simply throw an error at run time if it does). Same questions for methods below.

nalimilan: Yeah, there are many other things to fix in this file. I'd rather fix them separately in another PR.

bkamins: It is OK to leave it for later. I reacted because you explicitly named this PR "Accept any iterable ..." 😄 so I thought this is exactly the change you want to make in this PR.

nalimilan: Yeah the title is a bit too ambitious.

nalimilan:
OK to merge then?

PR #476, opened 2019-03-04
Allow mismatched element types for weighted means

### LINKED ISSUE #475: Method for weighted mean with mismatched types is gone
The fallback treats the first argument as a function apparently??
```julia
julia> mean(Union{Int,Missing}[1,2,3], weights([1,2,3]))
4.666666666666667

julia> @which mean(Union{Int,Missing}[1,2,3], weights([1,2,3]))
mean(f, A::AbstractArray) in Statistics at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/Statistics/src/Statistics.jl:76
```

ararslan:
We've recently inadvertently disallowed a mismatch between the array's element type and the element type of the weights when computing a weighted mean. This leads to bizarre behavior, as it dispatches to a nonsensical method where the array is treated as a function.

Fixes #475

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -466,11 +466,11 @@ w = rand(n)
 mean(x, weights(w))
 ```
 """
-mean(A::AbstractArray{T}, w::AbstractWeights{W};
-     dims::Union{Nothing,Int}=nothing) where {T<:Number,W<:Real} = _mean(A, w, dims)
-_mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Nothing) where {T<:Number,W<:Real} =
+mean(A::AbstractArray, w::AbstractWeights; dims::Union{Nothing,Int}=nothing) =
+    _mean(A, w, dims)
+_mean(A::AbstractArray, w::AbstractWeights, dims::Nothing) =
     sum(A, w) / sum(w)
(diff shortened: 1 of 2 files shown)


src/weights.jl
```diff
@@ -466,11 +466,11 @@ w = rand(n)
 mean(x, weights(w))
 ```
 """
-mean(A::AbstractArray{T}, w::AbstractWeights{W};
```
nalimilan: Better keep `W<:Real` for consistency as we do that everywhere. We could probably remove it, but let's do that separately (for example when moving these to Statistics).

nalimilan:
Good catch. But AFAICT the problem isn't the type mismatch, it's just that https://github.com/JuliaStats/StatsBase.jl/pull/442 added the `T<:Number` requirement (which previously applied only when `dims` was provided, which didn't really make sense).

src/weights.jl
ararslan: We don't in `mean!`, which is why I removed it--for consistency. :upside_down_face:

ararslan:
> But AFAICT the problem isn't the type mismatch ...

Oh yeah true. We can just change the commit message on merge to make it more accurate.

Issue #456, opened 2019-01-27
Strange histogram of mono-valued vector


GunnarFarneback:
This looks very strange to me:
```
julia> fit(Histogram, fill(0.5, 10))
Histogram{Int64,1,Tuple{StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}}
edges:
  1.5:2.0:3.5
weights: [0]
closed: left
isdensity: false
```
Traces back to
```
julia> StatsBase.histrange(0.5, 0.5, 5, :left)
1.5:2.0:3.5
```
Cc @simonbyrne

PR #479, opened 2019-03-06
handle `histrange(lo,hi,n)` when `lo~=hi`

smldis:
Fixes JuliaStats/StatsBase.jl#456.
It is trivial attempt that does not modify the other code path.

See also https://github.com/JuliaPlots/StatsMakie.jl/issues/62

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -40,46 +40,54 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where 
 end
 
 function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
-    if hi == lo
-        start = F(hi)
-        step = one(F)
-        divisor = one(F)
-        len = one(F)
-    else
-        bw = (F(hi) - F(lo)) / n
-        lbw = log10(bw)
-        if lbw >= 0
-            step = exp10(floor(lbw))
-            r = bw / step
-            if r <= 1.1
-                nothing
-            elseif r <= 2.2
-                step *= 2
-            elseif r <= 5.5
-                step *= 5
-            else
-                step *= 10
-            end
-            divisor = one(F)
-            start = step*floor(lo/step)
-            len = ceil((hi - start)/step)
+    closed == :left || closed ==:right || 
+        throw(ArgumentError("closed must be :left or :right"))
+    lo = float(lo)
+    hi = float(hi)
+    if abs(hi-lo) < n*10eps(F)
+        if closed == :left
+            hi = hi + n*10eps(F)
+        elseif closed ==:right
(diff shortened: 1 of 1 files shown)


src/hist.jl
```diff
@@ -40,46 +40,54 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
 end
 
 function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
-    if hi == lo
-        start = F(hi)
-        step = one(F)
+    closed == :left || closed ==:right ||
```
ararslan:
```suggestion
    closed === :left || closed === :right || 
```

src/hist.jl
```diff
@@ -40,46 +40,54 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
...
-    if hi == lo
-        start = F(hi)
-        step = one(F)
+    closed == :left || closed ==:right || 
+        throw(ArgumentError("closed must be :left or :right"))
+    lo = float(lo)
+    hi = float(hi)
+    if abs(hi-lo) < n*10eps(F)
+        if closed == :left
+            hi = hi + n*10eps(F)
```
ararslan:
```suggestion
            hi += n*10eps(F)
```

src/hist.jl
```diff
@@ -40,46 +40,54 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
...
+        elseif closed ==:right
+            lo = lo - n*10eps(F)
```
ararslan:
```suggestion
            lo -= n*10eps(F)
```

src/hist.jl
```diff
@@ -40,46 +40,54 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
...
+        else
+            # never happens
```
ararslan: I'd just delete the `else` block. The fact that this doesn't happen is apparent from the error check above.

src/hist.jl
```diff
@@ -40,46 +40,54 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
...
+        end
+    end
+
+    bw = (F(hi) - F(lo)) / n
+    lbw = log10(bw)
+    if lbw >= 0
+        step = exp10(floor(lbw))
+        r = bw / step
+        if r <= 1.1
+            nothing
```
ararslan: ? Why not just omit this condition and start the check with `r <= 2.2`?

smldis: I did not modify this code path, I will have a look at it

ararslan: Oh sorry, I misread the diff. Don't worry about it for now.

PR #480, opened 2019-03-29
Use project to build Documenter docs

nalimilan:
This should also fix building the docs, which haven't been deployed for one year (!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,13 @@ git:
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("StatsBase"); Pkg.test("StatsBase"; coverage=true)'
 after_success:
   - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder()); Codecov.submit(process_folder())';
-  - julia -e 'using Pkg; Pkg.add("Documenter"); include(joinpath("docs", "make.jl"))'
+jobs:
+  include:
+    - stage: "Documentation"
+      julia: 1.0
+      os: linux
+      script:
+        - julia --project=docs/ -e 'using Pkg; Pkg.instantiate();
+                                    Pkg.develop(PackageSpec(path=pwd()))'
+        - julia --project=docs/ docs/make.jl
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -17,4 +17,13 @@ git:
...
   - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder()); Codecov.submit(process_folder())';
-  - julia -e 'using Pkg; Pkg.add("Documenter"); include(joinpath("docs", "make.jl"))'
+jobs:
+  include:
+    - stage: "Documentation"
+      julia: 1.0
+      os: linux
+      script:
+        - julia --project=docs/ -e 'using Pkg; Pkg.instantiate();
+                                    Pkg.develop(PackageSpec(path=pwd()))'
```
ararslan: Develop should come before instantiate

docs/make.jl
```diff
@@ -1,7 +1,11 @@
 using Documenter, StatsBase, Statistics, Random
 
+# Workaround for JuliaLang/julia/pull/28625
+if Base.HOME_PROJECT[] !== nothing
+    Base.HOME_PROJECT[] = abspath(Base.HOME_PROJECT[])
+end
+
 makedocs(
-    format = :html,
```
ararslan: I think this should now be `format = Documenter.HTML()`

.travis.yml
nalimilan: Fixed. I just copied this from DataFrames. Any idea whether it really makes a difference?

docs/make.jl
nalimilan: That's the default now: https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.makedocs

.travis.yml
ararslan: Yes, if you instantiate first, some weird version resolution things can happen. I've run into them before and asked in #pkg-usage on Slack, and this was Fredrik's fix. I can't recall the full details of the problems it can cause though.

PR #484, opened 2019-04-07
Add convergenceexception information

galenlynch:
In order to add informative hints to `ConvergenceException`s, I have added a `info` field to `ConvergenceException` structs. If the `info` field is not empty, it will also be printed with the error message for `ConvergenceExceptions`. This is useful to inform the user of why he or she may be experiencing problems with convergence, and to suggest ways to avoid these errors. For an example of this, see JuliaStats/GLM.jl#303.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -459,8 +459,15 @@ ConvergenceException(iters, lastchange::T=NaN, tol::T=NaN) where {T<:Real} =
     ConvergenceException{T}(iters, lastchange, tol)
 
 function Base.showerror(io::IO, ce::ConvergenceException)
-    print(io, "failure to converge after $(ce.iters) iterations.")
+    print(io, "failure to converge after ", ce.iters, " iterations.")
     if !isnan(ce.lastchange)
-        print(io, " Last change ($(ce.lastchange)) was greater than tolerance ($(ce.tol)).")
+        print(
+            io,
+            " Last change (",
+            ce.lastchange,
+            ") was greater than tolerance (",
+            ce.tol,
+            ")."
+        )
     end
 end
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -447,9 +447,7 @@ struct ConvergenceException{T<:Real} <: Exception
     lastchange::T
     tol::T
     msg::String
-    function ConvergenceException{T}(
-        iters, lastchange::T, tol::T, msg::String
-    ) where T<:Real
+    function ConvergenceException{T}(iters, lastchange::T, tol::T, msg::String) where T<:Real
         if tol > lastchange
             throw(ArgumentError("Change must be greater than tol."))
         else
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -462,16 +462,9 @@ ConvergenceException(
     ConvergenceException{T}(iters, lastchange, tol, string(msg))
 
 function Base.showerror(io::IO, ce::ConvergenceException)
-    print(io, "failure to converge after ", ce.iters, " iterations.")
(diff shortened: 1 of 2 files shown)


src/statmodels.jl
```diff
@@ -446,21 +446,36 @@ struct ConvergenceException{T<:Real} <: Exception
...
-ConvergenceException(iters, lastchange::T=NaN, tol::T=NaN) where {T<:Real} =
-    ConvergenceException{T}(iters, lastchange, tol)
+ConvergenceException(
+    iters, lastchange::T=NaN, tol::T=NaN, msg::AbstractString = ""
+) where {T<:Real} =
+    ConvergenceException{T}(iters, lastchange, tol, string(msg))
 
 function Base.showerror(io::IO, ce::ConvergenceException)
-    print(io, "failure to converge after $(ce.iters) iterations.")
+    print(io, "failure to converge after ", ce.iters, " iterations.")
```
nalimilan: Why change this?

src/statmodels.jl
```diff
@@ -446,21 +446,36 @@ struct ConvergenceException{T<:Real} <: Exception
     iters::Int
     lastchange::T
     tol::T
-    function ConvergenceException{T}(iters, lastchange::T, tol::T) where T<:Real
+    msg::String
+    function ConvergenceException{T}(
```
nalimilan: Please use the same style as elsewhere, with arguments on the same line as parens.

test/statmodels.jl
```diff
@@ -36,5 +36,9 @@ x3   0.453058  0.72525 0.999172 0.5567
 @test sprint(showerror, ConvergenceException(10, 0.2, 0.1)) ==
     "failure to converge after 10 iterations. Last change (0.2) was greater than tolerance (0.1)."
 
+@test sprint(showerror, ConvergenceException(10, 0.2, 0.1, "Try changing maxIter.")) ==
+    "failure to converge after 10 iterations. Last change (0.2) was greater than tolerance (0.1). Try changing maxIter."
+
```
nalimilan: A single break is enough.

src/statmodels.jl
galenlynch: It makes an intermediate string  where it's not necessary to do so. It's mentioned in Julia's [performance tips](https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-string-interpolation-for-I/O-1).

src/statmodels.jl
galenlynch: If everything is on the same line, it's 92 characters long. 80 characters is the usual line length for source code. Where else would you like me to put a line break?

test/statmodels.jl
galenlynch: I don't understand what you mean.

src/statmodels.jl
nalimilan: Yes but we don't care about performance here, it's an error path.

src/statmodels.jl
nalimilan: 92 chars is the limit we use (as in Julia). Anyway you can perfectly break the line if needed even with that convention.

test/statmodels.jl
nalimilan: I mean better remove the second new empty line.

src/statmodels.jl
galenlynch: I put it all on the same line.

test/statmodels.jl
galenlynch: removed.

src/statmodels.jl
galenlynch: Reverted.

PR #490, opened 2019-05-02
Provide standardize API for 1D array

yuehhua:
I found that `standardize` API is just available for 2D array, and not for 1D array.

I just add supporting for 1D array.

### DIFF
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -9,7 +9,7 @@ abstract type AbstractDataTransform end
 Apply transformation `t` to vector or matrix `x` in place.
 """
 transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractVector{<:Real}) = transform!(t, reshape(x, :, 1))
+transform!(t::AbstractDataTransform, x::AbstractVector{<:Real}) = (transform!(t, reshape(x, :, 1)); x)
 
 """
     transform(t::AbstractDataTransform, x)
@@ -30,7 +30,7 @@ Perform an in-place reconstruction into an original data scale from a transforme
 vector or matrix `y` using `t` transformation.
 """
 reconstruct!(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) = reconstruct!(y, t, y)
-reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real}) = reconstruct!(t, reshape(y, :, 1))
+reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real}) = (reconstruct!(t, reshape(y, :, 1)); y)
 
 """
     reconstruct(t::AbstractDataTransform, y)
@@ -306,9 +306,7 @@ end
 
 function fit(::Type{UnitRangeTransform}, X::AbstractVector{<:Real};
              dims::Union{Integer,Nothing}=nothing, unit::Bool=true)
-    if dims == nothing
-        Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
-    elseif dims != 1
+    if dims != 1
         throw(DomainError(dims, "fit only accept dims=1 over a vector. Try fit(t, x, dims=1)."))
     end
 
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -9,11 +9,7 @@ abstract type AbstractDataTransform end
 Apply transformation `t` to vector or matrix `x` in place.
 """
 function transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real})
-    if t.dims == 1
-        transform!(x, t, x)
-    elseif t.dims == 2
-        transform!(x', t, x')
-    end
+    transform!(x, t, x)
     return x
 end
 
@@ -27,13 +23,7 @@ end
 
 Return a standardized vector or matrix `x` using `t` transformation.
 """
-function transform(t::AbstractDataTransform, x::AbstractMatrix{<:Real})
-    if t.dims == 1
-        return transform!(similar(x), t, x)
-    elseif t.dims == 2
-        return transform!(similar(x)', t, x')'
-    end
-end
+transform(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = transform!(similar(x), t, x)
 
 function transform(t::AbstractDataTransform, x::AbstractVector{<:Real})
     return vec(transform(t, reshape(x, :, 1)))
@@ -47,11 +37,7 @@ Perform an in-place reconstruction into an original data scale from a transforme
 vector or matrix `y` using `t` transformation.
 """
 function reconstruct!(t::AbstractDataTransform, y::AbstractMatrix{<:Real})
-    if t.dims == 1
-        reconstruct!(y, t, y)
-    elseif t.dims == 2
-        reconstruct!(y', t, y')
-    end
+    reconstruct!(y, t, y)
     return y
 end
 
@@ -66,13 +52,7 @@ end
 Return a reconstruction of an originally scaled data from a transformed vector
 or matrix `y` using `t` transformation.
 """
-function reconstruct(t::AbstractDataTransform, y::AbstractMatrix{<:Real})
-    if t.dims == 1
-        return reconstruct!(similar(y), t, y)
-    elseif t.dims == 2
-        return reconstruct!(similar(y)', t, y')'
-    end
-end
+reconstruct(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) = reconstruct!(similar(y), t, y)
 
 function reconstruct(t::AbstractDataTransform, y::AbstractVector{<:Real})
     return vec(reconstruct(t, reshape(y, :, 1)))
@@ -177,57 +157,67 @@ function fit(::Type{ZScoreTransform}, X::AbstractVector{<:Real};
 end
 
 function transform!(y::AbstractMatrix{<:Real}, t::ZScoreTransform, x::AbstractMatrix{<:Real})
-    l = t.len
-    size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y,1)
-    size(x,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    if t.dims == 1
+        l = t.len
+        size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
+        n = size(y,1)
+        size(x,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
 
-    m = t.mean
-    s = t.scale
+        m = t.mean
+        s = t.scale
 
-    if isempty(m)
-        if isempty(s)
-            if x !== y
-                copyto!(y, x)
+        if isempty(m)
+            if isempty(s)
+                if x !== y
+                    copyto!(y, x)
+                end
+            else
+                broadcast!(/, y, x, s')
             end
         else
-            broadcast!(/, y, x, s')
+            if isempty(s)
+                broadcast!(-, y, x, m')
+            else
+                broadcast!((x,m,s)->(x-m)/s, y, x, m', s')
+            end
         end
-    else
-        if isempty(s)
-            broadcast!(-, y, x, m')
-        else
-            broadcast!((x,m,s)->(x-m)/s, y, x, m', s')
-        end
+        return y
+    elseif t.dims == 2
+        t_ = ZScoreTransform(t.len, 1, t.mean, t.scale)
+        return transform!(y', t_, x')'
     end
-    return y
 end
 
 function reconstruct!(x::AbstractMatrix{<:Real}, t::ZScoreTransform, y::AbstractMatrix{<:Real})
-    l = t.len
-    size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y,1)
-    size(x,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    if t.dims == 1
+        l = t.len
+        size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
+        n = size(y,1)
+        size(x,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
 
-    m = t.mean
-    s = t.scale
+        m = t.mean
+        s = t.scale
 
-    if isempty(m)
-        if isempty(s)
-            if y !== x
-                copyto!(x, y)
+        if isempty(m)
+            if isempty(s)
+                if y !== x
+                    copyto!(x, y)
+                end
+            else
+                broadcast!(*, x, y, s')
             end
         else
-            broadcast!(*, x, y, s')
+            if isempty(s)
+                broadcast!(+, x, y, m')
+            else
+                broadcast!((y,m,s)->y*s+m, x, y, m', s')
+            end
         end
-    else
-        if isempty(s)
-            broadcast!(+, x, y, m')
-        else
-            broadcast!((y,m,s)->y*s+m, x, y, m', s')
-        end
+        return x
+    elseif t.dims == 2
+        t_ = ZScoreTransform(t.len, 1, t.mean, t.scale)
+        return reconstruct!(x', t_, y')'
     end
-    return x
 end
 
 """
@@ -351,37 +341,47 @@ function fit(::Type{UnitRangeTransform}, X::AbstractVector{<:Real};
 end
 
 function transform!(y::AbstractMatrix{<:Real}, t::UnitRangeTransform, x::AbstractMatrix{<:Real})
-    l = t.len
-    size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(x,1)
-    size(y,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    if t.dims == 1
+        l = t.len
+        size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
+        n = size(x,1)
+        size(y,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
 
-    tmin = t.min
-    tscale = t.scale
+        tmin = t.min
+        tscale = t.scale
 
-    if t.unit
-        broadcast!((x,s,m)->(x-m)*s, y, x, tscale', tmin')
-    else
-        broadcast!(*, y, x, tscale')
+        if t.unit
+            broadcast!((x,s,m)->(x-m)*s, y, x, tscale', tmin')
+        else
+            broadcast!(*, y, x, tscale')
+        end
+        return y
+    elseif t.dims == 2
+        t_ = UnitRangeTransform(t.len, 1, t.unit, t.min, t.scale)
+        return transform!(y', t_, x')'
     end
-    return y
 end
 
 function reconstruct!(x::AbstractMatrix{<:Real}, t::UnitRangeTransform, y::AbstractMatrix{<:Real})
-    l = t.len
-    size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y,1)
-    size(x,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    if t.dims == 1
+        l = t.len
+        size(x,2) == size(y,2) == l || throw(DimensionMismatch("Inconsistent dimensions."))
+        n = size(y,1)
+        size(x,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
 
-    tmin = t.min
-    tscale = t.scale
+        tmin = t.min
+        tscale = t.scale
 
-    if t.unit
-        broadcast!((y,s,m)->y/s+m, x, y, tscale', tmin')
-    else
-        broadcast!(/, x, y, tscale')
+        if t.unit
+            broadcast!((y,s,m)->y/s+m, x, y, tscale', tmin')
+        else
+            broadcast!(/, x, y, tscale')
+        end
+        return x
+    elseif t.dims == 2
+        t_ = UnitRangeTransform(t.len, 1, t.unit, t.min, t.scale)
+        return reconstruct!(x', t_, y')'
     end
-    return x
 end
 
 """
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -8,15 +8,8 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-function transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real})
-    transform!(x, t, x)
-    return x
-end
-
-function transform!(t::AbstractDataTransform, x::AbstractVector{<:Real})
-    transform!(t, reshape(x, :, 1))
-    return x
-end
+transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = transform!(x, t, x)
+transform!(t::AbstractDataTransform, x::AbstractVector{<:Real}) = transform!(t, reshape(x, :, 1))
 
 """
     transform(t::AbstractDataTransform, x)
@@ -36,15 +29,8 @@ end
 Perform an in-place reconstruction into an original data scale from a transformed
 vector or matrix `y` using `t` transformation.
 """
-function reconstruct!(t::AbstractDataTransform, y::AbstractMatrix{<:Real})
-    reconstruct!(y, t, y)
-    return y
-end
-
-function reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real})
-    reconstruct!(t, reshape(y, :, 1))
-    return y
-end
+reconstruct!(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) = reconstruct!(y, t, y)
+reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real}) = reconstruct!(t, reshape(y, :, 1))
 
 """
     reconstruct(t::AbstractDataTransform, y)
@@ -181,11 +167,11 @@ function transform!(y::AbstractMatrix{<:Real}, t::ZScoreTransform, x::AbstractMa
                 broadcast!((x,m,s)->(x-m)/s, y, x, m', s')
             end
         end
-        return y
     elseif t.dims == 2
         t_ = ZScoreTransform(t.len, 1, t.mean, t.scale)
-        return transform!(y', t_, x')'
+        transform!(y', t_, x')
     end
+    return y
 end
 
 function reconstruct!(x::AbstractMatrix{<:Real}, t::ZScoreTransform, y::AbstractMatrix{<:Real})
@@ -213,11 +199,11 @@ function reconstruct!(x::AbstractMatrix{<:Real}, t::ZScoreTransform, y::Abstract
                 broadcast!((y,m,s)->y*s+m, x, y, m', s')
             end
         end
-        return x
     elseif t.dims == 2
         t_ = ZScoreTransform(t.len, 1, t.mean, t.scale)
-        return reconstruct!(x', t_, y')'
+        reconstruct!(x', t_, y')
     end
+    return x
 end
 
 """
@@ -286,12 +272,12 @@ julia> StatsBase.transform(dt, X)
 function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
              dims::Union{Integer,Nothing}=nothing, unit::Bool=true)
     if dims == 1
-        l, tmin, tmax = _extract_info(X)
+        l, tmin, tmax = _compute_extrema(X)
     elseif dims == 2
-        l, tmin, tmax = _extract_info(X')
+        l, tmin, tmax = _compute_extrema(X')
     elseif dims == nothing
         Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
-        l, tmin, tmax = _extract_info(X')
+        l, tmin, tmax = _compute_extrema(X')
     else
         throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
     end
@@ -302,7 +288,7 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
     return UnitRangeTransform(l, dims, unit, tmin, tmax)
 end
 
-function _extract_info(X::AbstractMatrix{<:Real})
+function _compute_extrema(X::AbstractMatrix{<:Real})
     n, l = size(X)
     tmin = X[1, :]
     tmax = X[1, :]
@@ -326,18 +312,9 @@ function fit(::Type{UnitRangeTransform}, X::AbstractVector{<:Real};
         throw(DomainError(dims, "fit only accept dims=1 over a vector. Try fit(t, x, dims=1)."))
     end
 
-    n = length(X)
-    tmin = X[1]
-    tmax = X[1]
-    @inbounds for i = 2:n
-        if X[i] < tmin
-            tmin = X[i]
-        elseif X[i] > tmax
-            tmax = X[i]
-        end
-    end
+    l, tmin, tmax = _compute_extrema(reshape(X, :, 1))
     tmax = 1 / (tmax - tmin)
-    return UnitRangeTransform(1, dims, unit, [tmin], [tmax])
+    return UnitRangeTransform(1, dims, unit, vec(tmin), vec(tmax))
 end
 
 function transform!(y::AbstractMatrix{<:Real}, t::UnitRangeTransform, x::AbstractMatrix{<:Real})
@@ -355,11 +332,11 @@ function transform!(y::AbstractMatrix{<:Real}, t::UnitRangeTransform, x::Abstrac
         else
             broadcast!(*, y, x, tscale')
         end
-        return y
     elseif t.dims == 2
         t_ = UnitRangeTransform(t.len, 1, t.unit, t.min, t.scale)
-        return transform!(y', t_, x')'
+        transform!(y', t_, x')
     end
+    return y
 end
 
 function reconstruct!(x::AbstractMatrix{<:Real}, t::UnitRangeTransform, y::AbstractMatrix{<:Real})
@@ -377,11 +354,11 @@ function reconstruct!(x::AbstractMatrix{<:Real}, t::UnitRangeTransform, y::Abstr
         else
             broadcast!(/, x, y, tscale')
         end
-        return x
     elseif t.dims == 2
         t_ = UnitRangeTransform(t.len, 1, t.unit, t.min, t.scale)
-        return reconstruct!(x', t_, y')'
+        reconstruct!(x', t_, y')
     end
+    return x
 end
 
 """
@@ -392,9 +369,6 @@ end
 
 - `ZScoreTransform`
 - `UnitRangeTransform`
-
-Return a standardized matrix while the input of `X` is a `AbstractMatrix`.
-Return a standardized vector while the input of `X` is a `AbstractVector`.
 
 # Example
 
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -8,19 +8,20 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractVector{<:Real}) = (transform!(t, reshape(x, :, 1)); x)
+transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) =
+    transform!(x, t, x)
+transform!(t::AbstractDataTransform, x::AbstractVector{<:Real}) =
+    (transform!(t, reshape(x, :, 1)); x)
 
 """
     transform(t::AbstractDataTransform, x)
 
 Return a standardized vector or matrix `x` using `t` transformation.
 """
-transform(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = transform!(similar(x), t, x)
-
-function transform(t::AbstractDataTransform, x::AbstractVector{<:Real})
-    return vec(transform(t, reshape(x, :, 1)))
-end
+transform(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) =
+    transform!(similar(x), t, x)
+transform(t::AbstractDataTransform, x::AbstractVector{<:Real}) =
+    vec(transform(t, reshape(x, :, 1)))
 
 # reconstruct the original data from transformed values
 """
@@ -29,8 +30,10 @@ end
 Perform an in-place reconstruction into an original data scale from a transformed
 vector or matrix `y` using `t` transformation.
 """
-reconstruct!(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) = reconstruct!(y, t, y)
-reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real}) = (reconstruct!(t, reshape(y, :, 1)); y)
+reconstruct!(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) =
+    reconstruct!(y, t, y)
+reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real}) =
+    (reconstruct!(t, reshape(y, :, 1)); y)
 
 """
     reconstruct(t::AbstractDataTransform, y)
@@ -38,11 +41,10 @@ reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real}) = (reconstruct
 Return a reconstruction of an originally scaled data from a transformed vector
 or matrix `y` using `t` transformation.
 """
-reconstruct(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) = reconstruct!(similar(y), t, y)
-
-function reconstruct(t::AbstractDataTransform, y::AbstractVector{<:Real})
-    return vec(reconstruct(t, reshape(y, :, 1)))
-end
+reconstruct(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) =
+    reconstruct!(similar(y), t, y)
+reconstruct(t::AbstractDataTransform, y::AbstractVector{<:Real}) =
+    vec(reconstruct(t, reshape(y, :, 1)))
 
 """
     Standardization (Z-score transformation)
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -8,16 +8,14 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+transform!(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}) = transform!(x, t, x)
 
 """
     transform(t::AbstractDataTransform, x)
 
 Return a row-standardized vector or matrix `x` using `t` transformation.
 """
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(similar(x), t, x)
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(similar(x), t, x)
+transform(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}) = transform!(similar(x), t, x)
 
 # reconstruct the original data from transformed values
 """
@@ -26,8 +24,7 @@ transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(sim
 Perform an in-place reconstruction into an original data scale from a row-transformed
 vector or matrix `y` using `t` transformation.
 """
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,1}) = reconstruct!(y, t, y)
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,2}) = reconstruct!(y, t, y)
+reconstruct!(t::AbstractDataTransform, y::AbstractVecOrMat{<:Real}) = reconstruct!(y, t, y)
 
 """
     reconstruct(t::AbstractDataTransform, y)
@@ -35,13 +32,12 @@ reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,2}) = reconstruct
 Return a reconstruction of an originally scaled data from a row-transformed vector
 or matrix `y` using `t` transformation.
 """
-reconstruct(t::AbstractDataTransform, y::AbstractArray{<:Real,1})  = reconstruct!(similar(y), t, y)
-reconstruct(t::AbstractDataTransform, y::AbstractArray{<:Real,2})  = reconstruct!(similar(y), t, y)
+reconstruct(t::AbstractDataTransform, y::AbstractVecOrMat{<:Real})  = reconstruct!(similar(y), t, y)
 
 """
     Standardization (Z-score transformation)
 """
-struct ZScoreTransform{T <: Real} <: AbstractDataTransform
+struct ZScoreTransform{T<:Real} <: AbstractDataTransform
     dim::Int
     mean::Vector{T}
     scale::Vector{T}
@@ -97,7 +93,7 @@ julia> StatsBase.transform(dt, X)
  -1.0   0.0  1.0
 ```
 """
-function fit(::Type{ZScoreTransform}, X::AbstractArray{<:Real,2}; center::Bool = true, scale::Bool = true)
+function fit(::Type{ZScoreTransform}, X::AbstractArray{<:Real,2}; center::Bool=true, scale::Bool=true)
     d, n = size(X)
     n >= 2 || error("X must contain at least two columns.")
 
@@ -108,11 +104,15 @@ function fit(::Type{ZScoreTransform}, X::AbstractArray{<:Real,2}; center::Bool =
                               (scale ? vec(s) : zeros(T, 0)))
 end
 
+function fit(::Type{ZScoreTransform}, X::AbstractArray{<:Real,1}; kwargs...)
+    return fit(ZScoreTransform, X'; kwargs...)
+end
+
 function transform!(y::AbstractVecOrMat{<:Real}, t::ZScoreTransform, x::AbstractVecOrMat{<:Real})
     d = t.dim
-    size(x, 1) == size(y, 1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y, 2)
-    size(x, 2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    size(x,1) == size(y,1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
+    n = size(y,2)
+    size(x,2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
 
     m = t.mean
     s = t.scale
@@ -129,7 +129,7 @@ function transform!(y::AbstractVecOrMat{<:Real}, t::ZScoreTransform, x::Abstract
         if isempty(s)
             broadcast!(-, y, x, m)
         else
-            broadcast!((x, m, s)->(x - m) / s, y, x, m, s)
+            broadcast!((x,m,s)->(x-m)/s, y, x, m, s)
         end
     end
     return y
@@ -137,9 +137,9 @@ end
 
 function reconstruct!(x::AbstractVecOrMat{<:Real}, t::ZScoreTransform, y::AbstractVecOrMat{<:Real})
     d = t.dim
-    size(x, 1) == size(y, 1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y, 2)
-    size(x, 2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    size(x,1) == size(y,1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
+    n = size(y,2)
+    size(x,2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
 
     m = t.mean
     s = t.scale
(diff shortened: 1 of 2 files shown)


src/transformations.jl
```diff
@@ -301,6 +301,9 @@ julia> standardize(UnitRangeTransform, [0.0 -0.5 0.5; 0.0 1.0 2.0])
  0.0  0.5  1.0
 ```
 """
-function standardize(::Type{DT}, X::AbstractArray{<:Real,2}; kwargs...) where {DT<:AbstractDataTransform}
+function standardize(::Type{DT}, X::AbstractArray{<:Real,1}; kwargs...) where {DT <: AbstractDataTransform}
```
nalimilan: `AbstractVecOrMat{<:Real}` should replace both methods.

The docstring should also be updated, and tests added.

wildart: Correct. `AbstractVecOrMat{<:Real}` will do the trick in `standardize`.

yuehhua:
Correct the implementation, add tests and update docs.

nalimilan:
Can you also adjust `fit`, and revert all unrelated whitespace changes?

yuehhua:
> Can you also adjust `fit`, and revert all unrelated whitespace changes?

Do you mean to implement a `fit` which accept `AbstractArray{<:Real,1}`?

Sorry for adding unrelated whitespaces. My editor do so while I save.

wildart:
Yes, you should do that.

nalimilan:
Sorry, I hadn't realized that this is inconsistent with the fact that vectors are treated as column vectors in Julia. Can you change this?

I think this also illustrates a problem that I hadn't spotted when reviewing the original PR which added transformations: we shouldn't apply standardization to rows by default, as it's inconsistent with what e.g. `cov` does in Statistics. I think we should add a `dims` keyword argument and deprecate the current behavior, as I did in https://github.com/JuliaStats/Distances.jl/pull/121.

yuehhua:
Great! OK, Let's check our plan.

1. Add `dims` keyword argument as `cov` does in Statistics.

The `dims` behavior of `cov` is different from you did in JuliaStats/Distances.jl#121.

Suppose we have a **m-by-n** matrix `X`.

`cov(X, dims=1)` calculates along the first dimension, or in **column-wise fashion**, then gives **n-by-n** covariance matrix.

`cov(X, dims=2)` calculates along the second dimension, or in **row-wise fashion**, then gives **m-by-m** covariance matrix.

Similar to `cov`, we may have standardization API as follow:

`standardize(::Type{DT}, X::AbstractMatrix{<:Real}; kwargs...)` which has `dims=1` keyword argument to standardize `X` along the first dimension, then gives **column-standardized matrix**.

`standardize(::Type{DT}, X::AbstractMatrix{<:Real}; kwargs...)` with `dims=2` keyword argument  standardizes `X` along the second dimension, then gives **row-standardized matrix**.

2. Change the calculation behavior to column-wise fashion for calculation efficiency.

We could separate the API and implementation. API design would follow the design above. The implementation goes in column-wise fashion for calculation efficiency.

Thus, we provide the efficient implementation with user-friendly API.

The details of implementation:

For `dims=1`, column-standardized matrix can be calculated straight forward in column-wise fashion.

For `dims=2`, row-standardized matrix may be transposed before going to calculation in column-wise fashion.

The addition of `dims` keyword argument gives the minimal modification for user while deprecating the current behavior.

3. Provide the API and implementation for `Vector` fulfilling the purpose of this PR.

Could I change/redesign the behavior of whole transformations.jl to achieve the description above?

nalimilan:
> The `dims` behavior of `cov` is different from you did in [JuliaStats/Distances.jl#121](https://github.com/JuliaStats/Distances.jl/pull/121).
> 
> Suppose we have a **m-by-n** matrix `X`.
> 
> `cov(X, dims=1)` calculates along the first dimension, or in **column-wise fashion**, then gives **n-by-n** covariance matrix.
> 
> `cov(X, dims=2)` calculates along the second dimension, or in **row-wise fashion**, then gives **m-by-m** covariance matrix.

Yes, `cov` and `pairwise` are different since the former computes pairwise correlations between variables, but the latter computes pairwise distances between observations. So when you have an observations×variables matrix, you use `cov(X, dims=1)` and `pairwise(X, dims=1)`. (I guess a counter-argument to that decision is that in `cov` and `sum`, the `dims` argument indicates the dimension which will be collapsed and will therefore not appear in the resulting matrix, which isn't the case for `pairwise`. Ah, well... not sure which behavior is the most intuitive.)

> Similar to `cov`, we may have standardization API as follow:
> 
> `standardize(::Type{DT}, X::AbstractMatrix{<:Real}; kwargs...)` which has `dims=1` keyword argument to standardize `X` along the first dimension, then gives **column-standardized matrix**.
> 
> `standardize(::Type{DT}, X::AbstractMatrix{<:Real}; kwargs...)` with `dims=2` keyword argument standardizes `X` along the second dimension, then gives **row-standardized matrix**.

Yes, your proposal sounds consistent with how `cov` and `pairwise` work: use `dims=1` when observations are stored as rows, and `dims=2` when they are stored as columns. Standardizing with `dims=i` will then be equivalent to `(x .- mean(x, dims=i)) ./ std(x, dims=i)`.



>     1. Change the calculation behavior to column-wise fashion for calculation efficiency.
> 
> 
> We could separate the API and implementation. API design would follow the design above. The implementation goes in column-wise fashion for calculation efficiency.
> 
> Thus, we provide the efficient implementation with user-friendly API.
> 
> The details of implementation:
> 
> For `dims=1`, column-standardized matrix can be calculated straight forward in column-wise fashion.
> 
> For `dims=2`, row-standardized matrix may be transposed before going to calculation in column-wise fashion.

Yes, you can use `transpose` as I did at https://github.com/JuliaStats/Distances.jl/pull/121. But note that's a lazy operation: no transposed copy is made, which means the memory is still not accessed in the most efficient pattern. But making a copy would probably not be much faster since that would add some overhead (and use memory, which the user may not want).

Otherwise, the plan sounds good.

yuehhua:
> Yes, `cov` and `pairwise` are different since the former computes pairwise correlations between variables, but the latter computes pairwise distances between observations. So when you have an observations×variables matrix, you use `cov(X, dims=1)` and `pairwise(X, dims=1)`. (I guess a counter-argument to that decision is that in `cov` and `sum`, the `dims` argument indicates the dimension which will be collapsed and will therefore not appear in the resulting matrix, which isn't the case for `pairwise`. Ah, well... not sure which behavior is the most intuitive.)

A intuitive and user-friendly design is necessary, however, I also considering the consistency between APIs. So far, I saw the counter-arguments are also used in Flux.jl. I have no further guesses about this. Maybe some suggestions?

> Yes, you can use `transpose` as I did at [JuliaStats/Distances.jl#121](https://github.com/JuliaStats/Distances.jl/pull/121). But note that's a lazy operation: no transposed copy is made, which means the memory is still not accessed in the most efficient pattern. But making a copy would probably not be much faster since that would add some overhead (and use memory, which the user may not want).
> 
> Otherwise, the plan sounds good.

I also considered the scenarios you mentioned above. Transpose doesn't copy the matrix, while the deep copy may not get efficiency. So, I decide to implement it straight forward.

nalimilan:
> A intuitive and user-friendly design is necessary, however, I also considering the consistency between APIs. So far, I saw the counter-arguments are also used in Flux.jl. I have no further guesses about this. Maybe some suggestions?

What do you mean about Flux.jl? AFAIK the difference in conventions is only about defaults, which can be tackled separately from the meaning we assign to `dims=1` and `dims=2`, right? It's very hard to decide what's the best default since that's mostly an opposition between stats (observations×variables) and machine learning (variables×observations) people. A possible rule is to be consistent with Statistics since that's in the stdlib, but that's just one possible choice.

yuehhua:
For your reference, FluxML/Flux.jl#563

yuehhua:
I think that the design of arguments is not related to the domain (statistics/machine learning).
[Z-score in MATLAB](https://www.mathworks.com/help/stats/zscore.html) has the same behavior with Julia.
The reason behind the design might be the memory layout, while Julia and MATLAB are column major languages.

nalimilan:
> I think that the design of arguments is not related to the domain (statistics/machine learning).
> [Z-score in MATLAB](https://www.mathworks.com/help/stats/zscore.html) has the same behavior with Julia.
> The reason behind the design might be the memory layout, while Julia and MATLAB are column major languages.

That's fine with me, but the fact that the current method chose a different default indicates that not everybody feels that way. Anyway as a first step we should add a `dims` argument, then we can decide later whether to change the default. Indeed in column-major languages like Julia it's faster to standardize columns.

src/transformations.jl
```diff
@@ -8,16 +8,16 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}; dims::Int=1) = transform!(x, t, x, Val(dims))
+transform!(t::AbstractDataTransform, x::AbstractVector{<:Real}) = transform!(x, t, x)
```
nalimilan: Should also accept `dims`, and check that it's equal to `1`.

src/transformations.jl
```diff
@@ -97,22 +99,70 @@ julia> StatsBase.transform(dt, X)
...
 
-function transform!(y::AbstractVecOrMat{<:Real}, t::ZScoreTransform, x::AbstractVecOrMat{<:Real})
+function fit(::Type{ZScoreTransform}, X::AbstractVector{<:Real}; center::Bool=true, scale::Bool=true)
+    T = eltype(X)
+    m, s = mean_and_std(X)
+    return ZScoreTransform(1, (center ? [m] : zeros(T, 0)),
+                              (scale ? [s] : zeros(T, 0)))
+end
+
+function _fit(::Type{ZScoreTransform}, X::AbstractMatrix{<:Real}, ::Val{1}, center::Bool, scale::Bool)
```
nalimilan: Better use a branch inside a single `fit` method than dispatch on `::Val`: it will likely be faster, and inference doesn't depend on the value.

src/transformations.jl
```diff
@@ -8,16 +8,16 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}; dims::Int=1) = transform!(x, t, x, Val(dims))
```
nalimilan: For now the default should be `dims::Union{Integer,Nothing}=nothing`, and a deprecation warning in favor of `dims=2` should be printed when `dims === nothing`. Then we'll be able to switch the default later.

src/transformations.jl
```diff
@@ -97,22 +99,70 @@ julia> StatsBase.transform(dt, X)
...
-    n = size(y,2)
-    size(x,2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    size(x,2) == size(y,2) == d || throw(DimensionMismatch("Inconsistent dimensions."))
+    n = size(y,1)
+    size(x,1) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+
+    m = t.mean
+    s = t.scale
+
+    if isempty(m)
```
nalimilan: Instead of duplicating the code, just call `transform!` on the transpose. Same below.

src/transformations.jl
```diff
@@ -200,9 +280,9 @@ Fit a scaling parameters to `X` and return transformation description.
 
 # Keyword arguments
 
-* `center`: if `true` (the default) centere data around zero.
+* `dims`: if `1` (the default) fit standardization parameters in column-wise fashion. if `2` fit in row-wise fashion.
```
nalimilan:
```suggestion
* `dims`: if `1` (the default) fit standardization parameters in column-wise fashion; if `2` fit in row-wise fashion.
```

src/transformations.jl
yuehhua: OK

src/transformations.jl
yuehhua: OK

src/transformations.jl
yuehhua: Do you mean if else?

src/transformations.jl
yuehhua: OK

src/transformations.jl
nalimilan: Yes.

src/transformations.jl
yuehhua: Hummmmm........It seems redundant to a vector, because the `dims` has to be `1`, further with assertion check.

nalimilan: Yes, but that way you can pass `dims=1` and it works, whether `x` is a vector or a matrix. Also, I'm not completely sure it's redundant, since it could make sense to allow treating vectors like single-column matrices in the future, allowing `dims=2`: for example, `sum([1, 2], dims=2)` is allowed is the identity.

yuehhua: Oh, Cool!

yuehhua:
Something like this?

src/transformations.jl
```diff
@@ -26,17 +36,27 @@ transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(sim
 Perform an in-place reconstruction into an original data scale from a row-transformed
 vector or matrix `y` using `t` transformation.
 """
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,1}) = reconstruct!(y, t, y)
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,2}) = reconstruct!(y, t, y)
+function reconstruct!(t::AbstractDataTransform, y::AbstractVecOrMat{<:Real}; dims::Union{Integer,Nothing}=nothing)
```
wildart: Reconstruction shouldn't have a `dims` argument. It makes no sense. Information about which dimension was used in the transformation should come from the `DataTransform` object.

src/transformations.jl
```diff
@@ -8,16 +8,26 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+function transform!(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}; dims::Union{Integer,Nothing}=nothing)
+    if dims == 1
```
wildart: When `dims` defaults to `nothing`, you should call `depwarn` with some deprecation warning and set `dims` to 2

src/transformations.jl
```diff
@@ -8,16 +8,26 @@ abstract type AbstractDataTransform end
...
 
 """
     transform(t::AbstractDataTransform, x)
 
-Return a row-standardized vector or matrix `x` using `t` transformation.
+Return a standardized vector or matrix `x` using `t` transformation.
 """
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(similar(x), t, x)
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(similar(x), t, x)
+function transform(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}; dims::Union{Integer,Nothing}=nothing)
```
wildart: Same here

src/transformations.jl
yuehhua: OK

src/transformations.jl
yuehhua: OK

src/StatsBase.jl
```diff
@@ -2,7 +2,7 @@ __precompile__()
 
 module StatsBase
 
-import Base: length, isempty, eltype, values, sum, show, maximum, minimum, extrema
+import Base: length, isempty, eltype, values, sum, show, maximum, minimum, extrema, depwarn
```
nalimilan: `import` is for when you override a function. Just use `Base.depwarn` when you call it (adding it to `using` would run the risk of forgetting to remove it once the deprecation is removed).

src/transformations.jl
```diff
@@ -8,16 +8,30 @@ abstract type AbstractDataTransform end
...
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+function transform!(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}; dims::Union{Integer,Nothing}=nothing)
+    if dims == 1
+        return transform!(x, t, x)
+    elseif dims == 2
+        return transform!(x', t, x')'
+    elseif dims === nothing
+        depwarn("123", transform!(t, x; dims=2))
```
nalimilan: `"123"` isn't the final message, right?

src/transformations.jl
```diff
@@ -8,16 +8,30 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+function transform!(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}; dims::Union{Integer,Nothing}=nothing)
+    if dims == 1
+        return transform!(x, t, x)
+    elseif dims == 2
+        return transform!(x', t, x')'
```
nalimilan: Return `x` instead of calling `'`, which would return an `Adjoint` object wrapping it instead.

src/transformations.jl
```diff
@@ -8,16 +8,30 @@ abstract type AbstractDataTransform end
...
+    elseif dims === nothing
+        depwarn("123", transform!(t, x; dims=2))
+    end
+end
 
 """
     transform(t::AbstractDataTransform, x)
```
nalimilan: Need to document `dims` (here and elsewhere).

src/transformations.jl
```diff
@@ -26,17 +40,15 @@ transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(sim
 Perform an in-place reconstruction into an original data scale from a row-transformed
 vector or matrix `y` using `t` transformation.
 """
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,1}) = reconstruct!(y, t, y)
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,2}) = reconstruct!(y, t, y)
+function reconstruct!(t::AbstractDataTransform, y::AbstractVecOrMat{<:Real}) = reconstruct!(y, t, y)
```
nalimilan:
```suggestion
reconstruct!(t::AbstractDataTransform, y::AbstractVecOrMat{<:Real}) = reconstruct!(y, t, y)
```

src/transformations.jl
```diff
@@ -26,17 +40,15 @@ transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(sim
...
 
 """
     reconstruct(t::AbstractDataTransform, y)
 
 Return a reconstruction of an originally scaled data from a row-transformed vector
 or matrix `y` using `t` transformation.
 """
-reconstruct(t::AbstractDataTransform, y::AbstractArray{<:Real,1})  = reconstruct!(similar(y), t, y)
-reconstruct(t::AbstractDataTransform, y::AbstractArray{<:Real,2})  = reconstruct!(similar(y), t, y)
+function reconstruct(t::AbstractDataTransform, y::AbstractVecOrMat{<:Real}) = reconstruct!(similar(y), t, y)
```
nalimilan:
```suggestion
reconstruct(t::AbstractDataTransform, y::AbstractVecOrMat{<:Real}) = reconstruct!(similar(y), t, y)
```

src/transformations.jl
```diff
@@ -74,6 +86,8 @@ Fit standardization parameters to `X` and return a `ZScoreTransform` transformat
 
 # Keyword arguments
 
+* `dims`: if `1` (the default) fit standardization parameters in column-wise fashion. if `2` fit in row-wise fashion.
```
nalimilan:
```suggestion
* `dims`: if `1` (the default) fit standardization parameters in column-wise fashion;
  if `2` fit in row-wise fashion.
```

src/transformations.jl
```diff
@@ -97,22 +111,36 @@ julia> StatsBase.transform(dt, X)
...
+        d, n = size(X)
+        n >= 2 || error("X must contain at least two columns.")
+        m, s = mean_and_std(X, 2)
+    elseif dims === nothing
+        depwarn("123", fit(ZScoreTransform, X; dims=2, center=center, scale=scale))
+    end
     T = eltype(X)
-    m, s = mean_and_std(X, 2)
-
     return ZScoreTransform(d, (center ? vec(m) : zeros(T, 0)),
```
nalimilan: `dims` must be stored in `ZScoreTransform` as @wildart suggested.

src/transformations.jl
```diff
@@ -123,23 +151,75 @@ function transform!(y::AbstractVecOrMat{<:Real}, t::ZScoreTransform, x::Abstract
...
+    end
+    return y
+end
+
+function reconstruct!(x::AbstractMatrix{<:Real}, t::ZScoreTransform, y::AbstractMatrix{<:Real})
     d = t.dim
-    size(x,1) == size(y,1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y,2)
-    size(x,2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+    size(x,2) == size(y,2) == d || throw(DimensionMismatch("Inconsistent dimensions."))
```
nalimilan: I don't understand what this is supposed to do: it compares the size of a dimension with its number.

src/transformations.jl
```diff
@@ -223,17 +303,22 @@ julia> StatsBase.transform(dt, X)
...
-            if X[i, j] < tmin[i]
-                tmin[i] = X[i, j]
-            elseif X[i, j] > tmax[i]
-                tmax[i] = X[i, j]
+function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
+             dims::Union{Integer,Nothing}=nothing, unit::Bool=true)
+    if dims == 1
+        X_ = X
+    elseif dims == 2
+        X_ = X'
```
nalimilan: This is going to create a type instability which could hurt performance. Better put the code below in a separate function which would take `X_` to make sure the loop is type stable.

src/transformations.jl
yuehhua: Oh! I forget to change this.

src/transformations.jl
yuehhua: This exists before this PR, and I don't understand this, too.

nalimilan: Ah, sorry, I misread this. `d = t.dim` gives the number of rows/columns, depending on the dimension over which standardization happens. I thought you had added it to store `dims` (which is needed, as I noted above).

src/transformations.jl
```diff
@@ -4,20 +4,34 @@ abstract type AbstractDataTransform end
 
 # apply the transform
 """
-    transform!(t::AbstractDataTransform, x)
+    transform!(t::AbstractDataTransform, x; dims=nothing)
 
-Apply transformation `t` to vector or matrix `x` in place.
+Apply transformation `t` to vector or matrix `x` in place along the dims.
```
nalimilan:
```suggestion
Apply transformation `t` to vector or matrix `x` in place along dimension `dims`.
```

src/transformations.jl
```diff
@@ -4,20 +4,34 @@ abstract type AbstractDataTransform end
...
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+function transform!(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}; dims::Union{Integer,Nothing}=nothing)
+    if dims == 1
+        return transform!(x, t, x)
+    elseif dims == 2
+        return transform!(x', t, x')
+    elseif dims === nothing
+        Base.depwarn("Decreption of transform!(t, x) in favor of transform!(t, x; dims=2).", transform!(t, x; dims=2))
```
nalimilan:
```suggestion
        Base.depwarn("transform!(t, x) is deprecated: use transform!(t, x, dims=2) instead", :transform)
```

src/transformations.jl
```diff
@@ -4,20 +4,34 @@ abstract type AbstractDataTransform end
...
 """
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(similar(x), t, x)
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(similar(x), t, x)
+function transform(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real}; dims::Union{Integer,Nothing}=nothing)
+    if dims == 1
+        return transform!(similar(x), t, x)
+    elseif dims == 2
+        return transform!(similar(x)', t, x')
+    elseif dims === nothing
+        Base.depwarn("Decreption of transform(t, x) in favor of transform(t, x; dims=2).", transform(t, x; dims=2))
```
nalimilan:
```suggestion
        Base.depwarn("transform!(t, x) is deprecated: use transform!(t, x, dims=2) instead", :transform)
```

src/transformations.jl
nalimilan: Still an issue.

src/transformations.jl
```diff
@@ -64,7 +76,7 @@ function Base.getproperty(t::ZScoreTransform, p::Symbol)
 end
 
 """
-    fit(ZScoreTransform, X; center=true, scale=true)
+    fit(ZScoreTransform, X; dims=1, center=true, scale=true)
```
nalimilan: The default is still `nothing`/`2` currently. Same below.

src/transformations.jl
```diff
@@ -243,48 +329,99 @@ function fit(::Type{UnitRangeTransform}, X::AbstractArray{<:Real,2}; unit::Bool=
...
+    if t.unit
+        broadcast!((y,s,m)->y/s+m, x, y, tscale[1], tmin[1])
+    else
+        broadcast!(/, x, y, tscale[1])
     end
     return x
 end
 
 """
     standardize(DT, X; kwargs...)
```
nalimilan: Should take `dims` too.

src/transformations.jl
```diff
@@ -97,22 +112,36 @@ julia> StatsBase.transform(dt, X)
...
+    if dims == 1
+        n, d = size(X)
+        n >= 2 || error("X must contain at least two rows.")
+        m, s = mean_and_std(X, 1)
+    elseif dims == 2
+        d, n = size(X)
+        n >= 2 || error("X must contain at least two columns.")
+        m, s = mean_and_std(X, 2)
+    elseif dims === nothing
+        depwarn("123", fit(ZScoreTransform, X; dims=2, center=center, scale=scale))
```
nalimilan: Still needs fixing.

src/transformations.jl
```diff
@@ -4,20 +4,34 @@ abstract type AbstractDataTransform end
...
+        Base.depwarn("Decreption of transform!(t, x) in favor of transform!(t, x; dims=2).", transform!(t, x; dims=2))
+    end
+end
 
 """
-    transform(t::AbstractDataTransform, x)
+    transform(t::AbstractDataTransform, x; dims=nothing)
 
-Return a row-standardized vector or matrix `x` using `t` transformation.
+Return a standardized vector or matrix `x` using `t` transformation along the dims.
```
nalimilan:
```suggestion
Return a standardized vector or matrix `x` using `t` transformation along dimension `dims`.
```

src/transformations.jl
yuehhua: OK

yuehhua:
If we keep the `dims` in the type. That could be confused with original `dim`.
Is there a good way/naming for that?

src/transformations.jl
nalimilan: Yes.

nalimilan:
I guess `dim` could be renamed to `len`.

yuehhua:
If we store `dims` in the type, then `transform` does not need `dims` argument anymore, right?

Need `dims`:
* `fit`
* `standardize`

Doesn't need `dims`:
* `transform` and `transform!`
* `reconstruct` and `reconstruct`

So, we will have:

```
struct ZScoreTransform{T<:Real} <: AbstractDataTransform
    len::Int
    dims::Int
    mean::Vector{T}
    scale::Vector{T}

    ...
end
```

nalimilan:
Sounds right. I don't think `dims` needs to be in the type, as the `if dim == ...` branch should have a really negligible cost compared with the other computations.

yuehhua:
I found that adding `dims` to type restricts the transformation. Consider the following scenario:

```
julia> X = rand(3, 4)
3×4 Array{Float64,2}:
...

julia> t = fit(ZScoreTransform, X, dims=1)
ZScoreTransform{Float64}(4, 1, [0.444032, 0.400317, 0.564703, 0.509742], [0.169329, 0.17986, 0.363233, 0.355917])
```

I got the `ZScoreTransform` which will transform the `Array` along the `dims=1` and check the `len=4`.
This object transforms well on `X` itself but not on `X[:,1]`.

```
julia> StatsBase.transform(t, X[:,1])
ERROR: DimensionMismatch("Inconsistent dimensions.")
Stacktrace:
 [1] transform!(::Array{Float64,1}, ::ZScoreTransform{Float64}, ::Array{Float64,1}) at /media/pika/Workbench/workspace/StatsBase.jl/src/transformations.jl:172
 [2] transform(::ZScoreTransform{Float64}, ::Array{Float64,1}) at /media/pika/Workbench/workspace/StatsBase.jl/src/transformations.jl:26
 [3] top-level scope at none:0
```

The reason is inconsistent with the `len`. `ZScoreTransform` transform `X[:,1]` along the `dims=1` with `len=4`, but `X[:,1]` has `len=1`.
It means if we want a transformation fitted from a matrix and apply to a vector. That doesn't work well on this scenario.

nalimilan:
Hmm, I see this is in the tests, but I don't understand why we support that. IIUC, we assume that `t = fit(ZScoreTransform, X, dims=1); transform(t, X[:, 1])` means that we want to do `fit(ZScoreTransform, X[:, 1], dims=1); transform(t, X[:, 1])`? Choosing the first column sounds quite arbitrary to me, and possibly confusing if somebody calls `transform(t, X[:, 2])`. So probably better remove that.

yuehhua:
@nalimilan Is it good?

src/transformations.jl
nalimilan: This still applies:
```suggestion
        transform!(x', t, x')
        return x
```
Same below. Tests should be adapted of course.

src/transformations.jl
```diff
@@ -74,6 +95,9 @@ Fit standardization parameters to `X` and return a `ZScoreTransform` transformat
 
 # Keyword arguments
 
+* `dims`: if `1` fit standardization parameters in column-wise fashion;
+  if `2` fit in row-wise fashion. The default is `nothing`.
```
nalimilan:
```suggestion
  if `2` fit in row-wise fashion. The default is `nothing`, which is equivalent to `dims=2` with a deprecation warning.
```

src/transformations.jl
```diff
@@ -97,22 +121,69 @@ julia> StatsBase.transform(dt, X)
...
+    if dims == 1
+        n, l = size(X)
+        n >= 2 || error("X must contain at least two rows.")
+        m, s = mean_and_std(X, 1)
+    elseif dims == 2
+        l, n = size(X)
+        n >= 2 || error("X must contain at least two columns.")
+        m, s = mean_and_std(X, 2)
+    elseif dims === nothing
+        Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
```
nalimilan: Need to do the same thing as when `dims==2`. Same below.

Also need to throw an error if `dims` takes another value.

src/transformations.jl
```diff
@@ -97,22 +121,69 @@ julia> StatsBase.transform(dt, X)
...
+    end
+    return y
 end
 
-function transform!(y::AbstractVecOrMat{<:Real}, t::ZScoreTransform, x::AbstractVecOrMat{<:Real})
-    d = t.dim
-    size(x,1) == size(y,1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y,2)
-    size(x,2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+function transform!(y::AbstractVector{<:Real}, t::ZScoreTransform, x::AbstractVector{<:Real})
```
nalimilan: Rather than duplicating the function for `x::AbstractVector`, call `reshape(x, :, 1)` and call `transform!` on it to reuse the matrix method. Same below.

yuehhua:
In [the discussion](https://github.com/JuliaStats/StatsBase.jl/pull/490#discussion_r300565695), the matrix `x` (m-by-n) should be transposed after transform or reconstruct, otherwise, user cannot get the matrix with the same shape as the origin. They will get the transposed matrix (n-by-m). The transpose should be added before and after the operation. Furthermore, I ensured the returned matrix would not be the wrapped type of `LinearAlgebra.Adjoint`.

src/transformations.jl
yuehhua: Hmmmmmm, I need to add `depwarn` when `dims==2`?
If yes, what for?

nalimilan: I said "the same thing as when", not "the same thing when". I mean that you need to set `m` and `s`.

nalimilan:
When users call `transform!`, they want the original matrix `x` to contain the result, and the convention in Julia is that functions ending in `!` return their argument (that's just for convenience when chaining calls). The original matrix isn't affected at all by the transposition so that's not a correct approach (and it makes a copy).

src/transformations.jl
yuehhua: Oh, I got it.

src/transformations.jl
```diff
@@ -8,24 +8,32 @@ abstract type AbstractDataTransform end
...
-function transform!(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real})
-    if t.dims == 1
-        return transform!(x, t, x)
-    elseif t.dims == 2
-        return transform!(x', t, x')'
-    end
+transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = _transform!(x, t, x)
+
+function transform!(t::AbstractDataTransform, x::AbstractVector{<:Real})
+    x = reshape(x, :, 1)
```
nalimilan: Use another name than `x` (e.g. `xmat`), so that you can return the original `x` in the end; that's better for type stability anyway. Or simpler, just do `_transform!(x, t, reshape(x, :, 1))` directly.

src/transformations.jl
```diff
@@ -8,24 +8,32 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-function transform!(t::AbstractDataTransform, x::AbstractVecOrMat{<:Real})
-    if t.dims == 1
-        return transform!(x, t, x)
-    elseif t.dims == 2
-        return transform!(x', t, x')'
-    end
+transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = _transform!(x, t, x)
```
nalimilan: Why define `_transform` if that's completely equivalent to `transform!`?

test/transformations.jl
```diff
@@ -1,66 +1,182 @@
...
+    t = fit(ZScoreTransform, X, dims=1, center=false, scale=false)
     Y = transform(t, X)
     @test isa(t, AbstractDataTransform)
     @test isempty(t.mean)
     @test isempty(t.scale)
     @test isequal(X, Y)
-    @test transform(t, X[:,1]) ≈ Y[:,1]
-    @test reconstruct(t, Y[:,1]) ≈ X[:,1]
     @test reconstruct(t, Y) ≈ X
+    transform!(t, X)
```
nalimilan:
```suggestion
    @test transform!(t, X) === X
```
Same below.

src/transformations.jl
```diff
@@ -223,84 +289,125 @@ julia> StatsBase.transform(dt, X)
...
+        @inbounds tmax[i] = 1 / (tmax[i] - tmin[i])
+    end
+    return UnitRangeTransform(l, dims, unit, tmin, tmax)
+end
+
+function _extract_info(X::AbstractMatrix{<:Real})
+    n, l = size(X)
+    tmin = X[1, :]
+    tmax = X[1, :]
+    for j = 1:l
```
nalimilan: This was enough, right?
```suggestion
    for j = 2:l
```

src/transformations.jl
yuehhua: OK, now I totally understandwhat you said.

test/transformations.jl
yuehhua: OK

src/transformations.jl
yuehhua: This cause test failed. I think that the following `for i = 2:n` is enough.

src/transformations.jl
```diff
@@ -8,63 +8,106 @@ abstract type AbstractDataTransform end
...
+        transform!(x, t, x)
+    elseif t.dims == 2
+        transform!(x', t, x')
+    end
+    return x
+end
+
+function transform!(t::AbstractDataTransform, x::AbstractVector{<:Real})
+    transform!(t, reshape(x, :, 1))
+    return vec(x)
```
nalimilan:
```suggestion
    return x
```

src/transformations.jl
```diff
@@ -8,63 +8,106 @@ abstract type AbstractDataTransform end
...
+function reconstruct(t::AbstractDataTransform, y::AbstractMatrix{<:Real})
+    if t.dims == 1
+        return reconstruct!(similar(y), t, y)
+    elseif t.dims == 2
+        return reconstruct!(similar(y)', t, y')'
+    end
+end
+
+function reconstruct(t::AbstractDataTransform, y::AbstractVector{<:Real})
+    ymat = reconstruct(t, reshape(y, :, 1))
```
nalimilan:
```suggestion
    vec(reconstruct(t, reshape(y, :, 1)))
```
and remove next line.

src/transformations.jl
```diff
@@ -8,63 +8,106 @@ abstract type AbstractDataTransform end
...
+        reconstruct!(y, t, y)
+    elseif t.dims == 2
+        reconstruct!(y', t, y')
+    end
+    return y
+end
+
+function reconstruct!(t::AbstractDataTransform, y::AbstractVector{<:Real})
+    reconstruct!(t, reshape(y, :, 1))
+    return vec(y)
```
nalimilan:
```suggestion
    return y
```

src/transformations.jl
```diff
@@ -97,22 +143,46 @@ julia> StatsBase.transform(dt, X)
...
+    return ZScoreTransform(1, dims, (center ? [m] : zeros(T, 0)),
+                                    (scale ? [s] : zeros(T, 0)))
 end
 
-function transform!(y::AbstractVecOrMat{<:Real}, t::ZScoreTransform, x::AbstractVecOrMat{<:Real})
-    d = t.dim
-    size(x,1) == size(y,1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y,2)
-    size(x,2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+function transform!(y::AbstractMatrix{<:Real}, t::ZScoreTransform, x::AbstractMatrix{<:Real})
```
nalimilan: This function doesn't take into account `t.dims`? Same for the other one below.

src/transformations.jl
nalimilan: Ah sorry I didn't notice you had switched `i` and `j`.

src/transformations.jl
```diff
@@ -8,63 +8,106 @@ abstract type AbstractDataTransform end
...
+function transform(t::AbstractDataTransform, x::AbstractMatrix{<:Real})
+    if t.dims == 1
+        return transform!(similar(x), t, x)
+    elseif t.dims == 2
+        return transform!(similar(x)', t, x')'
+    end
+end
+
+function transform(t::AbstractDataTransform, x::AbstractVector{<:Real})
+    xmat = transform(t, reshape(x, :, 1))
```
nalimilan:
```suggestion
   vec(transform(t, reshape(x, :, 1)))
```
and remove next line.

src/transformations.jl
yuehhua: I move the `t.dims` check above (i.e. `transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real})` and related `transform`, `reconstruct!`, `reconstruct`).
Here, it would focus on dealing with column-major calculation.

yuehhua:
@nalimilan Is it good?

src/transformations.jl
nalimilan: But that method is publicly available, we can't assume people will call it and expect the column-major convention.

yuehhua: Maybe I didn't explain myself clearly. Those methods are publicly available. Users call those methods directly, and `t.dims` is specified. In that, further operations are dispatched to internals (i.e. `transform!(x, t, x)`) by `t.dims`. There is the only one version `transform!(x, t, x)` operates in column-major convention for performance. Therefore, users are allowed to use the operations they want, and the internals just conduct in column-major convention.

wildart: @nalimilan is correct. You need to move `t.dim` checks from `transfrom` to `transform!`.

yuehhua: OK, I see it. I didn't notice that `transform!(y, t, x)` is also public.

yuehhua: Do we need to public `transform!(y, t, x)`? It's not tested and lack of documentation, but public before.

wildart: `transform!` is public

wildart:
It is common for in-place update functions to pass as a first parameter a variable where the result of the operation will be stored. So, for `transform!(y, t, x)`, `y` argument is that variable. I do not see a point of creating some "private" functions when `transform!` & `reconstruct!` already exist. You just need to add proper dimension checks related for `t.dim` parameter.

yuehhua:
> It is common for in-place update functions to pass as a first parameter a variable where the result of the operation will be stored. So, for `transform!(y, t, x)`, `y` argument is that variable. I do not see a point of creating some "private" functions when `transform!` & `reconstruct!` already exist.
Yeah... I got your point. I know it is common way to go.

> You just need to add proper dimension checks related for `t.dim` parameter.
If I move dimensions checks from `transform` to `transfrom!` like this:
```
function transform!(y, t, x)
    if t.dims == 1
        (original part of code)
    elseif t.dims == 2
        (How can I implement here?)
    end
end
```

Implementing a similar algorithm in row is not applicable.
If I use recursion call `transfrom!(y', t, x')'`, `StackOverFlowError` occurs. Since `t.dims` is immutable, it never ends.
I cannot come out with a feasible solution to achieve this need.

nalimilan:
> If I use recursion call `transfrom!(y', t, x')'`, `StackOverFlowError` occurs. Since `t.dims` is immutable, it never ends.
> I cannot come out with a feasible solution to achieve this need.

You should be able to avoid this by creating a new transformation object with the expected `dim`.

yuehhua:
@nalimilan I committed. Would you please review it? Thank you.

src/transformations.jl
```diff
@@ -223,84 +283,135 @@ julia> StatsBase.transform(dt, X)
...
+function fit(::Type{UnitRangeTransform}, X::AbstractVector{<:Real};
+             dims::Union{Integer,Nothing}=nothing, unit::Bool=true)
+    if dims == nothing
+        Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
+    elseif dims != 1
+        throw(DomainError(dims, "fit only accept dims=1 over a vector. Try fit(t, x, dims=1)."))
     end
-    return UnitRangeTransform(d, unit, tmin, tmax)
+
+    n = length(X)
```
nalimilan: Can't you call `_extract_info` here?

src/transformations.jl
```diff
@@ -223,84 +283,135 @@ julia> StatsBase.transform(dt, X)
...
-        broadcast!(*, y, x, tscale)
+        if t.unit
+            broadcast!((x,s,m)->(x-m)*s, y, x, tscale', tmin')
+        else
+            broadcast!(*, y, x, tscale')
+        end
+        return y
+    elseif t.dims == 2
+        t_ = UnitRangeTransform(t.len, 1, t.unit, t.min, t.scale)
+        return transform!(y', t_, x')'
```
nalimilan: Same here.

src/transformations.jl
```diff
@@ -223,84 +283,135 @@ julia> StatsBase.transform(dt, X)
...
-        broadcast!(/, x, y, tscale)
+        if t.unit
+            broadcast!((y,s,m)->y/s+m, x, y, tscale', tmin')
+        else
+            broadcast!(/, x, y, tscale')
+        end
+        return x
+    elseif t.dims == 2
+        t_ = UnitRangeTransform(t.len, 1, t.unit, t.min, t.scale)
+        return reconstruct!(x', t_, y')'
```
nalimilan: Same here.

src/transformations.jl
```diff
@@ -223,84 +283,135 @@ julia> StatsBase.transform(dt, X)
...
+    standardize(DT, X; dims=nothing, kwargs...)
 
-Return a row-standardized matrix `X` using `DT` transformation which is a subtype of `AbstractDataTransform`:
+ Return a standardized copy of vector or matrix `X` along dimensions `dims`
+ using transformation `DT` which is a subtype of `AbstractDataTransform`:
 
 - `ZScoreTransform`
 - `UnitRangeTransform`
 
+Return a standardized matrix while the input of `X` is a `AbstractMatrix`.
```
nalimilan: These two lines sound redundant with the description above.

src/transformations.jl
```diff
@@ -8,63 +8,84 @@ abstract type AbstractDataTransform end
 
 Apply transformation `t` to vector or matrix `x` in place.
 """
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(x, t, x)
-transform!(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(x, t, x)
+function transform!(t::AbstractDataTransform, x::AbstractMatrix{<:Real})
```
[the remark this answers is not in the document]
yuehhua: OK.

src/transformations.jl
```diff
@@ -8,63 +8,84 @@ abstract type AbstractDataTransform end
...
 """
     reconstruct!(t::AbstractDataTransform, y)
 
-Perform an in-place reconstruction into an original data scale from a row-transformed
+Perform an in-place reconstruction into an original data scale from a transformed
 vector or matrix `y` using `t` transformation.
 """
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,1}) = reconstruct!(y, t, y)
-reconstruct!(t::AbstractDataTransform, y::AbstractArray{<:Real,2}) = reconstruct!(y, t, y)
+function reconstruct!(t::AbstractDataTransform, y::AbstractMatrix{<:Real})
```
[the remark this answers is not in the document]
yuehhua: OK.

src/transformations.jl
```diff
@@ -8,63 +8,84 @@ abstract type AbstractDataTransform end
...
 
-Return a reconstruction of an originally scaled data from a row-transformed vector
+Return a reconstruction of an originally scaled data from a transformed vector
 or matrix `y` using `t` transformation.
 """
-reconstruct(t::AbstractDataTransform, y::AbstractArray{<:Real,1})  = reconstruct!(similar(y), t, y)
-reconstruct(t::AbstractDataTransform, y::AbstractArray{<:Real,2})  = reconstruct!(similar(y), t, y)
+reconstruct(t::AbstractDataTransform, y::AbstractMatrix{<:Real}) = reconstruct!(similar(y), t, y)
+
+function reconstruct(t::AbstractDataTransform, y::AbstractVector{<:Real})
```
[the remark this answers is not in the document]
yuehhua: OK.

src/transformations.jl
yuehhua: OK.

src/transformations.jl
```diff
@@ -97,100 +121,135 @@ julia> StatsBase.transform(dt, X)
...
+            if isempty(s)
+                broadcast!(-, y, x, m')
+            else
+                broadcast!((x,m,s)->(x-m)/s, y, x, m', s')
+            end
         end
+        return y
+    elseif t.dims == 2
+        t_ = ZScoreTransform(t.len, 1, t.mean, t.scale)
+        return transform!(y', t_, x')'
```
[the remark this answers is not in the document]
yuehhua: OK.

src/transformations.jl
```diff
@@ -97,100 +121,135 @@ julia> StatsBase.transform(dt, X)
...
+            if isempty(s)
+                broadcast!(+, x, y, m')
+            else
+                broadcast!((y,m,s)->y*s+m, x, y, m', s')
+            end
         end
+        return x
+    elseif t.dims == 2
+        t_ = ZScoreTransform(t.len, 1, t.mean, t.scale)
+        return reconstruct!(x', t_, y')'
```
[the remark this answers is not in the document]
yuehhua: OK.

src/transformations.jl
```diff
@@ -223,84 +283,135 @@ julia> StatsBase.transform(dt, X)
...
+        throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
+    end
+
+    for i = 1:l
+        @inbounds tmax[i] = 1 / (tmax[i] - tmin[i])
+    end
+    return UnitRangeTransform(l, dims, unit, tmin, tmax)
+end
+
+function _extract_info(X::AbstractMatrix{<:Real})
```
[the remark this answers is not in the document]
yuehhua: OK.

src/transformations.jl
yuehhua: OK.

src/transformations.jl
yuehhua: OK.

src/transformations.jl
yuehhua: OK.

src/transformations.jl
yuehhua: OK.

src/transformations.jl
```diff
@@ -97,100 +121,135 @@ julia> StatsBase.transform(dt, X)
...
 
-function transform!(y::AbstractVecOrMat{<:Real}, t::ZScoreTransform, x::AbstractVecOrMat{<:Real})
-    d = t.dim
-    size(x,1) == size(y,1) == d || throw(DimensionMismatch("Inconsistent dimensions."))
-    n = size(y,2)
-    size(x,2) == n || throw(DimensionMismatch("Inconsistent dimensions."))
+function fit(::Type{ZScoreTransform}, X::AbstractVector{<:Real};
+             dims::Union{Integer,Nothing}=nothing, center::Bool=true, scale::Bool=true)
+    if dims == nothing
+        Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
```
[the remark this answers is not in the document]
yuehhua: OK.

src/transformations.jl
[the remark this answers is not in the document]
yuehhua: But it failed though ... [travis-ci](https://travis-ci.org/JuliaStats/StatsBase.jl/jobs/576461041)

src/transformations.jl
```diff
@@ -8,63 +8,84 @@ abstract type AbstractDataTransform end
...
     transform(t::AbstractDataTransform, x)
 
-Return a row-standardized vector or matrix `x` using `t` transformation.
+Return a standardized vector or matrix `x` using `t` transformation.
 """
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,1}) = transform!(similar(x), t, x)
-transform(t::AbstractDataTransform, x::AbstractArray{<:Real,2}) = transform!(similar(x), t, x)
+transform(t::AbstractDataTransform, x::AbstractMatrix{<:Real}) = transform!(similar(x), t, x)
+
+function transform(t::AbstractDataTransform, x::AbstractVector{<:Real})
```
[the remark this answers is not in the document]
yuehhua: Not going to reshape to matrix which means that I have to implement a 3-argument `transform!` for vector, right? Isn't it just the way I have been done before and got suggestion to change to use `reshape`? I got confused.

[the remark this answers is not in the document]
nalimilan: I think the best approach is to define the 3-argument `transform!` for vectors and have reshape the vector to a matrix and pass it to `transform!` for matrices. *But* the vector function should return the original vector, *not* the reshaped matrix. The reshaping is just an internal trick to avoid code duplication.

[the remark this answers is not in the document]
yuehhua: OK, that's clear.

src/transformations.jl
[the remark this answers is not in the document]
yuehhua: Fixed.

src/transformations.jl
```diff
@@ -97,100 +121,135 @@ julia> StatsBase.transform(dt, X)
...
+        Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
+    elseif dims != 1
+        throw(DomainError(dims, "fit only accept dims=1 over a vector. Try fit(t, x, dims=1)."))
+    end
 
-    m = t.mean
-    s = t.scale
+    T = eltype(X)
+    m, s = mean_and_std(X)
+    return ZScoreTransform(1, dims, (center ? [m] : zeros(T, 0)),
```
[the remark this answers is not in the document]
nalimilan: Has this been addressed?

test/transformations.jl
```diff
@@ -1,66 +1,182 @@
...
 @testset "Transformations" begin
+    # matrix
     X = rand(5, 8)
+    X_ = copy(X)
 
-    t = fit(ZScoreTransform, X; center=false, scale=false)
+    t = fit(ZScoreTransform, X, dims=1, center=false, scale=false)
     Y = transform(t, X)
     @test isa(t, AbstractDataTransform)
     @test isempty(t.mean)
```
yuehhua: @nalimilan I noticed that tests are covered and asked to be empty. I tried to change to `zeros(T, 1)`, then transformation and reconstruct failed.

src/transformations.jl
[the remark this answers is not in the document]
nalimilan: Ah, OK, it's just meant to create an empty vector. It would be more logical written as `T[]`, but that doesn't matter a lot.

Do you think the PR is ready for another (hopefully the last) round of review?

test/transformations.jl
nalimilan: OK, good.

src/transformations.jl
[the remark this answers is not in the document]
yuehhua: Sure, anytime.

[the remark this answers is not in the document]
yuehhua: Sure, anytime.

PR #494, opened 2019-05-11
added Jensen-Shannon divergence

nash1111:
added Jensen-Shannon divergence,
i'll add some other functions like mutual information later


nalimilan:
Thanks. I think the plan is to remove distance/divergence functions from StatsBase as they are defined in Distances.jl already. Would you open a PR there? Please also add tests and include the new function in the docs.

nalimilan:
Actually it seems `js_divergence` already exists in Distances.jl. :-/

nalimilan:
Should this be closed then?

nash1111:
yes, sorry for late reply

PR #495, opened 2019-05-23
add docstring to Histogram

jw3126:


jw3126:
@nalimilan could you review?

nalimilan:
Sorry for the delay (I'm not too familiar with the histogram code). I wonder whether we should document the fields, since they are supposed to be internal. Probably better attach the docstring to the constructors and to `fit`.

BTW, can you follow conventions detailed at https://docs.julialang.org/en/latest/manual/documentation/?

jw3126:
Are the fields private? It is important to know whether a histogram is a density or not. How can that info be private?

jw3126:
Not to mention `edges` and `weights`, no way they are private.

nalimilan:
In general in Julia all fields are private. You can perfectly access them, but there's no guaranty they won't change in the future. Accessor functions are defined when something is part of the public API.

That said, sometimes fields could be public too, and by overriding `getproperty` we can create "pseudo-fields" that may not reflect the internal fields, so accessor functions may not be required.

Anyway as I said I don't know the histogram code well, so I can't say what's the public API.

jw3126:
> In general in Julia all fields are private. You can perfectly access them, but there's no guaranty they won't change in the future. Accessor functions are defined when something is part of the public API.

That may be true in theory. In practice  I think you will break most packages that use `Histogram` if you e.g. rename the `weights` field.

src/hist.jl
```diff
@@ -115,6 +115,33 @@ end
...
 
 # N-dimensional histogram object
+"""
+The `Histogram` type represents data that has been tabulated into intervals
+(known as *bins*) along the real line, or in higher dimensions, over the real plane. Histograms can be fitted to data using the `fit` method.
+
+# Fields
+* edges: An iterator that contains the boundaries of the bins in each dimension.
+* weights: Array that contains the weight of each bin.
+* closed: The bins are (higher dimensional analogues) of half open intervals. The `closed` fieldis a symbol with value `:right` or `:left` which side if closed.
```
quinnj: `fieldis` => `field is`

`that indicates which side`

src/hist.jl
```diff
@@ -115,6 +115,33 @@ end
...
+
+julia> fit(Histogram, [2.],  1:3, closed=:right)
+Histogram{Int64,1,Tuple{UnitRange{Int64}}}
+edges:
+  1:3
+weights: [1, 0]
+closed: right
+isdensity: false
+```
+* isdensity: There are two interpretations of a `Histogram`. If `isdensity=false` the weight of a bin corresponds to amount of a quantity in the bin. If `isdensity=true` then it corresponds to the density (amount / volume) of the quantity in the bin.
```
quinnj: `corresponds to the amount of a `

src/hist.jl
```diff
@@ -115,6 +115,37 @@ end
 abstract type AbstractHistogram{T<:Real,N,E} end
 
 # N-dimensional histogram object
+"""
+    Histogram <: AbstractHistogram
+
+The `Histogram` type represents data that has been tabulated into intervals
+(known as *bins*) along the real line, or in higher dimensions, over a real space.
```
nalimilan: I've changed "plane" to "space" since AFAIK the former refers to a 2D space only. Right? (This was in the original docs.)

src/hist.jl
```diff
@@ -115,6 +115,37 @@ end
...
+* closed: A symbol with value `:right` or `:left` indicating on which side bins
+  (which half-open intervals or higher-dimensional analogues thereof) are closed.
+```jldoctest
+julia> fit(Histogram, [2.],  1:3, closed=:left)
+Histogram{Int64,1,Tuple{UnitRange{Int64}}}
+edges:
+  1:3
+weights: [0, 1]
+closed: left
+isdensity: false
```
nalimilan: That field should be documented too.

src/hist.jl
jw3126: Sure!

src/hist.jl
jw3126: I think I documented all fields, if you are missing `isdensity`, it is there, just a few lines below.

nalimilan: Ah, then it should be above the example, not below.

Also, the description of `isdensity` doesn't sound precise enough to me: "amount" is used for both `true` and `false`. AFAIK when `true` the weights sum to unity? Or maybe it's more subtle than that?

jw3126: > Ah, then it should be above the example, not below.

Not sure. The example is about the `closed` field

> Also, the description of isdensity doesn't sound precise enough to me: "amount" is used for both true and false.

Mhh one is  "amount" and the other is "amount per volume". Can you reread the docstring and suggest an concrete improvement?

> AFAIK when `true` the weights sum to unity? Or maybe it's more subtle than that?

I though so too. But then my code had strange bugs and I discovered the real meaning of `isdensity`. In fact `isdensity` has little relation to whether the histogram is normalized.

nalimilan: > Not sure. The example is about the `closed` field

Well, it breaks the list of arguments, and it also illustrates other fields (which don't have their own examples). AFAICT the convention is to put all examples at the end.

> Mhh one is "amount" and the other is "amount per volume". Can you reread the docstring and suggest an concrete improvement?

I'm not sure I can suggest an exact wording since as I said I don't know what this field means. But at least I can say that the description wasn't enough for me to understand that. :-) What is "volume" here? Also I'd say "amount of a quantity" is a pleonasm.

Maybe @simonbyrne can help.

jw3126: I will try an example. We have an area in which weasels live. The area is subdivided into a grid given by `edges`. Imagine non uniform grid, e.g `edges=([1,2,10], 1:3)`. We want a histogram of the weasel population and need to fill the `weights` matrix. One way is to fill the matrix with the numbers of weasels living in each cell. The other is to fill the matrix with the density of weasels in each cell. This first corresponds to `isdensity=true` the latter to `isdensity=false`.

nalimilan: OK, thanks. Then how can we define "density" in general? I guess it is the weights divided by the width (1D) or area (>1D) of the bin?

nalimilan: FWIW, [Numpy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html) says this:
> density: If False, the result will contain the number of samples in each bin. If True, the result is the value of the probability density function at the bin, normalized such that the integral over the range is 1. Note that the sum of the histogram values will not be equal to 1 unless bins of unity width are chosen; it is not a probability mass function.

jw3126: By denisity I mean number of weasels living in the bin / area of the bin.

jw3126: Ah nice it did not occur to me to look at numpy. In our case `isdensity=true` does not mean that the integral is one (which is good I think).

jw3126: I added an example, you think it clarifies things @nalimilan ?

PR #499, opened 2019-05-31
support Float32 for alias sampling

### LINKED ISSUE #158: Sampling with WeightVec and Float32
I encountered a bug when wanting to sample from weight vec of single-precision floats. Below is a minimal working example.

```
prob_32 = rand(Float32, 1000);
prob_64 = rand(Float64, 1000);
sample(1:1000, WeightVec(prob_64), 31);
sample(1:1000, WeightVec(prob_32), 31);  #both work
sample(1:1000, WeightVec(prob_64), 32); #this works
sample(1:1000, WeightVec(prob_32), 32); #this throws error

ERROR: MethodError: `make_alias_table!` has no method matching make_alias_table!(::Array{Float32,1}, ::Float32, ::Array{Float64,1}, ::Array{Int64,1})
Closest candidates are:
  make_alias_table!(::AbstractArray{Float64,1}, ::Float64, ::AbstractArray{Float64,1}, ::AbstractArray{Int64,1})
 in alias_sample! at /HOME/.julia/v0.4/StatsBase/src/sampling.jl:424
 in sample! at /HOME/.julia/v0.4/StatsBase/src/sampling.jl:515
```

I can get around this problem by sampling one item at a time, but seems like a bug to fix.

findmyway:
This fixes #158

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -451,9 +451,9 @@ end
 direct_sample!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
     direct_sample!(Random.GLOBAL_RNG, a, wv, x)
 
-function make_alias_table!(w::AbstractVector{Float64}, wsum::Float64,
-                           a::AbstractVector{Float64},
-                           alias::AbstractVector{Int})
+function make_alias_table!(w::AbstractVector{T}, wsum::S,
+                           a::AbstractVector{T},
+                           alias::AbstractVector{<:Integer}) where {S, T}
     # Arguments:
     #
     #   w [in]:         input weights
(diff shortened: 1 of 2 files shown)


bkamins:
See #506 for a fix which uses `AbstractWeights` which I think is a bit cleaner.

(and bump this PR as I am also getting this error but with `Int` weights)

findmyway:
Yeah, I added a limitation element type of `w` and `a` must be the same originally. If that's unnecessary, then I prefer your PR.

bkamins:
The problem is that:
```
julia> using StatsBase

julia> Weights([1,2,3])
3-element Weights{Int64,Int64,Array{Int64,1}}:
 1
 2
 3
```
creates `S` and `T` as `Int` in your PR, while `ap` must be able to hold floats`.

The best way to make sure that all works is to add a test case with integer weights in your PR.

The best way to get a type for `ap` is to make it `typeof(1*one(S)/one(T))`. This should be inferred in compile time, but it would be best to make sure that this is what really happens 😄.

The other change is that in my PR we pass a weight vector to `make_alias_table!` which I think is enough as it already puts a restriction on its element types (via its constructor), but this is minor - we can leave it as in your PR (the type of `a` needs to change only, and probably it is enough that you make it a subtype of `AbstractFloat` in the signature).

findmyway:
Sorry it's been a pretty long time. I can hardly remember what I tried to do here... 

@bkamins Could you be more specific about *what have to be fixed*?

devmotion:
Thank you for the PR @findmyway!

PR #500, opened 2019-06-27
extend ecdf to a weighted sample

johnczito:


src/empirical.jl
```diff
@@ -47,7 +50,12 @@ evaluate CDF values on other samples.
 `extrema`, `minimum`, and `maximum` are supported to for obtaining the range over which
 function is inside the interval ``(0,1)``; the function is defined for the whole real line.
 """
-ecdf(X::RealVector{T}) where T<:Real = ECDF(sort(X))
+function ecdf(X::RealVector{T}, w::AbstractWeights{W} = weights(ones(length(X)))) where {T, W <: Real}
```
nalimilan: Instead of allocating a `ones` vector, could you use `Float64[]` when no weights are provided? That way you don't need to allocate a vector for the unweighted case. Then functions should check whether `isempty(X.weights)` to decide what to do.

src/empirical.jl
```diff
@@ -17,22 +20,22 @@ function (ecdf::ECDF)(v::RealVector)
...
-    n = length(ecdf.sorted_values)
-    for x in ecdf.sorted_values
+    weightsum = sum(ecdf.weights)
+    for (j, x) in enumerate(ecdf.sorted_values)
         while i <= m && x > v[ord[i]]
             r[ord[i]] = r0
             i += 1
         end
-        r0 += 1
+        r0 += ecdf.weights.values[j]
```
nalimilan:
```suggestion
        r0 += ecdf.weights[j]
```

src/empirical.jl
```diff
@@ -3,12 +3,15 @@
...
-struct ECDF{T <: AbstractVector{<:Real}}
+struct ECDF{T <: AbstractVector{<:Real}, W <: AbstractWeights{<:Real}}
     sorted_values::T
+    weights::W
 end
 
 function (ecdf::ECDF)(x::Real)
-    searchsortedlast(ecdf.sorted_values, x) / length(ecdf.sorted_values)
+    weightsum = sum(ecdf.weights)
+    n         = searchsortedlast(ecdf.sorted_values, x)
```
nalimilan:
```suggestion
    n = searchsortedlast(ecdf.sorted_values, x)
```

src/empirical.jl
```diff
@@ -3,12 +3,15 @@
...
+    sum(ecdf.weights.values[1:n]) / weightsum
```
nalimilan:
```suggestion
    sum(view(ecdf.weights, 1:n)) / weightsum
```

src/empirical.jl
johnczito: Alright, I took a crack at it. Let me know what you think.

johnczito:
@nalimilan @rofinn Let me know what you want to see here for this to be considered for merging.

test/empirical.jl
```diff
@@ -13,3 +13,19 @@ using Test
...
+    y = [-1.96, -1.644854, -1.281552, -0.6744898, 0, 0.6744898, 1.281552, 1.644854, 1.96]
+    @test fnecdf.weights.values != w.values
+    @test fnecdf(y) ≈ map(fnecdf, y)
+    @test extrema(fnecdf) == (minimum(fnecdf), maximum(fnecdf)) == extrema(x)
+    fnecdf = ecdf([1.0, 0.5], weights([3, 1]))
+    @test fnecdf(0.75) == 0.25
+    @test extrema(fnecdf) == (minimum(fnecdf), maximum(fnecdf)) == (0.5, 1.0)
+
+    @test_throws MethodError ecdf(rand(8), rand(10))
+
```
nalimilan:
```suggestion
```

test/empirical.jl
```diff
@@ -13,3 +13,19 @@ using Test
...
     @test extrema(fnecdf) == (minimum(fnecdf), maximum(fnecdf)) == (0.5, 0.5)
 end
+
+@testset "Weighted ECDF" begin
+    x = randn(10000000)
+    w = weights(rand(10000000))
+    fnecdf = ecdf(x, w)
+    y = [-1.96, -1.644854, -1.281552, -0.6744898, 0, 0.6744898, 1.281552, 1.644854, 1.96]
+    @test fnecdf.weights.values != w.values
+    @test fnecdf(y) ≈ map(fnecdf, y)
```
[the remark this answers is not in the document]
johnczito: Right. I initially left it out because I wasn't sure about, but I think it's the same test. As in, since the weights are uniform and independent of the sample points, the empirical quantiles should be the same as an evenly weighted sample.

src/empirical.jl
```diff
@@ -47,7 +53,14 @@ evaluate CDF values on other samples.
 `extrema`, `minimum`, and `maximum` are supported to for obtaining the range over which
 function is inside the interval ``(0,1)``; the function is defined for the whole real line.
 """
-ecdf(X::RealVector{T}) where T<:Real = ECDF(sort(X))
+ecdf(X::RealVector{T}) where T<:Real = ECDF(sort(X), weights(Float64[]))
+
+function ecdf(X::RealVector{T}, w::AbstractWeights{W}) where {T, W <: Real}
```
nalimilan:
```suggestion
function ecdf(X::RealVector, w::AbstractVector{<:Real})
```

src/empirical.jl
```diff
@@ -47,7 +53,14 @@ evaluate CDF values on other samples.
 `extrema`, `minimum`, and `maximum` are supported to for obtaining the range over which
 function is inside the interval ``(0,1)``; the function is defined for the whole real line.
 """
-ecdf(X::RealVector{T}) where T<:Real = ECDF(sort(X))
+ecdf(X::RealVector{T}) where T<:Real = ECDF(sort(X), weights(Float64[]))
```
nalimilan:
```suggestion
ecdf(X::RealVector) = ECDF(sort(X), weights(Float64[]))
```

src/empirical.jl
```diff
@@ -47,7 +53,14 @@ evaluate CDF values on other samples.
...
+
+function ecdf(X::RealVector{T}, w::AbstractWeights{W}) where {T, W <: Real}
+    length(X) == length(w) || throw(ArgumentError("data and weight vectors must be the same size," *
+        "got $(length(X)) and $(length(w))"))
+    ord = sortperm(X)
+    ECDF(X[ord], weights(w.values[ord]))
```
nalimilan:
```suggestion
    ECDF(X[ord], weights(w[ord]))
```

johnczito:
@nalimilan Are you saying you want support for all three options: `(wt - 1)/T`, `wt/(T+1)`, or `wt/T`? Or you want to implement "the best" of those three (which may or may not be `wt/T`)?

nalimilan:
Sorry for the delay. I hoped @matthieugomez would have had the chance to comment. I'm not sure we need to support all three options right now, but I wonder whether different options would make sense depending on the particular `AbstractWeights` subtype that is passed. Ideally we could implement a single solution, but in `quantile` that's not what we did, for good reasons.

A good way to check whether a single method is possible is that for `FrequencyWeights` the result should be equivalent to calling the unweighted method after replicating each observation the number of times corresponding to its weight. On the contrary, for `ProbabilityWeights`, the result should not change if all weights are multiplied by an arbitrary value. If these two properties are incompatible for the ECDF, then we need several methods. Is that the case?

johnczito:
I've added unit tests that confirm that: (1) for integer weights, the weighted method is equivalent to the unweighted method where you replicate the observations according to the weights; (2) the output of the weighted method does not change if you rescale the weights arbitrarily.

Let me know what else you want to see for this to be considered for merging.

src/empirical.jl
```diff
@@ -47,7 +53,14 @@ evaluate CDF values on other samples.
 `extrema`, `minimum`, and `maximum` are supported to for obtaining the range over which
 function is inside the interval ``(0,1)``; the function is defined for the whole real line.
 """
-ecdf(X::RealVector{T}) where T<:Real = ECDF(sort(X))
+ecdf(X::RealVector) = ECDF(sort(X), weights(Float64[]))
+
+function ecdf(X::RealVector, w::AbstractVector{<:Real})
```
nalimilan: Actually we're planning to use a `weights` keyword argument in the future to pass weights to all functions. Could you change the PR to do that?

nalimilan:
OK, thanks, sounds good.

src/empirical.jl
johnczito: Changed.

johnczito:
@nalimilan Thanks for all of the feedback. What's the path to a merge?

matbesancon:
looks good to me :)

nalimilan:
Thanks @johnczito! Sorry for the delays, as you can see people are very busy so it's hard to get feedback.

PR #501, opened 2019-07-03
Bump to next non-breaking version

nickrobinson251:
Please may we release a version with the new `eweights` functionality added in https://github.com/JuliaStats/StatsBase.jl/pull/401

https://github.com/JuliaStats/StatsBase.jl/compare/v0.30.0...master

We're pre-1.0, and this is backwards compatible with `v0.30`, so just incremented to `0.30.1`
:)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
-version = "0.30.0"
+version = "0.30.1"
 
 [deps]
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"


rofinn:
I'd still vote to make it a minor release. I like to always reserve patch releases for bug fixes. Appveyor failure is a testing issue on fweights (`sqrt` on negative number) that should be fixed at some point.

nickrobinson251:
> I'd still vote to make it a minor release.

I am strongly against that in the pre-1.0 world, because having only two digits to play with, we should only distinguish breaking v non-breaking changes (and this is non-breaking).

Because 0.2 is considered incompatible with 0.3 etc., it's annoying to label backward compatible versions as incompatible. You end up with either:
- PkgA listing compat for a bunch of 0.Y versions for StatsBase in the Project.toml: this is painful as a maintainer, it does not distinguish actually breaking from non-breaking versions, and it's annoying to check if/when something should be dropped from a long compat list in the future
- PkgA listing only one StatsBase 0.Y version, e.g. 0.1, and PkgB listing only e.g. 0.2, but if 0.1 and 0.2 are compatible then we're going to get unnecessary "unsatsfiable requirement" errors when using PkgA and PkgB

ararslan:
This should still be a minor release. If someone wants to start depending on `eweights` in their code, they'll need to adjust their version of StatsBase anyway.

ararslan:
Closing in favor of #502.

PR #506, opened 2019-07-26
fix make_alias_table!

bkamins:
Use `AbstractWeights` in `make_alias_table!`. Currently it fails if weights are not `Float64` (a typical case is when they are `Int`).


bkamins:
dup of #499. Closing. Sorry.

PR #507, opened 2019-07-30
version bump

matbesancon:


bkamins:
the error on nightly seems unrelated. DataAPI.jl dependency does not version constrain as there is only one release so far and it includes `describe`.

matbesancon:
I fixed it (I think?) in the last PR, just a matter of adapting the tests

Issue #358, opened 2018-03-14
UnitWeights


Nosferican:
I would like to discuss adding a new type of weights for consistency and efficiency purposes. The proposed type is basically [UnitWeights](https://github.com/lbittarello/Microeconometrics.jl/blob/6887836da6e7ec9f0cb6e94a80c7db992cd4439b/src/data/weights.jl).

nalimilan:
I'm in favor in principle, but why would we need to actually store a vector of weights if we know they are unit weights? I would have thought we should use an implementation similar to https://github.com/JuliaStats/StatsBase.jl/issues/135.

Nosferican:
I would be fine with
```
struct UnitWeights <: AbstractWeights end
```
Then just specialize the methods for efficiency.

nalimilan:
If we add this, it should also implement `getindex` and `sum` so that it behaves like other weights types (and like vectors).

Nosferican:
Something like?
```
struct UnitWeights <: AbstractWeights
    sum::Int64
    UnitWeights(obj::AbstractVector) = new(length(obj))
end
getindex(obj::UnitWeights) = one(Float64)
sum(obj::UnitWeights) = getfield(obj, :sum)
```

nalimilan:
Mostly, though that's a bit more complex, see #135.

Nosferican:
FillArrays provides the `Ones` which are optimal for this.

lbittarello:
I have redesigned `UnitWeights` in Microeconometrics.jl (cf. [here](https://github.com/lbittarello/Microeconometrics.jl/blob/master/src/data/weights.jl)). It does not store a vector of ones any longer. We could perhaps port it into StatsBase if there is any interest and the implementation is fine. Let me know.

nalimilan:
Sure, please file a PR! Though looking at your implementation it appears to allow any value, while "unit" evokes weights equal to 1 to me, doesn't it?

lbittarello:
True. As I understand it, the current inner constructor should prevent the user from creating weights with a different value, but the user could later modify it. The following alternative implementation is safer, albeit marginally less efficient:

```
struct UnitWeights{S<:Real, T<:Real} <: AbstractWeights{S, T, V where V<:Vector{T}}
    sum::S
end

UnitWeights(::Type{T}, s::S) where {S, T} = UnitWeights{S, T}(s)
[...]

PR #515, opened 2019-08-22
Add unit weights

### LINKED ISSUE #135: A Ones type

lbittarello:
Closes #135, #358 and #516

### DIFF
--- a/docs/src/weights.md
+++ b/docs/src/weights.md
@@ -41,6 +41,17 @@ w = ProbabilityWeights([0.2, 0.1, 0.3])
 w = pweights([0.2, 0.1, 0.3])
 ```
 
+### `UnitWeights`
+
+Unit weights are a special case in which all observations are given a weight equal to `1`. Using such weights is equivalent to computing unweighted statistics.
+
+This type can notably be used when implementing an algorithm so that a only a weighted variant has to be written. The unweighted variant is then obtained by passing a `UnitWeights` object. This is very efficient since no weights vector is actually allocated.
+
+```julia
+w = uweights(Float64, 3)
+w = uweights(3)
+```
+
 ### `Weights`
 
 The `Weights` type describes a generic weights vector which does not support all operations possible for `FrequencyWeights`, `AnalyticWeights` and `ProbabilityWeights`.
@@ -114,15 +125,6 @@ julia> eweights(t, r, 0.3)
 
 NOTE: This is equivalent to `eweights(something.(indexin(t, r)), 0.3)`, which is saying that for each value in `t` return the corresponding index for that value in `r`.
 Since `indexin` returns `nothing` if there is no corresponding value from `t` in `r` we use `something` to eliminate that possibility.
-
-### `UnitWeights`
-
-Unit weights treat all observations equally.
-
-```julia
-w = uweights(Float64, 3)
-w = uweights(3)
-```
 
 ## Methods
 
--- a/docs/src/weights.md
+++ b/docs/src/weights.md
@@ -120,9 +120,8 @@ Since `indexin` returns `nothing` if there is no corresponding value from `t` in
 Unit weights treat all observations equally.
 
 ```julia
-w = UnitWeights{Float64}(3)
-w = weights(Float64, 3)
-w = weights(3)
+w = uweights(Float64, 3)
+w = uweights(3)
 ```
 
 ## Methods
--- a/docs/src/weights.md
+++ b/docs/src/weights.md
@@ -115,6 +115,16 @@ julia> eweights(t, r, 0.3)
 NOTE: This is equivalent to `eweights(something.(indexin(t, r)), 0.3)`, which is saying that for each value in `t` return the corresponding index for that value in `r`.
 Since `indexin` returns `nothing` if there is no corresponding value from `t` in `r` we use `something` to eliminate that possibility.
 
+### `UnitWeights`
+
+Unit weights treat all observations equally.
+
+```julia
+w = UnitWeights{Float64}(3)
+w = weights(Float64, 3)
+w = weights(3)
+```
+
 ## Methods
 
 `AbstractWeights` implements the following methods:
--- a/docs/src/weights.md
+++ b/docs/src/weights.md
@@ -48,8 +48,8 @@ Unit weights are a special case in which all observations are given a weight equ
 This type can notably be used when implementing an algorithm so that a only a weighted variant has to be written. The unweighted variant is then obtained by passing a `UnitWeights` object. This is very efficient since no weights vector is actually allocated.
 
 ```julia
-w = uweights(Float64, 3)
 w = uweights(3)
+w = uweights(Float64, 3)
 ```
 
 ### `Weights`
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -179,7 +179,7 @@ pweights(vs::RealArray) = ProbabilityWeights(vec(vs))
 """
     varcorrection(w::ProbabilityWeights, corrected=false)
 
-* `corrected=true`: ``\\frac{n}{(n - 1) \\sum w}`` where ``n`` equals `count(!iszero, w)`
+* `corrected=true`: ``\\frac{n}{(n - 1) \\sum w}``, where ``n`` equals `count(!iszero, w)`
 * `corrected=false`: ``\\frac{1}{\\sum w}``
 """
 @inline function varcorrection(w::ProbabilityWeights, corrected::Bool=false)
@@ -250,11 +250,9 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
 
 # NOTE: No variance correction is implemented for exponential weights
 
-struct UnitWeights{S<:Real, T<:Real} <: AbstractWeights{S, T, V where V<:Vector{T}}
-    sum::S
+struct UnitWeights{S<:Integer, T<:Real} <: AbstractWeights{S, T, V where V<:Vector{T}}
+    len::S
 end
-
-UnitWeights{T}(s::S) where {S, T} = UnitWeights{S, T}(s)
 
 @doc """
     UnitWeights{T}(s)
@@ -263,30 +261,21 @@ Construct a `UnitWeights` vector with length `s` and weight elements of type `T`
 All weight elements are identically one.
 """ UnitWeights
 
-eltype(wv::UnitWeights{S, T})    where {S, T} = T
-values(wv::UnitWeights{S, T})    where {S, T} = fill(one(T), length(wv))
-isempty(wv::UnitWeights{S, T})   where {S, T} = iszero(wv.sum)
-length(wv::UnitWeights{S, T})    where {S, T} = Int(wv.sum)
-Base.size(wv::UnitWeights{S, T}) where {S, T} = Tuple(length(wv))
+eltype(wv::UnitWeights{S, T}) where {S, T} = T
+values(wv::UnitWeights{S, T}) where {S, T} = fill(one(T), length(wv))
+sum(wv::UnitWeights{S, T}) where {S, T} = convert(T, length(wv))
+isempty(wv::UnitWeights) = iszero(wv.len)
+length(wv::UnitWeights) = wv.len
+Base.size(wv::UnitWeights) = Tuple(length(wv))
 
-function Base.getindex(wv::UnitWeights{S, T}, i::Int) where {S, T}
-    (i > 0) & (i <= length(wv)) ? one(T) : Base.throw_boundserror(wv, i)
+@propagate_inbounds function Base.getindex(wv::UnitWeights{S, T}, i::Integer) where {S, T}
+    @boundscheck checkbounds(wv,i)
+    one(T)
 end
 
-function Base.getindex(wv::UnitWeights{S, T}, i::AbstractRange{<:Int}) where {S, T}
-    if (minimum(i) > 0) & (maximum(i) <= length(wv))
-        return fill(one(T), length(i))
-    else
-        Base.throw_boundserror(wv, i)
-    end
-end
-
-function Base.getindex(wv::UnitWeights{S, T}, i::AbstractArray{<:Int}) where {S, T}
-    if (minimum(i) > 0) & (maximum(i) <= length(wv))
-        return fill(one(T), size(i))
-    else
-        Base.throw_boundserror(wv, i)
-    end
+@propagate_inbounds function Base.getindex(wv::UnitWeights{S, T}, i::AbstractArray{<:Int}) where {S, T}
+    @boundscheck checkbounds(wv,i)
+    fill(one(T), size(i))
 end
 
 Base.getindex(wv::UnitWeights{S, T}, ::Colon) where {S, T} = fill(one(T), length(wv))
@@ -313,8 +302,8 @@ uweights(s::S)            where S      = UnitWeights{S, S}(s)
 """
     varcorrection(w::UnitWeights, corrected=false)
 
-* `corrected=true`: ``\\frac{1}{\\sum w - \\sum {w^2} / \\sum w}``
-* `corrected=false`: ``\\frac{1}{\\sum w}``
+* `corrected=true`: ``\\frac{n}{n - 1}``, where ``n`` is the length of the weight vector
+* `corrected=false`: ``\\frac{1}{n}``, where ``n`` is the length of the weight vector
 """
 @inline function varcorrection(w::UnitWeights, corrected::Bool=false)
     corrected ? (1 / (w.sum - 1)) : (1 / w.sum)
@@ -354,7 +343,7 @@ Base.sum(v::AbstractArray, w::AbstractWeights) = dot(v, values(w))
 
 for v in (AbstractArray{<:Number}, BitArray, SparseArrays.SparseMatrixCSC, AbstractArray)
     @eval begin
-        function Base.sum(v::$v, w::UnitWeights) where {S,T}
+        function Base.sum(v::$v, w::UnitWeights)
             if length(v) != length(w)
                 throw(DimensionMismatch("array and weights have different lengths"))
             end
@@ -559,9 +548,9 @@ function wsum(A::AbstractArray{T}, w::AbstractVector{W}, dim::Int) where {T<:Num
     _wsum!(similar(A, wsumtype(T,W), Base.reduced_indices(axes(A), dim)), A, w, dim, true)
 end
 
-function wsum(A::AbstractArray{<:Number}, w::UnitWeights{S,T}, dim::Int) where {S,T}
+function wsum(A::AbstractArray{<:Number}, w::UnitWeights, dim::Int)
     (size(A, dim) != length(w)) && throw(DimensionMismatch("Inconsistent array dimension."))
-    return sum(A, dims = dim)
+    return sum(A, dims=dim)
 end
 
 # extended sum! and wsum
@@ -571,9 +560,9 @@ Base.sum!(R::AbstractArray, A::AbstractArray, w::AbstractWeights{<:Real}, dim::I
 
 Base.sum(A::AbstractArray{<:Number}, w::AbstractWeights{<:Real}, dim::Int) = wsum(A, values(w), dim)
 
-function Base.sum(A::AbstractArray{<:Number}, w::UnitWeights{S,T}, dim::Int) where {S,T}
+function Base.sum(A::AbstractArray{<:Number}, w::UnitWeights, dim::Int)
     (size(A, dim) != length(w)) && throw(DimensionMismatch("Inconsistent array dimension."))
-    return sum(A, dims = dim)
+    return sum(A, dims=dim)
 end
 
 ###### Weighted means #####
@@ -624,13 +613,13 @@ _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Int) where {T,W} =
     _mean!(similar(A, wmeantype(T, W), Base.reduced_indices(axes(A), dims)), A, w, dims)
 
 function StatsBase._mean(A::AbstractArray, w::UnitWeights{S,T}, dims::Nothing) where {S,T}
-    (length(A) != length(w)) && throw(DimensionMismatch("Inconsistent array dimension."))
+    length(A) != length(w) && throw(DimensionMismatch("Inconsistent array dimension."))
     return mean(A)
 end
 
-function StatsBase._mean(A::AbstractArray, w::UnitWeights{S,T}, dims::Int) where {S,T}
-    (length(A) != length(w)) && throw(DimensionMismatch("Inconsistent array dimension."))
-    return mean(A, dims = dims)
+function StatsBase._mean(A::AbstractArray, w::UnitWeights, dims::Int)
+    length(A) != length(w) && throw(DimensionMismatch("Inconsistent array dimension."))
+    return mean(A, dims=dims)
 end
 
 ###### Weighted quantile #####
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
 
 # NOTE: No variance correction is implemented for exponential weights
 
+struct UnitWeights{S<:Real, T<:Real} <: AbstractWeights{S, T, V where V<:Vector{T}}
+    sum::S
+end
+
+UnitWeights{T}(s::S) where {S, T} = UnitWeights{S, T}(s)
+
+@doc """
+    UnitWeights{T}(s)
+
+Construct a `UnitWeights` vector with length `s` and weight elements of type `T`.
+All weight elements are identically one.
+""" UnitWeights
+
+eltype(wv::UnitWeights{S, T})    where {S, T} = T
+values(wv::UnitWeights{S, T})    where {S, T} = fill(one(T), length(wv))
+isempty(wv::UnitWeights{S, T})   where {S, T} = iszero(wv.sum)
+length(wv::UnitWeights{S, T})    where {S, T} = Int(wv.sum)
+Base.size(wv::UnitWeights{S, T}) where {S, T} = Tuple(length(wv))
+
+function Base.getindex(wv::UnitWeights{S, T}, i::Int) where {S, T}
+    (i > 0) & (i <= length(wv)) ? one(T) : Base.throw_boundserror(wv, i)
+end
+
+function Base.getindex(wv::UnitWeights{S, T}, i::AbstractRange{<:Int}) where {S, T}
+    if (minimum(i) > 0) & (maximum(i) <= length(wv))
+        return fill(one(T), length(i))
+    else
+        Base.throw_boundserror(wv, i)
+    end
+end
+
+function Base.getindex(wv::UnitWeights{S, T}, i::AbstractArray{<:Int}) where {S, T}
+    if (minimum(i) > 0) & (maximum(i) <= length(wv))
+        return fill(one(T), size(i))
+    else
+        Base.throw_boundserror(wv, i)
+    end
+end
+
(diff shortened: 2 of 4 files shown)


src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+
+Construct a `UnitWeights` vector with length `s` and weight elements of type `T`.
+All weight elements are identically one.
+""" UnitWeights
+
+eltype(wv::UnitWeights{S, T})    where {S, T} = T
+values(wv::UnitWeights{S, T})    where {S, T} = fill(one(T), length(wv))
+isempty(wv::UnitWeights{S, T})   where {S, T} = iszero(wv.sum)
+length(wv::UnitWeights{S, T})    where {S, T} = Int(wv.sum)
+Base.size(wv::UnitWeights{S, T}) where {S, T} = Tuple(length(wv))
```
nalimilan: Import `size` for consistency.

src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+UnitWeights{T}(s::S) where {S, T} = UnitWeights{S, T}(s)
+
+@doc """
+    UnitWeights{T}(s)
+
+Construct a `UnitWeights` vector with length `s` and weight elements of type `T`.
+All weight elements are identically one.
+""" UnitWeights
+
+eltype(wv::UnitWeights{S, T})    where {S, T} = T
```
nalimilan: Actually I suspect this isn't needed for `AbstractVector`. I could be removed for other types.

src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+values(wv::UnitWeights{S, T})    where {S, T} = fill(one(T), length(wv))
+isempty(wv::UnitWeights{S, T})   where {S, T} = iszero(wv.sum)
```
nalimilan: No need for type parameters here now below.

src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+length(wv::UnitWeights{S, T})    where {S, T} = Int(wv.sum)
+Base.size(wv::UnitWeights{S, T}) where {S, T} = Tuple(length(wv))
+
+function Base.getindex(wv::UnitWeights{S, T}, i::Int) where {S, T}
```
nalimilan: Use  `Base.@propagate_inbounds` and put the bounds check inside `@boundscheck` (https://docs.julialang.org/en/latest/devdocs/boundscheck/). You can also write `i < 0 <= length(wv)`.

Same below.

src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
 
 # NOTE: No variance correction is implemented for exponential weights
 
+struct UnitWeights{S<:Real, T<:Real} <: AbstractWeights{S, T, V where V<:Vector{T}}
+    sum::S
```
nalimilan: Sum is actually the length here, so maybe better call it `len`? Also, `S` can probably be replaced with just `Int` given that it represents the length of a vector.

src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+values(wv::UnitWeights{S, T})    where {S, T} = fill(one(T), length(wv))
+isempty(wv::UnitWeights{S, T})   where {S, T} = iszero(wv.sum)
+length(wv::UnitWeights{S, T})    where {S, T} = Int(wv.sum)
+Base.size(wv::UnitWeights{S, T}) where {S, T} = Tuple(length(wv))
+
+function Base.getindex(wv::UnitWeights{S, T}, i::Int) where {S, T}
+    (i > 0) & (i <= length(wv)) ? one(T) : Base.throw_boundserror(wv, i)
+end
+
+function Base.getindex(wv::UnitWeights{S, T}, i::AbstractRange{<:Int}) where {S, T}
```
nalimilan: This definition is redundant with the previous one. Also, maybe we need to accept `<:Integer`?

src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+ 1.0
+```
+"""
+uweights(::Type{T}, s::S) where {S, T} = UnitWeights{S, T}(s)
+uweights(s::S)            where S      = UnitWeights{S, S}(s)
+
+"""
+    varcorrection(w::UnitWeights, corrected=false)
+
+* `corrected=true`: ``\\frac{1}{\\sum w - \\sum {w^2} / \\sum w}``
```
nalimilan: The formula doesn't reflect the code. Though I would expect all three formulas for frequency, probability and analytical weights to give the same result for this special case, right? If so, probably worth mentioning.

src/weights.jl
```diff
@@ -279,6 +352,17 @@ Base.sum(v::BitArray, w::AbstractWeights) = wsum(v, values(w))
 Base.sum(v::SparseArrays.SparseMatrixCSC, w::AbstractWeights) = wsum(v, values(w))
 Base.sum(v::AbstractArray, w::AbstractWeights) = dot(v, values(w))
 
+for v in (AbstractArray{<:Number}, BitArray, SparseArrays.SparseMatrixCSC, AbstractArray)
+    @eval begin
+        function Base.sum(v::$v, w::UnitWeights) where {S,T}
```
```suggestion
        function Base.sum(v::$v, w::UnitWeights)
```

src/weights.jl
```diff
@@ -475,13 +559,22 @@ function wsum(A::AbstractArray{T}, w::AbstractVector{W}, dim::Int) where {T<:Num
     _wsum!(similar(A, wsumtype(T,W), Base.reduced_indices(axes(A), dim)), A, w, dim, true)
 end
 
+function wsum(A::AbstractArray{<:Number}, w::UnitWeights{S,T}, dim::Int) where {S,T}
```
nalimilan: No need for parameters:
```suggestion
function wsum(A::AbstractArray{<:Number}, w::UnitWeights, dim::Int)
```

Same elsewhere.

src/weights.jl
```diff
@@ -475,13 +559,22 @@ function wsum(A::AbstractArray{T}, w::AbstractVector{W}, dim::Int) where {T<:Num
...
+    (size(A, dim) != length(w)) && throw(DimensionMismatch("Inconsistent array dimension."))
+    return sum(A, dims = dim)
```
nalimilan: No spaces around `=`. Same elsewhere.

src/weights.jl
```diff
@@ -475,13 +559,22 @@ function wsum(A::AbstractArray{T}, w::AbstractVector{W}, dim::Int) where {T<:Num
     _wsum!(similar(A, wsumtype(T,W), Base.reduced_indices(axes(A), dim)), A, w, dim, true)
 end
 
+function wsum(A::AbstractArray{<:Number}, w::UnitWeights{S,T}, dim::Int) where {S,T}
+    (size(A, dim) != length(w)) && throw(DimensionMismatch("Inconsistent array dimension."))
```
nalimilan: No need for parentheses:
```suggestion
    size(A, dim) != length(w) && throw(DimensionMismatch("Inconsistent array dimension."))
```
Same elsewhere.

src/weights.jl
```diff
@@ -530,6 +623,15 @@ _mean(A::AbstractArray, w::AbstractWeights, dims::Nothing) =
 _mean(A::AbstractArray{T}, w::AbstractWeights{W}, dims::Int) where {T,W} =
     _mean!(similar(A, wmeantype(T, W), Base.reduced_indices(axes(A), dims)), A, w, dims)
 
+function StatsBase._mean(A::AbstractArray, w::UnitWeights{S,T}, dims::Nothing) where {S,T}
+    (length(A) != length(w)) && throw(DimensionMismatch("Inconsistent array dimension."))
+    return mean(A)
```
nalimilan: I suspect you can drop this method an just pass `dims=dims` even when it's `nothing`.

src/weights.jl
```diff
@@ -250,6 +250,76 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+    if (minimum(i) > 0) & (maximum(i) <= length(wv))
+        return fill(one(T), size(i))
+    else
+        Base.throw_boundserror(wv, i)
+    end
+end
+
+Base.getindex(wv::UnitWeights{S, T}, ::Colon) where {S, T} = fill(one(T), length(wv))
+
+"""
```
nalimilan: I'm not sure we should add this method since it's very different from others: it doesn't accept a vector as input. The constructor may be enough.

src/weights.jl
lbittarello: So did I, but it doesn't actually work.

src/weights.jl
lbittarello: I lean towards keeping it. One can reasonably expect colon indexes to work on arrays. It works for other weight types too (because they have a `values` subfield).

src/weights.jl
lbittarello: You know, I got errors if I didn't use parentheses in earlier versions of Julia, so I've just used them ever since. They don't seem necessary any longer though.

src/weights.jl
lbittarello: I replaced `sum` by `len` and added a method to the function `sum`.

src/weights.jl
lbittarello: Sorry, I didn't know that ranges were arrays till @Nosferican told me so.

src/weights.jl
lbittarello: Sorry, mindless copying and pasting

src/weights.jl
lbittarello: `size` was not imported, though we'd already extended it in line 27. I will explicitly import it in the next pull request, when we simplify the weight types.

src/weights.jl
lbittarello: We use it in `quantile`, but it's easy to replace it. I will drop it in the next pull request.

src/weights.jl
nalimilan: Better do this now, commits should make sense in isolation.

src/weights.jl
nalimilan: What I mean is that there's a fallback in Base which gives the same result.

src/weights.jl
```diff
@@ -250,6 +250,65 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+
+eltype(wv::UnitWeights{S, T}) where {S, T} = T
+values(wv::UnitWeights{S, T}) where {S, T} = fill(one(T), length(wv))
+sum(wv::UnitWeights{S, T}) where {S, T} = convert(T, length(wv))
+isempty(wv::UnitWeights) = iszero(wv.len)
+length(wv::UnitWeights) = wv.len
+Base.size(wv::UnitWeights) = Tuple(length(wv))
+
+@propagate_inbounds function Base.getindex(wv::UnitWeights{S, T}, i::Integer) where {S, T}
+    @boundscheck checkbounds(wv,i)
```
```suggestion
    @boundscheck checkbounds(wv, i)
```

src/weights.jl
```diff
@@ -250,6 +250,65 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+    one(T)
+end
+
+@propagate_inbounds function Base.getindex(wv::UnitWeights{S, T}, i::AbstractArray{<:Int}) where {S, T}
+    @boundscheck checkbounds(wv,i)
```
```suggestion
    @boundscheck checkbounds(wv, i)
```

src/weights.jl
```diff
@@ -250,6 +250,65 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+isempty(wv::UnitWeights) = iszero(wv.len)
+length(wv::UnitWeights) = wv.len
+Base.size(wv::UnitWeights) = Tuple(length(wv))
+
+@propagate_inbounds function Base.getindex(wv::UnitWeights{S, T}, i::Integer) where {S, T}
+    @boundscheck checkbounds(wv,i)
+    one(T)
+end
+
+@propagate_inbounds function Base.getindex(wv::UnitWeights{S, T}, i::AbstractArray{<:Int}) where {S, T}
```
nalimilan: Actually I realize this definition isn't the most natural: we'd better return a new weights vector of the same type. But that's a separate issue since other types do the same thing.

src/weights.jl
nalimilan: Are we talking about the same thing? I don't see how colon indices are related. :-)

src/weights.jl
nalimilan: They're still there. ;-)

src/weights.jl
nalimilan: Too bad. Though it would be shorter to use a single function, with a condition `dims === nothing ? mean(A) : mean(A, dims=dims)`.

docs/src/weights.md
```diff
@@ -115,6 +115,15 @@ julia> eweights(t, r, 0.3)
 NOTE: This is equivalent to `eweights(something.(indexin(t, r)), 0.3)`, which is saying that for each value in `t` return the corresponding index for that value in `r`.
 Since `indexin` returns `nothing` if there is no corresponding value from `t` in `r` we use `something` to eliminate that possibility.
 
+### `UnitWeights`
+
+Unit weights treat all observations equally.
```
nalimilan: I suggest this:
```suggestion
Unit weights are a special case in which all observations are given a weight equal to `1`. Using such weights is equivalent to computing unweighted statistics.

This type can notably be used when implementing an algorithm so that a only a weighted variant has to be written. The unweighted variant is then obtained by passing a `UnitWeights` object. This is very efficient since no weights vector is actually allocated.
```

Also, I'd put this before `Weights`, which is the fallback type with less methods defined (no `varcorrection`).

src/weights.jl
```diff
@@ -250,6 +250,65 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+"""
+uweights(::Type{T}, s::S) where {S, T} = UnitWeights{S, T}(s)
+uweights(s::S)            where S      = UnitWeights{S, S}(s)
+
+"""
+    varcorrection(w::UnitWeights, corrected=false)
+
+* `corrected=true`: ``\\frac{n}{n - 1}``, where ``n`` is the length of the weight vector
+* `corrected=false`: ``\\frac{1}{n}``, where ``n`` is the length of the weight vector
+"""
```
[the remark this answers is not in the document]
nalimilan: Or maybe just "This definition is equivalent to the correction applied to unweighted data.".

src/weights.jl
lbittarello: Zut. I tried to be more thorough this time.

src/weights.jl
lbittarello: Sorry: your comment came just after the definition of `getindex(wv::UnitWeights{S, T}, ::Colon)`, so I assumed that was what it referred to. I don't know which method you have in mind then.

src/weights.jl
lbittarello: It isn't much shorter, because the methods have different boundary checks (which needed fixing, incidentally), so I left them as they are, but I am open to changing it.

nalimilan: Ah, sorry.

nalimilan:
Can you also add tests?

lbittarello:
Tests added!

test/weights.jl
```diff
@@ -447,6 +460,27 @@ end
...
+    @test sum(a, wt, 3) ≈ sum(a, dims=3)
+
+    @test mean(a, wt, dims=1) ≈ mean(a, dims=1)
+    @test mean(a, wt, dims=2) ≈ mean(a, dims=2)
+    @test mean(a, wt, dims=3) ≈ mean(a, dims=3)
+
+    @test_throws DimensionMismatch sum(a, wt, 4)
+    @test_throws DimensionMismatch mean(a, wt, dims=4)
+
+    @test quantile([1.0, 4.0, 6.0, 8.0, 10.0], uweights(5), [0.5]) ≈ [6.0]
```
nalimilan:
```suggestion
    @test quantile([1.0, 4.0, 6.0, 8.0, 10.0], uweights(5), 0.5) ≈ 6.0
```

src/weights.jl
nalimilan: OK, let's keep it, but better use a default value for `T` here, and require people to use the constructor to choose a different type: I don't think it should be too common.

The default `T` could be either `Int` or `Bool`. The latter is a bit weird (it gives `true`), but it is sometimes used for a "strong zero" and it always preserves the type with multiplication: `Int8(1) * true === Int8(1)`, which isn't the case with `1`. Not sure whether that's really useful though, since weighted algorithms likely need to widen `Int8` at some point, and for floats `Int` and `Bool` behave the same:`1.0f0 * 1 === 1.0f0`. So maybe the simpler `Int` is better.

src/weights.jl
```diff
@@ -250,6 +249,66 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
 
 # NOTE: No variance correction is implemented for exponential weights
 
+struct UnitWeights{S<:Integer, T<:Real} <: AbstractWeights{S, T, V where V<:Vector{T}}
+    len::S
```
nalimilan: As I noted, better hardcode this as `Int` for simplicity.

nalimilan:
Thanks. A few methods don't seem to be covered by tests: see dark red in the margin at https://codecov.io/gh/JuliaStats/StatsBase.jl/compare/416af5561d3efd11c78c82c269a9d7c7d26c681d...009b9d5b893bd9235dfd96895acfe1a3c1df81ea/diff.

src/weights.jl
Nosferican:
```
x = uweights(6)
x[:]
type UnitWeights has no fields values
```
vs
```
x[:] == fill(one(1, 6))
```
I think that is good.

lbittarello: The default constructor (`uweights(s)`) was already setting the element type to an integer type, since it used the type of `s`, but I made it more explicit in the latest commit.

lbittarello:
I added more tests.

As it turns out, we need `eltype` for unit weights or `show` throws an error (due to the absence of the `values` field), so I added it back.

test/weights.jl
lbittarello: I used an array here on purpose (because the next test doesn't), but I'm open to changing it if you prefer. We could actually only keep the test of `median`, since it calls `quantile(x, w, [0.5]) `.

nalimilan: Probably better test both forms for both methods then. You never know where bugs can happen.

src/weights.jl
```diff
@@ -279,23 +280,28 @@ end
...
 
 # Examples
 ```julia-repl
-julia> uweights(Float64, 3)
+julia> uweights(3)
 3-element UnitWeights{Int64,Float64}:
  1.0
  1.0
  1.0
+julia> uweights(Float64, 3)
```
nalimilan: Add empty line.

src/weights.jl
nalimilan: Still need to address this comment.

src/weights.jl
```diff
@@ -250,6 +249,72 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+    uweights(::Type{T}, s::Integer) where T<:Real
+    uweights(s::Integer)
+
+Construct a `UnitWeights` vector with length `s` and weight elements of type `T`.
+All weight elements are identically one.
+
+# Examples
+```julia-repl
+julia> uweights(3)
+3-element UnitWeights{Int64,Float64}:
```
nalimilan: Types don't seem consistent with what the code does. I guess this explains the CI failure.

src/weights.jl
```diff
@@ -250,6 +249,72 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+ 1.0
+ 1.0
+julia> uweights(Float64, 3)
+3-element UnitWeights{Int64,Float64}:
+ 1
+ 1
+ 1
+```
+"""
+uweights(::Type{T}, s::S) where {S, T} = UnitWeights{S, T}(s)
```
```suggestion
uweights(::Type{T}, s::S) where {S, T<:Real} = UnitWeights{S, T}(s)
```

src/weights.jl
```diff
@@ -250,6 +249,72 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+uweights(s::S)            where S      = UnitWeights{Int, S}(s)
```
nalimilan: It doesn't really make sense to use the type of the length to choose the type of the elements. Better just use `Int`.

nalimilan:
> As it turns out, we need `eltype` for unit weights or `show` throws an error (due to the absence of the `values` field), so I added it back.

What if you also remove the `eltype` methods for `AbstractWeights`?

lbittarello:
You're right. I'd messed up my tests. I've amended the last commit to eliminate the reintroduction of `eltype` and include additional revisions.

src/weights.jl
lbittarello: Sorry, the parameters were in the wrong order

src/weights.jl
lbittarello: Does it look better now?

test/weights.jl
```diff
@@ -96,6 +96,23 @@ end
...
+    wv = uweights(Float64, 3)
+    @test wv[1] === 1.
+    @test wv[1:3] == fill(1.0, 3)
+    @test wv[:] == fill(1.0, 3)
+    @test !isempty(wv)
+    @test length(wv) === 3
+    @test size(wv) === (3,)
+    @test sum(wv) === 3.
+    @test values(wv) == fill(1.0, 3)
+    @test varcorrection(wv) == 1/3
```
nalimilan:
```suggestion
    @test Base.varcorrection(wv) == 1/3
```
But it would also be nice to test `var`, just in case.

src/weights.jl
```diff
@@ -250,6 +249,72 @@ eweights(t::AbstractVector, r::AbstractRange, λ::Real) =
...
+
+@propagate_inbounds function Base.getindex(wv::UnitWeights{T}, i::AbstractArray{<:Int}) where T
+    @boundscheck checkbounds(wv, i)
+    fill(one(T), size(i))
+end
+
+Base.getindex(wv::UnitWeights{T}, ::Colon) where T = fill(one(T), length(wv))
+
+"""
+    uweights(::Type{T}, s::Integer) where T<:Real
```
nalimilan: Maybe put the one-arg method first?

test/weights.jl
lbittarello: It seems that this change broke the tests... I don't know if `varcorrection` is defined in base at all.

nalimilan: Sorry, I mean StatsBase.

src/weights.jl
```diff
@@ -300,8 +300,8 @@ julia> uweights(3)
  1
 ```
 """
-uweights(::Type{T}, s::Int) where {T<:Real} = UnitWeights{T}(s)
 uweights(s::Int)                            = UnitWeights{Int}(s)
+uweights(::Type{T}, s::Int) where {T<:Real} = UnitWeights{T}(s)
```
nalimilan: Also change order in the docstring?

src/weights.jl
```diff
@@ -279,6 +347,17 @@ Base.sum(v::BitArray, w::AbstractWeights) = wsum(v, values(w))
 Base.sum(v::SparseArrays.SparseMatrixCSC, w::AbstractWeights) = wsum(v, values(w))
 Base.sum(v::AbstractArray, w::AbstractWeights) = dot(v, values(w))
 
+for v in (AbstractArray{<:Number}, BitArray, SparseArrays.SparseMatrixCSC, AbstractArray)
+    @eval begin
+        function Base.sum(v::$v, w::UnitWeights)
+            if length(v) != length(w)
+                throw(DimensionMismatch("Inconsistent array dimension."))
```
oxinabox: Why does `UnitWeight` have length?
And if it has to have length then why do we need to actually check it?
Is this a useful feature vs it just always being considered the right size?
Like `LinearAlgebra.I`

lbittarello: That's an interesting point. Here's an example. In computing the variance, we obtain the normalization factor through `varcorrection`. It depends on the number of observations, so we need the length of the weight vector. We should also ensure that the user has not passed the wrong length lest we use the wrong normalization factor. There are ways around it (for instance, a method for `variance` that bypasses the weights), but we want to help developers avoid defining separate methods. More generally, I think that boundary checks at different points help users catch mistakes before they snowball.

nalimilan: Yes, it doesn't hurt to check lengths, and you can avoid it using `@inbounds`. I'd say that in general it's quite easy to get the appropriate length since it's the length of another input argument.

oxinabox: Calculating the length isn't free for general iterators.

lbittarello:
Happy to! It'll take me a couple of weeks though, as I'll be out of town.

src/weights.jl
nalimilan: Yeah, but in most cases we take arrays, right? Also you can always create an object of length `typemax` if you don't care about the checks.

oxinabox: Mostly isn't always. Isn't `dropmissing` after all is one of the cases of iterators that are not free to take then length of?

These checks just don't feel great to me.
What error are we catching?
And how does that error relate to weighting?

If one is checking one got the amount of data that was expected, then isn't that better done with an explicit check on the data gathering code? So a good error can be thrown?
I feel like this just adds work for the user of weights, and the best case is they get a confusing error message.

I'm all for defensive programming, but that means checking the function you are defining can work. Not just checking arbitrary things in the calling code.


> Also you can always create an object of length typemax if you don't care about the checks.

That doesn't work, does it? This check is checking the lengths are equal, not checking a `>`.

nalimilan: We don't support weights with `dropmissing` currently, the main reason being that in general we would need to skip weights corresponding to missing entries (instead of naively taking weights in their order of appearance).

> That doesn't work, does it? This check is checking the lengths are equal, not checking a `>`.

Right, it doesn't work here (I was thinking about custom functions which want to take a `UnitWeights` as default weights for simplicity of implementation). I guess we could allow something like `UnitWeights(nothing)` (or maybe `UnitWeights`) to create an object with a undefined length. But I still think it's nice to take a length by default, for checking, but also because that's required to be an actual `AbstractVector` (otherwise we need to error when calling `size`).

nalimilan:
Gentle bump about `value`: it would be nice to have that before we make a release to avoid changing (slightly) the API.

matthieugomez:
I know I am a bit late but why not import `FillArrays` and  define 
`UnitWeights{T} = AbstractWeights{Int, T, Ones{T}}`.

This would simplify the code. It would also make it more efficient. For instance,  `sqrt.(uweights(10)` creates a full vector in this pull request, whereas `sqrt.(Ones{Float64}(10))` creates a `FillArray`.

nalimilan:
The issue with that approach is that `UnitWeights` wouldn't be a concrete type, and objects created by `uweights` wouldn't be `<:AbstractWeights`.

PR #528, opened 2019-10-09
add responsename

matthieugomez:
Add `responsename`. Will be used by packages that print results of `RegressionModel`, such as `RegressionTables`.


src/StatsBase.jl
```diff
@@ -188,10 +188,11 @@ export
     islinear,
     leverage,
     loglikelihood,
-    meanresponse,
     modelmatrix,
     mss,
     response,
+    responsename,
+    meanresponse,
```
nalimilan: AFAICT names in this list are sorted lexicographically, so better keep the current order.

src/statmodels.jl
```diff
@@ -294,6 +294,12 @@ Return the model response (a.k.a. the dependent variable).
 """
 response(obj::RegressionModel) = error("response is not defined for $(typeof(obj)).")
 
+"""
+    responsename(obj::StatisticalModel)
+Return the name of the model response (a.k.a. the dependent variable).
```
```suggestion

Return the name of the model response (a.k.a. the dependent variable).
```

kleinschmidt:
I certainly would like to move towards requiring models to do their own "formula bookkeeping" and add getter/setter methods to the StatsBase/StatsModels API.  I don't think that's incompatible with this proposal here, since we could just add a fallback `responsename(mod) = responsename(formula(mod))` or whatever the getter function ends up being called.

nalimilan:
Yeah, that would be OK.

matthieugomez:
updated.

PR #529, opened 2019-10-09
add crossmodelmatrix

matthieugomez:
This is a  useful API for packages that compute robust standard errors. Different models may want to implement more efficient methods (accessors) since X'X is typically computed during the estimation.


docs/src/statmodels.md
```diff
@@ -39,6 +39,7 @@ fitted
 leverage
 meanresponse
 modelmatrix
+crossmodelmatrix
```
nalimilan: Please preserve lexicographic order.

src/statmodels.jl
```diff
@@ -308,6 +308,14 @@ Return the model matrix (a.k.a. the design matrix).
 """
 modelmatrix(obj::RegressionModel) = error("modelmatrix is not defined for $(typeof(obj)).")
 
+
```
```suggestion
```

src/statmodels.jl
```diff
@@ -308,6 +308,14 @@ Return the model matrix (a.k.a. the design matrix).
...
+"""
+    crossmodelmatrix(obj::RegressionModel)
+
+Return X'X where X is the model matrix
```
```suggestion
Return `X'X` where `X` is the model matrix of `obj`.
This function will return a pre-computed matrix stored in `obj` if possible.
```

matthieugomez:
updated.

matthieugomez:
Thanks!

PR #530, opened 2019-10-09
add covariance_estimator

matthieugomez:
Return the `CovarianceEstimator` used to compute the variance-covariance matrix

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -135,7 +135,7 @@ vcov(obj::StatisticalModel) = error("vcov is not defined for $(typeof(obj)).")
 """
     covariance_estimator(obj::StatisticalModel)
 
-Return the variance-covariance estimator used to compute the variance-variance matrix
+Return the `CovarianceEstimator` used to compute the variance-covariance matrix.
 """
 covariance_estimator(obj::StatisticalModel) = error("covariance_estimator is not defined for $(typeof(obj)).")
 
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -133,6 +133,13 @@ Return the variance-covariance matrix for the coefficients of the model.
 vcov(obj::StatisticalModel) = error("vcov is not defined for $(typeof(obj)).")
 
 """
+    covariance_estimator(obj::StatisticalModel)
+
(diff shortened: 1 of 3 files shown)


nalimilan:
Maybe we should just add a `CovarianceEstimator` method? Not sure. Cc: @kleinschmidt

kleinschmidt:
Does statsbase own CovarianceEstimator type?

nalimilan:
Yes: https://github.com/JuliaStats/StatsBase.jl/blob/24a80caf3b2a98adaed24b72e571c482928e15de/src/cov.jl#L158

kleinschmidt:
A few people on slack suggested that the constructor pun would be more appropriate when some kind of computation is involved, which (as far as I can tell) isn't the intended use case here, right?  This would just retrieve an existing `<:CovarianceEstimator`.  So I'm inclined to think that `covariance_estimator` seems more natural.

matthieugomez:
I'm personally indifferent, so let me know.

src/statmodels.jl
```diff
@@ -132,6 +132,13 @@ Return the variance-covariance matrix for the coefficients of the model.
 """
 vcov(obj::StatisticalModel) = error("vcov is not defined for $(typeof(obj)).")
 
+"""
+    covariance_estimator(obj::StatisticalModel)
+
+Return the variance-covariance estimator used to compute the variance-variance matrix
```
```suggestion
Return the `CovarianceEstimator` used to compute the variance-covariance matrix.
```

nalimilan:
Hard to decide... Another question if we decide to add a function: maybe it should be `vcov_estimator` since it goes with `vcov`?

matthieugomez:
Or use covariance instead of vcov ;)

nalimilan:
Yeah, well, that would be breaking, so... An even more appealing solution would be to overload `cov` from Base, since that's really the same thing. Then you'd call the function `covestimator`, which is a bit shorter.

matthieugomez:
I don't think it's that important actually. Maybe we should close until we have a better story for `covariance` `vcov` etc?

nalimilan:
You mean you don't really need that function? Then yes, maybe better close until we have a use case, that's always better to design good APIs.

matthieugomez:
Exactly

Issue #413, opened 2018-09-14
ECDF evaluated on NaN is 1.0


joshday:
I was looking at some strange results in my code when I discovered this:

```julia
julia> ecdf(randn(100))(NaN)
1.0
```

ararslan:
https://github.com/JuliaStats/StatsBase.jl/blob/master/src/empirical.jl#L18

```julia
julia> searchsortedlast(randn(10), NaN) / 10
1.0
```

joshday:
Some more ecdf fun:

```julia
julia> ecdf([1,2,NaN])(Inf)
0.6666666666666666
```

tpoisot:
What would be the general consensus with the following behavior:

```
ecdf(x) => ecdf(filter(!isnan, x))
ecdf(x)(NaN) => NaN
```

or something to that effect? I think it would be less surprising...

I would not be opposed to `ecdf(x)(Inf) => 1`, and `ecdf(x)(-Inf) => 0`, but I feel less strongly about these.

nalimilan:
The general convention in Julia is that `NaN` are never ignored, but propagate or throw errors. So `ecdf(x)` should probably just throw an error, or create a function which always returns `NaN`. `ecdf(x)(NaN) ` should return `NaN` too.

`ecdf(x)(Inf) == 1` and `ecdf(x)(-Inf) == 0` kind of make sense to me.

joshday:
+1 from me for:

- `ecdf(x)` throwing an error if `any(isnan, x)`
- `ecdf(x)(Inf)` == 1
- `ecdf(x)(-Inf)` == 0
- `isnan(ecdf(x)(NaN))`

I'll try to make a PR today.

tpoisot:
@joshday I think that makes sense -- users can always filter whatever they give to `ecdf` to remove `NaN` before.

joshday:
Different issue here but along the same lines.  How would folks expect `Inf` to work with `ecdf`?  Current behavior (matches R) is:

```julia
julia> ecdf([1,2,Inf])(Inf). # Works because Inf == Inf
1.0

julia> ecdf([1,2,-Inf])(-Inf)
0.3333333333333333
```

A case could be made to also disallow `Inf`/`-Inf` in creating ECDFs.

PR #537, opened 2019-11-25
better NaN handling with ecdf

joshday:
Ref: https://github.com/JuliaStats/StatsBase.jl/issues/413

New behaviors:

```julia
julia> ecdf([1,2,3])(NaN)
NaN

julia> ecdf([1,2,NaN])
ERROR: ArgumentError: ecdf can not include NaN values
Stacktrace:
 [1] #ecdf#113(::Weights{Float64,Float64,Array{Float64,1}}, ::typeof(ecdf), ::Array{Float64,1}) at /Users/joshday/.julia/dev/StatsBase/src/empirical.jl:57
 [2] ecdf(::Array{Float64,1}) at /Users/joshday/.julia/dev/StatsBase/src/empirical.jl:57
 [3] top-level scope at REPL[11]:1
```

Side note:  I didn't actually touch the Project.toml file...did Pkg make that change for me?

### DIFF
--- a/src/empirical.jl
+++ b/src/empirical.jl
@@ -1,6 +1,5 @@
 # Empirical estimation of CDF and PDF
 
-
 ## Empirical CDF
 
 struct ECDF{T <: AbstractVector{<:Real}, W <: AbstractWeights{<:Real}}
@@ -9,6 +8,7 @@ struct ECDF{T <: AbstractVector{<:Real}, W <: AbstractWeights{<:Real}}
 end
 
 function (ecdf::ECDF)(x::Real)
+    isnan(x) && return NaN
     n = searchsortedlast(ecdf.sorted_values, x)
     evenweights = isempty(ecdf.weights)
     weightsum = evenweights ? length(ecdf.sorted_values) : sum(ecdf.weights)
@@ -54,6 +54,7 @@ evaluate CDF values on other samples.
 function is inside the interval ``(0,1)``; the function is defined for the whole real line.
 """
 function ecdf(X::RealVector; weights::AbstractVector{<:Real}=Weights(Float64[]))
+    any(isnan, X) && throw(ArgumentError("ecdf can not include NaN values"))
     isempty(weights) || length(X) == length(weights) || throw(ArgumentError("data and weight vectors must be the same size," *
         "got $(length(X)) and $(length(weights))"))
     ord = sortperm(X)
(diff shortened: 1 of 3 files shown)


src/empirical.jl
```diff
@@ -9,6 +8,7 @@ struct ECDF{T <: AbstractVector{<:Real}, W <: AbstractWeights{<:Real}}
 end
 
 function (ecdf::ECDF)(x::Real)
+    isnan(x) && return NaN
```
andreasnoack: This should use the element type of `ECDF` to to ensure type stability of e.g. `Float32`s and `Dual` so probably something like `T(NaN)` for `ECDF{T}`.

joshday: Hmm, there are already other type instabilities present:

```julia
weightsum = evenweights ? length(ecdf.sorted_values) : sum(ecdf.weights)
```

I'm fine fixing these here or in a different PR

joshday: Never mind, fixes were easy.  I'll push in a sec

nalimilan: Can you also add tests, including `@inferred`?

joshday: Never mind again, way harder to handle NaN in a type stable way than I thought!  I'm gonna punt on this and ask this to be merged as-is.

nalimilan: The `weightsum` type instability is not visible outside of the function for `Float32` since `partialsum/weightsum` will return a `Float64`. So `NaN` is more correct than `oftype(T, NaN)`. Not sure about `Dual`, would need to try that.

But please add new tests.

nalimilan:
Sorry for saying this only now, but it would probably be better to only print a deprecation for now instead of throwing an error. That will possibly leave some time for users to adapt, and it won't require us to tag a new breaking release just for this. Breaking releases are painful as they require all dependent packages to update their upper bound, or users will keep using old StatsBase versions.

joshday:
Sure, every bug fix is technically breaking, but that's being awfully strict?  This is simply fixing incorrect results:

```julia
julia> ecdf([1,NaN])(1)
0.5
```

nalimilan:
Ah, right, I thought the current behavior just skipped `NaN`, but if it returns invalid results then turning that into an error is a strict improvement.

andreasnoack:
The test failure on 32 bit seems to be a time out because of deprecations

PR #543, opened 2019-12-03
Stop using the deprecated values function

andreasnoack:
There are currently many deprecation warnings because of the remaining use of `values` and it slows down CI quite a bit (and is just annoying).

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -48,7 +48,7 @@ function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange, wv:
     m0 = levels[1]
     m1 = levels[end]
     b = m0 - 1
-    w = values(wv)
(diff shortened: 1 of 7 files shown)


nalimilan:
Wait, no, most of these calls to `values` should actually just be removed. I guess they were added at a time when `AbstractWeights` where not `AbstractVectors` and therefore not indexable. For most weights types `convert` will be free, but for `UnitWeights` it will force the allocation of a new vector, while just using the weights vector will allow the compiler to optimize the loop.

andreasnoack:
I merged a bit fast but the deprecations were making CI extremely slow so I thought we better get rid of them and I believe it should be neutral wrt behavior.

nalimilan:
Yes in terms of behavior it should make no difference.

PR #546, opened 2019-12-19
trim and winsor now maintain original order

### LINKED ISSUE #437: winsor implicitly sorts data
Eg
```julia
julia> winsor(collect(5:-1:1), prop = 0.2)
5-element Array{Int64,1}:
 2
 2
 3
 4
 4
```
but this is not reflected in the documentation, which is
>     winsor(x; prop=0.0, count=0)
>
> Return a copy of `x` with either `count` or proportion `prop` of the lowest
elements of `x` replaced with the next-lowest, and an equal number of the
highest elements replaced with the previous-highest.  To compute the Winsorized
mean of `x` use `mean(winsor(x))`.

and is not apparent from the example in the docstring either (because it is already sorted).

Suggestion: perhaps `winsor` should **not** sort, as that would make it easier to eg winsorize a `DataFrame` column.

AsafManela:
trim and winsor now maintain original order. closes #437

### DIFF
--- a/src/robust.jl
+++ b/src/robust.jl
@@ -32,8 +32,10 @@ end
 """
     trim(x; prop=0.0, count=0)
 
-Return a generator of all elements of `x` that omits either `count` or proportion
-`prop` of the highest and lowest elements.  To compute the trimmed mean of `x` use
+Return an iterator of all elements of `x` that omits either `count` or proportion
+`prop` of the highest and lowest elements.  
+The number of trimmed elements could be smaller than specified if several 
+elements equal the lower or upper bound. To compute the trimmed mean of `x` use
 `mean(trim(x))`; to compute the variance use `trimvar(x)` (see [`trimvar`](@ref)).
 
 # Example
@@ -66,9 +68,11 @@ end
 """
     winsor(x; prop=0.0, count=0)
 
-Return a generator of all elements of `x` that replaces either `count` or
+Return an iterator of all elements of `x` that replaces either `count` or
 proportion `prop` of the highest elements with the previous-highest element  
 and an equal number of the lowest elements with the next-lowest element. 
+The number of replaced elements could be smaller than specified if several 
+elements equal the lower or upper bound.
 To compute the Winsorized mean of `x` use `mean(winsor(x))`.
 
 # Example
--- a/src/robust.jl
+++ b/src/robust.jl
@@ -7,6 +7,28 @@
 #############################
 
 # Trimmed set
+"Return the upper and lower bound elements used by `trim` and `winsor`"
+function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
+    n = length(x)
+    n > 0 || throw(ArgumentError("x can not be empty."))
+
+    if count == 0
+        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
+        count = floor(Int, n * prop)
+    else
+        prop == 0 || throw(ArgumentError("prop and count can not both be > 0."))
+        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
+    end
+
+    # indices for lowest count values
+    x2 = copy(x)
+    lo = partialsort!(x2, 1:count+1)[end]
+    # indices for largest count values
+    up = partialsort!(x2, n-count:n)[1]
+
+    up, lo
+end
+
 """
     trim(x; prop=0.0, count=0)
 
@@ -24,7 +46,9 @@ julia> trim([5,2,4,3,1], prop=0.2)
 ```
 """
 function trim(x::AbstractVector; prop::Real=0.0, count::Integer=0)
-    trim!(copy(x); prop=prop, count=count)
+    up, lo = uplo(x; prop=prop, count=count)
+
+    (xi for xi in x if lo <= xi <= up)
 end
 
 """
@@ -33,30 +57,9 @@ end
 A variant of [`trim`](@ref) that modifies `x` in place.
 """
 function trim!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
-    n = length(x)
-    n > 0 || throw(ArgumentError("x can not be empty."))
-
-    if count == 0
-        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
-        count = floor(Int, n * prop)
-    else
-        prop == 0 || throw(ArgumentError("prop and count can not both be > 0."))
-        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
-    end
-
-    # allocate vector of all x's indices
-    ixall = collect(1:n)
-    # indices for lowest count values
-    ixstart = partialsortperm!(ixall, x, 1:count; initialized=true)
-    # indices for largest count values
-    ixend = partialsortperm!(ixall, x, (n-count+1):n; initialized=true)
-    # indices for lowest and largest values
-    ixtrim = vcat(ixstart, ixend)
-    sort!(ixtrim)
-    unique!(ixtrim)
-    # trim them from x
-    deleteat!(x, ixtrim)
-
+    up, lo = uplo(x; prop=prop, count=count)
+    ix = (i for (i,xi) in enumerate(x) if lo > xi || xi > up)
+    deleteat!(x, ix)
     return x
 end
 
@@ -80,7 +83,9 @@ julia> winsor([5,2,3,4,1], prop=0.2)
 ```
 """
 function winsor(x::AbstractVector; prop::Real=0.0, count::Integer=0)
-    winsor!(copy(x); prop=prop, count=count)
+    up, lo = uplo(x; prop=prop, count=count)
+
+    (clamp(xi, lo, up) for xi in x)
 end
 
 """
@@ -89,30 +94,7 @@ end
 A variant of [`winsor`](@ref) that modifies vector `x` in place.
 """
 function winsor!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
-    n = length(x)
-    n > 0 || throw(ArgumentError("x can not be empty."))
-
-    if count == 0
-        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
-        count = floor(Int, n * prop)
-    else
-        prop == 0 || throw(ArgumentError("prop and count can not both be > 0."))
-        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
-    end
-
-    # allocate vector of all x's indices
-    ixall = collect(1:n)
-    # indices for lowest count values
-    ixstart = partialsortperm!(ixall, x, 1:(count+1); initialized=true)
-    # indices for largest count values
-    ixend = partialsortperm!(ixall, x, (n-count):n; initialized=true)
-
-    # set lowest to count+1's value
-    @inbounds x[ixstart[1:count]] .= x[ixstart[count+1]]
-
-    # set largest to n-count's value
-    @inbounds x[ixend[2:end]] .= x[ixend[1]]
-
+    copyto!(x, winsor(x; prop=prop, count=count))
     return x
 end
 
--- a/src/robust.jl
+++ b/src/robust.jl
@@ -32,13 +32,13 @@ end
 """
     trim(x; prop=0.0, count=0)
 
-Return a copy of `x` with either `count` or proportion `prop` of the highest
-and lowest elements removed.  To compute the trimmed mean of `x` use
+Return a generator of all elements of `x` that omits either `count` or proportion
+`prop` of the highest and lowest elements.  To compute the trimmed mean of `x` use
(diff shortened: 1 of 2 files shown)


src/robust.jl
```diff
@@ -44,10 +44,10 @@ function trim!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
         0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
     end
 
-    partialsort!(x, (n-count+1):n)
-    partialsort!(x, 1:count)
-    deleteat!(x, (n-count+1):n)
-    deleteat!(x, 1:count)
+    ix = vcat(partialsortperm(x, 1:count), partialsortperm(x, (n-count+1):n))
+    sort!(ix)
+    ix = unique(ix) # can be replaced by unique! starting julia 1.1
```
nalimilan: Since you know indices are sorted, a simple loop comparing values to previous ones and skipping them will probably be faster.

src/robust.jl
```diff
@@ -44,10 +44,10 @@ function trim!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
         0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
     end
 
-    partialsort!(x, (n-count+1):n)
-    partialsort!(x, 1:count)
-    deleteat!(x, (n-count+1):n)
-    deleteat!(x, 1:count)
+    ix = vcat(partialsortperm(x, 1:count), partialsortperm(x, (n-count+1):n))
```
nalimilan: Since `partialsortperm(x, 1:count)` is going to make a copy of `x`, better do it manually, and call `partialsortperm!` on the result twice. That way you avoid an allocation, and you take advantage of the partial sorting of the first call to avoid doing some work twice.

src/robust.jl
```diff
@@ -92,10 +92,11 @@ function winsor!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
         0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
     end
 
-    partialsort!(x, (n-count+1):n)
-    partialsort!(x, 1:count)
-    x[1:count] .= x[count+1]
-    x[n-count+1:end] .= x[n-count]
+    ix = partialsortperm(x, 1:(count+1))
+    x[ix[1:count]] .= x[ix[count+1]]
```
nalimilan:
```suggestion
    x[view(ix, 1:count)] .= x[ix[count+1]]
```

src/robust.jl
```diff
@@ -92,10 +92,11 @@ function winsor!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
...
+
+    ix = partialsortperm(x, (n-count):n)
+    x[ix[2:end]] .= x[ix[1]]
```
nalimilan:
```suggestion
    x[view(ix, 2:end)] .= x[ix[1]]
```

src/robust.jl
```diff
@@ -92,10 +92,11 @@ function winsor!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
         0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
     end
 
-    partialsort!(x, (n-count+1):n)
-    partialsort!(x, 1:count)
-    x[1:count] .= x[count+1]
-    x[n-count+1:end] .= x[n-count]
+    ix = partialsortperm(x, 1:(count+1))
```
nalimilan: Same remark here about using `partialsortperm!`.

src/robust.jl
AsafManela: done

src/robust.jl
AsafManela: perhaps because now this would be a view of a view, julia doesn't like this and it kill the process

src/robust.jl
AsafManela: same as above

src/robust.jl
AsafManela: fixed as for trim!

src/robust.jl
nalimilan: Well that's certainly not supposed to happen. Can you try starting Julia with `--check-bounds=yes`? Otherwise it would be worth reporting.

src/robust.jl
nalimilan: Have you tried this?

src/robust.jl
```diff
@@ -92,10 +100,18 @@ function winsor!(x::AbstractVector; prop::Real=0.0, count::Integer=0)
...
+    ixall = collect(1:n)
+    # indices for lowest count values
+    ixstart = partialsortperm!(ixall, x, 1:(count+1); initialized=true)
+    # indices for largest count values
+    ixend = partialsortperm!(ixall, x, (n-count):n; initialized=true)
+
+    # set lowest to count+1's value
+    @inbounds x[ixstart[1:count]] .= x[ixstart[count+1]]
+
+    # set largest to n-count)'s value
```
nalimilan:
```suggestion
    # set largest to n-count's value
```

nalimilan:
OK, I guess a 34% slowdown is still acceptable. Maybe my suggestion to replace `unique!` can reduce it a bit more.

nalimilan:
Sometimes I wonder whether it wouldn't make more sense to have these functions return iterators over the original data. If we only stored the minimum and maximum thresholds, we could just return generators that skip or replace values outside of the range on the fly. That would probably be quite faster when computing a single summary statistic (typically the mean), and you could easily call `collect` if you want a vector.

What do you think?

mschauer:
There isn't really a nice solution to it, because these would be iterators depending on properties of the full original data, so not playing nicely with iterators as inputs. But I would think an improvement in performance would justify this.

nalimilan:
> There isn't really a nice solution to it, because these would be iterators depending on properties of the full original data, so not playing nicely with iterators as inputs.

What do you mean?

mschauer:
We do not so often define iterators which require the input data to be collected or traversed,
but in forming the trimmed iterator quantiles of the full input are needed so the input iterator has to be collected or traversed before iteration can start.

nalimilan:
Yes, so as currently the input would be required to be an `AbstractVector`. I don't think that's a serious issue, as the same problem would exist if the functions returned vectors (they would have to collect the input first).

mschauer:
In this case no new structures would be needed, custom generators would suffice

```
(clamp(x, lo, up) for x in xs) # winsor
```
or
```
(x for x in xs if lo <= x <= up) #trim
```
for adequate values of `up`, `lo` could do.

AsafManela:
I fixed that typo, but also tried to implement the generator solution like this for winsor:
```julia
function winsor(x::AbstractVector; prop::Real=0.0, count::Integer=0)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))

    if count == 0
        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
        count = floor(Int, n * prop)
    else
        prop == 0 || throw(ArgumentError("prop and count can not both be > 0."))
        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
    end

    # allocate vector of all x's indices
    ixall = collect(1:n)
    # indices for lowest count values
    ixlo = partialsortperm!(ixall, x, count+1; initialized=true)
    # indices for largest count values
    ixup = partialsortperm!(ixall, x, n-count; initialized=true)

    @inbounds lo = x[ixlo]
    @inbounds up = x[ixup]

    (clamp(xi, lo, up) for xi in x)
end
```

Just creating the generator seems to be more expensive than the previous approach.

Version currently comitted to this PR:
```julia
julia> @btime winsor(v, prop=0.1);
  436.854 μs (13 allocations: 172.48 KiB)
```

Version using the above code with a generator:
```julia
julia> @btime winsor(v, prop=0.1);
  1.317 ms (8 allocations: 78.31 KiB)
julia> @btime collect(winsor(v, prop=0.1));
  1.342 ms (10 allocations: 156.52 KiB)
```

So I would suggest we just merge the PR as is.

nalimilan:
Thanks for trying. That's really surprising, as the creation of a generator is basically free... I turns out that calling `sortperm` on a range of indices is faster than passing a single value (I've filed https://github.com/JuliaLang/julia/issues/34233).

This variant is about as fast as the implementation in this PR:
```julia
function winsor2(x::AbstractVector; prop::Real=0.0, count::Integer=0)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))

    if count == 0
        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
        count = floor(Int, n * prop)
    else
        prop == 0 || throw(ArgumentError("prop and count can not both be > 0."))
        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
    end

    # allocate vector of all x's indices
    ixall = collect(1:n)
    # indices for lowest count values
    ixlo = partialsortperm!(ixall, x, 1:(count+1); initialized=true)[end]
    # indices for largest count values
    ixup = partialsortperm!(ixall, x, (n-count):n; initialized=true)[1]

    @inbounds lo = x[ixlo]
    @inbounds up = x[ixup]

    (clamp(xi, lo, up) for xi in x)
end
```
And this one is even faster:
```julia

function winsornew2(x::AbstractVector; prop::Real=0.0, count::Integer=0)
    n = length(x)
    n > 0 || throw(ArgumentError("x can not be empty."))

    if count == 0
        0 <= prop < 0.5 || throw(ArgumentError("prop must satisfy 0 ≤ prop < 0.5."))
        count = floor(Int, n * prop)
    else
        prop == 0 || throw(ArgumentError("prop and count can not both be > 0."))
        0 <= count < n/2 || throw(ArgumentError("count must satisfy 0 ≤ count < length(x)/2."))
    end

    # indices for lowest count values
    x2 = copy(x)
    lo = partialsort!(x2, 1:count+1)[end]
    # indices for largest count values
    up = partialsort!(x2, n-count:n)[1]

    (clamp(xi, lo, up) for xi in x)
end
```

Benchmarks computing mean:
```julia
julia> @btime mean(winsor(v, prop=0.1)); # From this PR
  575.010 μs (14 allocations: 172.50 KiB)

julia> @btime mean(winsornew(v, prop=0.1));
  572.053 μs (11 allocations: 78.45 KiB)

julia> @btime mean(winsornew2(v, prop=0.1));
  386.626 μs (5 allocations: 78.27 KiB)
```

AsafManela:
That explains it. Thanks for the better solution.
I implemented it for both trim and winsor and refactored to share some of the code.
New benchmarking suggests we get to have the cake and eat it too:
```julia
julia> @btime trim(v, prop=0.1);
  246.258 μs (5 allocations: 78.27 KiB)

julia> @btime winsor(v, prop=0.1);
  245.745 μs (4 allocations: 78.25 KiB)

julia> @btime collect(trim(v, prop=0.1));
  341.090 μs (19 allocations: 206.92 KiB)

julia> @btime collect(winsor(v, prop=0.1));
  252.993 μs (6 allocations: 156.45 KiB)
```

nalimilan:
Cool. We could even add a specialized `collect` method for `trim` in the future if we wanted to avoid the overhead due to `push!` (since we know the number of elements the vector can be allocated upfront).

Since moving to generators is breaking, we should probably wait a bit until we can bundle a series of breaking changes into a new minor release. If you want to fix to be available quickly, maybe drop the last commit from this PR so that we can merge it now, and open a new PR for the generator changes? As you prefer.

AsafManela:
No rush on my end. Please merge when you think appropriate.

nalimilan:
Fine with me.

Could you update the docstrings to reflect the fact that the functions no longer return a modified copy of the input? Examples can simply use `collect`. It would also be nice to specify that when several elements are equal to the threshold, they are all trimmed/clamped, meaning that the proportion or count might be adjusted (the passed value is only a lower bound). That wasn't the case before (which didn't make a lot of sense actually).

AsafManela:
I updated the examples, but I don't fully understand your second comment.
Clamp keeps lo <= x <= hi as is and only replaces strictly higher or lower elements. So elements equal to the thresholds are not trimmed or clamped. Right?

nalimilan:
Ah, right. Though my remark still holds in a different form: if there are several values equal to the threshold, none of them will be skipped/clamped, but in the previous implementations some of them could have been, and some of them not, if the specified proportion fell exactly in the middle of that group. Do you see what I mean?

src/robust.jl
```diff
@@ -66,14 +66,14 @@ end
 """
     winsor(x; prop=0.0, count=0)
 
-Return a copy of `x` with either `count` or proportion `prop` of the lowest
-elements of `x` replaced with the next-lowest, and an equal number of the
-highest elements replaced with the previous-highest.  To compute the Winsorized
-mean of `x` use `mean(winsor(x))`.
+Return a generator of all elements of `x` that replaces either `count` or
```
nalimilan: Maybe let's just say "an iterator", so that in the future we can return a more specific type than an iterator. For example we could return a special `AbstractVector` type. Same for `trim` (even if it cannot be a vector since computing indices is O(N)).

AsafManela:
I see. How about this:
"The number of trimmed / replaced elements could be smaller then specified if several elements equal the lower or upper bound."

nalimilan:
Sounds good! (But "then" -> "than")

Issue #549, opened 2020-01-17
R² generalization


getzze:
For generalized models, the docs mention that `:CoxSnell` is equivalent to R² for linear models:
https://github.com/JuliaStats/StatsBase.jl/blob/0faf210f5fb84d361bd101404aadaef54376fcbd/src/statmodels.jl#L225
But, it is the likelihood ratio index (`:MacFadden`) that gives the same definition for R² for a linear model and for a Normal GLM. Indeed, `log(L)` is the sum of squares for a Normal likelihood, but `L` is a Gaussian function.
I would suggest changing this sentence to:
"The likelihood ratio index matches the classical definition of R² for linear models."

[EDIT] It is not McFadden's likelihood ratio that corresponds to the linear R², but the _deviance_ ratio. See #550.

nalimilan:
@lbittarello You wrote this at #400. Do you remember the details?

lbittarello:
McFadden's R² doesn't match the classical R²  because the log likelihood is equal to `0.5 * [1 + ln(2π) + σ²]`, where `σ²` is the sum of squared residuals. The constant `1 + ln(2π)` distorts the ratio `ll / ll₀`. This term cancels out in Cox and Snell's R² (which uses `exp(ll - ll₀)` instead of `ll / ll₀`).

getzze:
Yes, McFadden uses loglikelihood and I always assumed that it was the same as the deviance. It works with the deviance ratio though, I will edit the original issue.
I created a PR to add the variant with the deviance ratio #550.

nalimilan:
@getzze Can you update https://github.com/JuliaStats/StatsBase.jl/pull/550 to keep the mention about Cox & Snell match the classical R² then?

nalimilan:
Closed by https://github.com/JuliaStats/StatsBase.jl/pull/550.

PR #550, opened 2020-01-21
Add :devianceratio variant to r² function

getzze:
Another generalization for R² to GLM is by using the deviance ratio (https://www.sciencedirect.com/science/article/pii/S0304407696018180). As deviance is zero for the saturated model, r² values are defined between 0 when the considered model is the null model, with the minimal number of parameters; and 1 when the model is the saturated model, with the maximal number of parameters.

Using this variant gives the same result ``mss/tss`` for OLS, which is mentioned in the last line of the doc. See https://github.com/JuliaStats/StatsBase.jl/issues/549

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -222,9 +222,9 @@ In the above formulas, ``L`` is the likelihood of the model,
 ``L_0`` is the likelihood of the null model (the model with only an intercept),
 ``D`` is the deviance of the model (from the saturated model),
 ``D_0`` is the deviance of the null model,
-``n`` is the number of observations.
+``n`` is the number of observations (given by [`nobs`](@ref)).
 
-The deviance ratio R² matches the classical definition of R² for linear models.
+The deviance ratio and the CoxSnell R²'s both match the classical definition of R² for linear models.
 """
 function r2(obj::StatisticalModel, variant::Symbol)
     loglikbased = [:McFadden, :CoxSnell, :Nagelkerke]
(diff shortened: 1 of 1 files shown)


src/statmodels.jl
```diff
@@ -216,13 +216,15 @@ Supported variants are:
...
 In the above formulas, ``L`` is the likelihood of the model,
 ``L_0`` is the likelihood of the null model (the model with only an intercept),
-``n`` is the number of observations, ``y_i`` are the responses,
-``\\hat{y}_i`` are fitted values and ``\\bar{y}`` is the average response.
+``D`` is the deviance of the model (from the saturated model),
+``D_0`` is the deviance of the null model,
+``n`` is the number of observations.
 
-Cox and Snell's R² should match the classical R² for linear models.
+The deviance ratio R² matches the classical definition of R² for linear models.
```
nalimilan: Ref #549.

src/statmodels.jl
```diff
@@ -233,8 +235,12 @@ function r2(obj::StatisticalModel, variant::Symbol)
         1 - exp(2 * (ll0 - ll) / nobs(obj))
     elseif variant == :Nagelkerke
         (1 - exp(2 * (ll0 - ll) / nobs(obj))) / (1 - exp(2 * ll0 / nobs(obj)))
+    elseif variant == :deviance_ratio
+        dev  = deviance(obj)
```
nalimilan: This variant doesn't need `ll` and `ll0`, so you should add a top-level `if` to avoid computing these.

src/statmodels.jl
```diff
@@ -259,19 +265,30 @@ adjr2(obj::StatisticalModel) = error("adjr2 is not defined for $(typeof(obj)).")
 Adjusted pseudo-coefficient of determination (adjusted pseudo R-squared).
 
 For nonlinear models, one of the several pseudo R² definitions must be chosen via `variant`.
-The only currently supported variant is `:MacFadden`, defined as ``1 - (\\log (L) - k)/\\log (L0)``.
+The only currently supported variants are `:MacFadden`, defined as ``1 - (\\log (L) - k)/\\log (L0)`` and
+`:deviance_ration`, defined as ``1 - (D/dof)/(D_0/dof_0)``.
```
nalimilan: Use `k` and `k_0` for consistency with the previous formula.

src/statmodels.jl
```diff
@@ -259,19 +265,30 @@ adjr2(obj::StatisticalModel) = error("adjr2 is not defined for $(typeof(obj)).")
...
     ll = loglikelihood(obj)
     ll0 = nullloglikelihood(obj)
     k = dof(obj)
     if variant == :McFadden
         1 - (ll - k)/ll0
+    elseif variant == :deviance_ratio
+        n = nobs(obj)
+        # Number of explanatory variables
+        p = dof(obj)
+        dev  = deviance(obj)/(n - p)
```
nalimilan: Just for clarity, I'd do `dev  = deviance(obj)`, and only divide at the end.

Also, same remark about not computing `ll` and `ll0`.

src/statmodels.jl
```diff
@@ -216,13 +216,15 @@ Supported variants are:
 - `:MacFadden` (a.k.a. likelihood ratio index), defined as ``1 - \\log (L)/\\log (L_0)``;
 - `:CoxSnell`, defined as ``1 - (L_0/L)^{2/n}``;
 - `:Nagelkerke`, defined as ``(1 - (L_0/L)^{2/n})/(1 - L_0^{2/n})``.
+- `:deviance_ratio`, defined as ``1 - D/D_0``.
```
nalimilan: I'd avoid the underscore for consistency with other variants:
```suggestion
- `:devianceratio`, defined as ``1 - D/D_0``.
```

PR #554, opened 2020-02-16
New version (~0.33.0~0.32.1) and update docs

### LINKED ISSUE #547: Tag a new version?
There has been a lot of changes since the last version. Would it be possible to tag a new one?

kleinschmidt:
fixes #547

The only substantive change here is that the docs are building with Documenter
0.24, I've updated some doctests, and I've changed .travis.yml to run CI on 1.3
and 1.4 instead of 1.2, and build docs using 1.3.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.32.0"
+version = "0.33.0"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.33.0"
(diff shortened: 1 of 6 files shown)


docs/Project.toml
```diff
@@ -1,5 +1,6 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
-Documenter = "~0.22"
+Documenter = "~0.24"
```
Nosferican: This should be

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.32.0"
+version = "0.33.0"
```
Nosferican: Is this really breaking? If it isn't breaking, it should `version = "0.32.1"` according to Pkg's semver (https://julialang.github.io/Pkg.jl/v1/toml-files/#The-version-field-1).

nalimilan:
Wait before merging, I had put some PRs on hold in case we wanted to make another non-breaking release.

kleinschmidt:
I've merged #548 ... anything else we should merge before a non-breaking release?

nalimilan:
I don't think so.

PR #559, opened 2020-02-24
Update sampling.jl

githubtomtom:
enable `make_alias_table!()` for all float types. It would be useful to enable e.g. `AliasTable()` in `Distributions.jl` to create object with element type other than `Float64`.


githubtomtom:
a related [pull request](https://github.com/JuliaStats/Distributions.jl/pull/1074) is made in `Distributions.jl`

PR #565, opened 2020-03-01
fix breaking zscore changes (backport to 0.32 branch)

### LINKED ISSUE #556: 0.32.1 introduces a breaking change
The constructor for `ZScoreTransform` changed between `v0.32.0` and `v0.32.1` and now requires an extra argument.

kleinschmidt:
fixes #556 

#490 introduced breaking changes to the constructor of the ZScoreTransform type, adding a fourth mandatory argument.  It also introduced an error in cases where the previous default (without `dims=`) was used for `fit` and everything the calls it.  This PR fixes those by adding an inner constructor with a depwarn for ZScoreTransform, and re-setting the default dims inside the call to `fit` when `dims=nothing`.

### DIFF
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -280,6 +280,7 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
         l, tmin, tmax = _compute_extrema(X')
     elseif dims == nothing
         Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
+        dims = 2
         l, tmin, tmax = _compute_extrema(X')
     else
         throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -63,9 +63,9 @@ struct ZScoreTransform{T<:Real} <: AbstractDataTransform
         new{T}(l, dims, m, s)
     end
 
-    function ZScoreTransform(dims::Int, m::Vector{T}, s::Vector{T}) where T
-        Base.depwarn("ZScoreTransform(dims, m, s) is deprecated: use ZScoreTransform(len, dims, m, s) instead", :ZScoreTransform)
-        ZScoreTransform(max(length(m), length(s)), dims, m, s)
+    function ZScoreTransform(l::Int, m::Vector{T}, s::Vector{T}) where T
+        Base.depwarn("ZScoreTransform(len, m, s) is deprecated: use ZScoreTransform(len, 2, m, s) instead", :ZScoreTransform)
+        ZScoreTransform(l, 2, m, s)
     end
 end
 
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -62,6 +62,11 @@ struct ZScoreTransform{T<:Real} <: AbstractDataTransform
         lens == l || lens == 0 || throw(DimensionMismatch("Inconsistent dimensions."))
         new{T}(l, dims, m, s)
     end
+
+    function ZScoreTransform(dims::Int, m::Vector{T}, s::Vector{T}) where T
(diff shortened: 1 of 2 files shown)


src/transformations.jl
```diff
@@ -62,6 +62,11 @@ struct ZScoreTransform{T<:Real} <: AbstractDataTransform
         lens == l || lens == 0 || throw(DimensionMismatch("Inconsistent dimensions."))
         new{T}(l, dims, m, s)
     end
+
+    function ZScoreTransform(dims::Int, m::Vector{T}, s::Vector{T}) where T
+        Base.depwarn("ZScoreTransform(dims, m, s) is deprecated: use ZScoreTransform(len, dims, m, s) instead", :ZScoreTransform)
+        ZScoreTransform(max(length(m), length(s)), dims, m, s)
+    end
```
nalimilan: I think you got the deprecation backwards: it's `dims` that was introduced by https://github.com/JuliaStats/StatsBase.jl/pull/490, not `len`:
```suggestion
    function ZScoreTransform(l::Int, m::Vector{T}, s::Vector{T}) where T
        Base.depwarn("ZScoreTransform(len, m, s) is deprecated: use ZScoreTransform(len, 2, m, s) instead", :ZScoreTransform)
        ZScoreTransform(l, 2, m, s)
    end
```
(BTW, this doesn't have to be an inner constructor, though it doesn't really make a difference I guess.)

kleinschmidt: Won't this cause problems if you try to apply this transformation to a vector though?

kleinschmidt: Although I guess that given this is not something users are likely to directly interact with very often then that's okay.

nalimilan: Well it should work the same as it did in 0.31 I guess: we just deprecated the implicit `dims`, but the behavior with `dims=2` isn't supposed to have changed IIRC.

src/transformations.jl
```diff
@@ -131,6 +137,7 @@ function fit(::Type{ZScoreTransform}, X::AbstractVector{<:Real};
              dims::Union{Integer,Nothing}=nothing, center::Bool=true, scale::Bool=true)
     if dims == nothing
         Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
+        dims = 1
```
nalimilan: Woops, shouldn't it be:
```suggestion
        dims = 2
```

kleinschmidt:
That will just result in an error. Depends how pedantic we want to be about the default being 2. I'd argue that that would still be breaking since code that worked before would result in an error.

nalimilan:
If there's a relatively simple way of making the old code work, then better implement it I guess.

nalimilan:
Sorry I think I misunderstood your point. Were you replying to my last suggestion? Anyway, AFAICT before #490 we didn't support standardizing vectors, so that probably doesn't matter.

kleinschmidt:
I'm not exactly sure whether vectors had first class support but there were transform etc. methods for `Array{<:Real, 1}` so that should still work to avoid breakage. But maybe I'm misunderstanding how vectors were handled before #490.

nalimilan:
Ah, yeah, the situation was weird at the time. Well, any deprecation is better than the current state, so choose the one you think is best.

kleinschmidt:
yeah now that I'm looking back at the state of things in v0.32 was pretty strange...I think providing a no-dims method that does something reasonable and provides a deprecation warning is a reasonable compromise between being a stickler about not changing the API at all (e.g., used to error on vector, still errors on a vector) and making the merged stuff useful.

kleinschmidt:
Actually, I changed my mind.  Previously (v0.32) `fit(ZScoreTransform, rand(100))` was a method error. Currently (v0.32.1) it's a different method error because `dims=nothing` is just passed onto the constructor.  With the current PR, it's a depwarn.  Removing the check for `dims==nothing` and just throwing a domain error saying "use `dims=1`" seems like a better choice overall, since it's more consistent with teh current behavior, the historical behavior, and still informative about how to fix the problem.

Issue #459, opened 2019-02-03
RealFP types too restrictive for signalcorr.jl methods


danmackinlay:
the methods in `signcorr.jl` require arrays of `RealFP` types;
This seems to me to be too restrictive since it, for example, the Dual types used by ForwardDiff.jl, and prevents differentiating through these functions.

Does anyone know a reason that we have this restriction? If I can get some clues about the reasons for this design choice I can propose a patch.

```julia
julia> using StatsBase:  autocov
julia> using ForwardDiff
julia> f2(x::Vector) = autocov(x, [0])[1]
julia> g2 = x -> ForwardDiff.gradient(f2, x);
julia> f2([1.0]), g2([1.0])

MethodError: no method matching fptype(::Type{ForwardDiff.Dual{ForwardDiff.Tag{typeof(f2),Float64},Float64,1}})
Closest candidates are:
  fptype(!Matched::Type{Complex{Float32}}) at /home/dan/.julia/packages/StatsBase/yNoh6/src/common.jl:27
  fptype(!Matched::Type{Complex{Float64}}) at /home/dan/.julia/packages/StatsBase/yNoh6/src/common.jl:28
  fptype(!Matched::Type{T<:Union{Bool, Float32, Int16, Int8, UInt16, UInt8}}) where T<:Union{Bool, Float32, Int16, Int8, UInt16, UInt8} at /home/dan/.julia/packages/StatsBase/yNoh6/src/common.jl:25
  ...
```

nalimilan:
This code is quite old, it probably just hasn't been updated. As you can see, the error is due to the lack of an appropriate `fptype` method). I guess we could add a fallback definition `fptype(::Type{T}) where {T} = float(T)`. Does that fix your problem? If so, feel free to make a pull request.

danmackinlay:
Hm - possibly. Thanks @nalimilan. I will check that for ForwardDiff. It will for sure help with things like BigFloats, which is another limitation the current code has.

Just so I can understand here, though, what we are doing with these `fptype` methods is trying to find the smallest float type such that there will be no precision lost when the covariance calculation is conducted in float arithmetic. Is that right?

If so, that looks like the [promotion methods built in to julia](https://docs.julialang.org/en/v1.1/manual/conversion-and-promotion/#Promotion-1). Is there a reason *not* to use those built in promotion methods? `signalcorr.jl` doesn't do anything very exotic, so the custom promotion logic is a bit weird.

[...]

PR #566, opened 2020-03-04
minor version

kleinschmidt:
The next release should be a minor version because breaking changes were introduced in #546.  This is a placeholder for bumping the minor version.  

Still to do: remove deprecations for transformations introduced in #490

PR #567, opened 2020-03-07
Remove fptype

devmotion:
This PR addresses https://github.com/JuliaStats/StatsBase.jl/issues/459 and fixes (at least parts of) https://github.com/TuringLang/MCMCChains.jl/issues/187.

`fptype` is replaced with `float` based on the element types of the input arguments. Additionally, many type restrictions are softened which allows to remove conversions of integer-valued arrays to floating point-valued arrays.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -20,13 +20,6 @@ const IntegerMatrix{T<:Integer} = AbstractArray{T,2}
 
 const RealFP = Union{Float32, Float64}
 
-## conversion from real to fp types
-
-fptype(::Type{T}) where {T<:Union{Float32,Bool,Int8,UInt8,Int16,UInt16}} = Float32
-fptype(::Type{T}) where {T<:Union{Float64,Int32,UInt32,Int64,UInt64,Int128,UInt128}} = Float64
-fptype(::Type{Complex{Float32}}) = Complex{Float32}
-fptype(::Type{Complex{Float64}}) = Complex{Float64}
-
 # A convenient typealias for deprecating default corrected Bool
 const DepBool = Union{Bool, Nothing}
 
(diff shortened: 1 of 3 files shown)


devmotion:
The second commit is needed to fix https://github.com/JuliaStats/StatsBase.jl/issues/459 completely.

devmotion:
Any comments or suggestions? I've confirmed that this PR actually fixes the downstream issue in MCMCChains https://github.com/TuringLang/MCMCChains.jl/issues/187.

src/common.jl
```diff
@@ -20,13 +20,6 @@ const IntegerMatrix{T<:Integer} = AbstractArray{T,2}
 
 const RealFP = Union{Float32, Float64}
 
-## conversion from real to fp types
-
-fptype(::Type{T}) where {T<:Union{Float32,Bool,Int8,UInt8,Int16,UInt16}} = Float32
```
nalimilan: `float(Int8) === Float64`, so this PR changes the behavior. This warrants a discussion at least. (Same for place where `typeof(zero(eltype(X))/1)` is used.

A less disruptive fix would be to just add a fallback `fptype(::Type{T}) where {T} = float(T)`.

devmotion: IMO the behaviour of `fptype` was completely unexpected and inconsistent with the default promotion logic in Julia that I've seen in other places (e.g., in `mean`). I expect floating point numbers of type `Float64` in all the cases `Bool`, `Int8` etc. since that is the default floating point number type in Julia and only of type `Float32` if the input type contains floating point numbers of type `Float32`.

Additionally, users can still provide output arrays and hence obtain results with floating point numbers of type `Float32` even in these cases.

In any case, it seems it is an additional argument for https://github.com/JuliaStats/StatsBase.jl/pull/566.

nalimilan: OK, everybody agrees with the change, so we'll tag a new minor release.

PR #570, opened 2020-03-25
Micro-change `ordinalrank`

mileslucas:
A micro-edit that changes to use `similar(x, Int)` in `ordinalrank`'s dispatch to `ordinalrank!` to improve type flexibility.


nalimilan:
Why not, but if we do this we should change all other rank functions. Do you have a use case in mind?

PR #576, opened 2020-05-07
fix legacy usage of RangeGenerator

rfourquet:
`RangeGenerator` is a pre-Julia-1.0 thing, and was maintained in `Random` only to reduce breakage of packages. But the (not so) new way of doing things is now `Sampler(rng, 1:n)`. This is supposed to select the best sampler type depending on `rng`. 
Fixes https://github.com/rfourquet/StableRNGs.jl/issues/3.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -5,12 +5,12 @@
 #
 ###########################################################
 
-using Random: RangeGenerator, Random.GLOBAL_RNG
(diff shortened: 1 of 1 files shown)


rfourquet:
This fails on Julia-nightly tests, but this is not related to this PR AFAICT (I get a similar failure locally on Julia-nightly with StatsBase@v0.33.0).

rfourquet:
Good to go?

PR #579, opened 2020-05-18
Fix wrong argument name in informationmatrix

### LINKED ISSUE #577: Typo in function `informationmatrix`
In the [definition of `informationmatrix`](https://github.com/JuliaStats/StatsBase.jl/blob/master/src/statmodels.jl#L118-L119), there is a typo, namely: the model is called `model` in the arguments but referred to as `obj` in the function's body.
This causes an `UndefVarError` when `informationmatrix` is called on a type wich it is not defined for.

nalimilan:
Also specify "of the model" for consistency with other functions.

Fixes #577.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -3,211 +3,214 @@
 abstract type StatisticalModel end
 
 """
-    coef(obj::StatisticalModel)
+    coef(model::StatisticalModel)
 
 Return the coefficients of the model.
 """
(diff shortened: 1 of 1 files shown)


andreasnoack:
Not a big deal but what about `obj` -> `model`? It's a little less general to use `model` here.

Any idea why tests are failing on nightly?

nalimilan:
OK, I've replaced all occurrences.

Not sure about the failures, but they might be due to the changes to random numbers.

PR #580, opened 2020-05-18
Check for length(wv), fixes #578

### LINKED ISSUE #578: Check for `length(wv)`
https://github.com/JuliaStats/StatsBase.jl/blob/0ea8e798c3d19609ed33b11311de5a2bd6ee9fd0/src/counts.jl#L46

I think you should check for the length of `wv`, since you're using `@inbounds` later

davidavdav:
As the title suggests...

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -37,7 +37,7 @@ end
 
 function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange, wv::WeightVec)
     k = length(levels)
-    length(r) == k || throw(DimensionMismatch())
+    length(r) == length(wv) == k || throw(DimensionMismatch())
 
     m0 = levels[1]
     m1 = levels[end]


nalimilan:
Thanks. Interestingly this makes tests fail in `addcounts!`, which isn't very reassuring... Could you check what's going on, and add a test using `@test_throws`?

davidavdav:
Sorry, haven't had the time yet to look into this

src/counts.jl
```diff
@@ -37,7 +37,7 @@ end
 
 function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange, wv::WeightVec)
     k = length(levels)
-    length(r) == k || throw(DimensionMismatch())
+    length(r) == length(wv) == k || throw(DimensionMismatch())
```
kleinschmidt, line 40: isn't this the wrong comparison?  the weights should have the same length indices as the `x` vector unless I'm misunderstanding this code

kleinschmidt:
```suggestion
    length(r) == k || throw(DimensionMismatch())
    length(wv) == length(x) || throw(DimensionMismatch())
```

nalimilan: Or even better:
```suggestion
    length(r) == k ||
        throw(DimensionMismatch("length of levels ($k) must match length of destination array ($(length(r)))"))
    length(wv) == length(x) ||
        throw(DimensionMismatch("length of weights vector ($(length(wv))) must match length of data vector ($(length(x))"))
```

Issue #574, opened 2020-04-28
arithmetics for Histogram objects


Moelf:
should this (and for `+, -, *`) be added for `Histogram` objects?
```
import Base:/

function /(h1::Histogram, h2::Histogram)
    @assert h1.edges == h2.edges
    Histogram(h1.edges, h1.weights ./ h2.weights)
end
```

PR #583, opened 2020-06-15
add Histogram arithmetics

Moelf:
#574

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -123,6 +123,9 @@ The `Histogram` type represents data that has been tabulated into intervals
 (known as *bins*) along the real line, or in higher dimensions, over a real space.
 Histograms can be fitted to data using the `fit` method.
 
+Histogram can be `+, -, *, /`ed with each other (when bins are the same) and with 
+`Number`s, which will change the weights of each bin.
+
 # Fields
 * edges: An iterator that contains the boundaries of the bins in each dimension.
 * weights: An array that contains the weight of each bin.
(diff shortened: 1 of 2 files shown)


Moelf:
bump, also I think adding `+-*/(h1:Histogram, n:Number)` is also reasonable

nalimilan:
Sounds reasonable, but I'm not familiar enough with that code to really confirm.

wildart:
Looks good.

src/hist.jl
```diff
@@ -122,6 +123,9 @@ The `Histogram` type represents data that has been tabulated into intervals
 (known as *bins*) along the real line, or in higher dimensions, over a real space.
 Histograms can be fitted to data using the `fit` method.
 
+Histogram can be `+, -, *, /`ed with each other (when bins are the same) and with 
+`Number`s, which will change the weights of each bin.
```
nalimilan, line 127:
```suggestion
`Histogram`s support `+`, `-`, `*` and `/` with other `Histogram`s (when bins are the same)
and with `Real`s, which will change the weights of each bin.
```

src/hist.jl
```diff
@@ -219,6 +223,11 @@ function show(io::IO, h::AbstractHistogram)
 end
 
 (==)(h1::Histogram,h2::Histogram) = (==)(h1.edges,h2.edges) && (==)(h1.weights,h2.weights) && (==)(h1.closed,h2.closed) && (==)(h1.isdensity,h2.isdensity)
+for op in (:+, :-, :*, :/)
+    @eval ($op)(h1::Histogram,h2::Histogram) = 
+    (==)(h1.edges,h2.edges) ? Histogram(h1.edges, broadcast($op, h1.weights, h2.weights)) : throw(DimensionMismatch("Bins don't match"))
```
nalimilan, line 228:
```suggestion
        (==)(h1.edges,h2.edges) ? Histogram(h1.edges, broadcast($op, h1.weights, h2.weights)) : throw(DimensionMismatch("The bins of the two histograms don't match"))
```

Moelf:
anything else needed?

Moelf:
bump

oschulz:
Gentle bump @andreasnoack,  @nalimilan

test/hist.jl
```diff
@@ -219,6 +219,21 @@ end
...
 
+@testset "Histogram arithmetics" begin
+    h1 = fit(Histogram, rand(100), 0:0.1:1)
+    h2 = fit(Histogram, rand(200), 0:0.1:1)
+    h3 = fit(Histogram, rand(100), 0:0.2:1)
+    @test (h1 + h2).weights == h1.weights .+ h2.weights
+    @test (h1 - h2).weights == h1.weights .- h2.weights
+    @test (h1 * h2).weights == h1.weights .* h2.weights
+    @test (h1 / h2).weights == h1.weights ./ h2.weights
+    @test (h1 + 1.5).weights == h1.weights .+ 1.5
```
nalimilan, line 230: Sorry to raise this only now, but what's the use case for adding a real value to all weights? That doesn't seem to make a lot of sense to me.

Moelf: you may have a "flat background", but ok, this is not terribly useful, I can live without it if you think it does harm

Moelf:
implemented in https://github.com/Moelf/FHist.jl

PR #584, opened 2020-06-16
Correct docstrings of autocor and autocov

fzierler:
If `x` is a vector then `autocov` and `autocor` as well as their mutating versions return a vector of length `length(lags)` and not a vector of the same length as `x`.


nalimilan:
Thanks!

PR #589, opened 2020-08-09
Enhance ranking code

alyst:
Small cleanups in the ranking code. Added `@inbounds` annotations for speed.
Common code is moved into `_rank()` method, which allows to:
  * correctly support n-dim arrays (n>1). If generic array input was not intended, I can tighten signatures to only allow `x::AbstractVector`
  * `sortperm()` kwargs are just passed through, so, in addition to *lt* and *rev*, ranking also supports *by* (so it could be applied to custom types) and *alg*
  * no function-generating macros are required to support missing values

### DIFF
--- a/src/ranking.jl
+++ b/src/ranking.jl
@@ -6,48 +6,63 @@
 # The implementations here follow this wikipedia page.
 #
 
-
 function _check_randparams(rks, x, p)
     n = length(rks)
     length(x) == length(p) == n || raise_dimerror()
     return n
 end
 
+# ranking helper function: calls sortperm(x) and then ranking method f!
+function _rank(f!, x::AbstractArray, R::Type=Int; sortkwargs...)
+    rks = similar(x, R)
+    ord = reshape(sortperm(vec(x); sortkwargs...), size(x))
+    return f!(rks, x, ord)
+end
 
+# ranking helper function for arrays with missing values
+function _rank(f!, x::AbstractArray{>: Missing}, R::Type=Int; sortkwargs...)
+    inds = findall(!ismissing, vec(x))
+    isempty(inds) && return missings(R, size(x))
+    T = nonmissingtype(eltype(x))
+    xv = Vector{T}(undef, length(inds))
+    @inbounds for (i, ind) in enumerate(inds)
+        xv[i] = x[ind]
+    end
+    rks = missings(R, size(x))
+    ordv = sortperm(xv; sortkwargs...)
+    f!(view(rks, inds), xv, ordv)
+    return rks
+end
 
 # Ordinal ranking ("1234 ranking") -- use the literal order resulted from sort
-function ordinalrank!(rks::AbstractArray, x::AbstractArray, p::IntegerArray)
-    n = _check_randparams(rks, x, p)
-
-    if n > 0
-        @inbounds for i in eachindex(p)
-            rks[p[i]] = i
-        end
+function _ordinalrank!(rks::AbstractArray, x::AbstractArray, p::IntegerArray)
+    _check_randparams(rks, x, p)
+    @inbounds for i in eachindex(p)
+        rks[p[i]] = i
     end
-
     return rks
 end
 
 
 """
-    ordinalrank(x; lt = isless, rev::Bool = false)
+    ordinalrank(x; sortkwargs...)
 
 Return the [ordinal ranking](https://en.wikipedia.org/wiki/Ranking#Ordinal_ranking_.28.221234.22_ranking.29)
 ("1234" ranking) of an array. The `lt` keyword allows providing a custom "less
 than" function; use `rev=true` to reverse the sorting order.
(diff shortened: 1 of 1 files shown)


src/ranking.jl
```diff
@@ -6,50 +6,63 @@
...
+        rks[p[i]] = i
     end
-
     return rks
 end
 
 
 """
-    ordinalrank(x; lt = isless, rev::Bool = false)
+    ordinalrank(x; sortkwargs...)
```
nalimilan, line 48: Repeating the list of supported arguments is more user-friendly IMO, and it doesn't take a lot of space.

src/ranking.jl
```diff
@@ -6,50 +6,63 @@
...
 
+# ranking helper function for arrays with missing values
+function _rank(f!, x::AbstractArray{>: Missing}, R::Type=Int; sortkwargs...)
+    inds = findall(!ismissing, vec(x))
+    isempty(inds) && return missings(R, size(x))
+    T = nonmissingtype(eltype(x))
+    xv = Vector{T}(undef, length(inds))
+    @inbounds for (i, ind) in enumerate(inds)
+        xv[i] = x[ind]
+    end
```
nalimilan, line 30: Why don't we just do `xv = view(x, inds)`? If making a copy is faster, then `xv = disallowmissing(view(x, inds))` as in the old code or `xv = convert(AbstractArray{T}, x)` is both shorter and much more efficient e.g. for `CategoricalArray`.

src/ranking.jl
```diff
@@ -6,50 +6,63 @@
...
+    _rank(_ordinalrank!, x; sortkwargs...)
 
 
 # Competition ranking ("1224" ranking) -- resolve tied ranks using min
-function competerank!(rks::AbstractArray, x::AbstractArray, p::IntegerArray)
+function _competerank!(rks::AbstractArray, x::AbstractArray, p::IntegerArray)
     n = _check_randparams(rks, x, p)
 
-    if n > 0
+    @inbounds if n > 0
```
nalimilan, line 65: While you're at it, maybe replace these `while` loops with `for i in 2:n` and so on? They are hard to read currently.

nalimilan (commented):
Sorry for the very slow reponse...

src/ranking.jl
alyst: I've expanded *lt*, *by* and *rev*, because they are more commonly used, and *alg* and *order* definitions are too long.
The docstring now explicitly refers to the `sort(x)` keywords in the description.

src/ranking.jl
```diff
@@ -45,12 +42,11 @@ end
...
 
 """
-    ordinalrank(x; sortkwargs...)
+    ordinalrank(x; lt=isless, by=identity, rev::Bool=false, ...)
 
 Return the [ordinal ranking](https://en.wikipedia.org/wiki/Ranking#Ordinal_ranking_.28.221234.22_ranking.29)
-("1234" ranking) of an array. The `lt` keyword allows providing a custom "less
-than" function; use `rev=true` to reverse the sorting order.
-All items in `x` are given distinct, successive ranks based on their
+("1234" ranking) of an array. Supports the same keyword arguments as `sort(x; sortkwargs...)`
```
nalimilan, line 48: `sortkwargs` isn't defined now. Same below.
```suggestion
("1234" ranking) of an array. Supports the same keyword arguments as the `sort`
```

alyst: Actually, in L48 I sort-of introduce the *sortkwargs*, because otherwise later in the docstring it would not be entirely clear that the keyword args are passed through to `sort` to define the element order. I think from the context (*supports the same keyword args as ...*) it should be clear that *sortkwargs* refers to the keyword arguments of *ordinalranks()*. But if you have a better suggestion, I'm eager to fix it.

nalimilan: OK. I would just have said "in the sorted vector" in the next sentence.

As you prefer, but if you keep this then please replace `sort(x)` with just `sort` below.

alyst: Yes, that sounds good. I also simplified a bit the language in the other ranking docstrings.

alyst:
CI failed on nightly, but it's not ranking-related.

@nalimilan Thanks for the review! There's also #590 if you have time :)

nalimilan (approved):

src/ranking.jl
```diff
@@ -6,65 +6,71 @@
...
+    rks = similar(x, R)
+    ord = reshape(sortperm(vec(x); sortkwargs...), size(x))
+    return f!(rks, x, ord)
+end
 
+# ranking helper function for arrays with missing values
+function _rank(f!, x::AbstractArray{>: Missing}, R::Type=Int; sortkwargs...)
+    inds = findall(!ismissing, vec(x))
+    isempty(inds) && return missings(R, size(x))
+    T = nonmissingtype(eltype(x))
```
alyst, line 26: just noticed that `T` is not used anymore, will remove this line

PR #590, opened 2020-08-09
Empirical CDF enhancements

alyst:
The PR improves the performance of *ECDF* and adds interpolation.
 * The major problem with the current code is that in the weighted case the partial sums have to be recalculated for each input. But this is not necessary, since all CDF values for weighted and non-weighted cases could be precalculated.
 * I suppose *ECDF* was written in pre-broadcast epoch, so it does not overload `Base.Broadcast.broadcasted()` for providing enhanced support for vectors. The PR deprecates `ecdf(v::AbstractVector)` in favor of standard dot notation. I've kept customized broadcasting, but now it just caches the CDF value of the last vector element.
 * Optionally, *ECDF* can enable interpolation (`ecdf(..., interpolate=true)`), which is handy for continuous empirical distributions. (It just linearly interpolates between the CDFs of adjacent values).

### DIFF
--- a/src/empirical.jl
+++ b/src/empirical.jl
@@ -2,18 +2,25 @@
 
 ## Empirical CDF
 
-struct ECDF{T <: Real, W <: Real}
-    sorted_values::Vector{Tuple{T, W}}
+struct ECDF{T <: Real, W <: Real, I}
+    sorted_values::Vector{Tuple{T, W, W, W}}
 end
 
 weighttype(::Type{<:ECDF{<:Any, W}}) where W = W
 weighttype(ecdf::ECDF) = weighttype(typeof(ecdf))
+isinterpolating(::Type{<:ECDF{<:Any, <:Any, I}}) where I = I
+isinterpolating(ecdf::ECDF) = isinterpolating(typeof(ecdf))
 
 function (ecdf::ECDF)(x::Real)
     isnan(x) && return NaN
     pos = searchsortedlast(ecdf.sorted_values, x, by=first)
     (pos == 0) && return zero(weighttype(ecdf))
-    @inbounds return ecdf.sorted_values[pos][2]
+    @inbounds val, cdf_val, val_invdelta, cdf_delta = ecdf.sorted_values[pos]
+    if !isinterpolating(ecdf) || (val == x) || (pos == length(ecdf.sorted_values))
+        return cdf_val
+    else
+        return muladd(cdf_delta, min((x - val) * val_invdelta, one(weighttype(ecdf))), cdf_val)
+    end
 end
 
 # broadcasts ecdf() over an array
@@ -39,7 +46,8 @@ evaluate CDF values on other samples.
 function is inside the interval ``(0,1)``; the function is defined for the whole real line.
 """
 function ecdf(X::RealVector;
-              weights::Union{Nothing, RealVector}=nothing)
+              weights::Union{Nothing, RealVector}=nothing,
+              interpolate::Bool=false)
     any(isnan, X) && throw(ArgumentError("ecdf can not include NaN values"))
     evenweights = isnothing(weights) || isempty(weights)
     evenweights || (length(X) == length(weights)) ||
@@ -52,14 +60,15 @@ function ecdf(X::RealVector;
     wsum = evenweights ? length(X) : sum(weights)
     ord = sortperm(X)
 
-    sorted_vals = sizehint!(Vector{Tuple{T, W}}(), length(X))
-    isempty(X) && return ECDF{T, W}(sorted_vals)
+    sorted_vals = sizehint!(Vector{Tuple{T, W, W, W}}(), length(X))
+    isempty(X) && return ECDF{T, W, interpolate}(sorted_vals)
 
     valprev = val = X[first(ord)]
     wsumprev = zero(W0)
     valw = zero(W0)
 
-    push_valprev!() = push!(sorted_vals, (valprev, min(wsumprev/wsum, one(W))))
+    push_valprev!() = push!(sorted_vals, (valprev, min(wsumprev/wsum, one(W)),
+                                          inv(val - valprev), valw/wsum))
 
     @inbounds for i in ord
         valnew = X[i]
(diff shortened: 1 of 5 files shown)


src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
-        r0 += evenweights ? 1 : ecdf.weights[j]
-        if i > m
-            break
-        end
-    end
-    while i <= m
-        r[ord[i]] = weightsum
-        i += 1
+# broadcasts ecdf() over an array
+# caches the last calculated value
```
nalimilan, line 54: What do you mean by "caches"?

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
-            i += 1
-        end
-        r0 += evenweights ? 1 : ecdf.weights[j]
-        if i > m
-            break
-        end
-    end
-    while i <= m
-        r[ord[i]] = weightsum
-        i += 1
```
nalimilan, line 39: So all this code was redundant with calling `ecdf` on each point in a loop? :-/

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
     end
-    return r / weightsum
+    return res
 end
 
 """
-    ecdf(X; weights::AbstractWeights)
+    ecdf(X::RealVector; weights::AbstractWeights=nothing, interpolate::Bool=false) -> ECDF
+
+Construct an *empirical cumulative distribution function* (ECDF) based on a vector of samples (`X`).
```
nalimilan, line 66:
```suggestion
Construct an *empirical cumulative distribution function* (ECDF) based on vector of samples `X`.
```

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
+Returns a callable [`ECDF`](@ref) object.
 
-Return an empirical cumulative distribution function (ECDF) based on a vector of samples
-given in `X`. Optionally providing `weights` returns a weighted ECDF.
+# Example
 
-Note: this function that returns a callable composite type, which can then be applied to
-evaluate CDF values on other samples.
```
nalimilan, line 51: It would be nice to keep a mention about the fact that one needs to use `broadcast` here (even if that's also in examples), as it's not totally obvious.

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
-    weights::W
+# whether to interpolate between the neighboring ECDF values
+isinterpolating(::Type{<:ECDF{<:Any, <:Any, I}}) where I = I
+isinterpolating(ecdf::ECDF) = isinterpolating(typeof(ecdf))
+
+function Base.show(io::IO, e::ECDF)
+    print(io, typeof(e))
+    print(io, "(")
+    print(io, length(e.sorted_values))
+    println(io, " values)")
```
nalimilan, line 37:
```suggestion
    print(io, "(")
    print(io, length(e.sorted_values))
    println(io, " values)")
```
To avoid making it look as if values was and argument to the constructor:
```suggestion
    print(io, " with ")
    print(io, length(e.sorted_values))
    println(io, " values")
```

Also maybe "points" is better than "values"? I.e. it's not the number of values in the input but the number of unique values, right?

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
+ECDF{Int64,Float64,false}(15 values)
+
+julia> xcdf(9)
+0.6
+```
+
+## Keyword arguments
+* `weights`: optional weights for each `X` sample
+* `interpolate`: when enabled, `ECDF(y)` will linearly interpolate the cumulative density between
+  the neighboring `X` samples (``x_i \\leq y < x_{i+1}``),
```
nalimilan, line 87:
```suggestion
  the neighboring values in `X` (``x_i \\leq y < x_{i+1}`` with ``x`` sorted values in `X`),
```

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
+over which the ECDF is inside the interval ``(0, 1)``.
 """
-function ecdf(X::RealVector; weights::AbstractVector{<:Real}=Weights(Float64[]))
+function ecdf(X::RealVector;
+              weights::Union{Nothing, RealVector}=nothing,
+              interpolate::Bool=false)
     any(isnan, X) && throw(ArgumentError("ecdf can not include NaN values"))
-    isempty(weights) || length(X) == length(weights) || throw(ArgumentError("data and weight vectors must be the same size," *
-        "got $(length(X)) and $(length(weights))"))
+    evenweights = isnothing(weights) || isempty(weights)
```
nalimilan, line 99: Inference works better with `===`:
```suggestion
    evenweights = weights === nothing || isempty(weights)
```

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
+    evenweights || (length(X) == length(weights)) ||
+        throw(ArgumentError("data and weight vectors must be the same size, " *
+                            "got $(length(X)) and $(length(weights))"))
+    T = eltype(X)
+    W0 = evenweights ? Int : eltype(weights)
+    W = isnothing(weights) ? Float64 : eltype(one(W0)/sum(weights))
```
nalimilan, line 105: Store `sum(weights)` to avoid calling it twice?

src/empirical.jl
```diff
@@ -1,68 +1,142 @@
...
+    wsumprev = zero(W0)
+    valw = zero(W0)
+
+    push_valprev!() = push!(sorted_vals, (valprev, min(wsumprev/wsum, one(W)),
+                                          inv(val - valprev), valw/wsum))
+
+    @inbounds for i in ord
+        valnew = X[i]
+        if (val != valnew) || (i == last(ord))
+            (wsumprev > 0) && push_valprev!()
```
nalimilan, line 123: Doesn't this anonymous function trigger boxing of these variables, making the loop quite slow?

test/empirical.jl
```diff
@@ -23,17 +34,25 @@ end
     fnecdf = ecdf(x, weights=w1)
     fnecdfalt = ecdf(x, weights=w2)
     @test fnecdf.sorted_values == fnecdfalt.sorted_values
-    @test fnecdf.weights == fnecdfalt.weights
-    @test fnecdf.weights != w1  #  check that w wasn't accidently modified in place
-    @test fnecdfalt.weights != w2
+    @test_skip fnecdf.weights == fnecdfalt.weights
+    @test_skip fnecdf.weights != w1  #  check that w wasn't accidently modified in place
+    @test_skip fnecdfalt.weights != w2
```
nalimilan, line 39: ?

test/empirical.jl
```diff
@@ -1,6 +1,12 @@
...
 
+function showstring(obj)
+    iobuf = IOBuffer()
+    show(iobuf, obj)
+    return String(take!(iobuf))
+end
+
 @testset "ECDF" begin
     x = randn(10000000)
     fnecdf = ecdf(x)
```
nalimilan, line 12: I'd change all of these to `ecdf.(x)`, and copy this to deprecated.jl.

nalimilan (commented):
Sorry for the long delay. This looks nice. I can't comment on the math but here are remarks about the code.

ParadaCarleton:
Is there anything here that's still worth it, or should I close?

PR #592, opened 2020-08-18
Test on current release and fix test failures

andreasnoack:
We didn't test on 1.5 because we'd hardcoded the minor release numbers. Hence, we hadn't noticed that tests didn't pass with the current Julia release version (1.5). There were some changes to the RNG in Julia 1.5 which broke the tests. Hence, I've added a test dependency on StableRNGs to ensure reproducible test results across Julia versions. For one of the tests, I also increased `n`. It should have little effect on the total test time.


andreasnoack:
I've enabled Windows testing on Travis since it now seems to work well. @nalimilan could you please disable Appveyor (and review)?

nalimilan:
It would be nice to keep testing on 32-bit. I think you need to use matrix in travis.yml.

andreasnoack:
> It would be nice to keep testing on 32-bit.

I agree.

> I think you need to use matrix in travis.yml.

It looks like just specifying `arch` "globally" now works.

nalimilan:
OK. You need a custom code to test on 32-bit Windows, but I always forget whether it’s useful or not. @ararslan knows.

andreasnoack:
Isn't 32 bit Linux sufficient?

ararslan:
Yeah I think 32-bit Linux is sufficient. Support for 32-bit Windows was bumped down a tier for Julia itself anyway.

ararslan:
As an aside, Travis must have fixed the thing where unsupported `os`/`arch` combinations are automatically skipped instead of requiring explicit `matrix: exclude:` stuff. That's nice, but it seems to be skipping 32-bit Windows, which is fine here but seems like a bug in general.

nalimilan (approved):
OK, thanks. I’ve disabled AppVeyor, can you also drop the badge from README.md ?

EDIT: I don’t appear to have the permission to delete the AppVeyor project. Not a big deal but if somebody wants to try...

PR #593, opened 2020-08-19
Allow DataStructures 0.18

andreasnoack:
Supersedes #591


devmotion:
I guess it should be fine to make a StatsBase 0.33.1 release with this change?

andreasnoack:
Yes. I'm not sure what else is on master though.

devmotion:
It seems the only other changes on master are the replacement of RangeGenerator with Sampler and fixes of CI and docstrings: https://github.com/JuliaStats/StatsBase.jl/compare/v0.33.0...master

PR #595, opened 2020-08-24
Fix transformation `fit` call with default variables

### LINKED ISSUE #594: `UndefVarError: l not defined` when calling `fit` with `dims = nothing` using StatsBase v0.33.0
```julia
julia> using StatsBase

julia> fit(ZScoreTransform, rand(3, 3))
ERROR: UndefVarError: l not defined
Stacktrace:
 [1] fit(::Type{ZScoreTransform}, ::Array{Float64,2}; dims::Nothing, center::Bool, scale::Bool) at /home/alir/.julia/packages/StatsBase/unDUx/src/transformations.jl:126
 [2] fit(::Type{ZScoreTransform}, ::Array{Float64,2}) at /home/alir/.julia/packages/StatsBase/unDUx/src/transformations.jl:111
 [3] top-level scope at REPL[2]:1
```

According to the docs (https://juliastats.org/StatsBase.jl/stable/transformations/#StatsBase.fit-Tuple{Type{ZScoreTransform},AbstractArray{#s4,2}%20where%20#s4%3C:Real}),

> The default is nothing, which is equivalent to dims=2 with a deprecation warning.

so either the docs need updating or `dims` should be made a required keyword argument.

wildart:
Fix #594

### DIFF
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -129,10 +129,8 @@ function fit(::Type{ZScoreTransform}, X::AbstractMatrix{<:Real};
 end
 
 function fit(::Type{ZScoreTransform}, X::AbstractVector{<:Real};
-             dims::Union{Integer,Nothing}=nothing, center::Bool=true, scale::Bool=true)
-    if dims == nothing
-        Base.depwarn("fit(t, x) is deprecated: use fit(t, x, dims=2) instead", :fit)
-    elseif dims != 1
+             dims::Integer=1, center::Bool=true, scale::Bool=true)
+    if dims != 1
         throw(DomainError(dims, "fit only accepts dims=1 over a vector. Try fit(t, x, dims=1)."))
     end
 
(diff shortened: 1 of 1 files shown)


kleinschmidt:
I wonder if we should back port this to the 0.32 series too (as with #565)

wildart:
I made `dims=1` to be the default value for vectors. Using `nothing` doesn't make sense in such case.

kleinschmidt:
Actually I think it's better to leave the default value as `nothing`, unless we want to make the `dims=` kwarg optional for vectors.

kleinschmidt:
The logic being that eventually (aka on the next breaking change release) there will be an error if `dims` is not specified.

wildart:
> there will be an error if dims is not specified.

After all, there is only one dimension in a vector. In future, this will be a default argument with value `1`. It is not that, we cannot figure out the parameter type, so the `dims` is required. We still rely on multiple dispatch. I understand necessity of the dimension selection for a matrix argument, but not for a vector. `mean` doesn't require `dims` parameter for the vectors nor many functions.  Why should any function operating on vectors require this parameter?

nalimilan:
For consistencty with Statistics, where `cov` and `cor` treat columns as variables, it would make sense to switch the default to `dims=1` at some point after removing the deprecation. Then the default would also be correct for vectors. I'm hesitant about whether we should make this change for vectors now in anticipation of the general switch or not.

wildart:
The functions with the matrix parameter have the warnings about `dims` parameter usage. But is this rely necessary to inform user use `dims=1` parameter with the vector input? Will `dims` be a positional non-default parameter?

nalimilan:
No, `dims` will remain a keyword argument AFAICT, that's consistent with Base, Statistics and Distances. I don't have a strong opinion about vectors, so let's hear what others think.

kleinschmidt:
If `dims=1` is going to be the default everywhere I have no problem making it the default for vectors as well; it's the only sensible choice for vectors so there's no sense in requiring users specify it.  It's not even clear to me that having it as a kwarg is even necessary except for consistency with the matrix methods

kleinschmidt:
Also just so I'm clear on the overall plan here: for v0.34 we'll make `dims=1` the default and remove the `dims=nothing` path, correct?

nalimilan:
Maybe first remove the deprecation, and wait for a release before switching the default?

Issue #409, opened 2018-09-07
var() and cov() give inconsistent values for weighted variance in Julia >= 0.6


yanncalec:
Let `x = 1:100` and `w=rand(100)` for example. Then `StatsBase.var(x)==StatsBase.cov(x)` is true, however the weighted variance `StatsBase.var(x, StatsBase.weights(w))` and `StatsBase.cov(x, StatsBase.weights(w))` are very different.

nalimilan:
The apparent inconsistency between `cov` and `var` is due to a dispatch issue: `w` is interpreted as the second array, not as a weights vector. We need to ensure the method taking weights is more specific than the fallback.

The inconsistency with the `dims` arguments is https://github.com/JuliaStats/StatsBase.jl/issues/408, combined with the dispatch problem.

ararslan:
A bit tangential, but ideally I'd like to see weights specified as a keyword argument. That would require defining it in the Statistics stdlib package though.

nalimilan:
Actually I don't think there's a bug here: `x` is a vector, and we only provide methods for matrices. Indeed the covariance of a vector doesn't make a lot of sense. The problem is just that the call treats `w` as the second vector rather than as weights. The best fix to that is to move these functions to the stdlib and make weights a keyword argument.

yanncalec:
@nalimilan : Yes I can confirm `cov` gives correct reweighted covariance matrix when applied on matrices, and it's a bit confusing of taking a vector as input. Also it would be nice if `cov` can compute the reweighted covariance between two different matrices as follows
```julia
(X.-mean(X,w,1))' * ((Y.-mean(Y,w,1)) .* w) / sum(w)
```

baggepinnen:
It is at least a big issue that this fails silently now, I just ran into it as well. Since `cov` is defined for a vector,   it would be nice to have `cov(v,w)` act similarily, or at least throw an error.

nalimilan:
Will be fixed by https://github.com/JuliaLang/julia/pull/31395.

MaximilianJHuber:
Is the above pull request a remedy for this issue too, or is is a different issue?
```
using Statistics, StatsBase
G = rand(10000, 4); 
StatsBase.cov(G, weights(rand(10000).^2)) # fine!
StatsBase.cov(@view(G[:,:]), weights(rand(10000).^2)) # yields only a column vector!
```
The `@view` changes behavior and yields a wrong object!

[...]

PR #599, opened 2020-09-13
Covariance-related functions for general AbstractArray

yha:
This is mostly changing `DenseMatrix` in method signatures to `AbstractMatrix`, and running the tests also on a sparse array and a custom-typed array.
Other changes:
- Specialized `_symmetrize!` for sparse matrices
- Fixed bug with zero variances in `cor2cov!` discovered when testing with sparse matrix
- Added `mean_and_cov(vector)`. I didn't add `mean_and_cov(vector, weights)` since that would need a fix to #409, which has a long discussion pointing to JuliaLang/Statistics.jl#2


src/cov.jl
```diff
@@ -110,19 +120,24 @@ end
...
     return m, covm(x, m, dims, corrected=corrected)
 end
-function mean_and_cov(x::DenseMatrix, wv::AbstractWeights, dims::Int=1;
+function mean_and_cov(x::AbstractMatrix, wv::AbstractWeights, dims::Int=1;
                       corrected::DepBool=nothing)
     m = mean(x, wv, dims=dims)
     return m, cov(x, wv, dims; corrected=depcheck(:mean_and_cov, corrected))
 end
 
+function mean_and_cov(x::AbstractVector; corrected::Bool=true)
```
nalimilan, line 136: Put this one above with the corresponding `AbstractMatrix` method?

test/cov.jl
```diff
@@ -87,27 +100,32 @@ weight_funcs = (weights, aweights, fweights, pweights)
         @testset "Mean and covariance" begin
             (m, C) = mean_and_cov(X; corrected=false)
             @test m == mean(X, dims=1)
-            @test C == cov(X, dims=1, corrected=false)
+            @test C ≈ cov(X, dims=1, corrected=false)
```
nalimilan, line 103: Any reason to use `≈` rather than `isequal`?

test/cov.jl
```diff
@@ -119,11 +137,12 @@ weight_funcs = (weights, aweights, fweights, pweights)
             cor1 = cor(X, wv1, 1)
             cor2 = cor(X, wv2, 2)
 
+
```
nalimilan, line 140:
```suggestion
```

src/cov.jl
```diff
@@ -148,8 +163,10 @@ standard deviations `s`.
 function cor2cov!(C::AbstractMatrix, s::AbstractArray)
     n = length(s)
     size(C) == (n, n) || throw(DimensionMismatch("inconsistent dimensions"))
-    for i in CartesianIndices(size(C))
-        @inbounds C[i] *= s[i[1]] * s[i[2]]
+    @inbounds for i in CartesianIndices(size(C))
+        si = s[i[1]] * s[i[2]]
+        # the covariance is 0 when si==0, although C[i] is NaN in this case
+        C[i] = iszero(si) ? zero(eltype(C)) : C[i] * si
```
nalimilan, line 169: In case `eltype(C)` isn't a concrete type:
```suggestion
        Ci = C[i]
        C[i] = iszero(si) ? zero(Ci) : Ci * si
```
Can you explain a bit more what happens here?

nalimilan (commented):
Thanks!

test/cov.jl
yha: Strict equality fails for sparse matrices because `cov` is specialized for `SparseMatrixCSC` in `Statistics` (https://github.com/JuliaLang/Statistics.jl/blob/b384104d35ff0e7cf311485607b177223ed72b9a/src/Statistics.jl#L1058), but `mean_and_cov` uses `covm` rather than `cov`.
I think if we want to achieve strict equality here, the fix should be in `Statistics` (specializing `covm` rather than `cov`)

test/cov.jl
yha: done

src/cov.jl
yha: If one of the variables has zero variance, its covariance with any other variable is 0, but the correlation is undefined, and the result of `cor` would have `NaN` in that element. In that case `Ci == NaN`, `si == 0`, but the correct output is 0, not `NaN`.
For example, 
```
X = [1:3 ones(3)]
cov(X) == [1 0; 0 0]
cor(X) == [1 NaN; NaN 1]
cor2cov(cor(X), std.(eachcol(X))) # previously [1 NaN; NaN 0], which is incorrect
```

yha:
Rebased

yha:
bump

PR #605, opened 2020-10-02
Make countmap support iterators

### LINKED ISSUE #581: `countmap` should accept any iterator
```
julia> x = [1, 2, 3]
3-element Array{Int64,1}:
 1
 2
 3

julia> countmap(skipmissing(x))
ERROR: MethodError: no method matching countmap(::Base.SkipMissing{Array{Int64,1}})
```

Moelf:
fix #581

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -390,7 +390,6 @@ countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(),
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
 countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :dict)
 countmap(x, wv::AbstractVector{W}) where {W<:Real} = addcounts!(Dict{eltype(x),W}(), collect(x), wv)
-countmap(x, wv) = countmap(collect(x), collect(wv))
 
 
 """
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -288,7 +288,7 @@ end
 # faster results and less memory usage. However we still wish to enable others
 # to write generic algorithms, therefore the methods below still accept the 
 # `alg` argument but it is ignored.
-function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
+function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x; alg = :ignored)
     sumx = sum(x)
     cm[true] = get(cm, true, 0) + sumx
     cm[false] = get(cm, false, 0) + length(x) - sumx
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -395,7 +395,7 @@ of occurrences.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
-countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :auto)
+countmap(x; alg = :auto) = addcounts!(Dict{eltype(x),Int}(), x; alg = alg)
 
 
 """
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -255,10 +255,10 @@ raw counts.
 - `:dict`:           use `Dict`-based method which is generally slower but uses less
                      RAM and is safe for any data type.
 """
-function addcounts!(cm::Dict{T}, x; alg = :auto) where T
+function addcounts!(cm::Dict, x; alg = :auto)
     # if it's safe to be sorted using radixsort then it should be faster
     # albeit using more RAM
-    if radixsort_safe(T) && (alg == :auto || alg == :radixsort)
+    if radixsort_safe(eltype(x)) && (alg == :auto || alg == :radixsort)
         addcounts_radixsort!(cm, x)
     elseif alg == :radixsort
         throw(ArgumentError("`alg = :radixsort` is chosen but type `radixsort_safe($T)` did not return `true`; use `alg = :auto` or `alg = :dict` instead"))
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -295,7 +295,19 @@ function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x::AbstractArray{Bool}; alg =
     cm[false] = get(cm, false, 0) + length(x) - sumx
     cm
 end
-_addcounts!(::Type{Bool}, cm::Dict{Bool}, x; alg = :ignored) = _addcounts!(Bool, cm, collect(x); alg = alg)
+
+#speailized for `Bool` iterator
+function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x; alg = :ignored)
+    sumx = 0
+    len = 0
+    for i in x
+        sumx += i
+        len += 1
+    end
+    cm[true] = get(cm, true, 0) + sumx
+    cm[false] = get(cm, false, 0) + len - sumx
+    cm
+end
 
 function _addcounts!(::Type{T}, cm::Dict{T}, x; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
     counts = zeros(Int, 2^(8sizeof(T)))
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -391,7 +391,6 @@ of occurrences.
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
 countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :auto)
-countmap(x, wv::AbstractVector{W}) where {W<:Real} = addcounts!(Dict{eltype(x),W}(), collect(x), wv)
 
 
 """
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -255,7 +255,7 @@ raw counts.
 - `:dict`:           use `Dict`-based method which is generally slower but uses less
                      RAM and is safe for any data type.
 """
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :auto) where T
+function addcounts!(cm::Dict{T}, x; alg = :auto) where T
     # if it's safe to be sorted using radixsort then it should be faster
     # albeit using more RAM
     if radixsort_safe(T) && (alg == :auto || alg == :radixsort)
@@ -269,7 +269,7 @@ function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :auto) where T
 end
 
 """Dict-based addcounts method"""
-function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
+function addcounts_dict!(cm::Dict{T}, x) where T
     for v in x
         index = ht_keyindex2!(cm, v)
         if index > 0
@@ -388,6 +388,9 @@ of occurrences.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
+countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :dict)
+countmap(x, wv::AbstractVector{W}) where {W<:Real} = addcounts!(Dict{eltype(x),W}(), collect(x), wv)
+countmap(x, wv) = countmap(collect(x), collect(wv))
 
 
 """
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -255,10 +255,12 @@ raw counts.
 - `:dict`:           use `Dict`-based method which is generally slower but uses less
                      RAM and is safe for any data type.
 """
-function addcounts!(cm::Dict, x; alg = :auto)
+addcounts!(cm::Dict, x; alg = :auto) = _addcounts!(eltype(x), cm, x)
+# manual dispatch for `x` being iterator
+function _addcounts!(::Type{T}, cm::Dict{T}, x; alg = :auto) where T
     # if it's safe to be sorted using radixsort then it should be faster
     # albeit using more RAM
-    if radixsort_safe(eltype(x)) && (alg == :auto || alg == :radixsort)
+    if radixsort_safe(T) && (alg == :auto || alg == :radixsort)
         addcounts_radixsort!(cm, x)
     elseif alg == :radixsort
         throw(ArgumentError("`alg = :radixsort` is chosen but type `radixsort_safe($T)` did not return `true`; use `alg = :auto` or `alg = :dict` instead"))
@@ -286,14 +288,14 @@ end
 # faster results and less memory usage. However we still wish to enable others
 # to write generic algorithms, therefore the methods below still accept the 
 # `alg` argument but it is ignored.
-function addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
+function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
     sumx = sum(x)
     cm[true] = get(cm, true, 0) + sumx
     cm[false] = get(cm, false, 0) + length(x) - sumx
     cm
 end
 
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
+function _addcounts!(::Type{T}, cm::Dict{T}, x::AbstractArray{T}; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
     counts = zeros(Int, 2^(8sizeof(T)))
 
     @inbounds for xi in x
@@ -394,15 +396,8 @@ of occurrences.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
-# fall-back for iterator `x`
-function countmap(x)
-    if eltype(x) <: Union{Bool, UInt8, UInt16, Int8, Int16}
-        # faster `addcounts!` specialized
-        addcounts!(Dict{eltype(x),Int}(), collect(x); alg = :auto)
-    else
-        addcounts!(Dict{eltype(x),Int}(), x; alg = :auto)
-    end
-end
+countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :auto)
+
 
 """
     proportionmap(x)
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -351,8 +351,12 @@ function addcounts_radixsort!(cm::Dict{T}, x::AbstractArray{T}) where T
     # It seems that sort is inferred in Julia 0.7.
     return _addcounts_radix_sort_loop!(cm, sx)
 end
-# general iterator form
-addcounts_radixsort!(cm::Dict{T}, x) where T = addcounts_radixsort!(cm, collect(x))
+
+# fall-back for `x` an iterator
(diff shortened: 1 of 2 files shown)


src/counts.jl
```diff
@@ -255,7 +255,7 @@ raw counts.
 - `:dict`:           use `Dict`-based method which is generally slower but uses less
                      RAM and is safe for any data type.
 """
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :auto) where T
+function addcounts!(cm::Dict{T}, x; alg = :auto) where T
     # if it's safe to be sorted using radixsort then it should be faster
     # albeit using more RAM
     if radixsort_safe(T) && (alg == :auto || alg == :radixsort)
```
nalimilan, line 261:
```suggestion
function addcounts!(cm::Dict, x; alg = :auto)
    # if it's safe to be sorted using radixsort then it should be faster
    # albeit using more RAM
    if radixsort_safe(eltype(x)) && (alg == :auto || alg == :radixsort)
```

nalimilan (commented):
Thanks!

src/counts.jl
```diff
@@ -388,6 +388,9 @@ of occurrences.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
+countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :dict)
+countmap(x, wv::AbstractVector{W}) where {W<:Real} = addcounts!(Dict{eltype(x),W}(), collect(x), wv)
+countmap(x, wv) = countmap(collect(x), collect(wv))
```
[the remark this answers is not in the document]
mschauer: It is also a case of https://docs.julialang.org/en/v1/manual/style-guide/#Handle-excess-argument-diversity-in-the-caller, the caller should collect.

src/counts.jl
```diff
@@ -388,6 +388,9 @@ of occurrences.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
+countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :dict)
```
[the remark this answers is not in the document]
mschauer: Should the default be auto or dict?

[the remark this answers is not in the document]
nalimilan: Ah right it should be `:auto`. I guess @Moelf used `:dict` because the PR doesn't currently extend `:dict` methods to general iterables. Is there a reason why this wouldn't be possible?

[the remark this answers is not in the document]
Moelf: the auto will land in this line where `length` is needed:
https://github.com/JuliaStats/StatsBase.jl/blob/983bf021be7521c399dfe4d25d174667344bc8fb/src/counts.jl#L358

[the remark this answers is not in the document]
nalimilan: This method is only used when weights are provided AFAICT (and I said I don't think we should allow passing two non-vector iterators in that case).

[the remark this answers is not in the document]
nalimilan: That said, `addcounts_radixsort` calls `sort`, which only works on vectors. But you could do `sort!(collect(x))` for other iterators.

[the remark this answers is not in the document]
Moelf: @nalimilan makes sense. I already removed the two non-vector methods. Now switched back to `:auto` by default.

src/counts.jl
```diff
@@ -352,6 +351,9 @@ function addcounts_radixsort!(cm::Dict{T}, x::AbstractArray{T}) where T
     # It seems that sort is inferred in Julia 0.7.
     return _addcounts_radix_sort_loop!(cm, sx)
 end
+# general iterator form
+addcounts_radixsort!(cm::Dict{T}, x) where T = addcounts_radixsort!(cm, collect(x))
```
mschauer, line 355: That's fine, because you basically collect into the workspace imho.

mschauer (approved):

src/counts.jl
nalimilan: But this is going to make an unnecessary copy. I think the best solution is to make the method above accept any iterator, and do:
```julia
cx = x isa AbstractArray ? copy(x) : collect(x)
sx = sort!(cx, alg = RadixSort)
```
(`collect` sometimes returns an `Array` when `copy` would be more efficient, e.g. for `BitArray`)

src/counts.jl
```diff
@@ -388,6 +390,8 @@ of occurrences.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
+countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :auto)
+countmap(x, wv::AbstractVector{W}) where {W<:Real} = addcounts!(Dict{eltype(x),W}(), collect(x), wv)
```
nalimilan, line 394: As I said currently the policy followed across the package is that we only provide weighted methods when both inputs are vectors. Better stick to that.
```suggestion
```

src/counts.jl
```diff
@@ -269,7 +269,7 @@ function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :auto) where T
 end
 
 """Dict-based addcounts method"""
-function addcounts_dict!(cm::Dict{T}, x::AbstractArray{T}) where T
+function addcounts_dict!(cm::Dict{T}, x) where T
```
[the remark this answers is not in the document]
nalimilan: Can you also update `addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)` and `addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}`?

Moelf:
@nalimilan isn't the non-copy method defined on 345L for `AbstractArray`? what's the problem of using collect here

src/counts.jl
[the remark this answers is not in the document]
Moelf: why those two need to be updated?

src/counts.jl
mschauer: Is there an inplace radix sort?

Moelf: https://github.com/Moelf/StatsBase.jl/blob/6e40ec06a1532eaddf6b3e86e9a8c7e01e8aa323/src/counts.jl#L345

it doesn't make unnecessary copy because this is only fall back when `x` is not an array. so instead of doing that manual dispatch on `cx=`, this looks fine to me

nalimilan:
> @nalimilan isn't the non-copy method defined on 345L for `AbstractArray`? what's the problem of using collect here

`sort` makes a copy, that's why I proposed using `sort!`.

src/counts.jl
nalimilan: Yes actually it's [defined in terms of `sort!`](https://github.com/JuliaCollections/SortingAlgorithms.jl/blob/11e7bf999847b39a651db86d7bebca93612f4f50/src/SortingAlgorithms.jl#L64), and `sort` just calls `sort!` on a copy.

nalimilan: When you call `addcounts_radixsort!(cm, collect(x))`, you allocate a new vector, and then inside `addcounts_radixsort!` `sort` makes another copy. This is redundant.

Moelf:
@nalimilan done. Last thing, what needs to be changed for `addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)` etc. ?

nalimilan:
Yes that's would be nice too, as it gives higher performance for these cases.

Moelf:
sorry, I just don't know what "change" were you referring to. They look like different specialized function to handle different `T` again.

nalimilan:
I mean that they should accept any `x` of the right `eltype`, not just `AbstractArray`. That requires a few dispatch tricks by passing `eltype(x)` and an argument and dispatching on that.

Moelf:
maybe we should just collect once at the very beginning at let the rest do their own thing....

Moelf:
@nalimilan how about this?

src/counts.jl
```diff
@@ -388,7 +394,15 @@ of occurrences.
 """
 countmap(x::AbstractArray{T}; alg = :auto) where {T} = addcounts!(Dict{T,Int}(), x; alg = alg)
 countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(Dict{T,W}(), x, wv)
-
+# fall-back for iterator `x`
+function countmap(x)
```
nalimilan, line 398: The problem with this approach is that if the user calls `addcounts!` directly (which is part of the public API), the fast method won't be used. I think the solution is to have `addcounts!(cm::Dict, x; alg = :ignored) = _addcounts!(eltype(x), cm, x)` and turn all existing `addcounts!` methods into `_addcounts!`, dispatching on the eltype.

Moelf: the problem is dispatch doesn't work on `alg=:ignored` keyword. idk, maybe for a different issue?

nalimilan: I think `alg=:ignored` is just used to make it clear that the value is ignored, but you can pass any value as it's not used. The idea is that for types for which we know a solution is clearly faster, we don't respect the user's request.

src/counts.jl
```diff
@@ -255,7 +255,9 @@ raw counts.
 - `:dict`:           use `Dict`-based method which is generally slower but uses less
                      RAM and is safe for any data type.
 """
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :auto) where T
+addcounts!(cm::Dict, x; alg = :auto) = _addcounts!(eltype(x), cm, x)
```
nalimilan, line 258:
```suggestion
addcounts!(cm::Dict, x; alg = :auto) = _addcounts!(eltype(x), cm, x, alg = alg)

```

src/counts.jl
```diff
@@ -255,7 +255,9 @@ raw counts.
...
+# manual dispatch for `x` being iterator
+function _addcounts!(::Type{T}, cm::Dict{T}, x; alg = :auto) where T
```
nalimilan, line 260:
```suggestion
function _addcounts!(::Type{T}, cm::Dict, x; alg = :auto) where T
```

src/counts.jl
```diff
@@ -286,14 +288,14 @@ end
 # faster results and less memory usage. However we still wish to enable others
 # to write generic algorithms, therefore the methods below still accept the 
 # `alg` argument but it is ignored.
-function addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
+function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
```
nalimilan, line 291:
```suggestion
function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x; alg = :ignored)
```

src/counts.jl
```diff
@@ -286,14 +288,14 @@ end
...
     sumx = sum(x)
     cm[true] = get(cm, true, 0) + sumx
     cm[false] = get(cm, false, 0) + length(x) - sumx
     cm
 end
 
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
+function _addcounts!(::Type{T}, cm::Dict{T}, x::AbstractArray{T}; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
```
nalimilan, line 298:
```suggestion
function _addcounts!(::Type{T}, cm::Dict{T}, x; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
```

src/counts.jl
```diff
@@ -394,15 +396,8 @@ of occurrences.
...
-# fall-back for iterator `x`
-function countmap(x)
-    if eltype(x) <: Union{Bool, UInt8, UInt16, Int8, Int16}
-        # faster `addcounts!` specialized
-        addcounts!(Dict{eltype(x),Int}(), collect(x); alg = :auto)
-    else
-        addcounts!(Dict{eltype(x),Int}(), x; alg = :auto)
-    end
-end
+countmap(x) = addcounts!(Dict{eltype(x),Int}(), x; alg = :auto)
```
nalimilan, line 399:
```suggestion
countmap(x; alg = :auto) = addcounts!(Dict{eltype(x),Int}(), x; alg = alg)
```
Also, replace `countmap(x::AbstractArray{T}; alg = :auto) ` with this method as it works for both.

src/counts.jl
```diff
@@ -353,6 +354,13 @@ function addcounts_radixsort!(cm::Dict{T}, x::AbstractArray{T}) where T
     return _addcounts_radix_sort_loop!(cm, sx)
 end
 
+# fall-back for `x` an iterator
+function addcounts_radixsort!(cm::Dict{T}, x) where T 
+    sx = sort!(collect(x), alg = RadixSort)
+    return _addcounts_radix_sort_loop!(cm, sx)
+end
+
```
nalimilan, line 362:
```suggestion
```

test/counts.jl
```diff
@@ -102,14 +102,24 @@ StatsBase.addcounts_radixsort!(cm,xx2)
...
+xx_missing = skipmissing([missing, "b", "a", "a", "b", "c"])
+cm_missing = countmap(xx_missing)
+
+@test cm_missing["a"] == 2
+@test cm_missing["b"] == 2
+@test cm_missing["c"] == 1
+
 @test cm == countmap(x, w)
 
 pm = proportionmap(x, weights(w))
```
nalimilan, line 125: Can you test general iterators below for `Bool` and small integer types too?

Moelf:
done. Manually adding special case for `x` that doesn't have length is quite annoying, maybe a re-factor is called....

src/counts.jl
```diff
@@ -286,14 +289,15 @@ end
...
 # to write generic algorithms, therefore the methods below still accept the 
 # `alg` argument but it is ignored.
-function addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
+function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
     sumx = sum(x)
     cm[true] = get(cm, true, 0) + sumx
     cm[false] = get(cm, false, 0) + length(x) - sumx
     cm
 end
+_addcounts!(::Type{Bool}, cm::Dict{Bool}, x; alg = :ignored) = _addcounts!(Bool, cm, collect(x); alg = alg)
```
nalimilan, line 298: Again this is going to allocate a copy that could be avoided. I can see two solutions:
- Extend the signature of the method below to `T <: Union{Bool, UInt8, UInt16, Int8, Int16}`. It seems to work fine, it's just slightly suboptimal as `counts` will be of length 256 -- but that's not a big deal.
- Copy the approach used by `_addcounts!(::Type{Bool}, cm::Dict{Bool}, x::AbstractArray{Bool})` above, but without using `length`. A simple `for` loop should be enough.

Moelf: if we're doing loops anyways, can't we just let it fall-back to `addcounts_dict!`?

src/counts.jl
```diff
@@ -286,14 +289,27 @@ end
...
-function addcounts!(cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
+function _addcounts!(::Type{Bool}, cm::Dict{Bool}, x::AbstractArray{Bool}; alg = :ignored)
     sumx = sum(x)
     cm[true] = get(cm, true, 0) + sumx
     cm[false] = get(cm, false, 0) + length(x) - sumx
     cm
 end
 
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :ignored) where T <: Union{UInt8, UInt16, Int8, Int16}
+#speailized for `Bool` iterator
```
nalimilan, line 299:
```suggestion
# specialized for `Bool` iterator
```

src/counts.jl
```diff
@@ -255,7 +255,10 @@ raw counts.
 - `:dict`:           use `Dict`-based method which is generally slower but uses less
                      RAM and is safe for any data type.
 """
-function addcounts!(cm::Dict{T}, x::AbstractArray{T}; alg = :auto) where T
+addcounts!(cm::Dict, x; alg = :auto) = _addcounts!(eltype(x), cm, x, alg = alg)
+
+# manual dispatch for `x` being iterator
```
nalimilan, line 260:
```suggestion
```
(this also applies to arrays -- anyway it's clear enough I think)

test/counts.jl
```diff
@@ -102,14 +102,24 @@ StatsBase.addcounts_radixsort!(cm,xx2)
...
 @test cm["a"] == 5.5
 @test cm["b"] == 4.5
 @test cm["c"] == 3.5
 
+xx_missing = skipmissing([missing, "b", "a", "a", "b", "c"])
+cm_missing = countmap(xx_missing)
+
+@test cm_missing["a"] == 2
+@test cm_missing["b"] == 2
+@test cm_missing["c"] == 1
```
nalimilan, line 121: Sorry, two last things with tests:
- This should be put above, with the equivalent test for vectors. Here it's in the middle of tests for the weighted variant.

- Also, could you check that `cm_missing isa Dict{String, Int}`, and that something like `countmap((i for i in x)` gives a `Dict{Any,Int}` instead? That's maybe not ideal (`String` would be better in both cases), but in terms of implementation we would have to call `collect` on the iterator first to find a more specific type when the iterator doesn't have one. Anyway, better test this kind of detail as end users can end up relying on it and we shouldn't change it later without noticing.

src/counts.jl
nalimilan: A dict lookup is much slower than an addition. TBH I'm not sure how common it is to call `countmap` on a vector of `Bool`, but since we provide this optimization for arrays...

test/counts.jl
```diff
@@ -90,6 +90,17 @@ pm = proportionmap(x)
...
+xx_missing = skipmissing(xx)
+cm_missing = countmap(xx_missing)
+@test cm_missing isa Dict{Int, Int}
+@test cm_missing == cm
+
+cm_any_itr = countmap((i for i in xx)) 
+@test cm_any_itr[1] == 200_000
+@test cm_any_itr[3] == 100_000
+@test cm_any_itr[6] == 100_000
+@test cm_any_itr isa Dict{Any,Int} #no knowledge about type
```
nalimilan, line 103: Would this work too?
```suggestion
@test cm_any_itr isa Dict{Any,Int} # no knowledge about type
@test cm_missing == cm
```

test/counts.jl
```diff
@@ -90,6 +90,17 @@ pm = proportionmap(x)
 xx = repeat([6, 1, 3, 1], outer=100_000)
```
nalimilan, line 90: Above, should test the general case with something like you did for `addcounts!`:
```julia
cm_missing = countmap(skipmissing(x))
@test cm_missing isa Dict{Int, Int}
@test cm_missing == cm

cm_any_itr = countmap((i for i in xx)) 
@test cm_any_itr isa Dict{Any,Int} #no knowledge about type
@test cm_any_itr == cm
```

test/counts.jl
```diff
@@ -90,6 +90,17 @@ pm = proportionmap(x)
...
 cm = countmap(xx)
 @test cm == Dict(1 => 200_000, 3 => 100_000, 6 => 100_000)
+#with iterator
+xx_missing = skipmissing(xx)
+cm_missing = countmap(xx_missing)
```
nalimilan, line 95:
```suggestion

# with iterator
cm_missing = countmap(skipmissing(xx))
```

test/counts.jl
Moelf: no because `cm`'s key has type info, so not `Any`?

Moelf: nvm, it does work

test/counts.jl
Moelf: not sure what you mean here.

nalimilan: I'm suggesting adding this code above, as currently you don't test `countmap` on iterators.

Moelf: aren't they on L95-L101?

nalimilan: These only cover the radix sort method.

test/counts.jl
```diff
@@ -80,6 +80,13 @@ cm = countmap(x)
 @test cm["a"] == 3
 @test cm["b"] == 2
 @test cm["c"] == 1
+
+# iterator, non-radixsort
+cm_missing = countmap(skipmissing(x))
+cm_itr = countmap((i for i in x))
+@test cm_missing == cm
+@test cm_itr == cm
```
nalimilan, line 88:
```suggestion
# iterator, non-radixsort
cm_missing = countmap(skipmissing(x))
cm_any_itr = countmap((i for i in x))
@test cm_missing == cm_any_itr == cm
@test cm_missing isa Dict{String, Int}
@test cm_any_itr isa Dict{String, Int}
```

nalimilan:
Thanks @Moelf!

PR #606, opened 2020-10-04
Fix tests on Julia 1.6

nalimilan:
`@printf(stdout,"<1e%2.2d", -4)` now correctly gives `"<1e-04"`, which allows the values to be always aligned (the column has always a 6-character width).

### DIFF
--- a/test/statmodels.jl
+++ b/test/statmodels.jl
@@ -30,7 +30,11 @@ m = rand(3,4)
 
 @test sprint(show, StatsBase.PValue(1.0)) == "1.0000"
 @test sprint(show, StatsBase.PValue(1e-1)) == "0.1000"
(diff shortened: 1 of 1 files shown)


mschauer (approved):

kleinschmidt:
This is causes test in StatsModels and GLM (and probably others) to break on 1.6.  Here's an example of one that's failing:

https://github.com/JuliaStats/StatsModels.jl/blob/0ba8a8826e84883e7281ef033fe68d5f861837f0/test/statsmodel.jl#L279-L286

These tests all expect the p value to be printed like `<1e-4`.

kleinschmidt:
Oh I see you've only changed the TEST here, not the actual printing.  I guess my question is why the printing changed in the first place?

mschauer:
Ref https://github.com/JuliaLang/julia/pull/32859

PR #608, opened 2020-10-09
Update empirical.md with link to example notebook

logankilpatrick:
CC @alanedelman


nalimilan:
I'm not a fan of pointing to an external notebook which is not even a stable published document. Since this shows Julia code, it would better be included directly in our manual.

PR #612, opened 2020-10-22
Add a no-op PValue(::PValue) constructor

ararslan:
`show(::CoefTable)` calls `PValue` on all values in the p-value column (if present). If things are constructed manually and you happen to already have your p-values as `PValue`s, printing the table will fail. It seems sort of silly to have this kind of definition, but it's nice to not get errors when trying to look at your `CoefTable` in the REPL.

cc @palday

### DIFF
--- a/test/statmodels.jl
+++ b/test/statmodels.jl
@@ -39,7 +39,7 @@ end
 @test sprint(show, PValue(NaN)) == "NaN"
 @test_throws ErrorException PValue(-0.1)
 @test_throws ErrorException PValue(1.1)
-@test PValue(PValue(PValue(0.05))) == PValue(0.05)
+@test PValue(PValue(0.05)) === PValue(0.05)
 
 @test sprint(showerror, ConvergenceException(10)) == "failure to converge after 10 iterations."
 
--- a/test/statmodels.jl
+++ b/test/statmodels.jl
@@ -1,4 +1,5 @@
 using StatsBase
+using StatsBase: PValue
 using Test, Random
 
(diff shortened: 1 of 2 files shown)


test/statmodels.jl
```diff
@@ -28,16 +29,17 @@ m = rand(3,4)
...
-    @test sprint(show, StatsBase.PValue(1e-5)) == "<1e-4"
+    @test sprint(show, PValue(1e-5)) == "<1e-4"
 end
-@test sprint(show, StatsBase.PValue(NaN)) == "NaN"
-@test_throws ErrorException StatsBase.PValue(-0.1)
-@test_throws ErrorException StatsBase.PValue(1.1)
+@test sprint(show, PValue(NaN)) == "NaN"
+@test_throws ErrorException PValue(-0.1)
+@test_throws ErrorException PValue(1.1)
+@test PValue(PValue(PValue(0.05))) == PValue(0.05)
```
nalimilan, line 42: Two sounds enough? :-D
```suggestion
@test PValue(PValue(0.05)) === PValue(0.05)
```

nalimilan (approved):

test/statmodels.jl
mschauer: More “importantly” PValue(PValue(0.05)) === PValue(0.05) implies PValue(PValue(PValue(0.05))) === PValue(0.05) but not vice versa :-D

ararslan: Three is so much more fun though :P

ararslan:
Windows CI failure is unrelated.

PR #614, opened 2020-10-29
add empty! for histogram

Moelf:
compare:
```
julia> @btime zero(h) setup=(h = fit(Histogram, rand(10^4), 0:0.01:1))
  50.156 ns (2 allocations: 976 bytes)
```

with empty:
```
julia> @btime empty!(h) setup=(h = fit(Histogram, rand(10^4), 0:0.01:1))
  12.697 ns (0 allocations: 0 bytes)
```

tests included.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -538,6 +538,15 @@ set to zero.
 Base.zero(h::Histogram{T,N,E}) where {T,N,E} =
     Histogram{T,N,E}(deepcopy(h.edges), zero(h.weights), h.closed, h.isdensity)
 
+"""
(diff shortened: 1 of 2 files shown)


nalimilan:
`empty!` is defined to remove all elements in a collection. That's not really what happens here, which is more similar to a hypothetical `zero!` function, right? Also if we added `empty!` we could also have to add `empty`.

Maybe we could use `hist .= 0` instead?

Moelf:
I like `.= 0` actually, we can extend it by saying "indexing a histogram is to index its bins" but I'm not sure if that's really intuitive

PR #618, opened 2020-11-26
fix float array of equal values

### LINKED ISSUE #616: Incorrect Histogram fit for arrays of equal Float64 values
```
julia> using StatsBase

julia> fit(Histogram, [1 for _ in 1:1000])
Histogram{Int64,1,Tuple{StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}}
edges:
  1.0:1.0:2.0
weights: [1000]
closed: left
isdensity: false

julia> fit(Histogram, [1.234 for _ in 1:1000])
Histogram{Int64,1,Tuple{StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}}
edges:
  2.234:2.0:4.234
weights: [0]
closed: left
isdensity: false
```

Fitting a histogram to all-equal integer values works, but not for all-equal float values – see the `0` weight in the second example. The bin edges seem to be computed incorrectly?

Moelf:
fix #616

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -96,6 +96,9 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
             len += one(F)
         end
     end
+    if lo==hi && !isinteger(hi)
+        divisor = zero(F)
+    end
     Base.floatrange(start,step,len,divisor)
 end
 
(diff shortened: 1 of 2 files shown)


Moelf:
This is a temporary hack, maybe the real solution is to use `StepRangeLen` that is now exported in base Julia.

PR #620, opened 2020-11-27
Tighten signature of weights constructor

nalimilan:
This ensures the error is thrown earlier for clarity, instead of doing it from the two-argument constructor.

See https://discourse.julialang.org/t/using-weights-from-statsbase-in-julia-and-using-an-array/50820.

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -13,7 +13,7 @@ macro weights(name)
             values::V
             sum::S
         end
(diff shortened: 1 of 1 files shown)


nalimilan:
Needs https://github.com/JuliaStats/StatsBase.jl/pull/621 for CI.

rofinn (approved):
Seems fine to me.

PR #621, opened 2020-11-27
Move to GitHub Actions for CI

nalimilan:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
+name: CI
+on:
+  - push
```
rofinn, line 3: Might want to limit this to

```
  push:
    branches: [master]
    tags: ["*"]
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
...
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.0'
+          - '1' # automatically expands to the latest stable 1.x release of Julia
+          - 'nightly'
```
rofinn, line 15: This might be noisy as there isn't an equivalent to allow failures.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
...
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+        arch:
+          - x64
+          - x86
+        exclude:
```
rofinn, line 23: Do we actually need to test on all combinations of julia version, os and architecture?

rofinn (approved):
LGTM, just a couple suggestions you might want to consider.

.github/workflows/ci.yml
nalimilan: Yeah that's a difficult point. AFAIK we used to do that until now. Another solution would be to do something like https://github.com/JuliaLang/Example.jl/pull/55. But given that StatsBase has a relatively low activity it may be OK to run many checks.

.github/workflows/ci.yml
nalimilan: We could have it in a separate step which is allowed to fail, like https://github.com/JuliaStats/StatsModels.jl/pull/199.

kleinschmidt: FWIW the statsmodels setup doesn't actually allow it to fail in the strict sense that you'll still get the red X on the commit if it's failing.  but you can at least see isolate it in a separate workflow so you can have the nice green badge.

rofinn: Yeah, I guess if you aren't editing these files that often that's probably fine.

.github/workflows/ci.yml
rofinn: Yeah, that's what we do. If you don't think StatsBase is gonna be that active going forward then I'm fine with leaving it.

nalimilan:
OK let's go with this, we can refine it later but at least we'll have CI again.

.github/workflows/ci.yml
nalimilan: Done.

Issue #619, opened 2020-11-26
ZScoreTransform and UnitRangeTransform do not support CuArray


sdewaele:
`ZScoreTransform` and `UnitRangeTransform` do not support `CuArrray`s. For example, with `StatsBase` v0.33.2:
```julia
using CUDA
using StatsBase

A = 5randn(2,4).+2
t = fit(ZScoreTransform,A;dims=2)
```
fails with
```julia
┌ Warning: Performing scalar operations on GPU arrays: This is very slow, consider disallowing these operations with `allowscalar(false)`
└ @ GPUArrays ~/.julia/packages/GPUArrays/jhRU7/src/host/indexing.jl:43
ERROR: MethodError: no method matching ZScoreTransform(::Int64, ::Int64, ::CuArray{Float32,1}, ::CuArray{Float32,1})
Closest candidates are:
  ZScoreTransform(::Int64, ::Int64, ::Array{T,1}, ::Array{T,1}) where T at /home/student/.julia/packages/StatsBase/EA8Mh/src/transformations.jl:58
Stacktrace:
 [1] fit(::Type{ZScoreTransform}, ::CuArray{Float32,2}; dims::Int64, center::Bool, scale::Bool) at /home/student/.julia/packages/StatsBase/EA8Mh/src/transformations.jl:127
 [2] top-level scope at REPL[4]:1
```
To fix this, the `ZScoreTransform` would have to allow `AbstractArray`s. Also, specifically for `CuArray`s, to prevent bad performance due to indexing, `mean_and_std` needs to be updated. See [this gist](https://gist.github.com/sdewaele/296d9547ef937af413b32b370044adc7) for a possible implementation. I can put together a PR if desired. One question with regards to this change: Is it okay to replace `stdm` by `Statistics.std` using the `mean` option? This will facilitate solving this issue.

See also https://github.com/JuliaGPU/CUDA.jl/issues/426.

sdewaele:
Fully resolved by PR #622 for `ZScoreTransform`. `UnitRangeTransform` has been generalized so it can handle `CuArray`. 

Implement `_compute_extrema` to make `UnitRangeTransform` fully functional for `CuArray`. Note that this implementation is not included in StatsBase because it is not desirable to take on a dependency on CUDA.jl for this.

PR #622, opened 2020-11-29
more generic ZScoreTransform, UnitRangeTranform to support CuArrays

sdewaele:
Adresses https://github.com/JuliaStats/StatsBase.jl/issues/619.

- [x]  Make transformations work for CuArray vectors. They currently only work CuArray matrices.
- [x]  Find an appropriate home for methods currently defined in "test/cuda.jl" during testing.

This code passes the tests and is ready for merge from my perspective.


src/transformations.jl
```diff
@@ -134,12 +133,24 @@ function fit(::Type{ZScoreTransform}, X::AbstractVector{<:Real};
         throw(DomainError(dims, "fit only accepts dims=1 over a vector. Try fit(t, x, dims=1)."))
     end
 
-    T = eltype(X)
-    m, s = mean_and_std(X)
-    return ZScoreTransform(1, dims, (center ? [m] : zeros(T, 0)),
-                                    (scale ? [s] : zeros(T, 0)))
+    return fit(ZScoreTransform, reshape(X, :, 1);dims=dims, center=center, scale=scale)
```
wildart, line 136: The vector structure of the input input will be lost, and `transform` will return a 1-column matrix instead of a vector.

sdewaele: Thanks for your review. I will fix this shortly. The issue is that when I stay closer to the original code, I got an inference error. I'll look for an alternate solution to work around that.

sdewaele: It looks like `transform` still works as expected, i.e. it returns a `Vector`.
```julia
julia> X = rand(3)
3-element Vector{Float64}:
 0.0220367
 0.019156
 0.572394

julia> t = fit(ZScoreTransform, X)
ZScoreTransform{Float64, Vector{Float64}}(1, 1, [0.204529], [0.318584])

julia> transform(t, X)
3-element Vector{Float64}:
 -0.572823
 -0.581866
  1.15469

julia> Xt = transform(t, X)
3-element Vector{Float64}:
 -0.572823
 -0.581866
  1.15469

julia> reconstruct(t, Xt)
3-element Vector{Float64}:
 0.0220367
 0.019156
 0.572394
```

This is achieved through the `transform` [method](https://github.com/JuliaStats/StatsBase.jl/blob/master/src/transformations.jl#L23-L24) that is specialized for vectors:

src/StatsBase.jl
```diff
@@ -215,6 +215,24 @@ export
     ZScoreTransform,       # the type to represent a z-score data transformation
     UnitRangeTransform     # the type to represent a 0-1 data transformation
 
+function __init__()
+    @require CUDA="052768ef-5323-5732-b1bb-66c8b64840ba" begin
+        using .CUDA
+        function _compute_extrema(X::Union{CuMatrix{<:Real},Adjoint{T,CuMatrix{T}} where T<:Real})
```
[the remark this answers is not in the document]
sdewaele: It is used [here](https://github.com/JuliaStats/StatsBase.jl/blob/1c3f3d7ae4fd8afae126443d15ddaf8029316618/src/transformations.jl#L274) in the `fit(::Type{UnitRangeTransform}, ...)` function. The existing implementation does not work (or, is very slow) for `CuArray`s because the array is indexed.

src/transformations.jl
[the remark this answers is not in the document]
sdewaele: Done

src/StatsBase.jl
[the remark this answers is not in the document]
nalimilan: Sorry, but I don't think using Requires is OK for a package like StatsBase, as it will increase load times of many packages. We should find better abstractions so that StatsBase doesn't have to be aware of CUDA.

Couldn't `_compute_extrema` be replaced with `extrema.(eachcol(X))`? Would that be enough to use optimized CUDA methods?

test/transformations.jl
```diff
@@ -2,21 +2,40 @@ using StatsBase
 import StatsBase: transform, reconstruct, transform!, reconstruct!
 using Statistics
 using Test
+cuda_available = try
+    using StatsBase.CUDA
+    true
+catch
+    @warn "CUDA not available - transformations not tested on GPU"
+    false
+end
```
nalimilan, line 11: These tests should probably better live in CUDA, as AFAIK CUDA isn't enabled when we run CI for StatsBase.

src/StatsBase.jl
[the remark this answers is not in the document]
sdewaele: I was not too happy with using Requires either. I have removed it now. I also replaced `_compute_extrema` by `extrema`. The caveat that `extrema` has not yet been implemented for CUDA arrays, but it is reasonable to expect that it will become available in the future; it is mentioned in https://github.com/JuliaGPU/CUDA.jl/issues/265#issue-649177087

src/transformations.jl
```diff
@@ -270,45 +266,22 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
...
-    else
-        throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
-    end
-
-    for i = 1:l
-        @inbounds tmax[i] = 1 / (tmax[i] - tmin[i])
-    end
+    dims ∈ [1,2] || throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
+    tmin_tmax = extrema(X; dims=dims)
+    tmin, tmax = [vec(getindex.(tmin_tmax, i)) for i in 1:2]
```
wildart, line 271: You can form a generator instead of allocating an array.
```julia
tmin, tmax = (vec(getindex.(tmin_tmax, i)) for i in 1:2)
```

wildart (approved):
Minor comment on allocation, otherwise looks good.

src/transformations.jl
sdewaele: Done. Thanks for the suggestion.

src/transformations.jl
```diff
@@ -270,45 +266,22 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
...
-    elseif dims == 2
-        l, tmin, tmax = _compute_extrema(X')
-    else
-        throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
-    end
-
-    for i = 1:l
-        @inbounds tmax[i] = 1 / (tmax[i] - tmin[i])
-    end
+    dims ∈ [1,2] || throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
```
nalimilan, line 269:
```suggestion
    dims ∈ (1, 2)] || throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
```

src/transformations.jl
```diff
@@ -270,45 +266,22 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
...
+    tmin_tmax = extrema(X; dims=dims)
+    tmin, tmax = (vec(getindex.(tmin_tmax, i)) for i in 1:2)
+    @inbounds @. tmax = 1 / (tmax - tmin)
```
nalimilan, line 272: `@inbounds` is probably not needed as broadcasting should take care of that.
```suggestion
    @. tmax = 1 / (tmax - tmin)
```

src/moments.jl
```diff
@@ -204,7 +204,6 @@ function mean_and_std(x::RealArray, dim::Int; corrected::Bool=true)
     m, s
 end
 
-
```
nalimilan, line 207: No need to change this.

src/transformations.jl
```diff
@@ -270,45 +266,22 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
...
-    else
-        throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
-    end
-
-    for i = 1:l
-        @inbounds tmax[i] = 1 / (tmax[i] - tmin[i])
-    end
+    dims ∈ [1,2] || throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
+    tmin_tmax = extrema(X; dims=dims)
+    tmin, tmax = (vec(getindex.(tmin_tmax, i)) for i in 1:2)
```
nalimilan, line 271: Ah so it's too bad that this forces copying the result to extract the min and max. Do you think CUDA supports `eachslice`, or would it be efficient to iterate over rows/columns and call `extrema` on them? Then we could store directly the result in the right format.

test/transformations.jl
```diff
@@ -13,10 +13,10 @@ using Test
     @test isa(t, AbstractDataTransform)
     @test isempty(t.mean)
     @test isempty(t.scale)
-    @test isequal(X, Y)
+    @test X == Y
```
nalimilan, line 16: Any reason to change these?

src/moments.jl
sdewaele: I restored the additional line.

src/transformations.jl
sdewaele: I have been experimenting a little bit, but have not found a good solution. This prevents the copying, but does not work with `CuArray`s because of the element-wise setindex:
```julia
l = size(X, dims==1 ? 2 : 1)
tmin, tmax = (similar(X, l) for _ in 1:2)
for (i, x) in enumerate(eachslice(X; dims=dims))
    tmin[i], tmax[i] = extrema(x) # per-element assignment does not work with CUDA
end
```
Then there is `zip` + `...` but I think the performance of that is even worse because of the additional penalty of `...`:
```julia
tmin, tmax = collect.(zip((extrema(x) for x in eachslice(X; dims=dims))...))
```
Also, now the outputs no longer end up in a `CuArray` which will mean the `transform!` will fail on `CuArray`.

test/transformations.jl
sdewaele: I incorrectly thought that `isequal` did not work for `CuArray`. I have restored it.

src/transformations.jl
nalimilan: Ah, too bad. Maybe better not implement this until CUDA supports it. Could you ask for advice there about the best way to achieve this without copying?

sdewaele: I asked the question on discourse, see [here](https://discourse.julialang.org/t/efficient-and-cuarray-compatible-implementation-of-statsbase-compute-extrema/51671).

nalimilan: Might also be worth posting a link on #gpu on Slack.

sdewaele: In the latest commit I added a new  `_compute_extrema` that eliminates the allocation. With that, it should now be as efficient as the original implementation for regular arrays. Also, this allow for a temporary solution for `CuArray`s by implementing as specialized `_compute_extrema` for `CuArray`s:
```julia
function  StatsBase._compute_extrema(X::CuArray; dims)
    tmin = vec(minimum(X; dims=dims))
    tmax = vec(maximum(X; dims=dims))
    return tmin, tmax
end
```
This is not optimal, but it works. Also, given the compute power of a GPU, I assume it will still give a significant performance improvement.

nalimilan: Cool. Though you'll need Compat 2.2 to use `eachslice` on Julia 1.0.

sdewaele:
I suggest to merge the PR now. It will enable:
1. `ZScoreTransform` has been made more generic and works for `CuArray`
2. `UnitRangeTransform` has been made more generic in preparation for using it for `CuArray`.

`ZScoreTransform` is anyway more important in my opinion, because it is used more frequently in machine learning.

After merging this PR request, users can now define `_compute_extrema` for `CuArrays` as described to make `UnitRangeTransform` work. This is not possible with the current definition of `UnitRangeTransform` because it is not generic enough.

P.S. The code check states that "Some checks were not successful". I don't think this has anything to do with this PR; it has failed on downloading the registry. Probably re-running the test will eliminate the error.

src/transformations.jl
```diff
@@ -270,45 +266,31 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
...
-    for j = 1:l
-        @inbounds for i = 2:n
-            if X[i, j] < tmin[j]
-                tmin[j] = X[i, j]
-            elseif X[i, j] > tmax[j]
-                tmax[j] = X[i, j]
-            end
-        end
+function _compute_extrema(X, dims)
+    dims==2 && return _compute_extrema(X', 1)
```
nalimilan, line 277:
```suggestion
    dims == 2 && return _compute_extrema(X', 1)
```

src/transformations.jl
```diff
@@ -270,45 +266,31 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
...
+    l = size(X, 2)
+    tmin, tmax = (similar(X, l) for _ in 1:2)
```
nalimilan, line 279:
```suggestion
    tmin = similar(X, l)
    tmax = similar(X, l)
```

src/transformations.jl
```diff
@@ -270,45 +266,31 @@ function fit(::Type{UnitRangeTransform}, X::AbstractMatrix{<:Real};
...
-    tmax = X[1, :]
-    for j = 1:l
-        @inbounds for i = 2:n
-            if X[i, j] < tmin[j]
-                tmin[j] = X[i, j]
-            elseif X[i, j] > tmax[j]
-                tmax[j] = X[i, j]
-            end
-        end
+function _compute_extrema(X, dims)
```
nalimilan, line 276: Better be explicit to make the code clearer:
```suggestion
function _compute_extrema(X::AbstractMatrix, dims::Integer)
```

nalimilan (commented):
Fine with me. Just a few stylistic comments.

nalimilan:
Thanks!

sdewaele:
You are welcome. Thanks for your thorough review @nalimilan and @wildart.

PR #624, opened 2020-12-04
Avoid random failures in test

nalimilan:
Negative weights can give errors. They should probably be disallowed but that's a more fundamental change.

### DIFF
--- a/test/moments.jl
+++ b/test/moments.jl
@@ -103,9 +103,9 @@ expected_std = sqrt.(expected_var)
     end
 end
 
-x = rand(5, 6)
-w1 = rand(5)
-w2 = rand(6)
+x = rand(5, 6)    
+w1 = [0.57, 5.10, 0.91, 1.72, 0.0]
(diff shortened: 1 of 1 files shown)


mschauer (approved):

mschauer:
Might be not about negative weights (the random numbers are non-negative), but issues with having less than 1 observation (in weight sum) when you estimate the standard deviation with correction and frequency weights.

nalimilan:
Ah indeed. Maybe we should check that in `varcorrection`.

mschauer:
Also the question what _is_ the right correction for weighted samples. I think we must have talked about this before, otherwise perhaps that wouldn't have come to my mind easily.

nalimilan:
What do you mean? We have different kinds of corrections which are appropriate for the corresponding kind of weight. But currently we don't check that the sum indicates that there's at least one observation, which is indeed required for the correction to make any sense.

PR #626, opened 2020-12-30
Stop using LinearAlgebra.copy_oftype

nalimilan:
This undocumented function does not actually ensure that the result is mutable.
Use the same approach as `Base.copymutable`, which relies only on public API.

This avoids breakage due to https://github.com/JuliaLang/julia/pull/28956 (see [this log](https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/d8c33b6_vs_ad7e59a/StatsBase.1.7.0-DEV-70af8d1ffd.log)).

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -274,7 +274,7 @@ function mad(x; center=nothing, normalize::Union{Bool, Nothing}=nothing, constan
     # Knowing the eltype allows allocating a single array able to hold both original values
     # and differences from the center, instead of two arrays
     S = isconcretetype(T) ? promote_type(T, typeof(middle(zero(T)))) : T
-    x2 = x isa AbstractArray ? LinearAlgebra.copy_oftype(x, S) : collect(S, x)
+    x2 = x isa AbstractArray ? copyto!(similar(x, S), x) : collect(S, x)
     c = center === nothing ? median!(x2) : center
     if isconcretetype(T)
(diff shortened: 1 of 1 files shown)


src/scalarstats.jl
```diff
@@ -274,7 +274,7 @@ function mad(x; center=nothing, normalize::Union{Bool, Nothing}=nothing, constan
     # Knowing the eltype allows allocating a single array able to hold both original values
     # and differences from the center, instead of two arrays
     S = isconcretetype(T) ? promote_type(T, typeof(middle(zero(T)))) : T
-    x2 = x isa AbstractArray ? LinearAlgebra.copy_oftype(x, S) : collect(S, x)
+    x2 = x isa AbstractArray ? copyto!(similar(x, S), x) : collect(S, x)
```
nalimilan, line 277: Note that this change assumes that when `S == T`, `copyto!(similar(x, S), x) ` is as fast as `copy(x)`. Not sure whether that holds: I could add another branch to ensure that this simple case is fast.

andreasnoack:
I'm mainly curious about what motivated this change. Is it worth adding as a test case?

nalimilan:
That's mainly to avoid breakage if JuliaLang/julia#28956 gets merged. There's no hurry. But in general it sounds better to avoid depending on unexported and undocumented functions from the stdlib.

PR #627, opened 2021-01-01
Add pairwise

nalimilan:
This generic method takes iterators of vectors and supports skipping missing values.
It is a more general version of `pairwise` in Distances.jl.
Since methods are compatible, both packages can override a common empty function defined in another API package.

See https://github.com/nalimilan/FreqTables.jl/pull/54. Cc: @dkarrasch @bkamins @quinnj


src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+"""
+    pairwise(f, x[, y], symmetric::Bool=false, skipmissing::Symbol=:none)
+
+Return a matrix holding the result of applying `f` to all possible pairs
+of vectors in iterators `x` and `y`. Rows correspond to
+vectors in `x` and columns to vectors in `y`. If `y` is omitted then a
+square matrix crossing `x` with itself is returned.
+
+As a special case, if `f` is `cor`, diagonal cells are set to 1 even in
+the presence `NaN` or `Inf` entries (but `missing` is propagated unless
```
bkamins, line 138: why `missing` is propagated on the diagonal for `cor`?

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+
+"""
+    pairwise(f, x[, y], symmetric::Bool=false, skipmissing::Symbol=:none)
+
+Return a matrix holding the result of applying `f` to all possible pairs
+of vectors in iterators `x` and `y`. Rows correspond to
+vectors in `x` and columns to vectors in `y`. If `y` is omitted then a
+square matrix crossing `x` with itself is returned.
+
+As a special case, if `f` is `cor`, diagonal cells are set to 1 even in
```
bkamins, line 137: diagonal - maybe clarify that this means when `y` is not passed?

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+the presence `NaN` or `Inf` entries (but `missing` is propagated unless
+`skipmissing` is different from `:none`).
+
+# Keyword arguments
+- `symmetric::Bool=false`: If `true`, `f` is only called to compute
+  for the lower triangle of the matrix, and these values are copied
+  to fill the upper triangle. Only possible when `y` is omitted.
```
bkamins, line 144: "possible" or "allowed"

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+  This is automatically set to `true` when `f` is `cor` or `cov`.
+- `skipmissing::Symbol=:none`: If `:none` (the default), missing values
+  in input vectors are passed to `f` without any modification.
+  Use `:pairwise` to skip entries with a `missing` value in either
+  of the two vectors passed to `f` for a given pair of vectors in `x` and `y`.
+  Use `:listwise` to skip entries with a `missing` value in any of the
+  vectors in `x` or `y`; note that this is likely to drop a large part of
```
bkamins, line 151: "likely" or "might" (very often in practical data sets `missing` are clustered together in rows)

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+
+As a special case, if `f` is `cor`, diagonal cells are set to 1 even in
+the presence `NaN` or `Inf` entries (but `missing` is propagated unless
+`skipmissing` is different from `:none`).
+
+# Keyword arguments
+- `symmetric::Bool=false`: If `true`, `f` is only called to compute
+  for the lower triangle of the matrix, and these values are copied
+  to fill the upper triangle. Only possible when `y` is omitted.
+  This is automatically set to `true` when `f` is `cor` or `cov`.
```
bkamins, line 145: it is "the default" in this case not "automatically set to `true` (as you can set it to `false` if you want as I see below)

src/pairwise.jl
bkamins: ah - actually it is when `y` is not passed or `x===y` - right?

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+    # that entries cannot be `missing` (similar to `skipmissing`)
+    # could offer better performance
+    return _pairwise!(Val(:none), res, f,
+                      [view(xi, nminds′) for xi in x],
+                      [view(yi, nminds′) for yi in y],
+                      symmetric)
+end
+
+function _pairwise(::Val{skipmissing}, f, x, y, symmetric::Bool) where {skipmissing}
+    inds = keys(first(x))
```
bkamins, line 82: have a special case when `isempty(x)` is true (the same for `y`). Do we have a test of this case?

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+        return (isconcretetype(eltype(res)) ? res : identity.(res))::Matrix{<:U}
+    else
+        return (isconcretetype(eltype(res)) ? res : identity.(res))
+    end
+end
+
+"""
+    pairwise(f, x[, y], symmetric::Bool=false, skipmissing::Symbol=:none)
+
+Return a matrix holding the result of applying `f` to all possible pairs
```
bkamins, line 132: is this a matrix or actually `Matrix`?

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+    y′ = y isa Union{AbstractArray, Tuple, NamedTuple} ? y : collect(y)
+    m = length(x)
+    n = length(y)
+
+    T = Core.Compiler.return_type(f, Tuple{eltype(x′), eltype(y′)})
+    Tsm = Core.Compiler.return_type((x, y) -> f(disallowmissing(x), disallowmissing(y)),
+                                     Tuple{eltype(x′), eltype(y′)})
+
+    if skipmissing === :none
+        res = Matrix{T}(undef, m, n)
```
bkamins, line 105: why not create `res` outside of the conditions?

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
+function _pairwise!(::Val{:none}, res::AbstractMatrix, f, x, y, symmetric::Bool)
+    m, n = size(res)
+    for j in 1:n, i in 1:m
+        symmetric && i > j && continue
+
+        # For performance, diagonal is special-cased
+        if f === cor && i == j && x[i] === y[j]
+            # If the type isn't concrete, 1 may not be converted to the right type
+            # and the final matrix will have an abstract eltype
+            # (missings are propagated via the second branch, but NaNs are ignored)
```
bkamins, line 10: are missings propagated via the second branch? It seems that this branch will be triggered even for missings

dkarrasch (commented):
Looks pretty good, only two comments about indexing/iterating `x` and `y` and one-based indexing into `res`.

If I haven't missed anything, then these methods here are more generic than the generic methods in Distances.jl. I believe one could map those to the `_pairwise!` "kernels" here.

src/pairwise.jl
nalimilan: This is tricky. `cor(x)` returns 1 whatever the input values (i.e. even if there are `NaN`s, `Inf`s or `missing`s). I find it kind of weird, though it's probably not dangerous since you're unlikely to use values on the diagonal.

The reason why `missing` is propagated is that at [line 38 above](https://github.com/JuliaStats/StatsBase.jl/pull/627/files/9c8b3a6599c88537a63f8e773fd111f536de699e#diff-c3666e661823c2d9a812e6b080021b2c08a66f16f114180905da0417d8aece27R38) I return `one(f(x[i], x[j]))` when the element type is not concrete, since `cor(x[i])` returns `1` when the input element type is `Int` despite `cor(x[i], x[i])` returning `1.0`. This should probably be fixed in Statistics.

But yeah, I guess returning `missing` isn't a good idea since that's inconsistent with Statistics and changing it there would be breaking. We can probably return `float(cor(x[i]))` and do the same in Statistics.

src/pairwise.jl
nalimilan: Yes technically it's when vectors are `===` (even if iterators themselves are not).

src/pairwise.jl
nalimilan: It's `Matrix`. I'll make the wording more precise.

src/pairwise.jl
nalimilan: I don't think I tested that indeed.

src/pairwise.jl
```diff
@@ -0,0 +1,166 @@
...
+            res[i, j] = isconcretetype(eltype(res)) ? 1 : one(f(x[i], y[j]))
+        else
+            res[i, j] = f(x[i], y[j])
```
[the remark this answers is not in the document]
nalimilan: Actually in the current implementation `pairwise` calls `collect` on the iterators so they are always `AbstractVector`s. The reason is that currently `eachcol` and `eachrow` are just generators, so `eltype` returns `Any` for them and that requires making a copy of the resulting matrix once we know the final eltype. Allocating vectors of length `m` and `n` is better than making a copy of an `m×n` matrix. But yes I guess I'd better make this code more generic just in case.

src/pairwise.jl
nalimilan: See https://github.com/JuliaStats/StatsBase.jl/pull/627/files/9c8b3a6599c88537a63f8e773fd111f536de699e#r550921778. I think this needs to be changed.

src/pairwise.jl
bkamins: If we change `Statistics` code to `cor(x::AbstractVector) = one(float(eltype(x)))` which I think we should (I will open a PR for this) then here it would be just `cor(x[i])` and for the time being what you propose is OK I think.

bkamins: https://github.com/JuliaLang/Statistics.jl/pull/61

src/pairwise.jl
bkamins: still - maybe it would be good to clarify, as "diagonal" is not fully clear here I think. Also it would be good to change `1` to `1.0` in the description (given the discussion below).

src/pairwise.jl
bkamins: As a second thought - in this PR it is a `Matrix`. But if someone adds a custom method to it it might not be a `Matrix` but `AbstractMatrix`. Do we want to allow for this?

src/pairwise.jl
bkamins: I mean - when `isempty(x)` is true then this code is incorrect (it will error, while it should return a value, but with empty dimension)

src/pairwise.jl
nalimilan: Yes probably better keep "matrix". Indeed in FreqTables (or elsewhere) I'd like to return a `NamedMatrix` when a Table is passed.

nalimilan:
I've pushed commits which should address all comments. I've also made two changes:
- Add `pairwise!` since it is standard practice to have in-place variants
- When `skipmissing=:none`, allow entries to be any objects, and not just vectors. This is more consistent with Distances.jl as @dkarrasch noted.

bkamins:
Nice. Just to go back for a second to the issue with missing and `NaN`. Let us confirm that we really want:
```
julia> x = [missing, missing, missing]
3-element Array{Missing,1}:
 missing
 missing
 missing

julia> pairwise(cor, Ref(x), Ref(x))
1×1 Array{Missing,2}:
 missing
```
as this is the first case when the docstring is not correct (but I think it is OK as you have no way to infer the type of "one").

The second case is:
```
julia> x = Union{Int, Missing}[missing, missing, missing]
3-element Array{Union{Missing, Int64},1}:
 missing
 missing
 missing

julia> pairwise(cor, Ref(x), Ref(x), skipmissing=:pairwise)
ERROR: ArgumentError: correlation only defined for non-empty vectors
```
(we error although we promise to return `1`, but again - maybe it is OK to leave this corner case as is)

nalimilan:
Good catch. Yes I'm not sure we can do much about it. Technically we return `one(missing)`, so the docstring isn't totally wrong in the first case. :-p I'm not sure it's worth mentioning as it sounds difficult to explain without making the sentence very complex, but at least I've added tests so that we don't change the behavior without noticing. This made me discover that in some cases we threw a `MethodError` because the destination was `Matrix{Union{}}` instead of letting `cor` throw the expected `ArgumentError`.

I've also found a simple way to use `promote_type` rather than `promote_typejoin` to choose the element type, while still having the function inferable. I think this makes more sense as it gives a concrete eltype (`Float64` in general) if you have columns of different types (which is common for tables), and it justifies using `pairwise` rather than just `map(f, Iterators.product(x, y))`.

bkamins (approved):

bkamins:
Looks good in general (but CI fails on Julia 1.0, so probably a separate code path for it is required).

src/pairwise.jl
```diff
@@ -0,0 +1,313 @@
...
+            dest[i, j] = dest[j, i]
+        end
+    end
+    return dest
+end
+
+function check_vectors(x, y, skipmissing::Symbol)
+    m = length(x)
+    n = length(y)
+    if !(all(xi -> xi isa AbstractVector, x) && all(yi -> yi isa AbstractVector, x))
```
dkarrasch, line 25:
```suggestion
    if !(all(xi -> xi isa AbstractVector, x) && all(yi -> yi isa AbstractVector, y))
```
?

dkarrasch (commented):
Indeed, looks good.

src/pairwise.jl
nalimilan: Woops. I'll always be surprised by the number of tests one needs to write to ensure all possible mistakes are covered! I've added another one.

nalimilan:
I've created a StatsAPI package, with a PR to define `pairwise`there: https://github.com/JuliaStats/StatsAPI.jl/pull/1

pdeffebach:
Could we implement a special case for `pairwise(cor, x::Vector{Number}, y::Vector{Number})` to allow scalar returns?

Or would that be a different function entirely.

nalimilan:
> Could we implement a special case for `pairwise(cor, x::Vector{Number}, y::Vector{Number})` to allow scalar returns?
> 
> Or would that be a different function entirely.

That would have to be another function (more akin to `skipmissings`), as this PR already gives a meaning to this signature. Vectors of numbers are treated like any other collection:
```julia
julia> pairwise(tuple, [1, 2], [3, 4])
2×2 Matrix{Tuple{Int64, Int64}}:
 (1, 3)  (1, 4)
 (2, 3)  (2, 4)
```

I've added a commit to import `pairwise` and `pairwise!` from StatsAPI. With https://github.com/JuliaStats/Distances.jl/pull/213, StatsBase and Distances will cohabit nicely without method ambiguities. I'll merge in a few days barring objections.

bkamins:
OK. Looks good.

wildart:
When is this expected to be merged? I'd like to use new APi in https://github.com/JuliaStats/MultivariateStats.jl/pull/148.

nalimilan:
https://github.com/JuliaRegistries/General/pull/35827

PR #629, opened 2021-01-05
Make CoefTable implement the Tables.jl interface

nalimilan:
This allows retrieving the contents of a `CoefTable` object in a convenient form, notably a `DataFrame`. To avoid introducing a dependency on Tables.jl, `CoefTable` has to iterate `NamedTuple`s, so that it implements the row-table interface implicitly. This is inefficient since `CoefTable` uses a column-based storage, but given the typical size of such tables it should not matter.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -420,6 +420,23 @@ mutable struct CoefTable
     end
 end
 
+Base.length(ct::CoefTable) = length(ct.cols[1])
+function Base.eltype(ct::CoefTable)
+    nmtype = isempty(ct.rownms) ? String : eltype(ct.rownms)
+    NamedTuple{(Symbol("Name"), Symbol.(ct.colnms)...),
+               Tuple{nmtype, eltype.(ct.cols)...}}
+end
+
+function Base.iterate(ct::CoefTable, i::Integer=1)
+    if i in 1:length(ct)
+        nm = isempty(ct.rownms) ? string('[', i, ']') : ct.rownms[i]
+        nt = eltype(ct)((nm, getindex.(ct.cols, Ref(i))...))
+        (nt, i+1)
+    else
+        nothing
+    end
+end
+
 """
 Show a p-value using 6 characters, either using the standard 0.XXXX
 representation or as <Xe-YY.
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -19,8 +19,12 @@ coefnames(model::StatisticalModel) = error("coefnames is not defined for $(typeo
 """
     coeftable(model::StatisticalModel; level::Real=0.95)
 
-Return a table of class `CoefTable` with coefficients and related statistics.
+Return a table with coefficients and related statistics of the model.
 `level` determines the level for confidence intervals (by default, 95%).
+
+The returned `CoefTable` object implements the
+[Tables.jl](https://github.com/JuliaData/Tables.jl/) interface, and can be
+converted e.g. to a `DataFrame` via `using DataFrames; DataFrame(coeftable(model))`.
 """
 coeftable(model::StatisticalModel) = error("coeftable is not defined for $(typeof(model)).")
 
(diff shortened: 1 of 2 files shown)


test/statmodels.jl
```diff
@@ -7,27 +7,50 @@ v2 = ["Good", "Great", "Bad"]
...
 [2]  0.368314  0.120781   0.815104  0.6699
 [3]  0.344454  0.179574   0.242208  0.4531
 ──────────────────────────────────────────"""
+@test length(ct) === 3
+@test eltype(ct) ==
+NamedTuple{(:Name, :Estimate, :Stderror, :df, :p),
+           Tuple{String,Float64,Float64,Float64,Float64}}
+@test collect(ct) == [
+    (Name = "[1]", Estimate = 0.11258244478647295, Stderror = 0.05664544616214151,
+     df = 0.38181274408522614, p = 0.8197779704008801)
```
nalimilan, line 48: I'm hesitant as to whether it's better to always create a "Name" column even when no names have been specified. That makes the result more predictable, but that also creates a column which is nearly useless...

bkamins: I think it is better to keep a fixed schema for this.

bkamins:
Can we add DataFrames.jl as [extras] dependency and just check if a `DataFrame` is correctly produced (it is not strictly necessary, but I would find it nice to have in the tests 😄).

nalimilan:
> Can we add DataFrames.jl as [extras] dependency and just check if a `DataFrame` is correctly produced (it is not strictly necessary, but I would find it nice to have in the tests smile).

Is that really useful? As long as we check that the interface is implemented everything should be OK.

test/statmodels.jl
```diff
@@ -7,27 +7,50 @@ v2 = ["Good", "Great", "Bad"]
...
      Estimate   Stderror        df       p
 ──────────────────────────────────────────
 [1]  0.112582  0.0566454  0.381813  0.8198
 [2]  0.368314  0.120781   0.815104  0.6699
 [3]  0.344454  0.179574   0.242208  0.4531
 ──────────────────────────────────────────"""
+@test length(ct) === 3
+@test eltype(ct) ==
+NamedTuple{(:Name, :Estimate, :Stderror, :df, :p),
+           Tuple{String,Float64,Float64,Float64,Float64}}
```
nalimilan, line 45:
```suggestion
    NamedTuple{(:Name, :Estimate, :Stderror, :df, :p),
               Tuple{String,Float64,Float64,Float64,Float64}}
```

test/statmodels.jl
nalimilan: Well there isn't really a fixed schema since each package determines the column names and types. Actually, "Names" would be the only column which is guaranteed to be present. :-)

bkamins:
> Is that really useful?

This was a soft suggestion - mainly to show users in the tests how this can be used. Alternatively maybe we could put some docstring about it. The point is to have somewhere a clear **visual** signal about what we allow (and `DataFrame` is very easy to recognize visually). This is a soft suggestion.

test/statmodels.jl
bkamins: Ah - in this case I would leave this out (i.e. rely on the schema provided by the package) and allow packages to specify the schema in full. Then actually we should check if packages provide a proper schema.

jerlich:
Thanks for being super responsive about this!

kleinschmidt (approved):

kleinschmidt:
I do think it could be a good example in the docs ("civilians" might not understand that Tables.jl integration means you can just call `DataFrame` on it...)

src/statmodels.jl
```diff
@@ -420,6 +420,23 @@ mutable struct CoefTable
     end
 end
 
+Base.length(ct::CoefTable) = length(ct.cols[1])
+function Base.eltype(ct::CoefTable)
+    nmtype = isempty(ct.rownms) ? String : eltype(ct.rownms)
+    NamedTuple{(Symbol("Name"), Symbol.(ct.colnms)...),
```
quinnj, line 426: The other option is we could return a custom "view" struct here instead of a NamedTuple. That could make it a bit more efficient and is a common pattern for other "table" types. For example, it's common to have something like:
```julia
struct CoefRow
    table::CoefTable
    rownumber::Int
end
```
Then on top of that you just need to implement `propertynames` and `getproperty` that correspond to the "row values" to be returned. The property names could be computed a single time in the initial `Base.iterate(ct::CoefTable)` method and passed as an additional piece of iterator state. It's obviously a bit more code/work, so maybe not worth it if these tables are always going to be small and simple anyway.

nalimilan: Interesting, thanks. Though these tables are always super small so as you say I'm not sure it's worth the additional complexity. If performance really mattered we should implement a column-wise approach anyway.

kleinschmidt (approved):

PR #632, opened 2021-01-12
Fix #631, summarystats returned NaN quantiles for arrays with mean 0

### LINKED ISSUE #631: summarystats returns NaN median/min/max/quantiles for any array with mean exactly 0
Because of this line https://github.com/JuliaStats/StatsBase.jl/blob/41669cd8dfeadce35db0c4e07ac7afe5d10fb957/src/scalarstats.jl#L663 whenever `mean(s) == 0` exactly (which is not even a meaningful check to make in the first place), the output is wrong for `summarystats(-1:1)`:

```
julia> summarystats(-1:1)
Summary Stats:
Length:         3
Missing Count:  0
Mean:           0.000000
Minimum:        NaN
1st Quartile:   NaN
Median:         NaN
3rd Quartile:   NaN
Maximum:        NaN
```

ikirill:
Removed check `mean(a) == 0`, added two tests

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -660,7 +660,7 @@ function summarystats(a::AbstractArray{T}) where T<:Union{Real,Missing}
     R = typeof(m)
     n = length(a)
     ns = length(s)
-    qs = if m == 0 || n == 0
+    qs = if n == 0
         R[NaN, NaN, NaN, NaN, NaN]
     elseif T >: Missing
         quantile!(s, [0.00, 0.25, 0.50, 0.75, 1.00])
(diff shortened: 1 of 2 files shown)


src/scalarstats.jl
```diff
@@ -660,7 +660,7 @@ function summarystats(a::AbstractArray{T}) where T<:Union{Real,Missing}
     R = typeof(m)
     n = length(a)
     ns = length(s)
-    qs = if m == 0 || n == 0
+    qs = if n == 0
```
nalimilan, line 663: BTW, this should probably be `ns` rather than `n`, if you feel like fixing another bug I introduced in that function... The difference should be visible when passing only missing values.

nalimilan (approved):
Thanks!

So apparently I introduced this nonsensical `m == 0` check at https://github.com/JuliaStats/StatsBase.jl/commit/0d1fe72d538d535c9b3b793fa868b93119ed0688, but I can't find any plausible explanation for that.

ikirill:
Sure, fixed.

nalimilan:
Thanks!

PR #640, opened 2021-01-24
Rewrote corkendall

PGS62:
See issue 634. New code is faster than the previous version. If both arguments are vectors the speed up is approx x 4, if one or both are matrices the speed up is approx x 7


PGS62:
I will close this pull request and make another one after fixing the errors that these tests discovered.

Issue #634, opened 2021-01-15
Speeding up Kendall Correlation `corkendall`


PGS62:
I need to calculate the Kendall correlation of an array of size approximately 1,000 x 10,000 and calculate some statistics of the resulting 10,000 x 10,000 correlation matrix.

The current implementation of `corkendall`, even though it uses Knight's algorithm, is quite slow in this case. I estimate about five hours on my PC. I hope I'll be able to get that time down using multiple threads. Should be straightforward no?

But first I had a go at improving the current not-threaded implementation and have achieved a 4 to 5 times speed improvement. I think the tests could be beefed up a bit too. An obvious way to do that is to test against a very simple order N^2 implementation.

Would a pull request containing my suggested replacement of the current `corkendall` likely be accepted? 

If it would help, I could make my code available as a standalone package as an interim step. 

Thanks,

Philip

nalimilan:
Well without more details about potential downsides, it's hard to say "no" to this kind of offer. :-) Please copy the code somewhere or make a PR if you want more specific comments.

PGS62:
OK, I will do that within the next few days. The code is already in GitHub in a private repository of mine, so when I'm happy to share it I will make the repo public, or I could make a pull request.

It's fair to raise the question about potential downsides. I don't think there are any downsides to my re-write, but it's always possible. For example, I need to look into performance testing against slightly "pathological" input data such as that with very large numbers of repeats. Another question is memory usage. My version does allocate more memory (as reported by `@btime`) and I confess I haven't really worried about that.

Since making my previous post I have written code that can compare the execution speed and results from arbitrary implementations of Kendall Tau, against randomly generated input data. That testing revealed some bugs in my code (no surprise, now fixed) but also a couple of minor ones in the existing code.

For example, `StatsBase.corkendall` fails if passed a single argument being a matrix of integers, but does not fail if passed two matrices of integers:

```
julia> X = rand(1:100,5,5)
5×5 Array{Int64,2}:
  59  44  25  11  20
 100  60  66   6  65
  85  36  18  11  32
  20  86  31  98  61
  39  69   1  90  75

julia> StatsBase.corkendall(X,X)
5×5 Array{Float64,2}:
  1.0       -0.6        0.2       -0.948683  0.0
 -0.6        1.0        0.2        0.527046  0.4
  0.2        0.2        1.0       -0.316228  0.0
 -0.948683   0.527046  -0.316228   1.0       0.105409
[...]

PR #647, opened 2021-01-26
Rewrote corkendall (issue 634)

PGS62:
New version of corkendall is approx 4 times faster if both arguments are vectors and 7 times faster if at least one is a matrix. See issue #634 for details. Passes all tests except against Julia nightly, which has unconnected failures (in hist.jl).

### DIFF
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -35,7 +35,7 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
 
 # Knight, William R. “A Computer Method for Calculating Kendall's Tau with Ungrouped Data.”
 # Journal of the American Statistical Association, vol. 61, no. 314, 1966, pp. 436–439.
-# JSTOR, www.jstor.org/stable/2282833. Accessed 15 Jan. 2021.
+# JSTOR, www.jstor.org/stable/2282833.
 function corkendall!(x::RealVector, y::RealVector, permx=sortperm(x))
     if any(isnan, x) || any(isnan, y) return NaN end
     n = length(x)
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -39,7 +39,7 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
 function corkendall!(x::RealVector, y::RealVector, permx=sortperm(x))
     if any(isnan, x) || any(isnan, y) return NaN end
     n = length(x)
-    if n ≠ length(y) error("Vectors must have same length") end
+    if n != length(y) error("Vectors must have same length") end
 
     # Initial sorting
     x[:] = x[permx]
@@ -49,7 +49,7 @@ function corkendall!(x::RealVector, y::RealVector, permx=sortperm(x))
     # ntiesx, ntiesy, ndoubleties are floats to avoid overflows on 32bit
     ntiesx, ntiesy, ndoubleties, k, nswaps = 0.0, 0.0, 0.0, 0, 0
 
-    @inbounds for i ∈ 2:n
+    @inbounds for i = 2:n
         if x[i - 1] == x[i]
             k += 1
         elseif k > 0
@@ -84,7 +84,7 @@ function countties(x::AbstractVector, lo::Integer, hi::Integer)
     # avoid overflows on 32 bit by using floats
     thistiecount, result = 0.0, 0.0
     (lo < 1 || hi > length(x)) && error("Bounds error")
-    @inbounds for i ∈ (lo + 1):hi
+    @inbounds for i = (lo + 1):hi
         if x[i] == x[i - 1]
             thistiecount += 1.0
         elseif thistiecount > 0
@@ -114,9 +114,9 @@ corkendall(x::RealVector, Y::RealMatrix) = (n = size(Y, 2); permx = sortperm(x);
 function corkendall(X::RealMatrix)
     n = size(X, 2)
     C = ones(float(eltype(X)), n, n)# avoids dependency on LinearAlgebra
-    @inbounds for j ∈ 2:n
+    @inbounds for j = 2:n
         permx = sortperm(X[:,j])
-        for i ∈ 1:j - 1
+        for i = 1:j - 1
             C[j,i] = corkendall!(X[:,j], X[:,i], permx)
             C[i,j] = C[j,i]
         end
@@ -128,9 +128,9 @@ function corkendall(X::RealMatrix, Y::RealMatrix)
     nr = size(X, 2)
     nc = size(Y, 2)
     C = zeros(float(eltype(X)), nr, nc)
-    @inbounds for j ∈ 1:nr
+    @inbounds for j = 1:nr
         permx = sortperm(X[:,j])
-        for i ∈ 1:nc
+        for i = 1:nc
             C[j,i] = corkendall!(X[:,j], Y[:,i], permx)
         end
     end
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -1,13 +1,13 @@
 # Rank-based correlations
-#
+# 
 # - Spearman's correlation
 # - Kendall's correlation
-#
+# 
 
 #######################################
-#
+# 
 #   Spearman correlation
-#
+# 
 #######################################
 
 """
@@ -28,9 +28,9 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
 
 
 #######################################
-# 
+#
 #   Kendall correlation
-# 
+#
 #######################################
 
 # Knight, William R. “A Computer Method for Calculating Kendall's Tau with Ungrouped Data.”
@@ -47,7 +47,7 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
 
     npairs = float(n) * (n - 1) / 2
     # ntiesx, ntiesy, ndoubleties are floats to avoid overflows on 32bit
-    ntiesx, ntiesy, ndoubleties, k, nswaps = 0.0, 0.0, 0.0, 0, 0
+    ntiesx, ntiesy, ndoubleties, k, nswaps = widen(0), widen(0), widen(0), widen(0), widen(0)
 
     @inbounds for i = 2:n
         if x[i - 1] == x[i]
@@ -57,14 +57,14 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
             # sorted first on x, then (where x values are tied) on y. Hence 
             # double ties can be counted by calling countties.
             sort!(view(y, (i - k - 1):(i - 1)))
-            ntiesx += float(k) * (k + 1) / 2
+            ntiesx += div(k * (k + 1), 2)
             ndoubleties += countties(y,  i - k - 1, i - 1)
             k = 0
         end
     end
     if k > 0
         sort!(view(y, ((n - k):n)))
-        ntiesx += float(k) * (k + 1) / 2
+        ntiesx += div(k * (k + 1), 2)
         ndoubleties += countties(y,  n - k, n)
     end
 
@@ -72,7 +72,7 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
     ntiesy = countties(y, 1, n)
 
     (npairs + ndoubleties - ntiesx - ntiesy - 2 * nswaps) /
-     sqrt((npairs - ntiesx) * (npairs - ntiesy))
+     sqrt(float(npairs - ntiesx) * float(npairs - ntiesy))
 end
 
 """
@@ -81,20 +81,20 @@ end
 Return the number of ties within `x[lo:hi]`. Assumes `x` is sorted. 
 """
 function countties(x::AbstractVector, lo::Integer, hi::Integer)
-    # avoid overflows on 32 bit by using floats
-    thistiecount, result = 0.0, 0.0
+    # avoid overflows on both 32 & 64 bit by using widen
+    thistiecount, result = widen(0), widen(0)
     (lo < 1 || hi > length(x)) && error("Bounds error")
     @inbounds for i = (lo + 1):hi
         if x[i] == x[i - 1]
-            thistiecount += 1.0
+            thistiecount += 1
         elseif thistiecount > 0
-            result += thistiecount * (thistiecount + 1) / 2
-            thistiecount = 0.0
+            result += div(thistiecount * (thistiecount + 1), 2)
+            thistiecount = 0
         end
     end
 
     if thistiecount > 0
-        result += thistiecount * (thistiecount + 1) / 2
+        result += div(thistiecount * (thistiecount + 1), 2)
     end
     result
 end
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -44,10 +44,11 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
     # Initial sorting
     permute!(x, permx)
     permute!(y, permx)
-
-    npairs = float(n) * (n - 1) / 2
-    # ntiesx, ntiesy, ndoubleties are floats to avoid overflows on 32bit
-    ntiesx, ntiesy, ndoubleties, k, nswaps = widen(0), widen(0), widen(0), widen(0), widen(0)
+    
+    # Use widen to avoid overflows on both 32bit and 64bit
+    npairs = widen(n) * (n - 1) / 2
+    ntiesx, ntiesy, ndoubleties, nswaps = widen(0), widen(0), widen(0), widen(0)
+    k = 0
 
     @inbounds for i = 2:n
         if x[i - 1] == x[i]
@@ -81,7 +82,7 @@ end
 Return the number of ties within `x[lo:hi]`. Assumes `x` is sorted. 
 """
 function countties(x::AbstractVector, lo::Integer, hi::Integer)
-    # avoid overflows on both 32 & 64 bit by using widen
+    # Use widen to avoid overflows on both 32bit and 64bit
     thistiecount, result = widen(0), widen(0)
     (lo < 1 || hi > length(x)) && error("Bounds error")
     @inbounds for i = (lo + 1):hi
@@ -113,7 +114,7 @@ corkendall(x::RealVector, Y::RealMatrix) = (n = size(Y, 2); permx = sortperm(x);
 
 function corkendall(X::RealMatrix)
     n = size(X, 2)
-    C = ones(float(eltype(X)), n, n)# avoids dependency on LinearAlgebra
+    C = ones(float(eltype(X)), n, n)# Avoids dependency on LinearAlgebra
     @inbounds for j = 2:n
         permx = sortperm(X[:,j])
         for i = 1:j - 1
@@ -152,8 +153,8 @@ This method is a copy-paste-edit of sort! in base/sort.jl (the method specialise
 but amended to return the bubblesort distance.
 """
 function msort!(v::AbstractVector, lo::Integer, hi::Integer, t=similar(v, 0))
-    # avoid overflow errors (if length(v)> 2^16) on 32 bit by using float
-    nswaps = 0.0
+    # Avoid overflow errors by using widen.
+    nswaps = widen(0)
     @inbounds if lo < hi
         hi - lo <= SMALL_THRESHOLD && return isort!(v, lo, hi)
 
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -64,7 +64,7 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
         end
     end
     if k > 0
-        sort!(view(y, ((n - k):n)))
+        sort!(view(y, (n - k):n))
         ntiesx += div(widen(k) * (k + 1), 2)
         ndoubleties += countties(y,  n - k, n)
     end
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -99,7 +99,7 @@ end
 
 function corkendall(X::RealMatrix)
     n = size(X, 2)
-    C = Matrix{float(eltype(X))}(I, n, n)
+    C = Matrix{Float64}(I, n, n)
     for j = 2:n
         permx = sortperm(X[:,j])
         for i = 1:j - 1
@@ -113,7 +113,7 @@ end
 function corkendall(X::RealMatrix, Y::RealMatrix)
     nr = size(X, 2)
     nc = size(Y, 2)
-    C = Matrix{float(eltype(X))}(undef, nr, nc)
+    C = Matrix{Float64}(undef, nr, nc)
     for j = 1:nr
         permx = sortperm(X[:,j])
         for i = 1:nc
@@ -140,7 +140,7 @@ function countties(x::AbstractVector, lo::Integer, hi::Integer)
             thistiecount += 1
         elseif thistiecount > 0
             result += div(thistiecount * (thistiecount + 1), 2)
-            thistiecount = w0
+            thistiecount = widen(0)
         end
     end
 
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -76,7 +76,7 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
 end
 
 """
-    countties(x::RealVector,lo::Int64,hi::Int64)
+    countties(x::RealVector, lo::Integer, hi::Integer)
 
 Assumes `x` is sorted. Returns the number of ties within `x[lo:hi]`.
 """
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -158,12 +158,12 @@ const SMALL_THRESHOLD = 64
 # merge_sort! copied from Julia Base
 # (commit 28330a2fef4d9d149ba0fd3ffa06347b50067647, dated 20 Sep 2020)
 """
-    merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t=similar(v, 0))    
+    merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t::AbstractVector=similar(v, 0))    
 
 Mutates `v` by sorting elements `x[lo:hi]` using the merge sort algorithm. 
 This method is a copy-paste-edit of sort! in base/sort.jl, amended to return the bubblesort distance.
 """
-function merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t=similar(v, 0))
+function merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t::AbstractVector=similar(v, 0))
     # Use of widen below prevents possible overflow errors when
     # length(v) exceeds 2^16 (32 bit) or 2^32 (64 bit)
     nswaps = widen(0)
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -47,7 +47,7 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
     
     # Use widen to avoid overflows on both 32bit and 64bit
     npairs = div(widen(n) * (n - 1), 2)
-    ntiesx, ndoubleties, nswaps = widen(0), widen(0), widen(0)
+    ntiesx = ndoubleties = nswaps = widen(0)
     k = 0
 
     @inbounds for i = 2:n
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -66,7 +66,7 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
     if k > 0
         sort!(view(y, (n - k):n))
         ntiesx += div(widen(k) * (k + 1), 2)
-        ndoubleties += countties(y,  n - k, n)
+        ndoubleties += countties(y, n - k, n)
     end
 
     nswaps = merge_sort!(y, 1, n)
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -57,7 +57,7 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
             # Sort the corresponding chunk of y, so the rows of hcat(x,y) are 
             # sorted first on x, then (where x values are tied) on y. Hence 
             # double ties can be counted by calling countties.
-            sort!(view(y, (i - k - 1):(i - 1)))  # Can't use wide integers here
+            sort!(view(y, (i - k - 1):(i - 1)))
             ntiesx += div(widen(k) * (k + 1), 2) # Must use wide integers here
             ndoubleties += countties(y,  i - k - 1, i - 1)
             k = 0
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -46,8 +46,8 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
     permute!(y, permx)
     
     # Use widen to avoid overflows on both 32bit and 64bit
-    npairs = widen(n) * (n - 1) / 2
-    ntiesx, ntiesy, ndoubleties, nswaps = widen(0), widen(0), widen(0), widen(0)
+    npairs = div(widen(n) * (n - 1), 2)
+    ntiesx, ndoubleties, nswaps = widen(0), widen(0), widen(0)
     k = 0
 
     @inbounds for i = 2:n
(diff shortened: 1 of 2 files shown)


src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
     end
-    return nSwaps
+    return nswaps
 end
 
+# This function is also copied from base/sort.jl
+midpoint(lo::T, hi::T) where T <: Integer = lo + ((hi - lo) >>> 0x01)
+midpoint(lo::Integer, hi::Integer) = midpoint(promote(lo, hi)...)
+
+# Copy was from https://github.com/JuliaLang/julia/commit/28330a2fef4d9d149ba0fd3ffa06347b50067647 dated 20 Sep 2020
```
nalimilan, line 198:
```suggestion
# Implementation copied from Julia Base
# (commit 28330a2fef4d9d149ba0fd3ffa06347b50067647, dated 20 Sep 2020)
```

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
+"""
+    isort!(v::AbstractVector, lo::Integer, hi::Integer)
+
+Mutates `v` by sorting elements `x[lo:hi]` using the insertion sort algorithm. 
+This method is a copy-paste-edit of sort! in base/sort.jl (the method specialised on InsertionSortAlg),
+amended to return the bubblesort distance.
+"""
+function isort!(v::AbstractVector, lo::Integer, hi::Integer)
```
nalimilan, line 206: Likewise, I would call this `insertion_sort!`.

test/rankcorr.jl
```diff
@@ -26,17 +26,63 @@ c22 = corspearman(x2, x2)
 
 # corkendall
 
-@test corkendall(x1, y) ≈ -0.105409255338946
-@test corkendall(x2, y) ≈ -0.117851130197758
+@test_throws ErrorException("Vectors must have same length") corkendall([1,2,3,4], [1,2,3])
+@test isnan(corkendall([1,2], [3,NaN]))
+@test isnan(corkendall([1,1,1], [1,2,3]))
```
nalimilan, line 31: Can you also test `Inf`, `-Inf` and `-0.0`?

test/rankcorr.jl
```diff
@@ -26,17 +26,63 @@ c22 = corspearman(x2, x2)
...
 
-@test corkendall(X, y) ≈ [-0.105409255338946, -0.117851130197758]
-@test corkendall(y, X) ≈ [-0.105409255338946 -0.117851130197758]
+@test corkendall(x1, y) == -1 / sqrt(90)
+@test corkendall(x2, y) == -1 / sqrt(72)
+@test corkendall(X, y) == [-1 / sqrt(90), -1 / sqrt(72)]
+@test corkendall(y, X) == [-1 / sqrt(90) -1 / sqrt(72)]
+
+n = 100_000
```
nalimilan, line 38: Please use a lower value if possible to avoid slowing down CI. Please also add comments to shortly explain what's the purpose of each series of tests when it's not obvious (e.g. overflow, small threshold...).

test/rankcorr.jl
```diff
@@ -26,17 +26,63 @@ c22 = corspearman(x2, x2)
...
+
 @test corkendall(X, X) ≈ [c11 c12; c12 c22]
 @test corkendall(X)    ≈ [c11 c12; c12 c22]
+
+@test corkendall(repeat(X, n), repeat(X, n)) ≈ [c11 c12; c12 c22]
+@test corkendall(repeat(X, n))               ≈ [c11 c12; c12 c22]
+
+@test corkendall(collect(1:n), collect(1:n)) == 1.0
+@test corkendall(collect(1:n), reverse(collect(1:n))) == -1.0
+@test isnan(corkendall(repeat([1], n), collect(1:n)))
```
nalimilan, line 60: Is `collect` really needed?

test/rankcorr.jl
```diff
@@ -26,17 +26,63 @@ c22 = corspearman(x2, x2)
...
+
+@test corkendall(repeat([0,1,1,0], n), repeat([1,0,1,0], n)) == 0.0
+
+z = [1  1  1;
+    1  1  2;
```
nalimilan, line 65: Fix indentation.

test/rankcorr.jl
```diff
@@ -26,17 +26,63 @@ c22 = corspearman(x2, x2)
...
+    1  2  2;
+    1  2  2;
+    1  2  1;
+    2  1  2;
+    1  1  2;
+    2  2  2]
+
+@test corkendall(z)   == [1 0 1 / 3; 0 1 0;1 / 3 0 1]
```
nalimilan, line 73: Here too. Also avoid spaces around `/` but use spaces after `;` for clarity.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
             i += 1
         end
     end
-    return nSwaps
+    return nswaps
 end
 
+# This function is also copied from base/sort.jl
+midpoint(lo::T, hi::T) where T <: Integer = lo + ((hi - lo) >>> 0x01)
+midpoint(lo::Integer, hi::Integer) = midpoint(promote(lo, hi)...)
```
nalimilan, line 196: Looks like you'll need a test with an array holding values of different types to test this line.

src/rankcorr.jl
```diff
@@ -28,71 +28,76 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
-            iU += 1
-        else
-            nU += iU*(iU - 1)
-            iU = 1
+    x[:] = x[permx]
+    y[:] = y[permx]
+
+    npairs = float(n) * (n - 1) / 2
+    # ntiesx, ntiesy, ndoubleties are floats to avoid overflows on 32bit
+    ntiesx, ntiesy, ndoubleties, k, nswaps = 0.0, 0.0, 0.0, 0, 0
```
nalimilan, line 50: AFAICT overflow can happen also on 64-bit with vectors larger than `sqrt(typemax(Int))`, right? `npairs` is also a float for this reason so the comment should mention it.

More fundamentally, are we sure that even `Float64` is large enough to avoid loss of precision in extreme cases? For example, imagine a vector with 1G entries (taking 8GB in RAM): IIUC it can happen that we will try to compute a number of the order of `1e9^2`, and yet `1e9^2 - (1e9^2+1) == 0` so everything is going to collapse in that case right? Maybe using `widen(Int)`/`widen(0)` would be safer, so that we use `Int128` on 64-bit machines?

nalimilan (commented):
Thanks. Looks mostly good, the main issue to sort out is overflow I think

src/rankcorr.jl
PGS62: Alternatively I think we could delete that second method of midpoint, I don't think it can ever be called. I will doublecheck and then either add the test you suggest or else delete the method.

test/rankcorr.jl
PGS62: I have a feeling that the not (agree not clear) spaces around the `/` might have been added by the "Julia: Format Document" command in VSCode.

test/rankcorr.jl
PGS62: I think so, since `corkendall` is not defined for `AbstractRange`:

```
julia> StatsBase.corkendall(1:10,1:10)
ERROR: setindex! not defined for StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}
```

src/rankcorr.jl
PGS62: Yes you are right. The issue was a nagging worry at the time, and I should have asked your advice about it.  There are (at least) two pitfalls that need to be avoided:
1. The type we use doesn't allow us to "count high enough" because of the `(x+1) == x` problem. I think a Float64 "stops working as a counter" at 2^53. So we must have `npairs <2^53` which in turn means we must have `length(x) < 2^27`. So that means we should abandon Float64 and go with your suggestion of `widen(Int)` with counters initialised to `widen(0)`.
2. The risk that the last line of function `corkendall` fails due to overflow problems.

I will make some changes and re-submit.

test/rankcorr.jl
PGS62: I've tidies up those spaces as requested. But yes, VSCode command  "Format Document" does seem to put the spaces in, I'll have to look into that...

src/rankcorr.jl
PGS62: Test was needed, so added it.

src/rankcorr.jl
```diff
@@ -28,71 +28,76 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
 
 
 #######################################
-#
+# 
 #   Kendall correlation
-#
+#
```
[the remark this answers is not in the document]
PGS62: ugh, that was VSCode auto formatter. Fixed now.

src/rankcorr.jl
PGS62: OK, that's submitted now, but tests are failing, and I think I see why. Was a mistake to initialise `k` to wide. Weird thing is that tests passed on my PC (Windows 64bit)

PGS62: Ah I see my prior comment looks a bit cryptic. I'd previously written a long comment basically agreeing with your analysis, and somehow that got lost. Sorry.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
-    xl = view(x, 1:n2)
-    xr = view(x, n2+1:n)
-    nsl = swaps!(xl)
-    nsr = swaps!(xr)
-    sort!(xl)
-    sort!(xr)
-    return nsl + nsr + mswaps(xl,xr)
-end
+# Tests appear to show that a value of 64 is optimal,
+# but note that the equivalent constant in base/sort.jl is 20.
```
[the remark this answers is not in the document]
PGS62: Yes, that's not a bad hypothesis. In my tests 64 was optimal, but not by much so a plot of SMALL_THRESHOLD vs speed showed a "flat-bottomed minimum" if that makes sense.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
-    C = Matrix{eltype(X)}(I, n, n)
-    for j = 2:n
-        for i = 1:j-1
-            C[i,j] = corkendall!(X[:,i],X[:,j])
-            C[j,i] = C[i,j]
+    C = ones(float(eltype(X)), n, n)# avoids dependency on LinearAlgebra
+    @inbounds for j = 2:n
+        permx = sortperm(X[:,j])
+        for i = 1:j - 1
+            C[j,i] = corkendall!(X[:,j], X[:,i], permx)
```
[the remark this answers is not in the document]
PGS62: I tried this and it doesn't work, the problem being that `Xj` gets mutated on the first pass through the loop hence the result is wrong for second and subsequent passes through the loop. In other words the copy really is necessary.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
+corkendall(x::RealVector, Y::RealMatrix) = (n = size(Y, 2); permx = sortperm(x); reshape(Float64[corkendall!(float(copy(x)), float(Y[:,i]), permx) for i in 1:n], 1, n))
 
 function corkendall(X::RealMatrix)
     n = size(X, 2)
-    C = Matrix{eltype(X)}(I, n, n)
-    for j = 2:n
-        for i = 1:j-1
-            C[i,j] = corkendall!(X[:,i],X[:,j])
-            C[j,i] = C[i,j]
+    C = ones(float(eltype(X)), n, n)# avoids dependency on LinearAlgebra
```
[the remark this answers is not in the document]
PGS62: OK, so I will revert to

`C = Matrix{eltype(X)}(I, n, n)`

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
 """
 corkendall(x::RealVector, y::RealVector) = corkendall!(float(copy(x)), float(copy(y)))
 
-corkendall(X::RealMatrix, y::RealVector) = Float64[corkendall!(float(X[:,i]), float(copy(y))) for i in 1:size(X, 2)]
-
-corkendall(x::RealVector, Y::RealMatrix) = (n = size(Y,2); reshape(Float64[corkendall!(float(copy(x)), float(Y[:,i])) for i in 1:n], 1, n))
+corkendall(X::RealMatrix, y::RealVector) = (permy = sortperm(y);Float64[corkendall!(float(copy(y)), float(X[:,i]), permy) for i in 1:size(X, 2)])
```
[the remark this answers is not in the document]
PGS62: I think you are correct that the `float` calls are not necessary. Also I don't think that call to `Float64` is necessary either. Both were inherited from the existing code. I will get rid of them, assuming that doesn't cause test failures.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
+            C[i,j] = C[j,i]
+        end
+    end
+    return C
+end
+
+function corkendall(X::RealMatrix, Y::RealMatrix)
+    nr = size(X, 2)
+    nc = size(Y, 2)
+    C = zeros(float(eltype(X)), nr, nc)
```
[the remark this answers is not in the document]
PGS62: Yes, I agree and will amend.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
 
 function corkendall(X::RealMatrix)
     n = size(X, 2)
-    C = Matrix{eltype(X)}(I, n, n)
-    for j = 2:n
-        for i = 1:j-1
-            C[i,j] = corkendall!(X[:,i],X[:,j])
-            C[j,i] = C[i,j]
+    C = ones(float(eltype(X)), n, n)# avoids dependency on LinearAlgebra
+    @inbounds for j = 2:n
```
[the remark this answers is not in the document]
PGS62: Yes, I agree and will amend.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
+        end
+    end
+    return C
+end
+
+function corkendall(X::RealMatrix, Y::RealMatrix)
+    nr = size(X, 2)
+    nc = size(Y, 2)
+    C = zeros(float(eltype(X)), nr, nc)
+    @inbounds for j = 1:nr
```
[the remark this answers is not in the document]
PGS62: Agreed.

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
-    xr = view(x, n2+1:n)
-    nsl = swaps!(xl)
-    nsr = swaps!(xr)
-    sort!(xl)
-    sort!(xr)
-    return nsl + nsr + mswaps(xl,xr)
-end
+# Tests appear to show that a value of 64 is optimal,
+# but note that the equivalent constant in base/sort.jl is 20.
+const SMALL_THRESHOLD  = 64
```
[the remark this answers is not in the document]
PGS62: agreed

src/rankcorr.jl
```diff
@@ -102,52 +107,118 @@ matrices or vectors.
...
-    i = 1
-    j = 1
-    nSwaps = 0
-    n = length(x)
-    while i <= n && j <= length(y)
-        if y[j] < x[i]
-            nSwaps += n - i + 1
+# Copy was from https://github.com/JuliaLang/julia/commit/28330a2fef4d9d149ba0fd3ffa06347b50067647 dated 20 Sep 2020
+"""
+    msort!(v::AbstractVector, lo::Integer, hi::Integer, t=similar(v, 0))
```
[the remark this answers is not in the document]
PGS62: Agreed

PGS62:
I will do another commit that incorporates all your suggestions except:

1. Suggestions you made to avoid copies where the copy was in fact necessary.
2. Enhancements to tests. I will look at those next.

Please don't merge the pull request just yet even if the next batch of tests pass.

Also I feel I've not been using GitHub quite as it's intended. I probably should have made more use of the "Commit Suggestion" buttons above, and I'm not sure what the "Add suggestion to batch" buttons do so I didn't use them either...

src/rankcorr.jl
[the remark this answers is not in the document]
nalimilan: Ah indeed. And what if you passed `permx = 1:length(Xj)` when `i > 1`? Wouldn't that be faster?

src/rankcorr.jl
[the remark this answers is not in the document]
PGS62: in fact reverted to 

`C = Matrix{float(eltype(X))}(I, n, n)`

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
-            iU += 1
-        else
-            nU += iU*(iU - 1)
-            iU = 1
+    permute!(x, permx)
+    permute!(y, permx)
+    
+    # Use widen to avoid overflows on both 32bit and 64bit
+    npairs = div(widen(n) * (n - 1), 2)
+    ntiesx, ndoubleties, nswaps = widen(0), widen(0), widen(0)
```
nalimilan, line 50:
```suggestion
    ntiesx = ndoubleties = nswaps = widen(0)
```

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
+    k = 0
+
+    @inbounds for i = 2:n
+        if x[i - 1] == x[i]
+            k += 1
+        elseif k > 0
+            # Sort the corresponding chunk of y, so the rows of hcat(x,y) are 
+            # sorted first on x, then (where x values are tied) on y. Hence 
+            # double ties can be counted by calling countties.
+            sort!(view(y, (i - k - 1):(i - 1)))  # Can't use wide integers here
```
nalimilan, line 60: What does the comment mean? We don't need wide integers here anyway.
```suggestion
            sort!(view(y, (i - k - 1):(i - 1)))
```

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
-    jV = 1
-    for i = 2:n
-        if x[i] == x[i-1] && y[i] == y[i-1]
-            iV += 1
-        else
-            nV += iV*(iV - 1)
-            iV = 1
-        end
+    if k > 0
+        sort!(view(y, ((n - k):n)))
```
nalimilan, line 67:
```suggestion
        sort!(view(y, (n - k):n))
```

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
+        ntiesx += div(widen(k) * (k + 1), 2)
+        ndoubleties += countties(y,  n - k, n)
```
nalimilan, line 69:
```suggestion
        ndoubleties += countties(y, n - k, n)
```

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
-    nD = div(n*(n - 1),2)
-    return (nD - nT - nU + nV - 2swaps!(y)) / (sqrt(nD - nT) * sqrt(nD - nU))
-end
+    nswaps = merge_sort!(y, 1, n)
+    ntiesy = countties(y, 1, n)
 
+    # Calls to float below prevent possible overflow errors when
+    # length(x) exceeds 77_936 (32 bit) or 5_107_605_667 (64 bit)
+    (npairs + ndoubleties - ntiesx - ntiesy - 2 * nswaps) /
+     sqrt(float(npairs - ntiesx) * float(npairs - ntiesy))
```
nalimilan, line 78: Is overflow still possible given that we use `widen`?

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
+function corkendall(x::RealVector, Y::RealMatrix)
+    n = size(Y, 2)
+    permx = sortperm(x)
+    return(reshape([corkendall!(copy(x), Y[:,i], permx) for i in 1:n], 1, n))
+end
 
 function corkendall(X::RealMatrix)
     n = size(X, 2)
-    C = Matrix{eltype(X)}(I, n, n)
+    C = Matrix{float(eltype(X))}(I, n, n)
```
nalimilan, line 102: Actually it's kind of weird that the existing code uses `eltype(X)`: for rank correlation, the element type doesn't matter (ranks are always `Int`). For example if values are using a Unitful type, we don't want the correlation to be expressed e.g. in meters (which is what would happen currently).  So I guess we'd better use `Float64`, which is what happens in most cases currently supported (except `BigInt` inputs).

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
+"""
+    countties(x::RealVector, lo::Integer, hi::Integer)
+
+Return the number of ties within `x[lo:hi]`. Assumes `x` is sorted. 
+"""
+function countties(x::AbstractVector, lo::Integer, hi::Integer)
+    # Use of widen below prevents possible overflow errors when
+    # length(x) exceeds 2^16 (32 bit) or 2^32 (64 bit)
+    w0 = widen(0)
+    thistiecount, result = w0, w0
```
nalimilan, line 137:
```suggestion
    thistiecount = result = widen(0)
```

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
-    sort!(xr)
-    return nsl + nsr + mswaps(xl,xr)
+"""
+    countties(x::RealVector, lo::Integer, hi::Integer)
+
+Return the number of ties within `x[lo:hi]`. Assumes `x` is sorted. 
+"""
+function countties(x::AbstractVector, lo::Integer, hi::Integer)
+    # Use of widen below prevents possible overflow errors when
+    # length(x) exceeds 2^16 (32 bit) or 2^32 (64 bit)
```
nalimilan, line 135: This comment would also be appropriate above, right?

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
+    w0 = widen(0)
+    thistiecount, result = w0, w0
+    checkbounds(x, lo:hi)
+    @inbounds for i = (lo + 1):hi
+        if x[i] == x[i - 1]
+            thistiecount += 1
+        elseif thistiecount > 0
+            result += div(thistiecount * (thistiecount + 1), 2)
```
nalimilan, line 143: Note that `div` might be costly, so if performance really matters in this loop it may be worth trying `>> 1` instead. Same in similar places.

src/rankcorr.jl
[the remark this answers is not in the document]
PGS62: If I understand your suggestion, you mean something like:

```julia
    n = size(X, 2)
    C = Matrix{float(eltype(X))}(I, n, n)
    for j = 2:n
        permx = sortperm(X[:,j])
        Xj = X[:,j]
        for i = 1:j - 1
            if i == 2
                permx = 1:length(Xj)
            end
            C[j,i] = corkendall!(Xj, X[:,i], permx)
            C[i,j] = C[j,i]
        end
    end
    return C
end

```

But that will yield incorrect answers. `Xj` would still be permuted on the first pass through the `for i = 1:j-1` loop so on the second pass through the loop the the order of elements in `Xj` has changed (they've been sorted) but the order of elements in `X[:j+1]` is unchanged. So the (Kendall) correlation `Xj` and `X[:j+1]` isn't the quantity we want to calculate. I don't think a changed value of `permx` can rescue that.

src/rankcorr.jl
```diff
@@ -33,121 +33,204 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
+
+# merge_sort! copied from Julia Base
+# (commit 28330a2fef4d9d149ba0fd3ffa06347b50067647, dated 20 Sep 2020)
+"""
+    merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t=similar(v, 0))    
+
+Mutates `v` by sorting elements `x[lo:hi]` using the merge sort algorithm. 
+This method is a copy-paste-edit of sort! in base/sort.jl, amended to return the bubblesort distance.
+"""
+function merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t=similar(v, 0))
```
nalimilan, line 166:
```suggestion
    merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t::AbstractVector=similar(v, 0))    

Mutates `v` by sorting elements `x[lo:hi]` using the merge sort algorithm. 
This method is a copy-paste-edit of sort! in base/sort.jl, amended to return the bubblesort distance.
"""
function merge_sort!(v::AbstractVector, lo::Integer, hi::Integer, t::AbstractVector=similar(v, 0))
```

test/rankcorr.jl
PGS62: Done

src/rankcorr.jl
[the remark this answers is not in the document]
nalimilan: Ah yes you would have to do `Xi = permute!(X[:,i], permx); corkendall!(Xj, Xi, 1:length(Xj))`. Right? Not sure that's really faster.

src/rankcorr.jl
PGS62: Yes, and that fact took me by surprise when 32 bit tests failed without those calls to float.

On 32 bit, `npairs` etc will be of type `Float64`. If `length(x) = 77_937` then

`npairs = 77937 * 77936 / 2 = 3_037_049_016`

so (absent ties) without the calls to `float` we would have the denominator being:

9_223_666_725_586_568_256

which is greater than (2^63)-1 = typemax(Int64) = 

9_223_372_036_854_775_807

On 64 bit the limit on `length(x)` is much greater at 5_107_605_667 (if my calculations are correct).

nalimilan: OK. Though on 32-bit `npairs` is `Int64`, not `Float64`, right?

Anyway, to avoid overflow, isn't the simplest approach to do `sqrt(npairs - ntiesx) * sqrt(npairs - ntiesy)`? Or even `... / sqrt(npairs - ntiesx) / sqrt(npairs - ntiesy)`.

src/rankcorr.jl
PGS62: OK, I've made that change and pushed it to GitHub

src/rankcorr.jl
PGS62: Ah yes, `Int64`, not `Float64`.

I think  writing `sqrt(float(npairs - ntiesx) * float(npairs - ntiesy))` has the advantage of not letting the result go very slighly outside the range -1 to +1 - wasn't that a minor bug that I fixed quite early on?

nalimilan: Not sure. But we can call `clamp` at the end of needed. Are you sure using `float` ensures the result is in the correct range?

PGS62: I'm reasonably sure, yes. If both `x` and `y` have the same ranking then it follows that `nswaps == 0`. If in addition there are no ties then the result (with the current code) becomes equivalent to evaluating:

`npairs/sqrt(npairs * npairs)`

and I would imagine that always evaluates to exactly 1.0 (since `npairs` is an integer)

if there _are_ ties then we have `tiesx == tiesy` and also `ndoubleties == tiesx^2` so the result becomes

`(npairs - tiesx)/sqrt((npairs-tiesx) * (npairs-tiesx))` and once again that will evaluate to exactly 1.0

I think a similar argument applies if `x` and `y` are in exactly opposite order except in that case `nswaps == 2npairs`

So the current code will definitely return exactly 1.0 or -1.0 when it should do so. Not quite the same as proving that the return from the function is always in the range [-1 to 1] I admit.

If we were to change to 

```julia
(npairs + ndoubleties - ntiesx - ntiesy - 2 * nswaps) /
     sqrt(npairs - ntiesx) / sqrt(npairs - ntiesy)

```

then you get
```
julia> corkendall(collect(1:5),collect(1:5))
0.9999999999999999

```
which of course can't be fixed by `clamp`. 

All this makes me think the current code is worth sticking with i.e.

```julia
  (npairs + ndoubleties - ntiesx - ntiesy - 2 * nswaps) /
     sqrt(float(npairs - ntiesx) * float(npairs - ntiesy))

```

nalimilan: Thanks for the analysis. I would have thought we could run into problems with values higher than `sqrt(maxintfloat(Float64))` due to rounding, but I can't find an example where it doesn't work.

src/rankcorr.jl
PGS62: I was curious so I tested replacing calls to `div( foo,2)` with `foo >> 1` using the same `speedtest` routine that I use to compare the old and new versions of `corkendall`.  I couldn't detect any speed difference.

Also `@btime` doesn't detect a difference:
```
function speedtest_divide()
  x = 1000
  @btime div($x,2)
  @btime $x >> 1
end

julia> KendallTau.speedtest_divide()
  0.001 ns (0 allocations: 0 bytes)
  0.001 ns (0 allocations: 0 bytes)

```
so I think we can stick with `div`

test/rankcorr.jl
PGS62: I've used a slightly lower value n = 78_000 then a bit later in the code redefined n to 100. I think test/rankcorr.jl executes in under a second (vs 7 seconds before).

test/rankcorr.jl
PGS62: That's done. Of course `Inf` and `-Inf` in input data are quite benign since they can be safely compared to finite values.

src/rankcorr.jl
```diff
@@ -28,71 +28,76 @@ corspearman(X::RealMatrix) = (Z = mapslices(tiedrank, X, dims=1); cor(Z, Z))
...
 
+"""
+    countties(x::RealVector,lo::Int64,hi::Int64)
+
+Assumes `x` is sorted. Returns the number of ties within `x[lo:hi]`.
+"""
+function countties(x::AbstractVector, lo::Integer, hi::Integer)
+    # avoid overflows on 32 bit by using floats
+    thistiecount, result = 0.0, 0.0
+    (lo < 1 || hi > length(x)) && error("Bounds error")
```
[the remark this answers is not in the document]
PGS62: Nice. I didn't know `checkbounds`

PGS62:
I've made changes you suggested to test/rankcorr.jl. 

I don't think there's any outstanding reason to not merge this PR now. Do you agree? I've clicked all the "Resolve conversation" buttons but mysteriously some of them still read "Resolve conversation" (rather than their other state of "Unresolve conversation"). Do both parties to a conversation need to click the button?

PGS62:
I would think that once #656 is merged in then all checks should pass for this PR as well. Any chance it can be merged in too?

nalimilan:
CI failures on nightly are only due to #656 so I'll merge. Thanks for this work! Once I'll have merged my `pairwise` PR I think it will be interesting to add specialized `pairwise` methods for `corkendall` to compute `sortperm(x)` only once for each variable. It would also be interesting to use multithreading, e.g. spawning one thread for each cell in the correlation table.

PGS62:
I enjoyed working on it and learned a great deal from your feedback, for which thank you!

I think one needs to be a bit careful with threading, on which I did do some work but never got to a stage that I was happy with. My original objective was to calculate `corkendall` for an input matrix with 10,000 columns so the output would be a 10,000 x 10,000 symmetric matrix. I couldn't get spawning 50,000,000 threads to work at all - Julia crashed. So I also tried one thread per column of the output (seemed to work OK ) and one thread per "chunk of columns" (this approach still a work in progress).

As we briefly discussed, I also need to be able to handle missing values where to calculate the correlation between two vectors one first filters out pairs for which either vector has missing elements. I hope the `sortperm` trick can be tweaked to work in this case, but I haven't thought about it yet.

nalimilan:
Yes we should probably never create much more tasks than `Threads.nthreads()`.

> As we briefly discussed, I also need to be able to handle missing values where to calculate the correlation between two vectors one first filters out pairs for which either vector has missing elements. I hope the `sortperm` trick can be tweaked to work in this case, but I haven't thought about it yet.

At least for the listwise method at https://github.com/JuliaStats/StatsBase.jl/pull/627 it shouldn't be hard since it calls `_pairwise!` one views of vectors without missing entries, so everything should work automatically. When skipping missing values in a pairwise fashion the implementation would be a bit trickier.

PR #648, opened 2021-01-27
Improve manual for data transformations

nalimilan:
`standardize` allows both Z-score normalization (a.k.a. as standardization) and unit range normalization.
This can be confusing, so avoid saying "standardization" without more explicit terms.

I wonder whether we should deprecate `standardize` in favor of overloading `LinearAlgebra.normalize` (which we already use for histograms).


mkborregaard (approved):
Great!

ararslan (approved):

Issue #518, opened 2019-09-07
Wrong weighted sum and mean results for complex numbers


aplavin:
Weighted `mean` gives wrong (conjugated) result for `complex` arrays: `mean([1 + 0.5im], weights([1])) == 1.0 - 0.5im`.

nalimilan:
Indeed. That's because we use `dot([1 + 0.5im], [1])` to compute the weighted sum (for performance reasons). I guess we could simply do `dot([1], [1 + 0.5im])` since only the first argument is conjugated, and we know weights are real.

Would you make a pull request to apply that change (in weights.jl) and add a test?

oschulz:
Just ran into this in another context (sum/mean on vectors of vectors): #252 

I think we should change the defintion of sum(A, w) to something like sum(A .* values(w)) or mapreduce(*, +, A, values(w)). I can do a PR, if there are no objections.

nalimilan:
The reason we use `dot` AFAICT is that it's much more efficient thanks to BLAS. It's not an issue to abuse it if we are sure it's correct in all cases in which we use it. What we can do is to have a method using it for arrays of numbers, and a fall back method using a slower but generally correct approach.

oschulz:
Sounds good. Regarding on where do draw the line - BLAS would only accelerate `Real`, right (resp. do the wrong thing for `Complex` for this use case)?

So we'd do

```julia
Base.sum(v::AbstractArray{<:Real}, w::AbstractWeights) = dot(v, values(w))

Base.sum(v::AbstractArray, w::AbstractWeights) = ...
```

right? @nalimilan, what do you think would be best for the generic implementation:

```julia
# Option 1:
sum(A .* values(w))

# Option 2:
sum(Base.Broadcast.Broadcasted(*, (A, values(w))))

# Option 3:
mapreduce(*, +, A, values(w))
```

oschulz:
I've been thinking - maybe we should drop use of `dot` altogether, though, even if it means a performance loss, since we'd gain precision. AFAIK, Julia's `sum()` uses pairwise summation, to keep sums stable when summing over values with large differences in magnitude. But I don't think BLAS  does fancy summing:

```
julia> A = Float32.(vcat(1.0, fill(1E-10, 10^6)))
julia> w = Weights(fill(Float32(1.0), length(A)))

julia> sum(A)
1.0000999f0

julia> sum(A, w), dot(A, values(w))
(1.0000969f0, 1.0000969f0)

[...]

PR #649, opened 2021-01-27

aplavin:
This fixes https://github.com/JuliaStats/StatsBase.jl/issues/518 following suggestion from @nalimilan. The linked issue also contains a long discussion on the best approach, but looks like no decision and fix followed from there. So here I just take the simplest approach to make the computations correct.

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -357,8 +357,8 @@ Base.:(==)(x::AbstractWeights, y::AbstractWeights)   = false
 
 Compute the weighted sum of an array `v` with weights `w`, optionally over the dimension `dim`.
 """
-wsum(v::AbstractVector, w::AbstractVector) = dot(v, w)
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks. That discussion is indeed quite long. After reading it again, I think that the best solution is to use `sum(Base.Broadcast.instantiate(Base.Broadcast.Broadcasted(*, (x, y))))` on Julia 1.4 and above for the generic, non-`Real` method (see https://github.com/JuliaStats/StatsBase.jl/issues/518#issuecomment-552009514). The simplest approach is probably to restrict the `dot` methods you have in this PR to `Real` (or `Number` of they are faster for `Complex` than broadcasting), and use `broadcast` in the general case.

andreasnoack:
Superseded by https://github.com/JuliaStats/StatsBase.jl/pull/658

PR #653, opened 2021-02-03
use docdeploy action instead of manual doc script

kleinschmidt:


nalimilan (approved):
I have no idea whether it's right, but if you say it is... :-p

nalimilan:
It worked! https://github.com/JuliaStats/StatsBase.jl/runs/1823082383?check_suite_focus=true

PR #658, opened 2021-02-10

aplavin:
Update to https://github.com/JuliaStats/StatsBase.jl/pull/649 following suggestion by @nalimilan


src/weights.jl
```diff
@@ -357,8 +357,9 @@ Base.:(==)(x::AbstractWeights, y::AbstractWeights)   = false
 
 Compute the weighted sum of an array `v` with weights `w`, optionally over the dimension `dim`.
 """
-wsum(v::AbstractVector, w::AbstractVector) = dot(v, w)
-wsum(v::AbstractArray, w::AbstractVector) = dot(vec(v), w)
+wsum(v::AbstractVector, w::AbstractVector) = sum(Base.Broadcast.instantiate(Base.Broadcast.Broadcasted(*, (v, w))))
```
nalimilan, line 360: AFAICT the first argument needs to be `v::AbstractArray`, right? (This will also need tests.)

nalimilan (commented):
Thanks. have you tried some benchmarking? `dot` doesn't seem to be faster anymore on a Julia 1.6, so I wonder whether we could always use `broadcast` when `VERSION >= v"1.6.0"` to be able to simplify code later.

```julia
julia> using LinearAlgebra, BenchmarkTools

julia> x = rand(10_000);

julia> xm = Vector{Union{Float64, Missing}}(x);

julia> w = rand(10_000);

julia> f(A, w) = sum(Base.Broadcast.instantiate(Base.Broadcast.Broadcasted(*, (A, w))))

julia> @btime dot(x, w);
  2.209 μs (1 allocation: 16 bytes)

julia> @btime f(x, w);
  2.472 μs (1 allocation: 16 bytes)

julia> @btime dot(xm, w);
  20.162 μs (1 allocation: 16 bytes)

julia> @btime f(xm, w);
  23.823 μs (1 allocation: 16 bytes)
```

src/weights.jl
aplavin: I don't really understand all this `Broadcast....` functions, just copied it from your suggestion. If you say it should work for `AbstractArray` as well, I can modify the signature.
As for tests, again not sure what's the proper test should be. A simple test with weighted sum of vector-of-vectors is already in this PR.

aplavin:
I may try to make it always go through the broadcasting implementation, as you say. But I'm not familiar with how to use the version check so that there is no runtime penalty, could you give an example?

nalimilan:
Sorry for the late reply. You can just do `if VERSION > v"X.Y.Z"` (if `@static if` if not in the global scope).

src/weights.jl
nalimilan: I was just thinking about a test that passes a matrix as first argument. The existing methods accepts them, so that should continue to be supported after this PR.

aplavin: Tried changing to `AbstractArray`, but this method fails for multidimensional arrays anyway: with `"arrays could not be broadcast to a common size; got a dimension with lengths 2 and 4`.

aplavin:
Bump without any changes...

The broadcast-based method doesn't work for empty collections:

>   Expression: wsum(Float64[], Float64[]) === 0.0
>   ArgumentError: reducing over an empty collection is not allowed

It also doesn't handle multidim arrays, see my comment above. So I didn't add the `VERSION >= 1.6` condition as you suggested.

src/weights.jl
nalimilan: Have you tried with `vec(v)` as in the original method?

nalimilan:
Another option that I used at https://github.com/JuliaLang/Statistics.jl/pull/94 is to simply do `vec(v)' * w`.

PR #661, opened 2021-02-16
Allow computing the correlation matrix from the covariance matrix only

ASaragga:
Following issue #652 in StatsBase.jl

PR #662, opened 2021-02-20
Allow computing the correlation matrix from the covariance matrix only

ASaragga:


ParadaCarleton:
Closed as duplicate of #661

PR #664, opened 2021-02-25
Markdown show method for CoefTable

palday:
This adds a `show(::IO, ::MIME"text/markdown", ::CoefTable)` method. It uses a lot of the spacing logic from the plaintext method so that the Markdown table still looks nice as text, but the alignment will be much simpler when rendered. Markdown only supports left, center and right alignment, so the decimal alignment won't work, but at least with the formatting from `TestStat` and `PValue`, the right alignment for those columns will do the trick.


src/statmodels.jl
```diff
@@ -519,6 +519,56 @@ function show(io::IO, ct::CoefTable)
...
+    # note that we use left-alignment for all columns besides the stats
+    # because Markdown doesn't have a native "align on decimal"
+
+    print(io, "|", " "^sum(A[1]))
+    for j in 1:length(colnms)
+        print(io, "  |", lpad(colnms[j], sum(A[j+1])))
+    end
+
+    println(io, "  |")
+    print(io, "|",rpad(":", sum(A[1])+2,'-'))
```
nalimilan, line 555:
```suggestion
    print(io, '|', rpad(':', sum(A[1])+2, '-'))
```

src/statmodels.jl
```diff
@@ -519,6 +519,56 @@ function show(io::IO, ct::CoefTable)
...
+         for i in 1:length(A)]
+    totwidth = sum(sum.(A)) + 2 * (length(A) - 1)
+
+    #   is non breaking space, which means
+    # that we can use the padding logic from the plaintext show method
+    # and get markdown that looks as nice visually as it will render
+    # note that we use left-alignment for all columns besides the stats
+    # because Markdown doesn't have a native "align on decimal"
+
+    print(io, "|", " "^sum(A[1]))
```
nalimilan, line 549:
```suggestion
    print(io, '|', ' '^sum(A[1]))
```

src/statmodels.jl
```diff
@@ -519,6 +519,56 @@ function show(io::IO, ct::CoefTable)
...
+        print(io, "  |", lpad(colnms[j], sum(A[j+1])))
+    end
+
+    println(io, "  |")
+    print(io, "|",rpad(":", sum(A[1])+2,'-'))
+    for j in 1:length(colnms)
+        _pad = j-1 in [ct.teststatcol; ct.pvalcol] ? rpad : lpad
+        print(io, "|", _pad(":", sum(A[j+1])+2,'-'))
+    end
+    println(io, "|")
```
nalimilan, line 560:
```suggestion
    println(io, '|')
```

src/statmodels.jl
```diff
@@ -519,6 +519,56 @@ function show(io::IO, ct::CoefTable)
...
+    print(io, "|", " "^sum(A[1]))
+    for j in 1:length(colnms)
+        print(io, "  |", lpad(colnms[j], sum(A[j+1])))
+    end
+
+    println(io, "  |")
+    print(io, "|",rpad(":", sum(A[1])+2,'-'))
+    for j in 1:length(colnms)
+        _pad = j-1 in [ct.teststatcol; ct.pvalcol] ? rpad : lpad
+        print(io, "|", _pad(":", sum(A[j+1])+2,'-'))
```
nalimilan, line 558:
```suggestion
        print(io, '|', _pad(':', sum(A[j+1])+2, '-'))
```

src/statmodels.jl
```diff
@@ -519,6 +519,56 @@ function show(io::IO, ct::CoefTable)
...
+    A = Base.alignment(io, mat, 1:size(mat, 1), 1:size(mat, 2),
+                       typemax(Int), typemax(Int), 3)
+    nmswidths = pushfirst!(length.(colnms), 0)
+    A = [nmswidths[i] > sum(A[i]) ? (A[i][1]+nmswidths[i]-sum(A[i]), A[i][2]) : A[i]
+         for i in 1:length(A)]
+    totwidth = sum(sum.(A)) + 2 * (length(A) - 1)
+
+    #   is non breaking space, which means
+    # that we can use the padding logic from the plaintext show method
+    # and get markdown that looks as nice visually as it will render
```
nalimilan, line 545: I don't understand this part. What have non breaking spaces to do with Markdown?

test/statmodels.jl
```diff
@@ -18,6 +18,14 @@ x1    1.45666         Good    1  -12.56  0.1200
 x2  -23.14            Great  56    0.13  0.3467
 x3    1.56734e-13     Bad     2    0.00  <1e-15
 ───────────────────────────────────────────────"""
+
+@test sprint(show, MIME"text/markdown"(), ct) == """
+|    |     Estimate  |Comments  |df  |     t  |     p  |
+|:---|--------------:|---------:|---:|-------:|:-------|
```
nalimilan, line 24: Is `:|:` intended?

nalimilan (commented):
Thanks!

test/statmodels.jl
ararslan: Markdown tables use which side the `:` is on to denote justification, so `t` has one justification and `p` has the opposite

src/statmodels.jl
palday: Markdown itself completely ignores non-breaking spaces, but:
- They prevent linebreaks in a plain text editor (but otherwise look like the REPL `show` behavior)
- Depending on how that's converted into HTML, non-breaking spaces *may* be preserved, which would preserve the decimal point alignment (but no guarantees on that front), which would otherwise be impossible in Common Markdown.

I'm happy to change that back to normal spaces. I think it might be more clever than functional. :smile:

nalimilan: Mmm, I don't think we want spaces to be preserved when converting to HTML. That would only preserve alignment when using a fixed width font, which isn't very common in HTML, and that would prevent text editors or spreadsheets from treating such columns as numeric.

So yeah I think it would be better not to be too clever here. :-)

palday: Cleverness removed. :smile:

test/statmodels.jl
```diff
@@ -18,6 +18,14 @@ x1    1.45666         Good    1  -12.56  0.1200
 x2  -23.14            Great  56    0.13  0.3467
 x3    1.56734e-13     Bad     2    0.00  <1e-15
 ───────────────────────────────────────────────"""
+
+@test sprint(show, MIME"text/markdown"(), ct) == """
+|    |     Estimate  |Comments  |df  |     t  |     p  |
```
nalimilan, line 23: Sorry I hadn't noticed this: why add two spaces after the column names rather than a single one? That would align them with the contents of the columns.

palday: I think I've fixed this, but if it becomes too painful, I'll swap over to using the `Markdown` stdlib (which would lose the decimal alignment but reduce complexity).

src/statmodels.jl
```diff
@@ -542,6 +542,56 @@ function show(io::IO, ct::CoefTable)
     nothing
 end
 
+show(mime::MIME"text/markdown", ct::CoefTable) = show(Base.stdout, mime, ct)
```
nalimilan, line 545: Why add this method? AFAICT this form isn't documented in `?show`, and it's not tested either.

src/statmodels.jl
```diff
@@ -542,6 +542,56 @@ function show(io::IO, ct::CoefTable)
...
+
+function show(io::IO, ::MIME"text/markdown", ct::CoefTable)
+    cols = ct.cols; rownms = ct.rownms; colnms = ct.colnms;
+    nc = length(cols)
+    nr = length(cols[1])
+    if length(rownms) == 0
+        rownms = [lpad("[$i]",floor(Integer, log10(nr))+3) for i in 1:nr]
```
nalimilan, line 552: Looks like this isn't tested either?

src/statmodels.jl
```diff
@@ -542,6 +542,56 @@ function show(io::IO, ct::CoefTable)
...
+    end
+    mat = [j == 1 ? NoQuote(rownms[i]) :
+           j-1 == ct.pvalcol ? NoQuote(sprint(show, PValue(cols[j-1][i]))) :
+           j-1 in ct.teststatcol ? TestStat(cols[j-1][i]) :
```
nalimilan, line 556: Why is `NoQuote(sprint(...))` needed for `PValue` but not for `TestStat`?

test/statmodels.jl
nalimilan: This is OK I think.

src/statmodels.jl
palday: Because of the `<` when printing small `PValue`.

src/statmodels.jl
palday: This code was literally copy-and-pasted from the plaintext `show` method.

src/statmodels.jl
palday: Convenience, but I'm not attached to it, so happy to delete.

nalimilan: `methods(show)` doesn't include any similar method, so I'd rather remove it.

src/statmodels.jl
nalimilan: Yeah, but this could be tested by passing zero rows IIUC.

nalimilan (approved):

palday:
@nalimilan Should I bump the patch number and immediately do a patch release?

nalimilan:
Yeah, why not.

PR #668, opened 2021-03-12
Make PValue and TestStat behave like Reals

palday:
- Add comparison operator methods for `PValue` and `TestStat`
- Add tests for `show(::TestStat)`
- Make `PValue` a subtype of `Real`

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -460,19 +460,6 @@ struct PValue <: Real
 end
 PValue(p::PValue) = p
 
-float(p::PValue) = float(p.v)
-
-for op in [:(==), :≈, :<, :≤, :>, :≥, :(isless)] # isless and < to place nice with NaN
-    @eval begin
-        Base.$op(p::PValue, y::Real) = $op(p.v, y)
-        Base.$op(y::Real, p::PValue) = $op(y, p.v)
-        Base.$op(p1::PValue, p2::PValue) = $op(p1.v, p2.v)
-    end
-end
-
-# necessary to avoid a method ambiguity with isless(::PValue, NaN)
-Base.isless(p::PValue, y::AbstractFloat) = isless(p.v, y)
-
 function show(io::IO, pv::PValue)
     v = pv.v
     if isnan(v)
@@ -492,18 +479,28 @@ end
 show(io::IO, x::TestStat) = @printf(io, "%.2f", x.v)
 TestStat(x::TestStat) = x
 
-float(x::TestStat) = float(x.v)
+float(x::Union{TestStat, PValue}) = float(x.v)
 
-for op in [:(==), :≈, :<, :≤, :>, :≥, :(isless)] # isless and < to place nice with NaN
+for op in [:(==), :<, :≤, :>, :≥, :(isless), :(isequal)] # isless and < to place nice with NaN
     @eval begin
-        Base.$op(x::TestStat, y::Real) = $op(x.v, y)
-        Base.$op(y::Real, x::TestStat) = $op(y, x.v)
-        Base.$op(x1::TestStat, x2::TestStat) = $op(x1.v, x2.v)
+        Base.$op(x::Union{TestStat, PValue}, y::Real) = $op(x.v, y)
+        Base.$op(y::Real, x::Union{TestStat, PValue}) = $op(y, x.v)
+        Base.$op(x1::Union{TestStat, PValue}, x2::Union{TestStat, PValue}) = $op(x1.v, x2.v)
     end
 end
 
 # necessary to avoid a method ambiguity with isless(::TestStat, NaN)
-Base.isless(x::TestStat, y::AbstractFloat) = isless(x.v, y)
+Base.isless(x::Union{TestStat, PValue}, y::AbstractFloat) = isless(x.v, y)
+Base.isequal(x::Union{TestStat, PValue}, y::AbstractFloat) = isequal(x.v, y)
+
+for op in [:≈, :(isapprox)]
+    @eval begin
+        Base.$op(x::Union{TestStat, PValue}, y::Real; kwargs...) = $op(x.v, y; kwargs...)
+        Base.$op(y::Real, x::Union{TestStat, PValue}; kwargs...) = $op(y, x.v; kwargs...)
+        Base.$op(x1::Union{TestStat, PValue}, x2::Union{TestStat, PValue}; kwargs...) = $op(x1.v, x2.v; kwargs...)
+    end
+end
+
 
 """Wrap a string so that show omits quotes"""
 struct NoQuote
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -493,13 +493,10 @@ end
 Base.isless(x::Union{TestStat, PValue}, y::AbstractFloat) = isless(x.v, y)
 Base.isequal(x::Union{TestStat, PValue}, y::AbstractFloat) = isequal(x.v, y)
 
-for op in [:≈, :(isapprox)]
-    @eval begin
-        Base.$op(x::Union{TestStat, PValue}, y::Real; kwargs...) = $op(x.v, y; kwargs...)
-        Base.$op(y::Real, x::Union{TestStat, PValue}; kwargs...) = $op(y, x.v; kwargs...)
-        Base.$op(x1::Union{TestStat, PValue}, x2::Union{TestStat, PValue}; kwargs...) = $op(x1.v, x2.v; kwargs...)
-    end
-end
+
+Base.isapprox(x::Union{TestStat, PValue}, y::Real; kwargs...) = isapprox(x.v, y; kwargs...)
+Base.isapprox(y::Real, x::Union{TestStat, PValue}; kwargs...) = isapprox(y, x.v; kwargs...)
+Base.isapprox(x1::Union{TestStat, PValue}, x2::Union{TestStat, PValue}; kwargs...) = isapprox(x1.v, x2.v; kwargs...)
 
 
 """Wrap a string so that show omits quotes"""
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -460,6 +460,14 @@ struct PValue
 end
 PValue(p::PValue) = p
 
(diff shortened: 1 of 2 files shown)


src/statmodels.jl
```diff
@@ -460,6 +460,14 @@ struct PValue
 end
 PValue(p::PValue) = p
 
+float(p::PValue) = float(p.v)
+
+for op in [Symbol("=="), :≈, :<, :≤, :>, :≥]
```
nalimilan, line 465:
```suggestion
for op in [:(==), :≈, :<, :≤, :>, :≥]
```
Maybe also define `isequal` and `isless`?

Actually I wonder whether it would be simpler to define `promote_type(::Type{PValue{S}}, y::T) where {S, T<:Real} = promote_type(S, T)`, making `S` a type parameter with `v::S`. Then I think we'll get all operations on numbers defined for free.

palday: You get `isequal` and `isless` for free with the symbolic forms (and I use `isequal` and `isless`) in the tests.

I was also thinking about whether `promote_type` might be the easier route. I'll try that later. :)

nalimilan: Ah yes so `<` falls back to `isless`, but it's not correct for `NaN`, right?

palday: Uh.... good question

palday: Oh wait, I know why I didn't do `promote_type`: that would be technically breaking, both for `PValue` and `TestStat`. We're pre 1.0 so another release isn't a big deal. But that might be something to get one more opinion on (e.g. @ararslan).

(I also want `PValue` and `TestStat` to have a consistent interface between themselves since they're both essentially just pretty printers for a semantically annotated `Real`.)

nalimilan: You mean it would be breaking if you do something like `[pval, 1.0]`? Yes that would be slightly breaking so that would probably warrant tagging a minor release.

palday: Yeah, and adding a type parameter to a struct also breaks (de)serialization.

nalimilan: Maybe let's keep your current approach then, and only use `promote_type` in another PR that would be merged only once we have other breaking changes to tag.

src/statmodels.jl
```diff
@@ -460,6 +460,14 @@ struct PValue
...
+    @eval begin
+        Base.$op(p::PValue, y::Real) = $op(p.v, y)
```
ararslan, line 467: Presumably you also want
```suggestion
        Base.$op(p::PValue, y::Real) = $op(p.v, y)
        Base.$op(y::Real, p::PValue) = $op(y, p.v)
        Base.$op(p1::PValue, p2::Real) = $op(p1.v, p2.v)
```
?

src/statmodels.jl
```diff
@@ -478,6 +486,14 @@ end
 
 show(io::IO, x::TestStat) = @printf(io, "%.2f", x.v)
 
+float(x::TestStat) = float(x.v)
+
+for op in [Symbol("=="), :≈, :<, :≤, :>, :≥]
+    @eval begin
+        Base.$op(x::TestStat, y::Real) = $op(x.v, y)
```
ararslan, line 493: Same as above

src/statmodels.jl
palday: As soon as this one (and CoefTable show method) lands, I'll open that PR :smile

src/statmodels.jl
```diff
@@ -460,6 +460,19 @@ struct PValue
 end
 PValue(p::PValue) = p
 
+float(p::PValue) = float(p.v)
+
+for op in [:(==), :≈, :<, :≤, :>, :≥, :(isless)] # isless and < to place nice with NaN
```
ararslan, line 465: I think you'll also want `isequal` for `NaN` handling. And I think the `isapprox` definition should be handled a bit differently, as we want to be able to forward keyword arguments as well (i.e. `atol`/`rtol`) when not using the infix shorthand.

palday: `isequal` seemed to work fine locally with `NaN`, but it's easy to add in and test. I've moved `isapprox` to handle the kwargs.

src/statmodels.jl
```diff
@@ -492,18 +479,28 @@ end
...
+        Base.$op(x1::Union{TestStat, PValue}, x2::Union{TestStat, PValue}) = $op(x1.v, x2.v)
     end
 end
 
 # necessary to avoid a method ambiguity with isless(::TestStat, NaN)
-Base.isless(x::TestStat, y::AbstractFloat) = isless(x.v, y)
+Base.isless(x::Union{TestStat, PValue}, y::AbstractFloat) = isless(x.v, y)
+Base.isequal(x::Union{TestStat, PValue}, y::AbstractFloat) = isequal(x.v, y)
+
+for op in [:≈, :(isapprox)]
```
ararslan, line 496: `(≈) === isapprox`, so this is extending twice

src/statmodels.jl
```diff
@@ -477,6 +477,27 @@ struct TestStat <: Real
...
+
+for op in [:(==), :<, :≤, :>, :≥, :(isless), :(isequal)] # isless and < to place nice with NaN
+    @eval begin
+        Base.$op(x::Union{TestStat, PValue}, y::Real) = $op(x.v, y)
+        Base.$op(y::Real, x::Union{TestStat, PValue}) = $op(y, x.v)
+        Base.$op(x1::Union{TestStat, PValue}, x2::Union{TestStat, PValue}) = $op(x1.v, x2.v)
+    end
+end
+
+# necessary to avoid a method ambiguity with isless(::TestStat, NaN)
```
ararslan, line 492: So you only need this direction, not with `NaN` as the first argument?

palday: Woops, fix incoming

ararslan (approved):

test/statmodels.jl
```diff
@@ -63,6 +63,73 @@ end
...
 @test_throws ErrorException PValue(1.1)
 @test PValue(PValue(0.05)) === PValue(0.05)
+@test isless(PValue(0.01), 0.05)
+@test isless(PValue(0.01), NaN) == isless(0.01, NaN)
+@test (PValue(0.01) < NaN) == (0.01 < NaN)
+@test isless(NaN, PValue(0.01)) == isless(NaN, 0.01)
+@test (NaN < PValue(0.01)) == (NaN < 0.01)
+@test isequal(NaN, PValue(0.01)) == isequal(NaN, 0.01)
+@test (NaN == PValue(0.01)) == (NaN == 0.01)
+@test isequal(PValue(0.01), NaN) == isequal(0.01, NaN)
```
nalimilan, line 73: A missing test is `isequal(PValue(NaN), NaN)`. Without it, `isequal` could fall back to `==` without tests failing.

palday: Added in #669

nalimilan:
Damn, this is really tricky. Since we now define `isequal`, it would be good to check that `hash` is also consistent. Otherwise putting `PValue` objects in a dict will give incorrect results.

AFAICT that's the case thanks to the `hash(x::Real, h::UInt)` fallback, but the only way to be sure is to test it. Defining `hash(x::PValue, h::UInt) = hash(x.val, h)` would probably be more efficient.

Out of curiosity, in what use case did you feel the need to this? :-) I realize that `PValue` and `TestValue` were originally essentially simple wrappers to customize printing internally which were not exposed to users, but we're now making them full-fledged types.

palday:
> Damn, this is really tricky. Since we now define `isequal`, it would be good to check that `hash` is also consistent. Otherwise putting `PValue` objects in a dict will give incorrect results.
> 
> AFAICT that's the case thanks to the `hash(x::Real, h::UInt)` fallback, but the only way to be sure is to test it. Defining `hash(x::PValue, h::UInt) = hash(x.val, h)` would probably be more efficient.

Hahahahah, I'll add this as well. And as soon as we get this and the `show` method tagged for a patch release, I'll open a PR for the promote-rule which will be breaking but just avoid so much of this hassle. :smile: 

> Out of curiosity, in what use case did you feel the need to this? :-) I realize that `PValue` and `TestValue` were originally essentially simple wrappers to customize printing internally which were not exposed to users, but we're now making them full-fledged types.

We had something at the dayjob where there was some conditional behavior based on a p-value threshold (I know, I don't like it either, but so much of the applied stats world still does that). In one instance, the p-values were already wrapped for pretty printing (and expensive to re-compute) and we had to add the filter at the very end. When doing that, I noticed that `TestStat` already subtyped `Real`, so I thought it made sense to just say "okay, these are `Real`s with pretty printing (and bounds checking for `PValue`), so let's make them behave that way".  I guess I'm not the only person who will say "give me all the rows in this `CoefTable` where p < 0.05", so it will be nice to have comparison operators Just Work™.

nalimilan:
> Hahahahah, I'll add this as well. And as soon as we get this and the `show` method tagged for a patch release, I'll open a PR for the promote-rule which will be breaking but just avoid so much of this hassle. smile

Thanks!

> We had something at the dayjob where there was some conditional behavior based on a p-value threshold (I know, I don't like it either, but so much of the applied stats world still does that). In one instance, the p-values were already wrapped for pretty printing (and expensive to re-compute) and we had to add the filter at the very end. When doing that, I noticed that `TestStat` already subtyped `Real`, so I thought it made sense to just say "okay, these are `Real`s with pretty printing (and bounds checking for `PValue`), so let's make them behave that way". I guess I'm not the only person who will say "give me all the rows in this `CoefTable` where p < 0.05", so it will be nice to have comparison operators Just Work™.

OK. But do you mean you get `PValue` objects when extracting values from `CoefTable` objects? I thought it was used only internally for printing, but never returned to users.

palday:
Uh.... I don't know. (We're not using `CoefTable` at that point in the actual example I was looking at.)

ararslan:
...come to think of it, how _did_ we end up with `PValues`? :thinking:

palday:
A custom `show` method working on a column table where we assumed that the element-level pretty printing was handled via `show` for that `eltype`.

And when this comparison wasn't implemented, everything broke.

PR #669, opened 2021-03-13
NaN test for Real-ness for PValue and TestStat

palday:


nalimilan:
Thanks! Can you also test things like `isless(1.0, PValue(NaN))`?

palday:
I've lost track of which things I'm missing. Just make a list and I'll add them. :smile:

test/statmodels.jl
```diff
@@ -100,15 +103,17 @@ end
...
+@test isless(0.01, TestStat(NaN)) == isless(0.01, NaN)
 @test (TestStat(0.01) < NaN) == (0.01 < NaN)
 @test isless(NaN, TestStat(0.01)) == isless(NaN, 0.01)
 @test (NaN < TestStat(0.01)) == (NaN < 0.01)
 @test isequal(TestStat(0.01), NaN) == isequal(0.01, NaN)
 @test (TestStat(0.01) == NaN) == (0.01 == NaN)
 @test isequal(NaN, TestStat(0.01)) == isequal(NaN, 0.01)
 @test (NaN == TestStat(0.01)) == (NaN == 0.01)
+@test isequal(PValue(NaN), NaN) == isequal(NaN, NaN)
+@test (PValue(NaN) == NaN) == (NaN == NaN)
```
nalimilan, line 115:
```suggestion
@test isequal(TestStat(NaN), NaN) == isequal(NaN, NaN)
@test (TestStat(NaN) == NaN) == (NaN == NaN)
```

nalimilan:
I think you just need to symmetrize all tests you have already added, to make sure all methods are tested in all cases. That's already the case for most of them, but not all.

test/statmodels.jl
palday: Woops. Copy-paste is hard, but that's actually a clue that I should just make this into a big loop....

nalimilan (approved):

PR #673, opened 2021-03-20
fixed corkendall(RealMatrix, RealVector)

### LINKED ISSUE #672: corkendall and corspearman return different types
When computing the correlation between a matrix and a vector, `corspearman` returns a `10x1` matrix while `corkendall` returns a 10-element vector. There should be consistent return types between these two functions.

Minimum example:

```julia
julia> z = rand(100, 100)

julia> x = rand(100)

julia> corspearman(z, x)
10×1 Matrix{Float64}:
[...]

julia corkendall(z, x)
10-element Vector{Float64}:
[...]
```

When switching the input order (`corkendall(x, z)` or `corspearman(x, z)`), the return type for both is a `1x10 Matrix`, so I believe that `corkendall(Matrix, Vector)` should return a Matrix with 1 column.

adknudson:
corkendall now returns a matrix with one column when operating on a Matrix-Vector input, the same behavior as `cor` and `corspearman`. The tests have been updated to reflect the change.

Fixes #672 and is related to #659

### DIFF
--- a/src/rankcorr.jl
+++ b/src/rankcorr.jl
@@ -143,8 +143,8 @@ function corkendall!(x::RealVector, y::RealVector, permx::AbstractVector{<:Integ
         if x[i - 1] == x[i]
             k += 1
         elseif k > 0
-            # Sort the corresponding chunk of y, so the rows of hcat(x,y) are 
-            # sorted first on x, then (where x values are tied) on y. Hence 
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks. Unfortunately this is breaking so we cannot merge this until the next breaking release.

PR #677, opened 2021-03-24
fix sequential sampling with replacement

### LINKED ISSUE #675: incorrect ordering for `sample(replace=true, ordered=true)`
[This line](https://github.com/JuliaStats/StatsBase.jl/blob/3f9030e72b21541df24bbb1e3ef823ef00d312e7/src/sampling.jl#L413) in the `sample!` function looks incorrect.   It assumes that "ordered" corresponds to sorting the samples in ascending order by value, but this is only true if the array `a` that you are sampling from is itself sorted, which is not required by `sample!`.   Instead, it should be sorting by the order in which the samples appeared in `a`.

For example compare the following incorrect result:
```jl
julia> sample!(10:-1:1, zeros(Int,5), replace=true, ordered=true)
5-element Array{Int64,1}:
 2
 3
 5
 6
 9
```
to the correct ordering obtained for sampling with `replace=false`:
```jl
julia> sample!(10:-1:1, zeros(Int,5), replace=false, ordered=true)
5-element Array{Int64,1}:
 8
 7
 6
 5
 4
```

A correct algorithm would be to sample indices (with replacement), sort the indices, and then use these indices to index into the original array.  In general this will require an auxiliary array, unless the type of `x` is such that it can store both indices and values of `a`.

stevengj:
Fixes #675.   I also noticed that *weighted* sampling *without* replacement had the same problem, and I fixed it there too.

To do:
- [x] Tests
- [x] Improved documentation of the `ordered` keyword`.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -891,7 +891,8 @@ efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::AbstractWeights, x::Abstra
 
 function sample!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray;
                  replace::Bool=true, ordered::Bool=false)
-    1 == firstindex(a) == firstindex(wv) == firstindex(x) || throw(ArgumentError("non 1-based arrays are not supported"))
+    1 == firstindex(a) == firstindex(wv) == firstindex(x) ||
+        throw(ArgumentError("non 1-based arrays are not supported"))
     n = length(a)
     k = length(x)
 
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -944,9 +944,9 @@ sample(a::AbstractArray, wv::AbstractWeights, dims::Dims;
 
 Select a weighted sample from an array `a` and store the result in `x`. Sampling
 probabilities are proportional to the weights given in `w`. `replace` dictates
-whether sampling is performed with replacement. . `ordered` dictates whether
-an ordered sample (also called a sequential sample)—a sample where
-items appear in the same order as in `a`—should be taken.
+whether sampling is performed with replacement. `ordered` dictates whether
+an ordered sample (also called a sequential sample, i.e. a sample where
+items appear in the same order as in `a`) should be taken.
 
 Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -79,7 +79,8 @@ sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractRange, x::AbstractArray) 
     sort!(sampler!(rng, a, x), rev=step(a)<0)
 
 # weighted case:
-sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
+sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray,
+                wv::AbstractWeights, x::AbstractArray) =
     sample_ordered!(rng, a, x) do rng, a, x
         sampler!(rng, a, wv, x)
     end
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -72,7 +72,7 @@ function direct_sample_ordered!(rng::AbstractRNG, a::AbstractArray, x::AbstractA
     return x
 end
 
-direct_sample_ordered!(rng::AbstractRNG, a::AbstractRange, x::AbstractArray) =
+direct_sample_ordered!(rng::AbstractRNG, a::AbstractRange, x::AbstractArray{<:Real}) =
     sort!(direct_sample!(rng, a, x), rev=step(a)<0)
 
 ### draw a pair of distinct integers in [1:n]
@@ -586,7 +586,7 @@ function direct_sample_ordered!(rng::AbstractRNG, a::AbstractArray, wv::Abstract
     return x
 end
 
-direct_sample_ordered!(rng::AbstractRNG, a::AbstractRange, wv::AbstractWeights, x::AbstractArray) =
+direct_sample_ordered!(rng::AbstractRNG, a::AbstractRange, wv::AbstractWeights, x::AbstractArray{<:Real}) =
     sort!(direct_sample!(rng, a, wv, x), rev=step(a)<0)
 
 function make_alias_table!(w::AbstractVector{Float64}, wsum::Float64,
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -514,8 +514,8 @@ Select a random, optionally weighted sample from an array `a` specifying
 the dimensions `dims` of the output array. Sampling probabilities are
 proportional to the weights given in `wv`, if provided. `replace` dictates
 whether sampling is performed with replacement. `ordered` dictates whether
-an ordered sample (also called a sequential sample)—a sample where
-items appear in the same order as in `a`—should be taken.
+an ordered sample (also called a sequential sample, i.e. a sample where
+items appear in the same order as in `a`) should be taken.
 
 Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -444,6 +444,7 @@ Optionally specify a random number generator `rng` as the first argument
 """
 function sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray;
                  replace::Bool=true, ordered::Bool=false)
+    1 == firstindex(a) == firstindex(x) || throw(ArgumentError("non 1-based arrays are not supported"))
     n = length(a)
     k = length(x)
     k == 0 && return x
@@ -887,6 +888,7 @@ efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::AbstractWeights, x::Abstra
 
 function sample!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray;
                  replace::Bool=true, ordered::Bool=false)
+    1 == firstindex(a) == firstindex(wv) == firstindex(x) || throw(ArgumentError("non 1-based arrays are not supported"))
     n = length(a)
     k = length(x)
 
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -64,7 +64,7 @@ function sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray, x::Abstra
     else
         indices = Array{Int}(undef, k)
         sort!(sampler!(rng, Base.OneTo(n), indices))
-        for i = 1:k
+        @inbounds for i = 1:k
             x[i] = a[indices[i]]
         end
     end
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -43,7 +43,7 @@ end
 direct_sample!(a::AbstractArray, x::AbstractArray) = direct_sample!(Random.GLOBAL_RNG, a, x)
 
 # check whether we can use T to store indices 1:n exactly
-for T in unique([Int,Int64,UInt64,Int128,UInt128])
+for T in unique!([Int,Int64,UInt64,Int128,UInt128])
     @eval isexact(n, ::Type{$T}) = true
 end
 isexact(n, ::Type{T}) where {T<:Integer} = n ≤ typemax(T)
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -54,7 +54,7 @@ storeindices(n, k, ::Type{T}) where {T<:Base.HWNumber}  = _storeindices(n, k, T)
 storeindices(n, k, T) = false
 
 # order results of a sampler that does not order automatically
-function sample_ordered!(sampler!::F, rng::AbstractRNG, a::AbstractArray, x::AbstractArray) where {F}
+function sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
     n, k = length(a), length(x)
     if storeindices(n, k, eltype(x))
         sort!(sampler!(rng, Base.OneTo(n), x), by=real, lt=<)
@@ -72,11 +72,11 @@ function sample_ordered!(sampler!::F, rng::AbstractRNG, a::AbstractArray, x::Abs
 end
 
 # special case of a range (or any sorted array) can be done more efficiently
-sample_ordered!(sampler!::F, rng::AbstractRNG, a::AbstractRange, x::AbstractArray) where {F} =
+sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractRange, x::AbstractArray) =
     sort!(sampler!(rng, a, x), rev=step(a)<0)
 
 # weighted case:
-sample_ordered!(sampler!::F, rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray) where {F} =
+sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
     sample_ordered!(rng, a, x) do rng, a, x
         sampler!(rng, a, wv, x)
     end
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -572,13 +572,13 @@ direct_sample!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
 function direct_sample_ordered!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray)
     n, k = length(a), length(x)
     if isexact(n, eltype(x))
-        sort!(direct_sample!(rng, Base.OneTo(n), wv, x), by=real)
+        sort!(sample!(rng, Base.OneTo(n), wv, x, replace=true, ordered=false), by=real)
         for i = 1:k
             x[i] = a[Int(x[i])]
         end
     else
         indices = Array{Int}(undef, k)
-        sort!(direct_sample!(rng, Base.OneTo(n), wv, indices))
+        sort!(sample!(rng, Base.OneTo(n), wv, indices, replace=true, ordered=false))
         for i = 1:k
             x[i] = a[indices[i]]
         end
@@ -587,7 +587,7 @@ function direct_sample_ordered!(rng::AbstractRNG, a::AbstractArray, wv::Abstract
 end
 
 direct_sample_ordered!(rng::AbstractRNG, a::AbstractRange, wv::AbstractWeights, x::AbstractArray{<:Real}) =
-    sort!(direct_sample!(rng, a, wv, x), rev=step(a)<0)
+    sort!(sample!(rng, a, wv, x, replace=true, ordered=false), rev=step(a)<0)
 
 function make_alias_table!(w::AbstractVector{Float64}, wsum::Float64,
                            a::AbstractVector{Float64},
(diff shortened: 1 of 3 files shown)


stevengj:
@nalimilan, what do you think?

src/sampling.jl
```diff
@@ -42,6 +42,47 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
 end
 direct_sample!(a::AbstractArray, x::AbstractArray) = direct_sample!(Random.GLOBAL_RNG, a, x)
 
+# check whether we can use T to store indices 1:n exactly
+for T in unique!([Int,UInt,Int64,UInt64,Int128,UInt128])
+    @eval isexact(n, ::Type{$T}) = true
+end
+isexact(n, ::Type{T}) where {T<:Integer} = n ≤ typemax(T)
```
nalimilan, line 49: AFAICT LLVM is smart enough to optimize the generic method statically, so the loop isn't needed:
```julia
julia> isexact(n, ::Type{T}) where {T<:Integer} = n ≤ typemax(T)
isexact (generic function with 1 method)

julia> @code_llvm isexact(2, Int)
;  @ REPL[20]:1 within `isexact'
define i8 @julia_isexact_1866(i64 signext %0) {
top:
  ret i8 1
}
```

src/sampling.jl
```diff
@@ -42,6 +42,47 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
...
+isexact(n, ::Type{BigInt}) = false # bigint is too expensive
+for T in (Float32, Float64)
```
nalimilan, line 51: Why not `Float16` too? Also, shouldn't there be a fallback definition for `AbstractFloat` just like for `Integer`?

BTW, have you checked performance? I wonder whether it really makes sense to store integer indices e.g. in a `Vector{Complex{Float64}}`. The memory savings may not be worth it if it's slower.

src/sampling.jl
```diff
@@ -42,6 +42,47 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
...
+    @eval isexact(n, ::Type{$T}) = n ≤ maxintfloat($T)
+end
+isexact(n, ::Type{Complex{T}}) where {T} = isexact(n, T)
+isexact(n, ::Type{Rational{T}}) where {T} = isexact(n, T)
+isexact(n, T) = false
+
+# order results of a sampler that does not order automatically
+function sample_ordered!(sampler!::F, rng::AbstractRNG, a::AbstractArray, x::AbstractArray) where {F}
```
nalimilan, line 59: Why do you force specialization on `F`? This happens automatically AFAIK.

src/sampling.jl
```diff
@@ -396,8 +437,9 @@ Draw a random sample of `length(x)` elements from an array `a`
 and store the result in `x`. A polyalgorithm is used for sampling.
 Sampling probabilities are proportional to the weights given in `wv`,
 if provided. `replace` dictates whether sampling is performed with
-replacement and `order` dictates whether an ordered sample, also called
-a sequential sample, should be taken.
+replacement. `ordered` dictates whether
+an ordered sample (also called a sequential sample)—a sample where
+items appear in the same order as in `a`—should be taken.
```
nalimilan, line 442: Parens plus dash might be a bit too much? :-)
```suggestion
an ordered sample (also called a sequential sample, i.e. a sample where
items appear in the same order as in `a`) should be taken.
```

src/sampling.jl
```diff
@@ -781,7 +825,7 @@ Noting `k=length(x)` and `n=length(a)`, this algorithm takes ``O(k \\log(k) \\lo
 processing time to draw ``k`` elements. It consumes ``O(k \\log(n / k))`` random numbers.
 """
 function efraimidis_aexpj_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
-                                         wv::AbstractWeights, x::AbstractArray)
+                                         wv::AbstractWeights, x::AbstractArray; ordered::Bool=false)
```
nalimilan, line 828:
```suggestion
                                         wv::AbstractWeights, x::AbstractArray;
                                         ordered::Bool=false)
```

src/sampling.jl
```diff
@@ -824,15 +868,22 @@ function efraimidis_aexpj_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
...
+        # fill output array with items in descending order
+        @inbounds for i in k:-1:1
+            x[i] = a[heappop!(pq).second]
+        end
     end
     return x
 end
-efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
-    efraimidis_aexpj_wsample_norep!(Random.GLOBAL_RNG, a, wv, x)
+efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray; ordered::Bool=false) =
```
nalimilan, line 885:
```suggestion
efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray;
                                ordered::Bool=false) =
```

test/wsampling.jl
```diff
@@ -5,19 +5,20 @@ Random.seed!(1234)
 
 #### weighted sample with replacement
 
-function check_wsample_wrep(a::AbstractArray, vrgn, wv::AbstractWeights, ptol::Real; ordered::Bool=false)
+function check_wsample_wrep(a::AbstractArray, vrgn, wv::AbstractWeights, ptol::Real; ordered::Bool=false, rev::Bool=false)
```
nalimilan, line 8:
```suggestion
function check_wsample_wrep(a::AbstractArray, vrgn, wv::AbstractWeights, ptol::Real;
                            ordered::Bool=false, rev::Bool=false)
```

test/wsampling.jl
```diff
@@ -45,13 +46,17 @@ check_wsample_wrep(a, (4, 7), wv, 5.0e-3; ordered=false)
...
+    r = T===Int ? r : T.(r)
+    aa = Int.(sample(r, wv, n; ordered=true))
+    check_wsample_wrep(aa, (4, 7), wv, 5.0e-3; ordered=true, rev=rev)
+end
 
 
 #### weighted sampling without replacement
 
-function check_wsample_norep(a::AbstractArray, vrgn, wv::AbstractWeights, ptol::Real; ordered::Bool=false)
+function check_wsample_norep(a::AbstractArray, vrgn, wv::AbstractWeights, ptol::Real; ordered::Bool=false, rev::Bool=false)
```
nalimilan, line 59:
```suggestion
function check_wsample_norep(a::AbstractArray, vrgn, wv::AbstractWeights, ptol::Real;
                             ordered::Bool=false, rev::Bool=false)
```

src/sampling.jl
```diff
@@ -42,6 +42,47 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
...
+isexact(n, ::Type{Complex{T}}) where {T} = isexact(n, T)
+isexact(n, ::Type{Rational{T}}) where {T} = isexact(n, T)
+isexact(n, T) = false
+
+# order results of a sampler that does not order automatically
+function sample_ordered!(sampler!::F, rng::AbstractRNG, a::AbstractArray, x::AbstractArray) where {F}
+    n, k = length(a), length(x)
+    if isexact(n, eltype(x))
+        sort!(sampler!(rng, Base.OneTo(n), x), by=real)
+        for i = 1:k
```
nalimilan, line 63: Maybe worth it?
```suggestion
        @inbounds for i = 1:k
```

src/sampling.jl
```diff
@@ -42,6 +42,47 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
...
+            x[i] = a[Int(x[i])]
+        end
+    else
+        indices = Array{Int}(undef, k)
+        sort!(sampler!(rng, Base.OneTo(n), indices))
+        for i = 1:k
```
nalimilan, line 69:
```suggestion
        @inbounds for i = 1:k
```

src/sampling.jl
```diff
@@ -42,6 +42,47 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
...
+            x[i] = a[indices[i]]
+        end
+    end
+    return x
+end
+
+# special case of a range (or any sorted array) can be done more efficiently
```
nalimilan, line 76: Better not mention sorted arrays, as this method only applies to ranges. Would it make sense to call `issorted(x)` for all array types and use this fast path when it's true?

nalimilan (commented):
I don't know this code at all but that sounds reasonable. Just a few minor comments.

src/sampling.jl
stevengj: I was worried that `Float16` might be too slow, and similarly for other `AbstractFloat` types; `Float32` and `Float64` are the only two that generally have hardware support.

src/sampling.jl
stevengj: It doesn't always happen automatically for function arguments: https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

(Of course, the specialization probably doesn't matter here since the inner loops occur inside the `sampler!` function.)

src/sampling.jl
stevengj: Sure.   It worries me a little that the code never checks for 1-based indices, however.  (But this applies to existing code too.)

src/sampling.jl
stevengj: `issorted` could throw if the array elements don't have a comparison.   But of course we could try `issorted` only for `<:Real` arrays, since it's just an optimization.

One worry is that if the array *isn't* sorted, it will actually be slower since it will pay the price of the `issorted` check in addition to what we are already doing.

src/sampling.jl
nalimilan: "Julia will always specialize when the argument is used within the method, but not if the argument is just passed through to another function." So I think we're safe. Actually I didn't know it was documented so clearly. :-)

src/sampling.jl
nalimilan: Yeah, it would be a good idea to add an assertion somewhere.

src/sampling.jl
nalimilan: I have no idea, but in general I think `issorted` should be quite cheap since unsorted arrays are very unlikely to have there first elements sorted, so typically only a small proportion of them will be read I guess?

src/sampling.jl
nalimilan: Yeah, it would be a good idea to add an assertion somewhere.

src/sampling.jl
stevengj: I ran some benchmarks, and (as you might expect) it depends on the length of `x`.

If you are only computing a few samples (e.g. 5), then it is faster to avoid the allocation.   If you are computing lots of samples (e.g. 1000) it's faster to allocate the indices.

I guess I could hard-code an approximate crossover point.

src/sampling.jl
stevengj: Done.

src/sampling.jl
stevengj: I removed the comment here and added a note elsewhere in `sample_ordered!` that this might be a possible future optimization.

stevengj:
I think I addressed all of the issues discussed above?

src/sampling.jl
```diff
@@ -42,43 +42,44 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
...
-isexact(n, ::Type{Rational{T}}) where {T} = isexact(n, T)
-isexact(n, T) = false
+# check whether we can use T to store indices 1:n exactly, and
+# use some heuristics to decide whether it is beneficial for k samples
+# (true for a subset of hardware-supported numeric types)
+_storeindices(n, k, ::Type{T}) where {T<:Integer} = n ≤ typemax(T)
+_storeindices(n, k, ::Type{T}) where {T<:Union{Float32,Float64}} = k < 22 && n ≤ maxintfloat(T)
+_storeindices(n, k, ::Type{Complex{T}}) where {T} = _storeindices(n, k, T)
+_storeindices(n, k, ::Type{Rational{T}}) where {T} = k < 16 && _storeindices(n, k, T)
+_storeindices(n, k, T) = false
```
nalimilan, line 52: Looks like this fallback isn't tested? I guess `BigFloat` would do?

src/sampling.jl
```diff
@@ -824,24 +871,35 @@ function efraimidis_aexpj_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
...
 end
-efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
-    efraimidis_aexpj_wsample_norep!(Random.GLOBAL_RNG, a, wv, x)
+efraimidis_aexpj_wsample_norep!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray;
+                                ordered::Bool=false) =
+    efraimidis_aexpj_wsample_norep!(Random.GLOBAL_RNG, a, wv, x; ordered=ordered)
 
 function sample!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray;
                  replace::Bool=true, ordered::Bool=false)
+    1 == firstindex(a) == firstindex(wv) == firstindex(x) || throw(ArgumentError("non 1-based arrays are not supported"))
```
nalimilan, line 894:
```suggestion
    1 == firstindex(a) == firstindex(wv) == firstindex(x) ||
        throw(ArgumentError("non 1-based arrays are not supported"))
```

src/sampling.jl
```diff
@@ -448,8 +492,9 @@ sample!(a::AbstractArray, x::AbstractArray; replace::Bool=true, ordered::Bool=fa
 Select a random, optionally weighted sample of size `n` from an array `a`
 using a polyalgorithm. Sampling probabilities are proportional to the weights
 given in `wv`, if provided. `replace` dictates whether sampling is performed
-with replacement and `order` dictates whether an ordered sample, also called
-a sequential sample, should be taken.
+with replacement. `ordered` dictates whether
+an ordered sample (also called a sequential sample)—a sample where
+items appear in the same order as in `a`—should be taken.
```
nalimilan, line 497:
```suggestion
an ordered sample (also called a sequential sample, i.e. a sample where
items appear in the same order as in `a`) should be taken.
```

src/sampling.jl
```diff
@@ -468,8 +513,9 @@ sample(a::AbstractArray, n::Integer; replace::Bool=true, ordered::Bool=false) =
 Select a random, optionally weighted sample from an array `a` specifying
 the dimensions `dims` of the output array. Sampling probabilities are
 proportional to the weights given in `wv`, if provided. `replace` dictates
-whether sampling is performed with replacement and `order` dictates whether
-an ordered sample, also called a sequential sample, should be taken.
+whether sampling is performed with replacement. `ordered` dictates whether
+an ordered sample (also called a sequential sample)—a sample where
+items appear in the same order as in `a`—should be taken.
```
nalimilan, line 518:
```suggestion
an ordered sample (also called a sequential sample, i.e. a sample where
items appear in the same order as in `a`) should be taken.
```

src/sampling.jl
```diff
@@ -889,8 +943,9 @@ sample(a::AbstractArray, wv::AbstractWeights, dims::Dims;
 
 Select a weighted sample from an array `a` and store the result in `x`. Sampling
 probabilities are proportional to the weights given in `w`. `replace` dictates
-whether sampling is performed with replacement and `order` dictates whether an
-ordered sample, also called a sequential sample, should be taken.
+whether sampling is performed with replacement. . `ordered` dictates whether
+an ordered sample (also called a sequential sample)—a sample where
+items appear in the same order as in `a`—should be taken.
```
nalimilan, line 948:
```suggestion
whether sampling is performed with replacement. `ordered` dictates whether
an ordered sample (also called a sequential sample, i.e. a sample where
items appear in the same order as in `a`) should be taken.
```

src/sampling.jl
```diff
@@ -923,8 +978,9 @@ wsample(a::AbstractArray, w::RealVector) = wsample(Random.GLOBAL_RNG, a, w)
 Select a weighted random sample of size `n` from `a` with probabilities proportional
 to the weights given in `w` if `a` is present, otherwise select a random sample of size
 `n` of the weights given in `w`. `replace` dictates whether sampling is performed with
-replacement and `order` dictates whether an ordered sample, also called a sequential
-sample, should be taken.
+replacement. `ordered` dictates whether
+an ordered sample (also called a sequential sample)—a sample where
+items appear in the same order as in `a`—should be taken.
```
nalimilan, line 983:
```suggestion
an ordered sample (also called a sequential sample, i.e. a sample where
items appear in the same order as in `a`) should be taken.
```

src/sampling.jl
```diff
@@ -42,6 +42,48 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
...
+    end
+    return x
+end
+
+# special case of a range can be done more efficiently
+sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractRange, x::AbstractArray) =
+    sort!(sampler!(rng, a, x), rev=step(a)<0)
+
+# weighted case:
+sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
```
nalimilan, line 82:
```suggestion
sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray,
                wv::AbstractWeights, x::AbstractArray) =
```

src/sampling.jl
```diff
@@ -396,21 +438,23 @@ Draw a random sample of `length(x)` elements from an array `a`
...
+replacement. `ordered` dictates whether
+an ordered sample (also called a sequential sample, i.e. a sample where
+items appear in the same order as in `a`) should be taken.
 
 Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
 """
 function sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray;
                  replace::Bool=true, ordered::Bool=false)
+    1 == firstindex(a) == firstindex(x) || throw(ArgumentError("non 1-based arrays are not supported"))
```
nalimilan, line 450:
```suggestion
    1 == firstindex(a) == firstindex(x) ||
        throw(ArgumentError("non 1-based arrays are not supported"))
```

src/sampling.jl
stevengj: `_storeindices` is only called for `Base.HWNumber` types, all of which I currently have methods for; I added this fallback in case `Base.HWNumber` someday adds more types that I haven't handled.

test/sampling.jl
```diff
@@ -68,11 +68,15 @@ test_rng_use(direct_sample!, 1:10, zeros(Int, 6))
...
+    r = rev ? reverse(3:12) : (3:12)
+    r = T===Int ? r : T.(r)
+    aa = Int.(sample(r, n; ordered=true))
+    check_sample_wrep(aa, (3, 12), 5.0e-3; ordered=true, rev=rev)
 
-a = sample(3:12, 10; ordered=true)
-check_sample_wrep(a, (3, 12), 0; ordered=true)
+    aa = Int.(sample(r, 10; ordered=true))
+    check_sample_wrep(aa, (3, 12), 0; ordered=true, rev=rev)
+end
```
stevengj, line 80:
```suggestion

@test StatsBase._storeindices(1, 1, BigFloat) == StatsBase._storeindices(1, 1, BigFloat) == false

```

src/sampling.jl
stevengj: I added a test by "manually" calling `_storeindices`.

nalimilan (approved):

nalimilan:
Thanks!

nalimilan:
https://github.com/JuliaRegistries/General/pull/33783

PR #678, opened 2021-03-25
Thow DomainError when sampling weights contain missing, Inf or NaN

### LINKED ISSUE #671: StatsBase.sample should warn about weights containing Inf/NaN
https://github.com/JuliaStats/StatsBase.jl/blob/08d4b77a4b42ef8cadb67b98da65e7fbd9959e0b/src/sampling.jl#L499-L500
Perhaps add something like
```julia
swv = sum(wv)
isfinite(swv) || throw(DomainError(swv, "weights sum to $swv"))
```

xiruizhao:
Fixes #671


nalimilan:
Thanks. Can you add a test?

xiruizhao:
I am not sure about this now. Should the check be done in `weights.jl`?  Should I also check for negative weights? Or should the library simply ask the callers to ensure valid inputs?

nalimilan:
Good question. In general it's true that non-finite or negative weights are probably not useful. Though it could be the case that in some applications it's OK to allow them, e.g. if observations with `NaN` weights should be skipped. That's hard to tell. Anyway that would be breaking so we would have first to print a deprecation warning. In the meantime I think it would be good to merge this PR as it fixes a real bug. But feel free to make a more general PR too.

(BTW, I forgot that the same check should be applied to `sample!` too.)

bkamins:
> Anyway that would be breaking so we would have first to print a deprecation warning.

I would say it is non-breaking if we decide that allowing `Inf` or `NaN` is a bug. I would say it is a bug as I cannot imagine any application when allowing it makes sense. The reason is that we guarantee to calculate `sum`. And `sum` is `Inf` or `NaN` in these cases - which is not well defined. So, by contract of `AbstractWeights` allowing `Inf` or `NaN` should not be allowed I think.

PR #682, opened 2021-04-15
corrected error message in corspearman

PGS62:
Currently the error thrown when `x` and `y` have a different number of rows states that the number of columns should match, so I corrected that to state that it's the number of _rows_ that need to match.


nalimilan (approved):
Thanks!

PR #685, opened 2021-04-21
pointwiseloglikelihood

palday:
This introduces `pointwiseloglikelihood`:

1. single argument variant for computing each observation's contribution to the log likelihood. This is useful for things like [Vuong's Test](https://en.wikipedia.org/wiki/Vuong%27s_closeness_test). Using this, we could add a generic `vuongtest` to `StatsModels.jl`. 
2. two argument variant for computing the log-likelihood of an observation under the model.  This can be used in various cross-validation regimes and other things examining out-of-sample error.

Assuming that this lands, I would write the associated methods for `GLM.jl` and `MixedModels.jl` and add the delegation for `TableRegressionModel` to `StatsModels.jl`.

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -77,6 +77,24 @@ nullloglikelihood(model::StatisticalModel) =
     error("nullloglikelihood is not defined for $(typeof(model)).")
 
 """
+    pointwiseloglikelihood(model::StatisticalModel)
+
+Return a vector of each observation's contribution to the log-likelihood of the model.
+
+In general, `sum(pointwiseloglikehood(model)) == loglikelihood(model)`.
+"""
+pointwiseloglikelihood(model::StatisticalModel) =
+    error("pointwiseloglikelihood(model::StatisticalModel) is not defined for $(typeof(model)).")
+
+"""
+    pointwiseloglikelihood(model::StatisticalModel, observation)
+
+Return the log-likelihood of `observation` contribution under `model`.
+"""
+pointwiseloglikelihood(model::StatisticalModel, observation) =
+    error("pointwiseloglikelihood(model::StatisticalModel, observation) is not defined for $(typeof(model)).")
+
+"""
     score(model::StatisticalModel)
 
 Return the score of the model, that is the gradient of the
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -193,6 +193,7 @@ export
     nobs,
     nulldeviance,
     nullloglikelihood,
+    pointwiseloglikelihood,
     rss,
     score,
     stderror,


src/statmodels.jl
```diff
@@ -76,6 +76,24 @@ This is usually the model containing only the intercept.
...
+"""
+pointwiseloglikelihood(model::StatisticalModel) =
+    error("pointwiseloglikelihood(model::StatisticalModel) is not defined for $(typeof(model)).")
+
+"""
+    pointwiseloglikelihood(model::StatisticalModel, observation)
+
+Return the log-likelihood of `observation` contribution under `model`.
+"""
+pointwiseloglikelihood(model::StatisticalModel, observation) =
```
nalimilan, line 94: Do you expect observation to be anything other than integer?

nalimilan:
Makes sense, but maybe we could find a simpler name? :-)

src/statmodels.jl
palday: Yes -- I don't expect it to be simply an index, it could be e.g. a vector corresponding to a row in the model matrix in a linear model.`

palday:
> Makes sense, but maybe we could find a simpler name? :-)

I actually was thinking this earlier and asked @ararslan but he was strangely okay with a looooooong name. `contribution` and `elementwise` also aren't simpler than `pointwise`

nalimilan:
What if we just added a two-argument method to `log-likelihood`?

palday:
I thought about this and that works when you're specifying `observation`, but if you want just the pointwise likelihood for all observations that the model used to fit it? I guess we could add add `::Val{true}` thing to enable pointwise computation. I don't like having a kwarg for pointwise because then you have type instability.

palday:
@nalimilan What about?

```julia
"""
    loglikelihood(model::StatisticalModel, ::Val{true})

Return a vector of each observation's contribution to the log-likelihood of the model.
In other words, this is the vector of the pointwise log-likelihood contributions.


In general, `sum(loglikehood(model, Val(true))) == loglikelihood(model)`.
"""
loglikelihood(model::StatisticalModel, ::Val{true}) =
    error("loglikelihood(model::StatisticalModel, ::Val{true}) is not defined for $(typeof(model)).")

"""
    loglikelihood(model::StatisticalModel, observation)

Return the log-likelihood of `observation` contribution under `model`.
"""
loglikelihood(model::StatisticalModel, observation) =
    error("loglikelihood(model::StatisticalModel, observation) is not defined for $(typeof(model)).")
```

nalimilan:
`Val(true)` is quite ugly. Maybe use `:` for this?

src/statmodels.jl
```diff
@@ -76,6 +76,26 @@ This is usually the model containing only the intercept.
...
+
+In general, `sum(loglikehood(model, :)) == loglikelihood(model)`.
+"""
+loglikelihood(model::StatisticalModel, ::Colon) =
+    error("loglikelihood(model::StatisticalModel, ::Colon) is not defined for $(typeof(model)).")
+
+"""
+    loglikelihood(model::StatisticalModel, observation)
+
+Return the log-likelihood of `observation` contribution under `model`.
```
nalimilan, line 94: That sentence feels off to me. Something like this?
```suggestion
Return the contribution of `observation` to the log-likelihood of `model`.
```

src/statmodels.jl
```diff
@@ -76,6 +76,26 @@ This is usually the model containing only the intercept.
 nullloglikelihood(model::StatisticalModel) =
     error("nullloglikelihood is not defined for $(typeof(model)).")
 
+"""
+    loglikelihood(model::StatisticalModel, ::Colon)
+
+Return a vector of each observation's contribution to the log-likelihood of the model.
+In other words, this is the vector of the pointwise log-likelihood contributions.
+
```
nalimilan, line 84:
```suggestion
```

PR #688, opened 2021-04-23
Preserve eltype where possible for moments

### LINKED ISSUE #680: std(::AbstractVector{Float32}, ::AnalyticWeights{Float32}) gives a Float64
Setup:
```julia
julia> using StatsBase, Statistics

julia> xs = Float32[1, 2, 3, 4, 5];

julia> ws = AnalyticWeights(Float32[1, 1, 1, 1, 1]);

julia> eltype(ws)
Float32
```

Now trying some operations

```julia
julia> typeof(std(xs))  # As expected
Float32


julia> typeof(mean(xs, ws))  # as expected
Float32

julia> typeof(std(xs, ws))  # Not expected Should be a Float32
Float64
```

I was pretty sure this used to work to give a `Float32`
if so, I can't work-out when it changed.

palday:
Fixes #680.

### DIFF
--- a/src/moments.jl
+++ b/src/moments.jl
@@ -42,7 +42,7 @@ function var(v::RealArray, w::AbstractWeights; mean=nothing,
                   corrected::DepBool=nothing)
     corrected = depcheck(:var, corrected)
 
-    if mean == nothing
+    if mean === nothing
         varm(v, w, Statistics.mean(v, w); corrected=corrected)
     else
         varm(v, w, mean; corrected=corrected)
@@ -82,17 +82,17 @@ function var!(R::AbstractArray, A::RealArray, w::AbstractWeights, dims::Int;
     end
 end
 
-function varm(A::RealArray, w::AbstractWeights, M::RealArray, dim::Int;
-                   corrected::DepBool=nothing)
+function varm(A::RealArray{T}, w::AbstractWeights, M::RealArray, dim::Int;
+                   corrected::DepBool=nothing) where T
     corrected = depcheck(:varm, corrected)
-    varm!(similar(A, Float64, Base.reduced_indices(axes(A), dim)), A, w, M,
+    varm!(similar(A, promote_type(T, eltype(w)), Base.reduced_indices(axes(A), dim)), A, w, M,
                dim; corrected=corrected)
 end
 
-function var(A::RealArray, w::AbstractWeights, dim::Int; mean=nothing,
-                  corrected::DepBool=nothing)
+function var(A::RealArray{T}, w::AbstractWeights, dim::Int; mean=nothing,
+                  corrected::DepBool=nothing) where T
     corrected = depcheck(:var, corrected)
-    var!(similar(A, Float64, Base.reduced_indices(axes(A), dim)), A, w, dim;
+    var!(similar(A, promote_type(T, eltype(w)), Base.reduced_indices(axes(A), dim)), A, w, dim;
          mean=mean, corrected=corrected)
 end
 
@@ -221,9 +221,9 @@ end
 
 
 ##### General central moment
-function _moment2(v::RealArray, m::Real; corrected=false)
+function _moment2(v::RealArray{T}, m::Real; corrected=false) where T
     n = length(v)
-    s = 0.0
+    s = zero(T)
     for i = 1:n
         @inbounds z = v[i] - m
         s += z * z
@@ -231,9 +231,9 @@ function _moment2(v::RealArray, m::Real; corrected=false)
     varcorrection(n, corrected) * s
 end
 
-function _moment2(v::RealArray, wv::AbstractWeights, m::Real; corrected=false)
+function _moment2(v::RealArray{T}, wv::AbstractWeights, m::Real; corrected=false) where T
     n = length(v)
-    s = 0.0
+    s = zero(promote_type(T, eltype(wv)))
     for i = 1:n
         @inbounds z = v[i] - m
         @inbounds s += (z * z) * wv[i]
@@ -242,9 +242,9 @@ function _moment2(v::RealArray, wv::AbstractWeights, m::Real; corrected=false)
     varcorrection(wv, corrected) * s
 end
 
-function _moment3(v::RealArray, m::Real)
+function _moment3(v::RealArray{T}, m::Real) where T
     n = length(v)
-    s = 0.0
+    s = zero(T)
     for i = 1:n
         @inbounds z = v[i] - m
         s += z * z * z
@@ -252,9 +252,9 @@ function _moment3(v::RealArray, m::Real)
     s / n
 end
 
-function _moment3(v::RealArray, wv::AbstractWeights, m::Real)
+function _moment3(v::RealArray{T}, wv::AbstractWeights, m::Real; corrected=false) where T
     n = length(v)
-    s = 0.0
+    s = zero(promote_type(T, eltype(wv)))
     for i = 1:n
         @inbounds z = v[i] - m
         @inbounds s += (z * z * z) * wv[i]
@@ -262,9 +262,9 @@ function _moment3(v::RealArray, wv::AbstractWeights, m::Real)
     s / sum(wv)
 end
 
-function _moment4(v::RealArray, m::Real)
+function _moment4(v::RealArray{T}, m::Real) where T
     n = length(v)
-    s = 0.0
+    s = zero(T)
     for i = 1:n
         @inbounds z = v[i] - m
         s += abs2(z * z)
@@ -272,9 +272,9 @@ function _moment4(v::RealArray, m::Real)
     s / n
 end
 
-function _moment4(v::RealArray, wv::AbstractWeights, m::Real)
+function _moment4(v::RealArray{T}, wv::AbstractWeights, m::Real; corrected=false) where T
     n = length(v)
-    s = 0.0
+    s = zero(promote_type(T, eltype(wv)))
     for i = 1:n
         @inbounds z = v[i] - m
         @inbounds s += abs2(z * z) * wv[i]
@@ -282,9 +282,9 @@ function _moment4(v::RealArray, wv::AbstractWeights, m::Real)
     s / sum(wv)
 end
 
-function _momentk(v::RealArray, k::Int, m::Real)
+function _momentk(v::RealArray{T}, k::Int, m::Real) where T
     n = length(v)
-    s = 0.0
+    s = zero(T)
     for i = 1:n
         @inbounds z = v[i] - m
         s += (z ^ k)
@@ -292,9 +292,9 @@ function _momentk(v::RealArray, k::Int, m::Real)
     s / n
 end
 
-function _momentk(v::RealArray, k::Int, wv::AbstractWeights, m::Real)
+function _momentk(v::RealArray{T}, k::Int, wv::AbstractWeights, m::Real) where T
     n = length(v)
-    s = 0.0
+    s = zero(promote_type(T, eltype(wv)))
     for i = 1:n
         @inbounds z = v[i] - m
         @inbounds s += (z ^ k) * wv[i]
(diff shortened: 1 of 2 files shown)


oxinabox (approved):

src/moments.jl
```diff
@@ -221,19 +221,19 @@ end
 
 
 ##### General central moment
-function _moment2(v::RealArray, m::Real; corrected=false)
+function _moment2(v::RealArray{T}, m::Real; corrected=false) where T
```
ararslan, line 224: Rather than forcing specialization, I would just replace the use of `T` below with `eltype(v)`. Same for all of the other methods here.

src/moments.jl
```diff
@@ -242,59 +242,59 @@ function _moment2(v::RealArray, wv::AbstractWeights, m::Real; corrected=false)
...
+    s = zero(T)
     for i = 1:n
         @inbounds z = v[i] - m
         s += z * z * z
     end
     s / n
 end
 
-function _moment3(v::RealArray, wv::AbstractWeights, m::Real)
+function _moment3(v::RealArray{T}, wv::AbstractWeights, m::Real; corrected=false) where T
```
ararslan, line 255: The added keyword argument here seems not to be used?

src/moments.jl
```diff
@@ -419,7 +419,7 @@ function kurtosis(v::RealArray, wv::AbstractWeights, m::Real)
     sw = sum(wv)
     cm4 /= sw
     cm2 /= sw
-    return (cm4 / (cm2 * cm2)) - 3.0
+    return (cm4 / (cm2 * cm2)) - 3 * one(T)
```
ararslan, line 422:
```suggestion
    return (cm4 / (cm2 * cm2)) - 3
```

src/moments.jl
mschauer: “Forcing specialization”? Isn’t Julia always specializing?

src/moments.jl
palday: copy-paste error

src/moments.jl
palday: @ararslan that's how I originally had it, but I wanted to get some additional type stability because I figured that moments might be something in a tight loop. It turned out harder to have that specialization on the weight vector (which is why it's not specialized) because the type parameterization is complicated and nested.

@mschauer I think the trick here is that it's parametric types and how much specialization is done on the type parameter.  

I'm happy to have the simpler variant with `eltype`.

palday:
The failure on nightly is an upstream bug (https://github.com/JuliaLang/julia/issues/40609).

src/moments.jl
```diff
@@ -223,7 +223,7 @@ end
 ##### General central moment
 function _moment2(v::RealArray, m::Real; corrected=false)
     n = length(v)
-    s = 0.0
+    s = zero(eltype(v))
```
ararslan, line 226:
```suggestion
    s = zero(promote_type(eltype(v), typeof(m))
```

ararslan: I think this applies for all of changes here.

src/moments.jl
```diff
@@ -264,17 +264,17 @@ end
...
     for i = 1:n
         @inbounds z = v[i] - m
         s += abs2(z * z)
     end
     s / n
 end
 
-function _moment4(v::RealArray, wv::AbstractWeights, m::Real)
+function _moment4(v::RealArray, wv::AbstractWeights, m::Real; corrected=false)
```
ararslan, line 275:
```suggestion
function _moment4(v::RealArray, wv::AbstractWeights, m::Real)
```

src/moments.jl
palday: Everybody wants a promotion...

src/moments.jl
```diff
@@ -85,14 +85,16 @@ end
 function varm(A::RealArray, w::AbstractWeights, M::RealArray, dim::Int;
                    corrected::DepBool=nothing)
     corrected = depcheck(:varm, corrected)
-    varm!(similar(A, Float64, Base.reduced_indices(axes(A), dim)), A, w, M,
+    T = promote_type(promote_type(eltype(A), eltype(w)), eltype(M))
```
ararslan, line 88: `promote_type` has a vararg method, no need to nest calls

palday: If only that were in its docstring :angry:

ararslan:
This looks good but the prior uses of `Float64` make me worried that the code may have issues around under/overflow when using other types.

palday:
Ugh, I'm conflicted on that front. Because that would be breaking, but technically a user hitting overflow on their custom narrow types is expected behavior. @nalimilan do we have enough for another breaking pre-1.0 release?

nalimilan:
We can break things if we want to tag 1.0. Though ideally I'd rather move these to Statistics, but that work has stalled (https://github.com/JuliaLang/julia/pull/31395).

src/moments.jl
```diff
@@ -233,7 +236,8 @@ end
 
 function _moment2(v::RealArray, wv::AbstractWeights, m::Real; corrected=false)
     n = length(v)
-    s = 0.0
+    T = promote_type(eltype(v), eltype(wv), typeof(m))
+    s = zero(T)
```
nalimilan, line 240: To ensure type stability, how about following exactly what the loop does and using `T = typeof(zero(eltype(v) - zero(m))^2 * zero(eltype(w)))`? That's what I did in my first attempt to moving these functions to Statistics: https://github.com/JuliaLang/julia/pull/31395/files#diff-6b82a0c4e60cbe68048e35375987a86688afb4aea60dda8f2d2d4938dc845776R375

palday:
Okay, so on the breaking / non breaking front:

1. Did the API ever specify the return type for `mean` and the various moments? If not, this is not breaking in terms of type behavior.
2. In terms of numeric behavior, Julia generally keeps things at narrow types even at the risk of over-/underflow. As such, keeping the narrowest type that would be possible for the arithmetic operations in these computations seems fine. In other words, the new return type and potential numeric implications are expected in the Julia ecosystem. 
3. The potential numeric issues with all of the moments here are possible regardless of type width -- we do accumulate then divide without sorting first, so
    - we have issues with computations where things differ in scale and the smaller values accumulate enough to be relevant on the larger scale, but consistently round down when the initial values in the array are large (sorting solves this but is computationally expensive)
    - we can have overflow in the accumulation step (batching the accumulate+divide on subsets solves this but potentially causes performance and numeric issues from the extra divisions)
4. The way for best numerical accuracy would be sort, do the math in a higher precision and cast down. But this also seems unacceptable -- if a user choose to keep everything in single or half precision, then we should assume there is a reason for that. If they really need that level of numerical stability, then they should probably just implement their own moments.

In other words: given that we're not going to change the algorithm used (which I think is reasonable), the potential numeric issues from the eltype preservation should not be considered breaking. Perhaps the best solution is to add to each of the docstrings:

```markdown
!!! note
    The computation is done in the narrowest type for which the underlying 
    arithmetic operations are defined based on the element types of the 
    vectors and the mean. For higher numerical accuracy and to reduce the risk
    of overflow when dealing with values near the limits of the type (or many 
    values whose sum is near the limits of the type), we recommend casting to 
    a broader type / higher precision first.
```

src/moments.jl
```diff
@@ -281,8 +276,7 @@ end
 
 function _moment4(v::RealArray, wv::AbstractWeights, m::Real)
     n = length(v)
-    T = promote_type(eltype(v), eltype(wv), typeof(m))
-    s = zero(T)
+    s = zero(eltype(wv))  * (zero(eltype(v)) - zero(m))^2
```
nalimilan, line 279:
```suggestion
    s = zero(eltype(wv)) * (zero(eltype(v)) - zero(m))^4
```

src/moments.jl
```diff
@@ -236,8 +235,7 @@ end
 
 function _moment2(v::RealArray, wv::AbstractWeights, m::Real; corrected=false)
     n = length(v)
-    T = promote_type(eltype(v), eltype(wv), typeof(m))
-    s = zero(T)
+    s = zero(eltype(wv))  * (zero(eltype(v)) - zero(m))^2
```
nalimilan, line 238:
```suggestion
    s = zero(eltype(wv)) * (zero(eltype(v)) - zero(m))^2
```

src/moments.jl
```diff
@@ -248,8 +246,7 @@ end
 
 function _moment3(v::RealArray, m::Real)
     n = length(v)
-    T = promote_type(eltype(v), typeof(m))
-    s = zero(T)
+    s = zero(eltype(v)) - zero(m)
```
nalimilan, line 249:
```suggestion
    s = (zero(eltype(v)) - zero(m))^3
```

src/moments.jl
```diff
@@ -225,8 +225,7 @@ end
 ##### General central moment
 function _moment2(v::RealArray, m::Real; corrected=false)
     n = length(v)
-    T = promote_type(eltype(v), typeof(m))
-    s = zero(T)
+    s = zero(eltype(v)) - zero(m)
```
nalimilan, line 228:
```suggestion
    s = (zero(eltype(v)) - zero(m))^2
```

src/moments.jl
```diff
@@ -259,8 +256,7 @@ end
 
 function _moment3(v::RealArray, wv::AbstractWeights, m::Real)
     n = length(v)
-    T = promote_type(eltype(v), eltype(wv), typeof(m))
-    s = zero(T)
+    s = zero(eltype(wv))  * (zero(eltype(v)) - zero(m))^2
```
nalimilan, line 259:
```suggestion
    s = zero(eltype(wv)) * (zero(eltype(v)) - zero(m))^3
```

src/moments.jl
```diff
@@ -270,8 +266,7 @@ end
 
 function _moment4(v::RealArray, m::Real)
     n = length(v)
-    T = promote_type(eltype(v), typeof(m))
-    s = zero(T)
+    s = zero(eltype(v)) - zero(m)
```
nalimilan, line 269:
```suggestion
    s = (zero(eltype(v)) - zero(m))^4
```

src/moments.jl
```diff
@@ -303,8 +296,7 @@ end
 
 function _momentk(v::RealArray, k::Int, wv::AbstractWeights, m::Real)
     n = length(v)
-    T = promote_type(eltype(v), eltype(wv), typeof(m))
-    s = zero(T)
+    s = zero(eltype(wv))  * (zero(eltype(v)) - zero(m))^2
```
nalimilan, line 299:
```suggestion
    s = zero(eltype(wv)) * (zero(eltype(v)) - zero(m))^k
```

src/moments.jl
```diff
@@ -85,14 +85,16 @@ end
 function varm(A::RealArray, w::AbstractWeights, M::RealArray, dim::Int;
                    corrected::DepBool=nothing)
     corrected = depcheck(:varm, corrected)
-    varm!(similar(A, Float64, Base.reduced_indices(axes(A), dim)), A, w, M,
+    T = promote_type(eltype(A), eltype(w), eltype(M))
```
nalimilan, line 88: Likewise, here and below it should be possible to use `zero` to choose the appropriate type? That's not supported currently given the signature, but that would allow e.g. computing the standard deviation of Unitful types.

nalimilan:
Yeah it's not super likely to break user code. Though we could collect the few changes which are breaking (some of them are marked with the [breaking](https://github.com/JuliaStats/StatsBase.jl/labels/breaking) label) and take that occasion to tag 1.0.

src/moments.jl
```diff
@@ -341,8 +348,9 @@ specifying a weighting vector `wv` and a center `m`.
 """
 function skewness(v::RealArray, m::Real)
     n = length(v)
-    cm2 = 0.0   # empirical 2nd centered moment (variance)
-    cm3 = 0.0   # empirical 3rd centered moment
+    T = promote_type(eltype(v), typeof(m))
```
nalimilan, line 351: Use the same strategy here and below too?

PR #698, opened 2021-07-13
API for VIF

palday:
Variance Inflation Factor -- if we ever move `StatisticalModel` and `RegressionModel` to StatsModels, then this should go there, but until then, I'm guessing this is the best place. Next step would be implementation for GLM.jl and MixedModels.jl.

This is just a sketch, if anybody has a better idea, let me know.

@nalimilan  I've tagged you as a reviewer since this is mostly an API design thing and you seem to have a really good (and tough :wink:) perspective on such things.


src/statmodels.jl
```diff
@@ -423,6 +423,29 @@ Return all parameters of a model.
 params(model) = error("params is not defined for $(typeof(model))")
 function params! end
 
+"""
+    vif(model, param)
+
+Return the variance inflation factor for a parameter in a model.
```
nalimilan, line 429: How about saying "coefficient" rather than "parameter"? That's how we call these in general (function `coef` and type `CoefTable`, while `params` returns distributions parameters IIRC).

src/statmodels.jl
```diff
@@ -423,6 +423,29 @@ Return all parameters of a model.
...
+
+The [variance inflation factor (VIF)](https://en.wikipedia.org/wiki/Variance_inflation_factor) measures
+the increase in the variance of a parameter's estimate in a model with multiple parameters relative to
+the variance in of a paremeter's estimate in a model containing only that parameter.
```
nalimilan, line 433:
```suggestion
the variance of a parameter's estimate in a model containing only that parameter.
```

src/statmodels.jl
```diff
@@ -423,6 +423,29 @@ Return all parameters of a model.
 params(model) = error("params is not defined for $(typeof(model))")
 function params! end
 
+"""
+    vif(model, param)
```
nalimilan, line 427:
```suggestion
    vif(model::StatisticalModel, param)
```

src/statmodels.jl
```diff
@@ -423,6 +423,29 @@ Return all parameters of a model.
...
+
+
+"""
+    viftable(model)
+
+Return the variance inflation factor for all parameters in a model as a `CoefTable`.
+
+See also [`vif`](@ref)
+"""
+viftable(model::StatisticalModel) =
```
nalimilan, line 446: I'm hesitant about adding this given that we have the `coeftable` function with the same name as the `CoefTable` type. If we add a function for the VIF, we should add `pvaluetable`, `tvaluetable`, `confinttable` and so on. How about adding a keyword argument `vif=false` to `coeftable` instead?

nalimilan (commented):
Thanks, that sounds useful. Yes, I'm tough regarding API issues. :-p

It would be good to prepare at least one implementation in a package before merging this, just to ensure the result suits our needs before committing to this API forever.

See also https://github.com/JuliaStats/GLM.jl/issues/428.

src/statmodels.jl
nalimilan: Would it make sense to restrict this to `::RegressionModel`? AFAIK the VIF is only used for linear regression, right?

palday:
> It would be good to prepare at least one implementation in a package before merging this, just to ensure the result suits our needs before committing to this API forever.

Yeah, I wanted to do this in GLM.jl and MixedModels.jl, but it's much easier to have methods in those packages be deprecated than here if we don't like the API.

src/statmodels.jl
palday: I think it's used for any model falling within the generalized linear model and there might even be meaningful definitions for a broader class of models. But it's much easier to start narrow and broaden, so I'm fine restricting this to `RegressionModel` for now.

src/statmodels.jl
palday: Hmmm. Tough one. I see your point, but I'm not sure I like having the kwarg in `coeftable` because it makes it sound like it's returning the usual `CoefTable` with an extra column for for VIF. 

My motivation for having a table method is:
- A lot of the VIF computation is common to all model params, so having all VIFs in one call might save some overhead
- Specifying parameters can be awkward (especially for interaction terms), so this is a decent way to dodge that for people who want VIF for overall diagnostics
- Often people want VIF for all parameters anyway, so this makes that convenient.

Maybe having a second `vif` method without the `param` argument would be better? Still have type stability then and it matches the behavior in other common languages.

src/statmodels.jl
palday: :+1: Yeah, `coef` is also a better match if we restrict to `RegressionModel`.

src/statmodels.jl
nalimilan: Very tricky. Using a single-argument `vif` method is appealing, but that would be inconsistent with `coef`, which returns a vector of coefficients. I think this is sub-optimal and we should return a named vector like in R. That sounds more appropriate than a `CoefTable` when there's a single value, and that can be changed without breaking since the result is still an `AbstractArray`.

Unfortunately, so far we haven't standardized on a package for arrays with names. Maybe StatsBase could remain agnostic regarding the exact kind of vector which is returned, but packages could choose to return e.g. a `NamedArray`, `AxisArray`, `KeyedArray` or similar type?

palday: Huh, we could do single-argument `vif` as `AbstractArray` (then implementations are free to choose a named array of some type if they so desire), but then also have a separate function for extracting names. I'm tempted to have `vifnames(model::RegressionModel) = coefnames(model)`, thus giving a reasonable default that type owners are free to override. This would be a reasonable default because the VIFs are computed for each `coef`.  I would omit two-argument `vif` for now -- it might potentially be useful in the future (computing the VIFs isn't quite "free" in the same way that extracting the coefficients for a fitted model is), having `vif` behavior match `coef` behavior seems reasonable.

nalimilan: Why would `vifnames` return something different from `coefnames`? Couldn't a single function be enough?

palday: It seems that you generally don't compute the VIF for the intercept term, which means that 
```julia
vifnames(model) = filter(!=("(Intercept)"), coefnames(model))
```

nalimilan: (Sorry for the delay.) That's a good point, but I'm not sure that justifies adding a method to the API if that's just for convenience. Maybe we can just expect users to skip the intercept manually? Anyway, as I noted above, I hope that we will be able to return a vector with names from `vif` and `coef` instead of adding functions to get names.

palday: Yeah, then I think we can just document that the `vif` filters out the `(Intercept)` term. Now I know enough to mock up a real VIF implementation for GLM or MixedModels and then we can see how much we like the API when applied downstream. :smile:

ararslan:
Yes?

nalimilan:
> @nalimilan I have a pretty general implementation over in MixedModelsExtras ([palday/MixedModelsExtras.jl#11](https://github.com/palday/MixedModelsExtras.jl/pull/11)) that I want to use as an API testing ground. Let me know if that API works for you. If so, we could migrate most of the code and method-name ownership to StatsBase/StatsAPI for `vif` and to StatsModels for `gvif` (which depends on having the model formula and hence needs `TableRegressionModel`).

Cool. Unfortunately this cannot live in StatsAPI as it calls `StatsBase.cov2cor!`. But we have moved almost all model-related functions out of StatsBase now, so `vif` would be kind of lonely there. Would it be OK for you to put it in StatsModels? Otherwise, we could put an empty definition in StatsAPI and have each package define a method for its custom type, even if that's suboptimal.

palday:
@nalimilan I would say put the `[g]vif` stub in wherever `RegressionModel` now lives and a general implementation in StatsModels.jl that other packages can optionally overload. Mostly I just want it to be as accessible as possible and since it was possible to implement `vif` using only functionality from StatsBase for a type owned by StatsBase, I thought that was good place for it. I think you have a better view of the entire ecosystem, so tell me where to put PRs and I'll open them. From my perspective, we have (noting that the stubs and default definitions don't necessarily need to appear in the same spots):

1. `vif` stub -- StatsAPI?
2. `vif(::RegressionModel)` default definition -- StatsBase or StatsModels?
3. `gvif` stub -- unsure, but probably same place as `vif` stub
4. `gvif(::RegressionModel; scale=false)` default definition. Technically requires a `formula` method to be defined, which `TableRegressionModel` and `MixedModel` both have, but which isn't AFAIK defined for general `RegressionModel`. So then StatsModels? 
4. `termnames`: should IMHO definitely be in StatsModels though @kleinschmidt may have thoughts and ideas about how to better define it / alternatives.

mschauer:
> Yes?

Nothing, I just remembered that you were looking for something like this is think, so I thought I point it out.

ararslan:
Thanks, I appreciate it :smile:

ararslan:
> Unfortunately this cannot live in StatsAPI as it calls `StatsBase.cov2cor!`.

`cov2cor!` is actually from Statistics (`Statistics.cov2cor! === StatsBase.cov2cor!`), so AFAICT it should be fine in StatsAPI.

nalimilan:
> `cov2cor!` is actually from Statistics (`Statistics.cov2cor! === StatsBase.cov2cor!`), so AFAICT it should be fine in StatsAPI.

Ah yes good catch. Though we kind of have the plan to move Statistics out of the stdlib and merge it with StatsBase so soon the problem would be the same. Also, the definition of `gvif` is not completely trivial so it wouldn't be appropriate in StatsAPI.

@palday Your plan sounds fine. Let's put stubs in StatsAPI, and discuss with @kleinschmidt whether it would be appropriate to put definitions in StatsModels.

nalimilan:
@palday Any plans to implement what you proposed? :-)

PR #699, opened 2021-07-13
fix tests for change in default RNG on nightly

palday:


test/sampling.jl
```diff
@@ -219,11 +219,14 @@ wv = Weights([zeros(5); 1:4; -1])
 @test_throws ErrorException sample(a, wv, 1, replace=false)
 
 #### weighted sampling with dimension
-Random.seed!(1);
 
-@test sample([1, 2], Weights([1, 1]), (2,2)) == ones(2,2)
+# shape
+@test size(sample([1, 2], Weights([1, 1]), (2,2))) == (2, 2)
```
ararslan, line 224: IMO testing the shape is not that useful, as the requested shape is literally part of the function call.

palday: that's pretty much what the test did previously ... I think part of it is determine that it is indeed returning the requested shape

palday: but I guess also that is tested in the other tests as well, so we could ditch it

ararslan: Nothing better than deleting tests if they fail on nightly

ararslan (approved):

PR #701, opened 2021-07-25
Added Minimal Support for Reliability Scores (aka Crombach's alpha)

storopoli:
This is a minimal implementation. It is based on a covariance matrix.
I created a struct `Reliability` to hold the total reliability score and the calculations for each reliability score if a certain item (i.e. column) was dropped from the covariance matrix.

There is a set of test functions and I used the [Wikipedia](https://en.wikipedia.org/wiki/Cronbach%27s_alpha) example for a covariance matrix both in the tests and in the docstrinngs.

This is my first PR in a Julia package, so please let me know what I can do improve...

Some worthy mentions:

R's `psych::alpha` has the following output:

```R
Reliability analysis   
Call: alpha(x = covmatrix)
  raw_alpha std.alpha G6(smc) average_r S/N median_r
      0.78      0.78    0.74      0.47 3.5     0.45
 Reliability if an item is dropped:
     raw_alpha std.alpha G6(smc) average_r S/N   var.r med.r
var1      0.71      0.71    0.62      0.45 2.4 0.00022  0.44
var2      0.72      0.72    0.64      0.46 2.6 0.01126  0.44
var3      0.71      0.71    0.63      0.45 2.4 0.00806  0.43
var4      0.77      0.77    0.69      0.53 3.3 0.00321  0.54
 Item statistics 
          r r.cor r.drop
var1   0.80  0.71   0.62
var2   0.79  0.68   0.60
var3   0.80  0.71   0.62
var4   0.72  0.56   0.50
```

If we were to include variable names we would have to include some `Tables.jl` or `DataFrames.jl` dependency.

Also, I am only calculating the "vanilla" Crombach's alpha. The `std.alpha`, `6(smc)`, `average_r S/N`, `var.r` and `med.r` were not implemented.


src/StatsBase.jl
```diff
@@ -212,7 +212,11 @@ export
     standardize,
     AbstractDataTransform, # the type to represent a abstract data transformation
     ZScoreTransform,       # the type to represent a z-score data transformation
-    UnitRangeTransform     # the type to represent a 0-1 data transformation
+    UnitRangeTransform,    # the type to represent a 0-1 data transformation
+
+    # Reliability
+    Reliability,           # the type to represent reliability scores
+    alpha                  # function to compute Crombach's alpha
```
bkamins, line 219: I would prefer some longer name as `alpha` seems too common

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
+using LinearAlgebra: Diagonal
+
+struct Reliability
+    alpha::Float64
+    dropped::Vector{Pair{Int64, Float64}}
```
bkamins, line 5:
```suggestion
    dropped::Vector{Pair{Int, Float64}}
```

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+end
+
+function Base.show(io::IO, x::Reliability)
+    println(io, "Reliability for all items: $(round(x.alpha; digits=3))")
```
bkamins, line 9: I usually like 4 digits better :).

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+item 2: 0.761
+item 3: 0.771
+item 4: 0.783
+```
+"""
+function alpha(covmatrix::AbstractMatrix{T}) where T <: Real
+    alpha = _alpha(covmatrix)
+    k = size(covmatrix, 2)
+    dropped = Vector{Pair{Int64, Float64}}(undef, k)
+    for i ∈ 1:k
```
storopoli, line 67: Do we think we need to annotate this `for` loop with a `@simd` macro for performance?

src/reliability.jl
storopoli: ok changed

src/StatsBase.jl
storopoli: changed to `crombach_alpha`

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+    println(io, "Reliability for all items: $(round(x.alpha; digits=3))")
+    println(io, "")
+    println(io, "Reliability if an item is dropped:")
+    for i ∈ x.dropped
+        println(io, "item $(i.first): $(round(i.second; digits=3))")
+    end
+end
+
+function _alpha(covmatrix::AbstractMatrix{T}) where T <: Real
+    k = size(covmatrix, 2)
```
bkamins, line 18: I would check if the matrix is square and symmetric (approximately).

bkamins: and with non-negative entries.
I assume that checking if it is a covariance matrix is an overkill.

src/StatsBase.jl
```diff
@@ -212,7 +212,11 @@ export
     standardize,
     AbstractDataTransform, # the type to represent a abstract data transformation
     ZScoreTransform,       # the type to represent a z-score data transformation
-    UnitRangeTransform     # the type to represent a 0-1 data transformation
+    UnitRangeTransform,    # the type to represent a 0-1 data transformation
+
+    # Reliability
+    Reliability,           # the type to represent reliability scores
+    crombach_alpha                  # function to compute Crombach's alpha
```
bkamins, line 219: alignment of the comment

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+    for i ∈ x.dropped
+        println(io, "item $(i.first): $(round(i.second; digits=4))")
+    end
+end
+
+function _crombach_alpha(covmatrix::AbstractMatrix{T}) where T <: Real
+    k = size(covmatrix, 2)
+    σ = sum(covmatrix)
+    σ_ij = sum(covmatrix - Diagonal(covmatrix)) / (k * (k - 1))
+    ρ = k^2 * σ_ij / σ
```
bkamins, line 21: Why not something like `k / (k - 1) * (1 - sum(i -> covmatrix[i, i], 1:k)/ σ)`. I have not checked but it seems it will be more numerically stable.

src/reliability.jl
bkamins: and that `k>1`.

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+```
+
+where k is the number of items, i.e. columns; σᵢⱼ denote the average of
+the inter-item covariances; and  σ²ₓ consists of item variances and inter-item
+covariances.
+
+Returns a `Reliability` object that holds:
+
+* `alpha`: the reliability score for all items, i.e. columns, in `comatrix`; and
+* `dropped`: A `Vector{Pair{item, score}}` reliability scores if a specific item, i.e. column, is dropped from `comatrix`.
```
bkamins, line 42: I would say "variable", as you drop both column and row - right?

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+item 1: 0.75
+item 2: 0.761
+item 3: 0.771
+item 4: 0.783
+```
+"""
+function crombach_alpha(covmatrix::AbstractMatrix{T}) where T <: Real
+    alpha = _crombach_alpha(covmatrix)
+    k = size(covmatrix, 2)
+    dropped = Vector{Pair{Int64, Float64}}(undef, k)
```
bkamins, line 66:
```suggestion
    dropped = Vector{Pair{Int, Float64}}(undef, k)
```

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+    @simd for i ∈ 1:k
```
bkamins, line 67: I think `@simd` and `@inbounds` is an overkill for such a simple function.

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+        reduced_covmatrix = covmatrix[1:end .!= i, 1:end .!= i]
```
bkamins, line 68: this will be very inefficient.

Seeing the code now I would do something like the following:
```
v = vec(sum(covmatrix, dims=1))
σ = sum(v)
for i in axes(v, 1)
    v[i][i] -= covmatrix[i, i]
end
σ_ij = sum(v)
```
and then you can compute crombach alpha from it. The point is that do compute dropped for variable `i` you just need to do:
```
σ - 2*v[i] - covmatrix[i,i]
σ_ij - 2*v[i]
```
as inputs to calculation. It will be much faster as you avoid almost all allocations.

Is the idea clear?

src/reliability.jl
```diff
@@ -0,0 +1,72 @@
...
+        @inbounds dropped[i] = Pair{Int64, Float64}(i, _crombach_alpha(reduced_covmatrix))
+    end
+    return Reliability(alpha, dropped)
```
bkamins, line 71: you need to make sure that in `dropped` you have values converted to `Float64`. Alternatively you might write a more general solution that would have parameter `T and just dynamically determine it based on the type of the computation result (a most visible change would be for `Rational` values and `BigFloat` potentially).

bkamins:
In general maybe it would be also good to add other reliability measures like https://en.wikipedia.org/wiki/Congeneric_reliability.
Also for the Cronbach alpha maybe it should be reported if the data satisfies tau-equivalence?

src/reliability.jl
storopoli: Something like:

```julia
using LinearAlgebra: cholesky

@assert issuccess(cholesky(cox_matrix, Val(true)))
```

src/reliability.jl
storopoli: Great, Wikipedia also list this formula. So we can use it...

nalimilan:
Thanks for the PR. Unfortunately I don't have time to review it right now, but I just wanted to note that the API design is faced with similar challenges as things like pairwise correlation and distances: how to pass variables (matrix, iterator of vectors, and/or Tables.jl object), and whether/how/when to report variable names. See https://github.com/JuliaStats/StatsBase.jl/pull/627 and https://github.com/nalimilan/FreqTables.jl/pull/54.

storopoli:
Great to see that this kind of API is being worked on. I will work on @bkamins suggestions, and make it ready once you are ready to review.

On quick question: Do any of you think we should expect a covariance matrix or should we expect a matrix and do the covariance matrix construction inside the `crombach_alpha` function?

@bkamins could you point me towards on how to implement a test if the the data satisfies tau-equivalence? I'm not a psychometrics specialist...

bkamins:
> Do any of you think we should expect a covariance matrix or should we expect a matrix and do the covariance matrix construction inside the `crombach_alpha` function?

I would assume this function should take a covariance matrix. However, then the challenge is with column names as @nalimilan said, as currently we do not have a set standard AFAICT how to pass them.

> how to implement a test if the the data satisfies tau-equivalence

all off-diagonal values should be the same. But maybe - if other software does not do this check we can also skip it and rely on the user to test it.

storopoli:
Ok, I also agree that the user should pass a covariance matrix.

Regarding tau equivalence, R's `psych` cannot handle that in the `alpha` function. It is instead something that only SEMs packages deals with. See [here](https://stats.stackexchange.com/questions/255004/cronbach-alpha-testing-for-tau-equivalence)

src/reliability.jl
storopoli: Here is a great discussion... Generally when you perform a Crombach's alpha (or any other reliability measures) you are using the nomenclature item and latent variable. So I avoided the word "variable" and opted for item and column of the covariance matrix. I think that if we include "variable" in the docstring would bring unavoid confusion.

storopoli: I will add "row" also

src/reliability.jl
storopoli: The code is not clear. I think that we need to do allocations because we cannot change inplace the covmatrix in the for-loop. This is because we need the matrix without a row/column each time to calculate what the `crombach_alpha` would be without that row/column.

Or am I getting this wrong.

bkamins:
>  It is instead something that only SEMs packages deals with.

That is why I have said above that we might consider adding other reliability scores. But we can leave it for later and skip checking the assumptions of the method here.

src/reliability.jl
bkamins: This is what this part:
```
σ - 2*v[i] - covmatrix[i,i]
σ_ij - 2*v[i]
```
is doing. It is calculating the parts of the `crombach_alpha` parameters without `i`-th row and column

storopoli: My deepest apologies, but I could not get the idea. I changed the code in the for loop to avoid an allocation... But I could not get your ideia.

bkamins: I can see that you have changed the formula you use so I also change it in my code.

Here is the implementation - you can check that it produces the same results:
```
function ca(covmatrix::AbstractMatrix{T}) where T <: Real
    isposdef(covmatrix) || throw(ArgumentError("Covariance matrix is not positive definite!"))
    k = size(covmatrix, 2)
    k > 1  || throw(ArgumentError("Covariance matrix has only one columnn!"))
    v = vec(sum(covmatrix, dims=1))
    σ = sum(v)
    for i in axes(v, 1)
        v[i] -= covmatrix[i, i]
    end
    σ_diag = sum(i -> covmatrix[i, i], 1:k)

    alpha = k / (k - 1) * (1 - σ_diag / σ)
    if k > 2
        dropped = Vector{Pair{Int, AbstractFloat}}(undef, k)
        for i ∈ 1:k
            dropped[i] = i => (k - 1) / (k - 2) * (1 - (σ_diag - covmatrix[i,i]) / (σ - 2*v[i] - covmatrix[i,i]))
        end
    else
        # if k = 2 do not produce dropped; this has to be also
        # correctly handled in show
        dropped = Pair{Int, AbstractFloat}[]
    end
    return (alpha, dropped) # I am not using your result struct as it might change its implementation
end
```

Now consider analyzing a larger matrix, eg. with 1000 items (unlikely, but shows the difference):
```
julia> z = Matrix{Float64}(I, 1000, 1000);

julia> @time ca(z); # my code
  0.024561 seconds (1.01 k allocations: 7.668 MiB)

julia> @time crombach_alpha(z); # your code
 29.676325 seconds (13.01 k allocations: 14.888 GiB, 4.79% gc time)
```

storopoli: Oh now I get it! Thanks! you've omitted the whole `_crombach()` auxiliary function. That's great. I will add the implementation as you've suggested. Huge gains of performance...

storopoli:
Ok, I think all suggestions and reviews were addressed.

One thing we need to decide is to whether use the `struct` to hold the results and offer a nice print from the `Base.show` (current implementation) or just return a tuple (with a `Dict` or vector of `Pair`) and let the user deal with that.

src/reliability.jl
bkamins: Ah - sorry for being brief!

I it is a bad habit of working with @nalimilan. I have been cooperating with him for over 3 years, and we seem to have converged to a state where we pass only minimal required information, and the rest is inferred from our common experience 😄.

Kind of https://www.huffingtonpost.com.au/2017/08/02/facebook-shuts-down-ai-robot-after-it-creates-its-own-language_a_23058978/.

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
+using LinearAlgebra: Diagonal, isposdef
+
+struct Reliability
+    alpha::AbstractFloat
+    dropped::Vector{Pair{Int, AbstractFloat}}
+end
```
bkamins, line 6: I will add suggestions below to reflect this change:
```suggestion
struct Reliability{T <: Real}
    alpha::T
    dropped::Vector{Pair{Int, T}}
end
```
In particular note that `Rational` is not a subtype of `AbstractFloat` (we might add a test where covariance matrix has `Rational` `eltype`).

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
...
+
+function Base.show(io::IO, x::Reliability)
+    println(io, "Reliability for all items: $(round(x.alpha; digits=4))")
+    println(io, "")
```
bkamins, line 10:
```suggestion
    isempty(x.dropped) && return
    println(io, "")
```

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
...
+Reliability for all items: 0.814
+
+Reliability if an item is dropped:
+item 1: 0.75
+item 2: 0.761
+item 3: 0.771
+item 4: 0.783
+```
+"""
+function crombach_alpha(covmatrix::AbstractMatrix{T}) where T <: Real
```
bkamins, line 54:
```suggestion
function crombach_alpha(covmatrix::AbstractMatrix{<:Real})
```

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
...
+    println(io, "Reliability for all items: $(round(x.alpha; digits=4))")
+    println(io, "")
+    println(io, "Reliability if an item is dropped:")
+    for i ∈ x.dropped
+        println(io, "item $(i.first): $(round(i.second; digits=4))")
+    end
+end
+
+"""
+    crombach_alpha(covmatrix::AbstractMatrix{T}) where T <: Real
```
bkamins, line 18:
```suggestion
    crombach_alpha(covmatrix::AbstractMatrix{<:Real})
```

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
...
+        dropped = Vector{Pair{Int, AbstractFloat}}(undef, k)
+        for i ∈ 1:k
+            dropped[i] = i => (k - 1) / (k - 2) * (1 - (σ_diag - covmatrix[i,i]) / (σ - 2*v[i] - covmatrix[i,i]))
+        end
+    else
+        # if k = 2 do not produce dropped; this has to be also
+        # correctly handled in show
+        dropped = Pair{Int, AbstractFloat}[]
+    end
+    return Reliability(alpha, dropped)
```
bkamins, line 76: ah - now I see that the definition above should just work given the changes I have proposed.

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
...
+    v = vec(sum(covmatrix, dims=1))
+    σ = sum(v)
+    for i in axes(v, 1)
+        v[i] -= covmatrix[i, i]
+    end
+    σ_diag = sum(i -> covmatrix[i, i], 1:k)
+
+    alpha = k / (k - 1) * (1 - σ_diag / σ)
+    if k > 2
+        dropped = Vector{Pair{Int, AbstractFloat}}(undef, k)
```
bkamins, line 67:
```suggestion
        dropped = Vector{Pair{Int, typeof(alpha)}}(undef, k)
```

test/reliability.jl
```diff
@@ -0,0 +1,15 @@
...
+             6 11 6 6;
+             6 6 12 6;
+             6 6 6 13]
+    reliabity_x = crombach_alpha(cov_X)
+    @test reliability_x.alpha ≈ 0.8135593220338981
+    @test reliabity_x.dropped[1].second ≈ 0.75
+    @test reliabity_x.dropped[2].second ≈ 0.7605633802816901
+    @test reliabity_x.dropped[3].second ≈ 0.7714285714285715
+    @test reliabity_x.dropped[4].second ≈ 0.782608695652174
+end # @testset "StatsBase.Reliability"
```
bkamins, line 15: as commented above. I think we need more tests for:
* `k=1`
* `k=2`
* `BigFloat` and `Rational` entries in covariance matrix

To make sure we get what is expected.

Thank you!

src/reliability.jl
storopoli: Nice I didn’t knew we could use this syntax!

test/reliability.jl
storopoli: Agreed, will include.

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+```jldoctest
+julia> using StatsBase
+
+julia> cov_X = [10 6 6 6;
+                6 11 6 6;
+                6 6 12 6;
+                6 6 6 13];
+
+julia> crombach_alpha(cov_X)
+Reliability for all items: 0.814
```
bkamins, line 46: probably the output here needs to be updated given the new rounding equal to 4.

storopoli:
Added the tests and fixed the typos in the `@testset`. CI should run OK now.

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
+using LinearAlgebra: Diagonal, isposdef
+
+struct Reliability{T <: Real}
+    alpha::T
+    dropped::Vector{Pair{Int, T}}
+end
+
+function Base.show(io::IO, x::Reliability)
```
bkamins, line 8: maybe add a test of `show` (you can check in DataFrames.jl repo how testing of display can be done)?

test/reliability.jl
```diff
@@ -0,0 +1,42 @@
...
+    reliability_X = crombach_alpha(cov_X)
+    @test reliability_X.alpha ≈ 0.8135593220338981
+    @test reliability_X.dropped[1].second ≈ 0.75
+    @test reliability_X.dropped[2].second ≈ 0.7605633802816901
+    @test reliability_X.dropped[3].second ≈ 0.7714285714285715
+    @test reliability_X.dropped[4].second ≈ 0.782608695652174
+
+    # testing Rational
+    cov_rational = 1 .// cov_X
+    reliability_rational = crombach_alpha(cov_X)
```
bkamins, line 19:
```suggestion
    reliability_rational = crombach_alpha(cov_X)
    @test reliability_bigfloat isa Reliability{Rational{Int}}
```

test/reliability.jl
```diff
@@ -0,0 +1,42 @@
+using StatsBase
+using LinearAlgebra, Random, Test
+
+@testset "StatsBase.Reliability" begin
+    # basic vanilla test
+    cov_X = [10 6 6 6;
+             6 11 6 6;
+             6 6 12 6;
+             6 6 6 13]
+    reliability_X = crombach_alpha(cov_X)
```
bkamins, line 10:
```suggestion
    reliability_X = crombach_alpha(cov_X)
    @test reliability_bigfloat isa Reliability{Float64}
```

test/reliability.jl
```diff
@@ -0,0 +1,42 @@
...
+    @test reliability_X.alpha ≈ 0.8135593220338981
+    @test reliability_X.dropped[1].second ≈ 0.75
+    @test reliability_X.dropped[2].second ≈ 0.7605633802816901
+    @test reliability_X.dropped[3].second ≈ 0.7714285714285715
+    @test reliability_X.dropped[4].second ≈ 0.782608695652174
+
+    # testing Rational
+    cov_rational = 1 .// cov_X
```
bkamins, line 18:
```suggestion
    cov_rational = cov_X .// 1
```

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+    k = size(covmatrix, 2)
+    k > 1  || throw(ArgumentError("Covariance matrix has only one columnn!"))
+    v = vec(sum(covmatrix, dims=1))
+    σ = sum(v)
+    for i in axes(v, 1)
+        v[i] -= covmatrix[i, i]
+    end
+    σ_diag = sum(i -> covmatrix[i, i], 1:k)
+
+    alpha = k / (k - 1) * (1 - σ_diag / σ)
```
bkamins, line 66: This is a correct order of operations to get a correct return type:
```suggestion
    alpha = k * (1 - σ_diag / σ) / (k - 1)
```

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+    if k > 2
+        dropped = Vector{Pair{Int, typeof(alpha)}}(undef, k)
+        for i ∈ 1:k
+            dropped[i] = i => (k - 1) / (k - 2) * (1 - (σ_diag - covmatrix[i,i]) / (σ - 2*v[i] - covmatrix[i,i]))
```
bkamins, line 70: same here:
```suggestion
            dropped[i] = i => (k - 1) * (1 - (σ_diag - covmatrix[i,i]) / (σ - 2*v[i] - covmatrix[i,i])) / (k - 2)
```

test/reliability.jl
```diff
@@ -0,0 +1,42 @@
...
+    reliability_rational = crombach_alpha(cov_X)
+    @test reliability_rational.alpha ≈ 0.8135593220338981
+    @test reliability_rational.dropped[1].second ≈ 0.75
+    @test reliability_rational.dropped[2].second ≈ 0.7605633802816901
+    @test reliability_rational.dropped[3].second ≈ 0.7714285714285715
+    @test reliability_rational.dropped[4].second ≈ 0.782608695652174
```
bkamins, line 24:
```suggestion
    @test reliability_rational.alpha == 48 // 59
    @test reliability_rational.dropped[1].second == 3 // 4
    @test reliability_rational.dropped[2].second == 54 // 71
    @test reliability_rational.dropped[3].second == 27 // 35
    @test reliability_rational.dropped[4].second == 18 // 23
```

bkamins: this is a point of making `Rational` parametric - e.g. if someone wants to make a precise computation without rounding it is done as expected.

bkamins:
You can add suggestions to batches and commit them in one-shot (this lowers the cost of running CI).

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
+using LinearAlgebra: Diagonal, isposdef
```
bkamins, line 1:
```suggestion
using LinearAlgebra: isposdef
```

test/reliability.jl
```diff
@@ -0,0 +1,45 @@
...
+using LinearAlgebra, Random, Test
+
+@testset "StatsBase.Reliability" begin
+    # basic vanilla test
+    cov_X = [10 6 6 6;
+             6 11 6 6;
+             6 6 12 6;
+             6 6 6 13]
+    reliability_X = crombach_alpha(cov_X)
+    @test reliability_bigfloat isa Reliability{Float64}
```
bkamins, line 11:
```suggestion
    @test reliability_X isa Reliability{Float64}
```

test/reliability.jl
```diff
@@ -0,0 +1,45 @@
...
+    @test reliability_X.alpha ≈ 0.8135593220338981
+    @test reliability_X.dropped[1].second ≈ 0.75
+    @test reliability_X.dropped[2].second ≈ 0.7605633802816901
+    @test reliability_X.dropped[3].second ≈ 0.7714285714285715
+    @test reliability_X.dropped[4].second ≈ 0.782608695652174
+
+    # testing Rational
+    cov_rational = cov_X .// 1
+    reliability_rational = crombach_alpha(cov_X)
+    @test reliability_bigfloat isa Reliability{Rational{Int}}
```
bkamins:
```suggestion
    @test reliability_rational isa Reliability{Rational{Int}}
```

test/reliability.jl
```diff
@@ -0,0 +1,45 @@
...
+    @test reliability_rational.alpha == 48 // 59
+    @test reliability_rational.dropped[1].second == 3 // 4
+    @test reliability_rational.dropped[2].second == 54 // 71
+    @test reliability_rational.dropped[3].second == 27 // 35
+    @test reliability_rational.dropped[4].second == 18 // 23
+
+    # testing BigFloat
+    cov_bigfloat = BigFloat.(cov_X)
+    reliability_bigfloat = crombach_alpha(cov_X)
```
bkamins:
```suggestion
    reliability_bigfloat = crombach_alpha(cov_bigfloat)
```

bkamins (approved):
Nice - it looks good now. I hope you enjoyed it :). Let us wait for other reviewers now.

storopoli:
Yeah I learned a ton, thanks! I had huge respect for you and all of your work. After this PR it increased orders of magnitude!

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
+using LinearAlgebra: isposdef
+
+struct Reliability{T <: Real}
+    alpha::T
+    dropped::Vector{Pair{Int, T}}
```
bkamins, line 5: The first parameter here should be `Any` not `Int`
```suggestion
    dropped::Vector{Pair{Any, T}}
```
I think.

This would allow passing names to items. The `Int` we have now is not useful (it is just a position in a vector so it is redundant).

@storopoli - I would wait with working on this on @nalimilan to comment about the API and passing item names here.

@nalimilan - I believe that the computational part in this PR is finished. We just need to decide on the way we handle the item names.

mschauer:
Bump

src/StatsBase.jl
```diff
@@ -212,7 +212,11 @@ export
     standardize,
     AbstractDataTransform, # the type to represent a abstract data transformation
     ZScoreTransform,       # the type to represent a z-score data transformation
-    UnitRangeTransform     # the type to represent a 0-1 data transformation
+    UnitRangeTransform,    # the type to represent a 0-1 data transformation
+
+    # Reliability
+    Reliability,           # the type to represent reliability scores
+    crombach_alpha         # function to compute Crombach's alpha
```
nalimilan, line 219: AFAIK the name is Cro**n**bach, right? :-D

Also, most of the function names in StatsBase do not have an underscore, so it would be more consistent to skip it here (like e.g. `cooksdistance`). Or should we use a name based on "reliability" or "tau" instead? The discrepancy between the function name and the `Reliability` type isn't ideal.

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+end
+
+function Base.show(io::IO, x::Reliability)
+    println(io, "Reliability for all items: $(round(x.alpha; digits=4))")
```
nalimilan, line 9: I suggest using `@printf` to ensure that the printed value always has 4 digits. Using `round`, fewer digits will be printed they are enough (like for 1.0), which is less clear for users:
```suggestion
    @printf(io, "Reliability for all items: %.4f", x.alpha)
```
(Also needs `using Printf` somewhere if that's not already loaded.)

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
+using LinearAlgebra: isposdef
```
nalimilan, line 1: Given that this is at the top level, it will affect other files, so better put it in src/StatsBase.jl.

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+    dropped::Vector{Pair{Int, T}}
+end
+
+function Base.show(io::IO, x::Reliability)
+    println(io, "Reliability for all items: $(round(x.alpha; digits=4))")
+    isempty(x.dropped) && return
+    println(io, "")
+    println(io, "Reliability if an item is dropped:")
+    for i ∈ x.dropped
+        println(io, "item $(i.first): $(round(i.second; digits=4))")
```
nalimilan, line 14:
```suggestion
        @printf(io, "item %i: %.4f\n, i.first, i.second")
```

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+    end
+end
+
+"""
+    crombach_alpha(covmatrix::AbstractMatrix{<:Real})
+
+Calculate Crombach's alpha (1951) from a covariance matrix `covmatrix` according to
+the Wikipedia formula (https://en.wikipedia.org/wiki/Cronbach%27s_alpha):
```
nalimilan, line 22:
```suggestion
the [formula](https://en.wikipedia.org/wiki/Cronbach%27s_alpha):
```
(Wikipedia isn't an authoritative nor stable source)

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+```math
+\\rho = \\frac{k}{k-1} (1 - \\frac{\\sum^k_{i=1} \\sigma^2_i}{\\sigma^2_X})
+```
+
+where k is the number of items, i.e. columns; σᵢ denotes item variance;
+and σ²ₓ consists of item variances and inter-item covariances.
+
+Returns a `Reliability` object that holds:
+
+* `alpha`: the reliability score for all items, i.e. columns, in `comatrix`; and
```
nalimilan, line 33:
```suggestion
* `alpha`: the reliability score for all items, i.e. columns, in `covmatrix`; and
```

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+* `dropped`: A `Vector{Pair{item, score}}` reliability scores if a specific item, i.e. column/row, is dropped from `comatrix`.
```
nalimilan, line 34:
```suggestion
* `dropped`: A `Vector{Pair{item, score}}` giving reliability scores if a specific item,
  i.e. column, is dropped from `covmatrix`.
```

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+
+Reliability if an item is dropped:
+item 1: 0.75
+item 2: 0.7606
+item 3: 0.7714
+item 4: 0.7836
+```
+"""
+function crombach_alpha(covmatrix::AbstractMatrix{<:Real})
+    isposdef(covmatrix) || throw(ArgumentError("Covariance matrix is not positive definite!"))
```
nalimilan, line 56:
```suggestion
    isposdef(covmatrix) || throw(ArgumentError("Covariance matrix must be positive definite."))
```
It would be nice for users to check that the matrix is square first: saying that it's not positive-definite can be intimidating when the error is more trivial than that.

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+    k = size(covmatrix, 2)
+    k > 1  || throw(ArgumentError("Covariance matrix has only one columnn!"))
```
nalimilan, line 58:
```suggestion
    k > 1  || throw(ArgumentError("Covariance matrix must have more than one column."))
```

src/reliability.jl
nalimilan: I suggest we only store a `Vector{T}` of values for now. Names can be stored in a separate field if we add support for them later. Anyway it's more practical to store values separately from names than having to access the fields in a pair.

src/reliability.jl
```diff
@@ -0,0 +1,78 @@
...
+
+Calculate Crombach's alpha (1951) from a covariance matrix `covmatrix` according to
+the Wikipedia formula (https://en.wikipedia.org/wiki/Cronbach%27s_alpha):
+
+```math
+\\rho = \\frac{k}{k-1} (1 - \\frac{\\sum^k_{i=1} \\sigma^2_i}{\\sigma^2_X})
+```
+
+where k is the number of items, i.e. columns; σᵢ denotes item variance;
+and σ²ₓ consists of item variances and inter-item covariances.
```
nalimilan, line 29:
```suggestion
where ``k`` is the number of items, i.e. columns; ``\\sigma_i`` denotes item variance;
and ``\\sigma^2_i`` consists of item variances and inter-item covariances.
```

test/reliability.jl
```diff
@@ -0,0 +1,60 @@
...
+
+    # testing corner cases
+    @test_throws MethodError crombach_alpha([1.0, 2.0])
+    cov_k2 = [10 6;
+              6 11]
+    reliability_k2 = crombach_alpha(cov_k2)
+    @test reliability_k2.alpha ≈ 0.7272727272727273
+    @test isempty(reliability_k2.dropped)
+
+    # testing Base.show
```
nalimilan, line 46: Also test the case where there are only two columns?

test/reliability.jl
```diff
@@ -0,0 +1,60 @@
...
+    reliability_bigfloat = crombach_alpha(cov_bigfloat)
+    @test reliability_bigfloat isa Reliability{BigFloat}
+    @test reliability_bigfloat.alpha ≈ 0.8135593220338981
+    @test reliability_bigfloat.dropped[1].second ≈ 0.75
+    @test reliability_bigfloat.dropped[2].second ≈ 0.7605633802816901
+    @test reliability_bigfloat.dropped[3].second ≈ 0.7714285714285715
+    @test reliability_bigfloat.dropped[4].second ≈ 0.782608695652174
+
+    # testing corner cases
+    @test_throws MethodError crombach_alpha([1.0, 2.0])
```
nalimilan, line 39: Also test the two error paths where matrix isn't positive-definite and when it has zero or one columns.

nalimilan (commented):
Sorry foe the delay!

src/reliability.jl
storopoli: It is already in `StatsBase.jl`

test/reliability.jl
storopoli: Done, I also included `missing` in one of the tests

src/reliability.jl
storopoli: Done.

src/StatsBase.jl
storopoli: Changed to `crombachalpha`

storopoli:
Thanks @nalimilan I've implemented the desired tests, converted the `Vector` of `Pair`s to a simple `Vector`, changed the function naming to `crombachalpha` and changed the `round` printing stuff to `Printf`.

src/StatsBase.jl
```diff
@@ -11,6 +11,7 @@ using Missings
 
 using Statistics
 using LinearAlgebra
+import LinearAlgebra: isposdef
```
nalimilan, line 14: Actually this probably isn't needed given the previous line.
```suggestion
```

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
...
+        v[i] -= covmatrix[i, i]
+    end
+    σ_diag = sum(i -> covmatrix[i, i], 1:k)
+
+    alpha = k * (1 - σ_diag / σ) / (k - 1)
+    if k > 2
+        dropped = Vector{typeof(alpha)}(undef, k)
+        for i in eachindex(dropped)
+            dropped[i] = (k - 1) * (1 - (σ_diag - covmatrix[i,i]) / (σ - 2*v[i] - covmatrix[i,i])) / (k - 2)
+        end
```
nalimilan, line 70: Even simpler:
```suggestion
        dropped = typeof(alpha)[(k - 1) * (1 - (σ_diag - covmatrix[i, i]) / (σ - 2*v[i] - covmatrix[i, i])) / (k - 2)
                                for i in eachindex(dropped)]
```

src/reliability.jl
```diff
@@ -43,20 +41,20 @@ julia> cov_X = [10 6 6 6;
...
 item 3: 0.7714
 item 4: 0.7836
 ```
 """
-function crombach_alpha(covmatrix::AbstractMatrix{<:Real})
+function crombachalpha(covmatrix::AbstractMatrix{<:Real})
     isposdef(covmatrix) || throw(ArgumentError("Covariance matrix must be positive definite."))
     k = size(covmatrix, 2)
-    k > 1  || throw(ArgumentError("Covariance matrix must have more than one column."))
+    k >= 1  || throw(ArgumentError("Covariance matrix must have more than one column."))
```
nalimilan, line 57: Is this change intended? The error message doesn't match the check now.

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
...
+             6 11 6 6;
+             6 6 12 6;
+             6 6 6 13]
+    reliability_X = crombachalpha(cov_X)
+    @test reliability_X isa Reliability{Float64}
+    @test reliability_X.alpha ≈ 0.8135593220338981
+    @test reliability_X.dropped[1] ≈ 0.75
+    @test reliability_X.dropped[2] ≈ 0.7605633802816901
+    @test reliability_X.dropped[3] ≈ 0.7714285714285715
+    @test reliability_X.dropped[4] ≈ 0.782608695652174
```
nalimilan, line 16:
```suggestion
    @test reliability_X.dropped ≈
        [0.75, 0.7605633802816901, 0.7714285714285715, 0.782608695652174]
```

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
...
+
+    # testing Rational
+    cov_rational = cov_X .// 1
+    reliability_rational = crombachalpha(cov_rational)
+    @test reliability_rational isa Reliability{Rational{Int}}
+    @test reliability_rational.alpha == 48 // 59
+    @test reliability_rational.dropped[1] == 3 // 4
+    @test reliability_rational.dropped[2] == 54 // 71
+    @test reliability_rational.dropped[3] == 27 // 35
+    @test reliability_rational.dropped[4] == 18 // 23
```
nalimilan, line 26:
```suggestion
    @test reliability_rational.dropped ==
        [3 // 4, 54 // 71, 27 // 35, 18 // 23]
```

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
...
+
+    # testing BigFloat
+    cov_bigfloat = BigFloat.(cov_X)
+    reliability_bigfloat = crombachalpha(cov_bigfloat)
+    @test reliability_bigfloat isa Reliability{BigFloat}
+    @test reliability_bigfloat.alpha ≈ 0.8135593220338981
+    @test reliability_bigfloat.dropped[1] ≈ 0.75
+    @test reliability_bigfloat.dropped[2] ≈ 0.7605633802816901
+    @test reliability_bigfloat.dropped[3] ≈ 0.7714285714285715
+    @test reliability_bigfloat.dropped[4] ≈ 0.782608695652174
```
nalimilan, line 36:
```suggestion
    @test reliability_bigfloat.dropped ≈
        [0.75, 0.7605633802816901, 0.7714285714285715, 0.782608695652174]
```

test/reliability.jl
```diff
@@ -7,52 +7,71 @@ using LinearAlgebra, Random, Test
...
+
     # testing Base.show
-    reliability_X = crombach_alpha(cov_X)
+    reliability_X = crombachalpha(cov_X)
     io = IOBuffer()
     show(io, reliability_X)
     str = String(take!(io))
     @test str == """
         Reliability for all items: 0.8136
-
```
nalimilan: Sorry, I didn't mean to remove this empty line in the output. Just add `\n` at the end of the string passed to `@printf` to restore it.

src/StatsBase.jl
nalimilan: Thanks. It should still be called Cronbach though. :-)

Also, how about calling the struct `CronbachAlpha` for consistency? `Reliability` refers to something more general AFAICT.

src/reliability.jl
```diff
@@ -0,0 +1,77 @@
...
+\\rho = \\frac{k}{k-1} (1 - \\frac{\\sum^k_{i=1} \\sigma^2_i}{\\sigma^2_X})
+```
+
+where ``k`` is the number of items, i.e. columns; ``\\sigma_i`` denotes item variance;
+and ``\\sigma^2_i`` consists of item variances and inter-item covariances.
+
+Returns a `Reliability` object that holds:
+
+* `alpha`: the reliability score for all items, i.e. columns, in `covmatrix`; and
+* `dropped`: A `Vector{Pair{item, score}}` giving reliability scores if a specific item,
```
nalimilan, line 32:
```suggestion
* `dropped`: A vector giving reliability scores if a specific item,
```

src/reliability.jl
nalimilan: What about my second comment?

test/reliability.jl
nalimilan: This case is still not tested for the `show` method AFAICT.

nalimilan (commented):
Thanks. Just a few more comments.

Can you also add a reference to the function in the docs so that it's included in the manual? Maybe a new section under Scalar Statistics?

src/reliability.jl
storopoli: Yes, you are right. I made a terrible confusion. Let's revert:

```suggestion
    k > 1  || throw(ArgumentError("Covariance matrix must have more than one column."))
```

test/reliability.jl
storopoli: Sorry I don't understand. This is how DataFrames.jl does the Base.show testing. How would we improve/change this?

nalimilan: I just mean that since the printing is different when the input matrix has only two columns, the output should also be tested for this special case.

test/reliability.jl
storopoli: > Thanks. Just a few more comments.
> 
> Can you also add a reference to the function in the docs so that it's included in the manual? Maybe a new section under Scalar Statistics?

I am finishing it up the `.md` file for documentation. I also noticed a huge mistake. It's Cronbach not Crombach! (https://en.wikipedia.org/wiki/Cronbach%27s_alpha). I will change the function naming accordingly...

src/StatsBase.jl
storopoli: Sure `CrombachAlpha`

test/reliability.jl
storopoli: Oh test Base.show for k=2 (2-col covmatrix)! OK

storopoli:
Ready for a review. See if the docs implementation is enough (never done it before, but followed from the examples in code)

src/reliability.jl
```diff
@@ -0,0 +1,75 @@
...
+    println(io, "Reliability if an item is dropped:")
+    for (idx, val) in enumerate(x.dropped)
+        @printf(io, "item %i: %.4f\n", idx, val)
+    end
+end
+
+"""
+    cronbachalpha(covmatrix::AbstractMatrix{<:Real})
+
+Calculate cronbach's alpha (1951) from a covariance matrix `covmatrix` according to
```
nalimilan, line 19:
```suggestion
Calculate Cronbach's alpha (1951) from a covariance matrix `covmatrix` according to
```

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
...
+
+
+    # testing Base.show
+    cronbach_X = cronbachalpha(cov_X)
+    io = IOBuffer()
+    show(io, cronbach_X)
+    str = String(take!(io))
+    @test str == """
+        Reliability for all items: 0.8136\n
+        Reliability if an item is dropped:
```
nalimilan, line 67:
```suggestion
        Reliability for all items: 0.8136

        Reliability if an item is dropped:
```

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
...
+        item 1: 0.7500
+        item 2: 0.7606
+        item 3: 0.7714
+        item 4: 0.7826
+        """
+    # for two columns
+    io = IOBuffer()
+    show(io, cronbach_k2)
+    str = String(take!(io))
+    @test str == "Reliability for all items: 0.7273"
```
nalimilan, line 77: A final `\n` should be printed (consistent with the k > 2 case).

src/reliability.jl
```diff
@@ -0,0 +1,75 @@
...
+\\rho = \\frac{k}{k-1} (1 - \\frac{\\sum^k_{i=1} \\sigma^2_i}{\\sigma^2_X})
+```
+
+where ``k`` is the number of items, i.e. columns; ``\\sigma_i`` denotes item variance;
+and ``\\sigma^2_i`` consists of item variances and inter-item covariances.
+
+Returns a `CronbachAlpha` object that holds:
+
+* `alpha`: the reliability score for all items, i.e. columns, in `covmatrix`; and
+* `dropped`: A vector giving reliability scores if a specific item,
```
nalimilan, line 32:
```suggestion
* `dropped`: a vector giving reliability scores if a specific item,
```

src/reliability.jl
```diff
@@ -0,0 +1,75 @@
+struct CronbachAlpha{T <: Real}
+    alpha::T
+    dropped::Vector{T}
+end
+
+function Base.show(io::IO, x::CronbachAlpha)
+    @printf(io, "Reliability for all items: %.4f", x.alpha)
```
nalimilan, line 7: Maybe we'd better just say "Cronbach's alpha" rather than "reliability" here and below? Otherwise, nothing in the output indicates what's exactly the coefficient that we're reporting.

nalimilan (commented):
> See if the docs implementation is enough (never done it before, but followed from the examples in code)

Thanks, looks good.

test/reliability.jl
storopoli: Done!

src/reliability.jl
storopoli: Done!

storopoli:
@nalimilan see if now is fine. I've replaced all "reliability" to "Cronbach's alpha"

bkamins:
Looks good for me.

bkamins:
(I am OOO so I have not checked if the documentation renders properly)

src/StatsBase.jl
```diff
@@ -212,7 +212,11 @@ export
     standardize,
     AbstractDataTransform, # the type to represent a abstract data transformation
     ZScoreTransform,       # the type to represent a z-score data transformation
-    UnitRangeTransform     # the type to represent a 0-1 data transformation
+    UnitRangeTransform,    # the type to represent a 0-1 data transformation
+
+    # Reliability
```
nalimilan, line 217:
```suggestion
    # reliability
```

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
+using StatsBase
+using LinearAlgebra, Random, Test
+
+@testset "StatsBase.CronbachAlpha" begin
```
nalimilan, line 4:
```suggestion
@testset "Cronbach's Alpha" begin
```

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
...
+        item 3: 0.7714
+        item 4: 0.7826
+        """
+    # for two columns
+    io = IOBuffer()
+    show(io, cronbach_k2)
+    str = String(take!(io))
+    @test str == "Cronbach's alpha for all items: 0.7273\n"
+
+end # @testset "StatsBase.CronbachAlpha"
```
nalimilan, line 79:
```suggestion
end # @testset "Cronbach's Alpha"
```

src/reliability.jl
```diff
@@ -0,0 +1,74 @@
...
+                6 6 6 13];
+
+julia> cronbachalpha(cov_X)
+Cronbach's alpha for all items: 0.8136
+
+Cronbach's alpha if an item is dropped:
+item 1: 0.7500
+item 2: 0.7606
+item 3: 0.7714
+item 4: 0.7836
```
nalimilan, line 50: To fix doctest failure:
```suggestion
item 4: 0.7826
```

test/reliability.jl
```diff
@@ -0,0 +1,79 @@
...
+
+
+    # testing Base.show
+    cronbach_X = cronbachalpha(cov_X)
+    io = IOBuffer()
+    show(io, cronbach_X)
+    str = String(take!(io))
+    @test str == """
+        Cronbach's alpha for all items: 0.8136\n
+        Cronbach's alpha if an item is dropped:
```
nalimilan, line 67:
```suggestion
        Cronbach's alpha for all items: 0.8136

        Cronbach's alpha if an item is dropped:
```

nalimilan:
Thanks!

storopoli:
Great! My Pleasure!

PR #706, opened 2021-08-04
Fix _addcounts_radix_sort for empty input array

bosonbaas:
This PR introduces a test to keep `_addcounts_radix_sort_loop!` from attempting to index an empty array. This error was causing `countmap` to fail for an input of an empty array (a test which demonstrates this has also been added).

Definitely happy to change this if the fix should be added in another place.

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -336,6 +336,9 @@ const BaseRadixSortSafeTypes = Union{Int8, Int16, Int32, Int64, Int128,
 radixsort_safe(::Type{T}) where T = T<:BaseRadixSortSafeTypes
 
 function _addcounts_radix_sort_loop!(cm::Dict{T}, sx::AbstractArray{T}) where T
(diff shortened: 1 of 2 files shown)


src/counts.jl
```diff
@@ -336,6 +336,9 @@ const BaseRadixSortSafeTypes = Union{Int8, Int16, Int32, Int64, Int128,
 radixsort_safe(::Type{T}) where T = T<:BaseRadixSortSafeTypes
 
 function _addcounts_radix_sort_loop!(cm::Dict{T}, sx::AbstractArray{T}) where T
+    if isempty(sx)
+      return cm
+    end
```
nalimilan, line 341:
```suggestion
    isempty(sx) && return cm
```

nalimilan (approved):
Thanks, just a minor stylistic suggestion!

PR #714, opened 2021-09-01
Fix type instability of `entropy` and generalize `crossentropy` and `kldivergence`

devmotion:
While I was working on a PR to Distributions, I noticed that `entropy` is not type stable:
```julia
julia> @inferred entropy([0, 1])
ERROR: return type Float64 does not match inferred return type Union{Float64, Int64}
```
This PR fixes the issue by using `LogExpFunctions.xlogx` instead of `x -> x > 0 ? zero(x) : z * log(x)`. Moreover, it bases `crossentropy` and `kldivergence` on `LogExpFunctions.xlogy` and generalizes them to non-Float64 return types and arguments with different element types.


nalimilan:
Thanks, that indeed simplifies the code a lot. Have you checked that performance doesn't regress? Also, does it still work for empty inputs?

devmotion:
> Thanks, that indeed simplifies the code a lot. Have you checked that performance doesn't regress?
> Also, does it still work for empty inputs?

With the latest release:
```julia
julia> using StatsBase, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p, q = rand(100_000), rand(100_000);

julia> @btime entropy($p);
  753.493 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($p, $q);
  779.563 μs (0 allocations: 0 bytes)

julia> @btime kldivergence($p, $q);
  1.540 ms (0 allocations: 0 bytes)

julia> pint, qint = rand((0, 1), 100_000), rand((0, 1), 100_000);

julia> @btime entropy($pint);
  819.110 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($pint, $qint);
  910.370 μs (0 allocations: 0 bytes)

julia> @btime kldivergence($pint, $qint);
  948.375 μs (0 allocations: 0 bytes)

julia> entropy(Float64[])
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
...

julia> crossentropy(Float64[], Float64[])
-0.0

julia> kldivergence(Float64[], Float64[])
0.0
```

With this PR:
```julia
julia> using StatsBase, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p, q = rand(100_000), rand(100_000);

julia> @btime entropy($p);
  837.509 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($p, $q);
  1.133 ms (0 allocations: 0 bytes)

julia> @btime kldivergence($p, $q);
  1.680 ms (0 allocations: 0 bytes)

julia> pint, qint = rand((0, 1), 100_000), rand((0, 1), 100_000);

julia> @btime entropy($pint);
  1.116 ms (0 allocations: 0 bytes)

julia> @btime crossentropy($pint, $qint);
  1.210 ms (0 allocations: 0 bytes)

julia> @btime kldivergence($pint, $qint);
  1.725 ms (0 allocations: 0 bytes)

julia> entropy(Float64[])
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
...

julia> crossentropy(Float64[], Float64[])
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
...

julia> kldivergence(Float64[], Float64[])
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
...
```

So there seems to be some performance regression and empty inputs don't work but also don't work with `entropy` in the latest release. I will look into why there is a performance drop. For `entropy` I just updated the mapping function in `sum` with `xlogx`, so I wonder why it is slower than before - maybe it could be caused by the use of `ifelse` in LogExpFunctions? If empty inputs should be supported, I could just add a check and compute the return type with `zero(eltype(p))` etc. I assume (this would also fix the problem in the latest release I guess).

devmotion:
On a second thought, it is not completely clear to me anymore if it is reasonable to define `entropy`, `kldivergence` or `crossentropy` for empty collections, and what values one would expect.

nalimilan:
Given that these are sums, isn't it logical to return zero for empty inputs? Anyway, throwing an error instead would be breaking so we would at least need to keep returning zero, possible with a deprecation warning.

devmotion:
`entropy` already throws an error but indeed it would be a breaking change for `crossentropy` and `kldivergence`.

My main motivation for throwing an error would be that empty vectors don't represent probability distributions.

devmotion:
So what's the plan here? Return a typestable 0 for empty inputs also for `entropy`, as currently done for `crossentropy` and `kldivergence`? Or keep throwing a (possibly more meaningful) error in the case of `entropy`?

mschauer:
> My main motivation for throwing an error would be that empty vectors don't represent probability distributions.

I find that convincing.

nalimilan:
As you prefer, but I just think we shouldn't add new errors for now to avoid being breaking. A deprecation warning would be OK, with a PR to turn them into errors in the next breaking release.

BTW should we check that the values sum to one if we want to ensure they are probability distributions (with a way to skip the check)?

src/scalarstats.jl
```diff
@@ -584,21 +584,12 @@ end
...
-    z = zero(T)
-    for i = 1:length(p)
-        @inbounds pi = p[i]
-        @inbounds qi = q[i]
-        if pi > z
-            s += pi * log(qi)
-        end
-    end
-    return -s
+    return - sum(xlogy(pi, qi) for (pi, qi) in zip(p, q))
```
nalimilan, line 589: Note that using lazy broadcasting (on recent julia versions) is more efficient and has better precision as it uses pairwise summation. Maybe that would fix the performance regression?
```julia
    return - sum(Broadcast.instantiate(Broadcast.broadcasted(xlogx, p, q)))
```

devmotion: I've seen the Julia PR and played around with `sum(Brodcast....)` in some local branch of Distributions but in my benchmarks there another implementation was actually faster. I don't remember the details right now and I assume it was not really comparable to this example here though. I'll check if it helps but I guess we have to live with a slight performance regression - the existing implementations are not type stable and less general, so it is not completely fair to compare the performance of the bugfixes in this PR to the performance of the incorrect existing implementation (of course, we should still try to improve the performance of the bugfixes as much as possible).

nalimilan: Improved precision may still be valuable anyway.

devmotion:
I reran the benchmarks with Julia 1.6.3 and also compared it with the `sum(Broadcast....)` suggestion. There is still a performance regression (switching from `ifelse` to ternary operators in LogExpFunctions helped a bit but did not resolve them) and it seems the comparison between the PR and `sum(Broadcast...)` is not conclusive and none of the approaches is clearly better than the other.

Without this PR:
```julia
julia> using StatsBase, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p, q = rand(100_000), rand(100_000);

julia> @btime entropy($p);
  732.629 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($p, $q);
  766.728 μs (0 allocations: 0 bytes)

julia> @btime kldivergence($p, $q);
  1.441 ms (0 allocations: 0 bytes)

julia> pint, qint = rand((0, 1), 100_000), rand((0, 1), 100_000);

julia> @btime entropy($pint);
  788.951 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($pint, $qint);
  895.965 μs (0 allocations: 0 bytes)

julia> @btime kldivergence($pint, $qint);
  919.979 μs (0 allocations: 0 bytes)
```

With this PR (commit https://github.com/JuliaStats/StatsBase.jl/pull/714/commits/97d3bfadae8d560c6d0aeda799cedcaffeaf677a):
```julia
julia> using StatsBase, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p, q = rand(100_000), rand(100_000);

julia> @btime entropy($p);
  763.036 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($p, $q);
  1.100 ms (0 allocations: 0 bytes)

julia> @btime kldivergence($p, $q);
  1.557 ms (0 allocations: 0 bytes)

julia> pint, qint = rand((0, 1), 100_000), rand((0, 1), 100_000);

julia> @btime entropy($pint);
  996.637 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($pint, $qint);
  1.178 ms (0 allocations: 0 bytes)

julia> @btime kldivergence($pint, $qint);
  1.604 ms (0 allocations: 0 bytes)
```

With `sum(Broadcast....)`:
```julia
julia> using StatsBase, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p, q = rand(100_000), rand(100_000);

julia> @btime entropy($p);
  763.060 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($p, $q);
  899.686 μs (0 allocations: 0 bytes)

julia> @btime kldivergence($p, $q);
  1.582 ms (0 allocations: 0 bytes)

julia> pint, qint = rand((0, 1), 100_000), rand((0, 1), 100_000);

julia> @btime entropy($pint);
  997.401 μs (0 allocations: 0 bytes)

julia> @btime crossentropy($pint, $qint);
  1.312 ms (0 allocations: 0 bytes)

julia> @btime kldivergence($pint, $qint);
  1.666 ms (0 allocations: 0 bytes)
```

devmotion:
I updated the PR, it handles (and deprecates) empty collections of probabilities and uses pairwise summation now for `crossentropy` and `kldivergence`.

src/scalarstats.jl
```diff
@@ -532,7 +532,12 @@ Compute the entropy of a collection of probabilities `p`,
 optionally specifying a real number `b` such that the entropy is scaled by `1/log(b)`.
 Elements with probability 0 or 1 add 0 to the entropy.
 """
-entropy(p) = -sum(pᵢ -> iszero(pᵢ) ? zero(pᵢ) : pᵢ * log(pᵢ), p)
+function entropy(p)
+    if isempty(p)
+        throw(ArgumentError("empty collections of probabilities are not supported"))
```
nalimilan, line 537: Maybe more explicit?
```suggestion
        throw(ArgumentError("empty collections are not supported as they do not " *
                            "represent a proper probability distribution""))
```

src/scalarstats.jl
```diff
@@ -584,21 +589,27 @@ end
...
+        # return zero for empty arrays
+        return xlogy(zero(eltype(p)), zero(eltype(q)))
     end
-    return -s
+
+    # use pairwise summation (https://github.com/JuliaLang/julia/pull/31020)
+    broadcasted = Broadcast.broadcasted(xlogy, vec(p), vec(q))
+    result = - sum(Broadcast.instantiate(broadcasted))
+
+    return result
```
nalimilan, line 609:
```suggestion
    return - sum(Broadcast.instantiate(broadcasted))
```

devmotion:
Any additional comments or suggestions?

PR #722, opened 2021-10-02
make counting more robust to input datatype

### LINKED ISSUE #721: _addcounts_radix_sort_loop! unsafe for OffsetVectors
This is because _addcounts_radix_sort_loop! uses absolute indexing starting at 1.
```julia
using StatsBase
using OffsetArrays

v = OffsetVector([5,6,7], -2)

display(countmap(v; alg=:dict))
#=
Dict{Int64, Int64} with 3 entries:
  5 => 1
  6 => 1
  7 => 1
=#
display(countmap(v; alg=:radixsort))
#=
Dict{Int64, Int64} with 2 entries:
  7                   => 1
  7205759405604732928 => 1
=#
```

### LINKED ISSUE #796: Stop using SortingAlgorithms.RadixSort in count.jl
SortingAlgorithms.RadixSort is [slow for small inputs](https://github.com/JuliaCollections/SortingAlgorithms.jl/issues/50) and broken for [very large inputs](https://github.com/JuliaCollections/SortingAlgorithms.jl/issues/52) and for [some offset arrays](https://github.com/JuliaCollections/SortingAlgorithms.jl/issues/56). As of Julia 1.9, Base's sorting policy includes RadixSort automatically for appropriate inputs so it is unlikely that these SortingAlgorithms issues will be fixed with a high priority. We should probably switch to base's default sorting for at least v1.9+ and possibly for all Julia versions, depending on how we value support for OffsetArrays vs performance.

```julia
using StatsBase, BenchmarkTools

short = zeros(Int, 5);
short[1] = typemax(Int)
@btime countmap(short)
#=
  27.573 μs (20 allocations: 194.27 KiB) (that's a lot of allocations!)
Dict{Int64, Int64} with 2 entries:
  0                   => 4
  9223372036854775807 => 1
=#

long = zeros(UInt32, 2^32+1);
long[1] = typemax(UInt32)
countmap(long)
#=
ERROR: InexactError: trunc(UInt32, 4294967296)
Stacktrace:
  [1] throw_inexacterror(f::Symbol, #unused#::Type{UInt32}, val::Int64)
    @ Core ./boot.jl:612
  [2] checked_trunc_uint
    @ ./boot.jl:642 [inlined]
  [3] toUInt32
    @ ./boot.jl:726 [inlined]
  [4] UInt32
    @ ./boot.jl:766 [inlined]
  [5] convert
    @ ./number.jl:7 [inlined]
  [6] setindex!
    @ ./array.jl:905 [inlined]
  [7] sort!(vs::Vector{UInt32}, lo::Int64, hi::Int64, ::SortingAlgorithms.RadixSortAlg, o::Base.Order.ForwardOrdering, ts::Vector{UInt32})
    @ SortingAlgorithms ~/.julia/packages/SortingAlgorithms/PEcBU/src/SortingAlgorithms.jl:84
  [8] sort!
    @ ~/.julia/packages/SortingAlgorithms/PEcBU/src/SortingAlgorithms.jl:66 [inlined]
  [9] sort!(v::Vector{UInt32}, alg::SortingAlgorithms.RadixSortAlg, order::Base.Order.ForwardOrdering)
    @ Base.Sort ./sort.jl:664
 [10] sort!(v::Vector{UInt32}; alg::SortingAlgorithms.RadixSortAlg, lt::Function, by::Function, rev::Nothing, order::Base.Order.ForwardOrdering)
    @ Base.Sort ./sort.jl:725
 [11] #sort#9
    @ ./sort.jl:772 [inlined]
 [12] addcounts_radixsort!
    @ ~/.julia/packages/StatsBase/pJqvO/src/counts.jl:363 [inlined]
 [13] _addcounts!(::Type{UInt32}, cm::Dict{UInt32, Int64}, x::Vector{UInt32}; alg::Symbol)
    @ StatsBase ~/.julia/packages/StatsBase/pJqvO/src/counts.jl:264
 [14] #addcounts!#92
    @ ~/.julia/packages/StatsBase/pJqvO/src/counts.jl:258 [inlined]
 [15] #countmap#97
    @ ~/.julia/packages/StatsBase/pJqvO/src/counts.jl:411 [inlined]
 [16] countmap(x::Vector{UInt32})
    @ StatsBase ~/.julia/packages/StatsBase/pJqvO/src/counts.jl:411
 [17] top-level scope
    @ REPL[4]:1
=#

### PR DESCRIPTION
**Pull request by LilithHafner**
Fixes #721
Fixes #796

### CODE REVIEW

File: src/counts.jl [hunk 1]
```diff
@@ -337,23 +337,22 @@ radixsort_safe(::Type{T}) where T = T<:BaseRadixSortSafeTypes
...
+            cm[last_sx] = get(cm, last_sx, 0) + i - start_i
             last_sx = sxi
-            tmpcount = get(cm, last_sx, 0) + 1
+            start_i = i
         end
     end
 
-    cm[sx[end]] = tmpcount
+    last_sx = last(sx)
+    cm[last_sx] = get(cm, last_sx, 0) + length(sx) + 1 - start_i
```
nalimilan, line 355: Isn't this equivalent?
```suggestion
    cm[sxi] = get(cm, sxi, 0) + length(sx) + 1 - start_i
```

nalimilan (commented):
Thanks! Could you add tests using OffsetArrays to ensure it works and we don't introduce regressions in the future? While you're at it, it would be nice to test the dict-based method too if it works.

src/counts.jl
LilithHafner: I believe `sxi` is not in scope. It gives me an "UndefVarError: sxi not defined".

src/counts.jl
```diff
@@ -19,21 +19,21 @@ end
...
+    m0 = first(levels)
+    m1 = last(levels)
     b = m0 - 1
 
-    @inbounds for i in 1 : length(x)
-        xi = x[i]
+    @inbounds for xi in x
         if m0 <= xi <= m1
             r[xi - b] += 1
         end
```
kalmarek, line 39: I'm not sure this actually changes the correctness that much. In case of incorrect indices in `x` this will still segfault or in the wrong case silently corrupt data. Shouldn't `@inbounds` be called only if `x`s are sanitized?

I'd suggest something along those lines:

```julia
Base.@propagate_inbounds function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
    @boundscheck checkbounds(r, axes(levels)...)
    m0, m1 = first(levels), last(levels)
    b = m0 - 1 # are we sure this is a valid index?
    for xi in x
        if m0 <= xi <= m1
            r[xi - b] += 1
        end
    end
    return r
end
```
where the caller of `addcounts!` takes responsibility for `@inbounds`-ing the call.

LilithHafner: > incorrect indices in `x`

I'm not sure what you mean here. Could you give an example input that would cause a segfault?

src/counts.jl
```diff
@@ -110,26 +121,25 @@ proportions(x::IntegerArray, wv::AbstractWeights) = proportions(x, span(x), wv)
...
-    size(r) == (kx, ky) || throw(DimensionMismatch())
+    axes(x) == axes(y) ||
+        throw(DimensionMismatch("x and y must have the same axes, but got $(axes(x)) and $(axes(y))"))
 
-    mx0 = xlevels[1]
-    mx1 = xlevels[end]
-    my0 = ylevels[1]
-    my1 = ylevels[end]
+    axes(r) == (axes(xlevels, 1), axes(ylevels, 1)) ||
+        throw(DimensionMismatch("axes(r) must correspond to the axes of levels, but got $(axes(r)) ≠ $(axes(xlevels, 1)), $(axes(ylevels, 1))"))
```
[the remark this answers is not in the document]
LilithHafner:
```suggestion
    checkbounds(r, axes(xlevels, 1), axes(xlevels, 1)) ||
        throw(DimensionMismatch("axes(r) must correspond to the axes of levels, but got $(axes(r)) ≠ $(axes(xlevels, 1)), $(axes(ylevels, 1))"))
```
Yep, thanks!

src/counts.jl
```diff
@@ -110,26 +121,25 @@ proportions(x::IntegerArray, wv::AbstractWeights) = proportions(x, span(x), wv)
...
-    length(y) == n || throw(DimensionMismatch())
+    # add counts of pairs from zip(x,y) to r
 
     xlevels, ylevels = levels
 
-    kx = length(xlevels)
-    ky = length(ylevels)
-    size(r) == (kx, ky) || throw(DimensionMismatch())
+    axes(x) == axes(y) ||
+        throw(DimensionMismatch("x and y must have the same axes, but got $(axes(x)) and $(axes(y))"))
```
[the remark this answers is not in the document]
LilithHafner: Agreed.
```suggestion
```

src/counts.jl
```diff
@@ -110,26 +121,25 @@ proportions(x::IntegerArray, wv::AbstractWeights) = proportions(x, span(x), wv)
...
+    mx0 = first(xlevels)
+    mx1 = last(xlevels)
+    my0 = first(ylevels)
+    my1 = last(ylevels)
 
     bx = mx0 - 1
     by = my0 - 1
 
-    for i = 1:n
+    for i in eachindex(x, y)
```
LilithHafner, line 142: This is supposed to work any time `x` and `y` have the same length. Failing on OffsetVectors with different offsets is acceptable. Failing on row matrix + column matrix is not.
```suggestion
    for i in eachindex(vec(x), vec(y))
```

src/counts.jl
```diff
@@ -141,28 +151,32 @@ end
...
+
+    xv, yv = vec(x), vec(y) # discard shape because weights() discards shape
 
     xlevels, ylevels = levels
 
-    kx = length(xlevels)
-    ky = length(ylevels)
-    size(r) == (kx, ky) || throw(DimensionMismatch())
+    axes(r) == (axes(xlevels, 1), axes(ylevels, 1)) || throw(DimensionMismatch(
+        "axes(r) must correspond to the axes of levels, got $(axes(r)) ≠ $(axes(xlevels, 1)), $(axes(ylevels, 1))"))
```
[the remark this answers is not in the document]
LilithHafner:
```suggestion
    checkbounds(r, axes(xlevels, 1), axes(ylevels, 1))
```

LilithHafner (commented):
Sorry I abandoned this PR for so long. It was getting to be too much for me to keep track of and stay interested in. Hopefully we should be able to finish it up now, though.

src/counts.jl
LilithHafner: `b` is not an index, it's something we subtract from `xi` to get an index. That said, it's a clever observation that `b` is possibly invalid. If `levels` was, for example, a `UnitRange` of `SaferIntegers.UInt`, which throw errors instead of overflowing, then this may be an error. However, we know that `eltype` of levels must be `Int`, and so we can rely on the arithmetic trick the original authors used here to simplify the inner loop.

We could rely on the compiler to factor that subtraction out and manually inline `b`, but I don't see much advantage to including that in this PR.

LilithHafner: > Shouldn't `@inbounds` be called only if `x`s are sanitized?

Sanitization happens on line 37 `if m0 <= xi <= m1`

src/weights.jl
```diff
@@ -21,6 +21,7 @@ length(wv::AbstractWeights) = length(wv.values)
 sum(wv::AbstractWeights) = wv.sum
 isempty(wv::AbstractWeights) = isempty(wv.values)
 size(wv::AbstractWeights) = size(wv.values)
+Base.axes(wv::AbstractWeights) = Base.axes(wv.values)
```
devmotion, line 24: Is this definition correct? We only support `Integer` indices in `getindex` and `Int` in `setindex!` below it seems, so non-standard axes that don't operate with `Integer`s will yield unsupported indices it seems.

src/counts.jl
kalmarek: I think that e.g. `1` may not be a valid index for `r`. Then for `xi=m0` we try to access `r` at `1`?
(and by a segfault I mean invalid read/write to memory)

LilithHafner: We know from line 37 that `first(levels) <= xi <= last(levels)`.
Subtracting `b = first(levels) - 1` from all sides, we get `1 <= xi-b <= last(levels)-first(levels)+1 = length(levels)`
`levels` is an `IntUnitRange` which is a subtype of `UnitRange`. `UnitRange`s use one-based indexing. Thus `xi-b` is in bounds in `levels`.
We know from line 30 that `axes(levels)` is in bounds in `r`, so anything that is in bounds in `levels` must also be in bounds in `r`.
Therefore `@inbounds r[xi - b]` is safe.

It would be possible to subtract off `firstindex(levels)` and extend support to `AbstractUnitRange{<:Integer}`, but this PR is concerned with correctness, not completeness.

kalmarek: ok, so this is correct since we check that indices of `levels` fit into `r`, and the former happens indexes through `OneTo(n)`. TIL ;)

src/counts.jl
```diff
@@ -19,21 +19,21 @@ end
     addcounts!(r, x, levels::UnitRange{<:Int}, [wv::AbstractWeights])
```
bkamins, line 19:
```suggestion
    addcounts!(r, x, levels::UnitRange{<:Integer}, [wv::AbstractWeights])
```

src/counts.jl
```diff
@@ -19,21 +19,21 @@ end
...
+If a weighting vector `wv` is specified, the sum of weights is used rather than the
+raw counts.
 """
 function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
-    # add counts of integers from x to r
+    # add counts of integers from x that fall within levels to r
 
-    k = length(levels)
-    length(r) == k || throw(DimensionMismatch())
+    checkbounds(r, axes(levels)...)
```
bkamins, line 30: I think
```suggestion
    (firstindex(r) == 1) && length(r) == k) || throw(DimensionMismatch())
```
better expresses the intention in the original code. Also maybe this condition should be noted in the documentation.

src/counts.jl
```diff
@@ -42,15 +42,21 @@ function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
...
+
+    checkbounds(r, axes(levels)...)
+
+    m0 = first(levels)
+    m1 = last(levels)
     b = m0 - 1
 
-    @inbounds for i in 1 : length(x)
-        xi = x[i]
+    @inbounds for i in eachindex(xv, wv)
```
bkamins, line 58: this will fail if `x` is `OffsetArray`:
```
julia> eachindex(vec(OffsetArray(1:4, 3)), Weights(1:4))
ERROR: DimensionMismatch
```

src/counts.jl
```diff
@@ -42,15 +42,21 @@ function addcounts!(r::AbstractArray, x::IntegerArray, levels::IntUnitRange)
...
+    # add wv weighted counts of integers from x that fall within levels to r
+
+    length(x) == length(wv) ||
+        throw(DimensionMismatch("x and wv must have the same length, got $(length(x)) and $(length(wv))"))
 
-    m0 = levels[1]
-    m1 = levels[end]
+    xv = vec(x) # discard shape because weights() discards shape
+
+    checkbounds(r, axes(levels)...)
```
bkamins, line 52: same comment as above

src/weights.jl
LilithHafner: The `AbstractArray` interface requires that `axes` returns a tuple of `AbstractUnitRange{<:Integer}`s, so we can assume that their `eltype` will be an `Integer`. From the `axes` docstring: "Each of the indices has to be an `AbstractUnitRange{<:Integer}`".

As for `Integers` other than `Int`, As long as we implement `getindex` and `setindex` for `Int` as required by the AbstractArray interface, Julia's Base libraries will convert other `Integer`s to `Int` as needed.

For example
```julia
using StatsBase
w = Weights([4,2,6])
w[0x2] = 1
display(w)
#=
3-element Weights{Int64, Int64, Vector{Int64}}:
 4
 1
 6
=#
```

The previous implicit definition (`axes(wv::AbstractWeights) = (Base.OneTo(length(wv)),)`) was _not_ correct

```julia
using StatsBase, OffsetArrays
w = Weights(OffsetVector([4,2,6], 2))
display(w)
#=
3-element Weights{Int64, Int64, OffsetVector{Int64, Vector{Int64}}}:
          3 # unsafe read
 4972085568 # unsafe read
          4
=#
```
On this PR
```julia
using StatsBase, OffsetArrays
w = Weights(OffsetVector([4,2,6], 2))
display(w)
#=
3-element Weights{Int64, Int64, OffsetVector{Int64, Vector{Int64}}} with indices 3:5:
 4
 2
 6
=#
```

src/counts.jl
LilithHafner: Correct. This probably belongs in a separate PR b.c. we don't change the type signature here, but I'll just commit it anyway because it's only 1 loc.

src/counts.jl
LilithHafner: Previously, addcounts of an OffsetArray and a non-offset Weights resulted in a segfault. In general, I am opposed to silently supporting operations with mismatched indices, and in this particular case, because it is not a regression to throw a dimension mismatch instead of segfaulting, I'd like to call that error out of scope.

src/counts.jl
LilithHafner: Yes, the previous version errored when `r` was larger than `levels` and if we keep that restriction, it should be noted in the documentation.

However, it feels more natural to me _not_ to keep that restriction, especially since it is not documented. So long as what we are adding to `r` fits into `r`, this method will work and other entries of `r` will not be affected. This looser restriction and guarantee not to mess with other entries of `r` is implicit and need not be mentioned in the docstring.

bkamins: The issue is broader. With your code if `r` is `OffsetArray` then indices from `1` to length(levels)` must be valid in it. But then you are adding values to `r` in indices starting from `1`, which in general does not have to be first element of `r`. The docstring does not specify where the elements are put into `r`. There can be two options:
* starting from index `1` or
* starting from beginning of `r`

That is why I propose to add `firstindex(r)==1` condition which resolves this issue as then both options are the same.
I agree that we could allow `r` to be longer than `length(levels)` (this is a change, but I think it is acceptable and makes sense).

src/counts.jl
bkamins: OK - I thought you wanted to fix OffsetArrays.jl issues in this PR, but agreed it can go to a separate PR since what you propose now is safe.

However note that in:
> `xv = vec(x)` # discard shape because weights() discards shape

you already decide on the interface you mention - choosing to support "position in sequence" interpretation (if `x` is higher dimensional you implicitly perform this matching)

LilithHafner: That choice was forced to avoid a regression on `addcounts(zeros(Int, 5), rand(1:5, 3, 3), Weights(rand(9)))`.

It is also an implicit conversion to linear indexing, which is somewhat different than an implicit alignment of mismatched indices.

> OK - I thought you wanted to fix OffsetArrays.jl issues in this PR

I don't see this as an issue:

```julia
julia> w = Weights([4,2,6]);

julia> ow = Weights(OffsetVector([4,2,6], 2));

julia> x = [3,1,2];

julia> ox = OffsetVector([3,1,2], 2);

julia> addcounts!(zeros(Int, 3), x, 1:3, w)
3-element Vector{Int64}:
 2
 6
 4

julia> addcounts!(zeros(Int, 3), ox, 1:3, ow)
3-element Vector{Int64}:
 2
 6
 4

julia> addcounts!(zeros(Int, 3), ox, 1:3, w)
ERROR: DimensionMismatch: all inputs to eachindex must have the same axes, got (OffsetArrays.IdOffsetRange(values=3:5, indices=3:5),) and (Base.OneTo(3),)
Stacktrace:
 [1] throw_eachindex_mismatch_indices(::IndexCartesian, ::Tuple{OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}}}, ::Vararg{Any})
   @ Base ./abstractarray.jl:292
 [2] eachindex
   @ ./multidimensional.jl:388 [inlined]
 [3] eachindex
   @ ./abstractarray.jl:334 [inlined]
 [4] addcounts!(r::Vector{Int64}, x::OffsetVector{Int64, Vector{Int64}}, levels::UnitRange{Int64}, wv::Weights{Int64, Int64, Vector{Int64}})
   @ StatsBase ~/.julia/dev/StatsBase/src/counts.jl:58
 [5] top-level scope
   @ REPL[24]:1

julia> addcounts!(zeros(Int, 3), x, 1:3, ow)
ERROR: DimensionMismatch: all inputs to eachindex must have the same axes, got (Base.OneTo(3),) and (OffsetArrays.IdOffsetRange(values=3:5, indices=3:5),)
Stacktrace:
 [1] throw_eachindex_mismatch_indices(::IndexCartesian, ::Tuple{Base.OneTo{Int64}}, ::Vararg{Any})
   @ Base ./abstractarray.jl:292
 [2] eachindex
   @ ./multidimensional.jl:388 [inlined]
 [3] eachindex
   @ ./abstractarray.jl:334 [inlined]
 [4] addcounts!(r::Vector{Int64}, x::Vector{Int64}, levels::UnitRange{Int64}, wv::Weights{Int64, Int64, OffsetVector{Int64, Vector{Int64}}})
   @ StatsBase ~/.julia/dev/StatsBase/src/counts.jl:58
 [5] top-level scope
   @ REPL[25]:1

julia> 

```

src/counts.jl
```diff
@@ -19,21 +19,21 @@ end
     addcounts!(r, x, levels::UnitRange{<:Int}, [wv::AbstractWeights])
 
 Add the number of occurrences in `x` of each value in `levels` to an existing
-array `r`. If a weighting vector `wv` is specified, the sum of weights is used
-rather than the raw counts.
+array `r`.
```
LilithHafner, line 22:
```suggestion
array `r`. For each `xi ∈ x`, if `xi == levels[j]`, then we increment `r[j]`.
```

bkamins:
@LilithHafner - thank you for comments in both points. I would leave the decision to maintainers which way to go (I understand your perspective, which is more flexible than my original thinking).

Side note:
```
ow = Weights(OffsetVector([4,2,6], 2));
```
also needs to be fixed. I have opened https://github.com/JuliaStats/StatsBase.jl/pull/791 to track this.

LilithHafner:
On this branch we have
```julia
julia> ow = Weights(OffsetVector([4,2,6], 2))
3-element Weights{Int64, Int64, OffsetVector{Int64, Vector{Int64}}} with indices 3:5:
 4
 2
 6
```

src/weights.jl
bkamins: Given the discussion in #791 I would propose to add tests of the consequences of this change (e.g. making sure that `getindex`, setindex!`, `firstindex`, `lastindex`, `isequal` etc. work correctly for non-1-based case)

bkamins:
If we want to support in StatsBase.jl non-1-based indexing in general then I think given @LilithHafner comment:

> In general, it is a nontrivial interface decision whether to silent support mismatched indices.

We should make this decision and clearly document it (as you see in my comments my original preference was to be more restrictive so that we would not have to make this decision, but I am OK to be more flexible and allow non-1-based vectors but then I think we should rethink the interfaces so that they correctly reflect this).

src/counts.jl
nalimilan: Option 1 sounds fine, the way `r` is used should have been explained better anyway in the current version (one could have imagined that the value of each level was used as to index into `r` rather than starting at 1). Allowing `r` to be longer than `length(levels)` sounds OK too as long as it's documented.

nalimilan:
I'd prefer throwing errors when passing arrays with mismatched indices, but there are some cases where we silently treated weight arrays as vectors whatever their dimensionality. So we have to preserve these, but in other cases we can be strict.

src/counts.jl
```diff
@@ -369,18 +380,23 @@ function addcounts_radixsort!(cm::Dict{T}, x::AbstractArray{T}) where T
...
 
 function addcounts!(cm::Dict{T}, x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real}
-    n = length(x)
-    length(wv) == n || throw(DimensionMismatch())
+    # add wv weighted counts of integers from x to cm
+
+    length(x) == length(wv) ||
+        throw(DimensionMismatch("x and wv must have the same length, got $(length(x)) and $(length(wv))"))
+
+    xv = vec(x) # discard shape because weights() discards shape
```
nalimilan, line 394: Thinking about this again, maybe we should deprecate this behavior by adding a separate method like this:
```julia
@deprecate addcounts!(cm::Dict{T}, x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcounts!(cm, vec(x), wv)
```
And restrict the signature of this method to `x::AbstractVector{T}`.

bkamins:
x-ref: https://github.com/JuliaStats/StatsBase.jl/pull/790#discussion_r875603821 (as I think we need to make a general design decision about how we want to handle non-1-based indexing that next is consistently implemented in the package)

LilithHafner:
I propose that any time two arrays (including array and weights) are supposed to line up, we call `eachindex(a, b)`. And we add `@depricated` methods to preserve backward compatibility.

src/counts.jl
kalmarek: > then indices from `1` to `length(levels)` must be valid in it

@bkamins this is actually already checked by `checkbounds(r, axes(levels)...)` above, since the latter is `OneTo(...)`. But given that I raised similar issue two days ago (in comments just below) there might be some merit in documenting it ;)

bkamins: Yes - it is chcecked. My point is that it should be documented.

LilithHafner: @kalmarek 
> But given that I raised similar issue two days ago (in comments just below) there might be some merit in documenting it ;)

Could you link that issue? I'm not seeing it and want to make sure the documentation I added addresses it.

src/counts.jl
getzze: To allow different index styles (but same length) we could do:

```for (xi, wi) in zip(xv, wv)```

Comment (nalimilan) [reply src/counts.jl:58]: `zip` is too tolerant, it even allows inputs of different sizes and silently ignores extra elements in the other argument. As mentioned in the main discussion, we should probably be strict about having matching indices or things could be really confusing.

File: src/counts.jl (hunk 9 above)
Comment (LilithHafner) [reply src/counts.jl:30]: For completeness, here is the link: https://github.com/JuliaStats/StatsBase.jl/pull/722#discussion_r874519428

File: src/weights.jl (hunk 7 above)
Comment (bkamins) [reply src/weights.jl:24]: if StatsBase.jl maintainers are OK to merge this PR without these tests then I can add them later.

**Comment by nalimilan** [comment]
Is there anything left to decide here?

**Comment by LilithHafner** [comment]
Bump, I'd love to get this merged once folks are happy with it. It hits a lot of correctness issues and is already 8 months old.

**Comment by bkamins** [comment]
I am not a StatsBase.jl maintainer, but most likely maintainers are waiting for CI to pass before doing a final review.
What are the reasons of bouds errors for OffsetArrays.jl on nightly?

**Comment by LilithHafner** [comment]
Oh, I wasn't paying attention to those failed tests because they also fail on StatsBase master. I'll look into it.

**Comment by bkamins** [comment]
Ah - if this is unrelated then the failing tests should of course be fixed in a separate PR.

**Comment by LilithHafner** [comment]
There are unrelated failures and also failures due to tests added in this PR. This PR tests counting with OffsetArrays  (though only on v1.9+) and counting uses SortingAlgorihtms.RadixSort which [fails on some OffsetArrays](https://github.com/JuliaCollections/SortingAlgorithms.jl/issues/56). One solution would be to use `Base.Sort.DEFAULT_UNSTABLE` which should be reliable* and as of 1.9 is also as fast as SortingAlgorihtms.RadixSort and much faster in many common special cases (e.g. inputs shorter than 1000 elements).

We could switch from `SortingAlgorithms.RadixSort` to `Base.Sorrt.DEFAULT_UNSTABLE` for all Julia versions (+ correctness; + performance for collections < 500-1000 elements; - performance for large collections on julia < v1.9) or just for v1.9+. 

*Coincidentally, Base.Sort.sort! is also [broken](https://github.com/JuliaLang/julia/issues/45568) for some OffsetArrays right now.

**Comment by nalimilan** [comment]
Ah yes it would be good to move to the new radix sort implementation in Base on Julia >= 1.9. But I'd keep SortingAlgorithms on older releases to avoid regressions (1.9 isn't even close to be released). We can just call `Base.require_on_based_indexing` on Julia < 1.9.

**Comment by LilithHafner** [comment]
I did that and also had to limit OffsetArrays tests to 1.9+ rather than 1.6+ because they now give an "addcounts_radixsort! requires either one based indexing or Julia 1.9. Use `alg = :dict` as an alternative." error. SortingAlgorithms.RadixSort only sometimes segfaults and the StatsBase teest cases don't hit those cases which is why they previously passed on 1.6.

**Comment by LilithHafner** [comment]
CI Tests now fail in the same way in this PR that they do in master. (something about random numbers)

File: src/counts.jl [hunk 14]
```diff
@@ -369,9 +369,18 @@ function _addcounts_radix_sort_loop!(cm::Dict{T}, sx::AbstractVector{T}) where T
     return cm
 end
 
+function _alg(x::AbstractArray)
+    @static if VERSION >= v"1.9.0-DEV"
+        Base.DEFAULT_UNSTABLE
+    else
+        firstindex(x) == 1 || error("addcounts_radixsort! requires either one based indexing or Julia 1.9. Use `alg = :dict` as an alternative.")
```
nalimilan, line 376:
```suggestion
        firstindex(x) == 1 ||
            throw(ArgumentError("alg=:radixsort requires either one based indexing or Julia >= 1.9. "
                                "Use `alg = :dict` as an alternative."))
```

src/counts.jl
```diff
@@ -413,10 +442,27 @@ countmap(x::AbstractArray{T}, wv::AbstractVector{W}) where {T,W<:Real} = addcoun
...
+                     algorithm to sort the input vector which will generally lead to
+                     shorter running time. However the radix sort algorithm creates a
+                     copy of the input vector and hence uses more RAM. Choose `:dict`
+                     if the amount of available RAM is a limitation.
+
+- `:dict`:           use `Dict`-based method which is generally slower but uses less
+                     RAM and is safe for any data type.
 """
-proportionmap(x::AbstractArray) = _normalize_countmap(countmap(x), length(x))
+proportionmap(x::AbstractArray; alg = :auto) = _normalize_countmap(countmap(x; alg = alg), length(x))
```
nalimilan, line 467: It's nice to add this argument, but can you add a test that uses it then? Or just leave this change out for now.

nalimilan (commented):
OK, thanks. Sorry, I have one more request, as new code should really be tested.

src/counts.jl
LilithHafner: Good catch! I tested on offset arrays but didn't test specified alg.

src/counts.jl
```diff
@@ -335,32 +347,42 @@ const BaseRadixSortSafeTypes = Union{Int8, Int16, Int32, Int64, Int128,
...
     return cm
 end
 
+function _alg(x::AbstractArray)
+    @static if VERSION >= v"1.9.0-DEV"
+        Base.DEFAULT_UNSTABLE
+    else
+        firstindex(x) == 1 ||
+            throw(ArgumentError("alg=:radixsort requires either one based indexing or Julia >= 1.9. "
+                                "Use `alg = :dict` as an alternative."))
```
LilithHafner, line 378:
```suggestion
            throw(ArgumentError("alg=:radixsort requires either one based indexing or Julia >= 1.9. "
                              * "Use `alg = :dict` as an alternative."))
```

src/counts.jl
```diff
@@ -335,32 +347,42 @@ const BaseRadixSortSafeTypes = Union{Int8, Int16, Int32, Int64, Int128,
...
 
     return cm
 end
 
+function _alg(x::AbstractArray)
+    @static if VERSION >= v"1.9.0-DEV"
+        Base.DEFAULT_UNSTABLE
+    else
+        firstindex(x) == 1 ||
+            throw(ArgumentError("alg=:radixsort requires either one based indexing or Julia >= 1.9. "
```
LilithHafner, line 377:
```suggestion
            throw(ArgumentError("alg = :radixsort requires either one based indexing or Julia >= 1.9. "
```

src/counts.jl
LilithHafner: I just left the change out

nalimilan:
Thanks!

LilithHafner:
Wow! It actually happened! At 8 months and 8 participants, this is certainly the largest PR I've been a substantial part of. It builds my confidence in these systems of voluntary open-source code review to see that even if something takes this long, it still merges eventually if it's technically sound and provides a contribution that justifies the effort.

Thank you to everyone who helped see this through!

nalimilan:
Glad you enjoyed it! :-p

LilithHafner:
Now I know what you meant when you said

> Thanks. Unfortunately I'm afraid the rabbit hole is even deeper. :-)

Issue #561, opened 2020-02-27
Views of weights do not work well


oxinabox:
Views of weights are not `<:AbstractWeight`
They are only `<:AbstractVector`

This messes with some dispatches.

For example:
```
julia> w = eweights(20,0.3);

julia> mean(1:10, w[1:10])
7.957353087579495

julia> mean(1:10, @view w[1:10])
ERROR: MethodError: objects of type UnitRange{Int64} are not callable

julia> mean_and_std(1:10, @view w[1:10])
ERROR: MethodError: no method matching mean_and_std(::UnitRange{Int64}, ::SubArray{Float64,1,Weights{Float64,Float64,Array{Float64,1}},Tuple{UnitRange{Int64}},false})
```

Work-around is to use:
`AnalyticWeights(@view w[1:10])`

A possible solution is to overload the functions the create views and make then return subtypes of `AbstractWeight`

nalimilan:
For views to be `AbstractWeights`, we will need not only to override `view`, but also to allow each `AbstractWeights` type to wrap any kind of array (including `SubArray`).

ParadaCarleton:
Closed as duplicate.

PR #723, opened 2021-10-05
Make `view(::AbstractWeights, ...)` return an `AbstractWeights`

### LINKED ISSUE #719: Views of weight vectors should be weight vectors of views
quoting @nalimilan 😄

nalimilan:
This is necessary to preserve the information regarding the type of weights.

Fixes #719 and #561.

Cc: @bkamins, @oxinabox

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -37,6 +37,12 @@ end
 
 Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values), sum(wv))
 
+@propagate_inbounds function Base.view(wv::W, inds...) where {W <: AbstractWeights}
+    @boundscheck checkbounds(wv, inds...)
+    @inbounds v = view(wv.values, inds...)
+    W(v, sum(v))
+end
+
 @propagate_inbounds function Base.setindex!(wv::AbstractWeights, v::Real, i::Int)
     s = v - wv[i]
     wv.values[i] = v
--- a/test/weights.jl
+++ b/test/weights.jl
@@ -37,7 +37,7 @@ weight_funcs = (weights, aweights, fweights, pweights)
     @test sum(sa, wv) === 7.0
 end
 
-@testset "$f, setindex!" for f in weight_funcs
+@testset "$f, getindex, view, setindex!" for f in weight_funcs
     w = [1., 2., 3.]
     wv = f(w)
 
@@ -45,6 +45,17 @@ end
     @test wv[1] === 1.
     @test sum(wv) === 6.
     @test wv == w
+
+    @test wv[[1, 3]] == w[[1, 3]]
+    @test typeof(wv[[1, 3]]) === typeof(wv)
+    @test sum( wv[[1, 3]]) === sum(w[[1, 3]])
+
+    # Check view
+    @test view(wv, [1, 3]) == view(wv, [true, false, true]) == w[[1, 3]]
+    @test typeof(view(wv, [1, 3])) === typeof(view(wv, [true, false, true])) === typeof(wv)
+    @test sum(view(wv, [1, 3])) === sum(view(wv, [true, false, true])) === sum(w[[1, 3]])
+    @test_throws BoundsError view(wv, [1, 5])
+    @test_throws BoundsError view(wv, [true, false, true, true])
 
     # Test setindex! success
     @test (wv[1] = 4) === 4             # setindex! returns original val


oxinabox (approved):
LGTM and makes sense.

I totally forgot that issue.

I assume the math for this kind of slicing already have been checked when `getindex` for this kind of slicing was permitted?

Looking at https://github.com/JuliaStats/StatsBase.jl/blob/2faa6e80b7966b915086d8cd5a4a4d89a2126db5/src/moments.jl
I am not sure that this kind of slicing should be allowed for `ProbabilityWeight`s ?
But I am not at all an expert on this math; where as I assume you are.
So if you've thought it through then this should be all good.

nalimilan:
Yes it's fine for frequency weights and analytic weights. For probability weights, it's a complex matter, but it's better to allow people to use weights for a subsample than to completely disallow it (otherwise you wouldn't be able to use them in the presence of missing values at all).

bkamins (approved):

bkamins:
What functions we provide are `ProbabilityWeights` aware?

If I understand things correctly `ProbabilityWeights` vector just denotes what were the inverse probabilities of sampling given values. If you subset such a vector these inverse probabilities do not change (the population from which you have drawn the sample remains the same - you just have a smaller sample from it). @nalimilan - is there anything more to it (I have not taken part in the discussions when `ProbabilityWeights` were designed)

nalimilan:
> What functions we provide are `ProbabilityWeights` aware?

Essentially `var` with `corrected=true`. Ideally GLM.jl will also support them in addition to frequency weights (that's easy, just divide them by their mean).

> If I understand things correctly `ProbabilityWeights` vector just denotes what were the inverse probabilities of sampling given values. If you subset such a vector these inverse probabilities do not change (the population from which you have drawn the sample remains the same - you just have a smaller sample from it). @nalimilan - is there anything more to it (I have not taken part in the discussions when `ProbabilityWeights` were designed)

Well that's right if you take a random subsample, but if you select a nonrandom subsample (which is the most common case) weights will not be exactly correct IIUC. But in practice it's better to use somewhat incorrect weights than no weights at all (often the difference won't be that large). Also, if you take a subsample based on a variable which was used as a strata to construct the weights then the result will be OK. The correct way to handle this in general is to use software designed to take into account complex survey designs (like design in R or svy in Stata). See for example https://www.restore.ac.uk/PEAS/subgroups.php.

bkamins:
We also need to take care of the fact that `AbstractWeigths` must be `AbstractVector`, so we need to make sure we handle the following correctly:
```
julia> w = Weights([1,2,3])
3-element Weights{Int64, Int64, Vector{Int64}}:
 1
 2
 3

julia> vw = @view w[1]
0-dimensional view(::Weights{Int64, Int64, Vector{Int64}}, 1) with eltype Int64:
1
```
as this expression drops a dimension, so `vw` cannot be `AbstractWeights` (maybe such views should be disallowed?).

bkamins:
Relatedly `similar` on `Weights` produces a `Vector` - not sure if this is intended.

bkamins:
Maybe `similar` is OK, given that `parent` returns the underlying vector, so `Weights` is a kind of view already?

nalimilan:
I've pushed a commit to make `view` return an `AbstractWeights` view into the `AbstractWeights`. In addition to that change, to ensure that mutating the parent doesn't make the view's precomputed sum inconsistent with its contents, I added a `sum` method which always recomputes the sum from the actual contents. In theory we could avoid this by storing a state in the parent indicating that it was mutated since the last time `sum` was called on a view. Not sure whether it's worth it.

I also added another commit making `copy` and `wv[:]` return an `AbstractWeights` object.

src/weights.jl
```diff
@@ -337,8 +358,12 @@ end
 
 for w in (AnalyticWeights, FrequencyWeights, ProbabilityWeights, Weights)
     @eval begin
-        Base.isequal(x::$w, y::$w) = isequal(x.sum, y.sum) && isequal(x.values, y.values)
-        Base.:(==)(x::$w, y::$w)   = (x.sum == y.sum) && (x.values == y.values)
+        Base.isequal(x::$w, y::$w) =
+            (x.values isa SubArray || y.values isa SubArray || isequal(x.values, y.values)) &&
+                isequal(x.values, y.values)
+        Base.:(==)(x::$w, y::$w)   =
+            (x.values isa SubArray || y.values isa SubArray || (x.sum == y.sum)) &&
```
nalimilan, line 365:
```suggestion
            (x.values isa SubArray || y.values isa SubArray || x.sum == y.sum) &&
```

src/weights.jl
```diff
@@ -14,6 +14,7 @@ macro weights(name)
             sum::S
         end
         $(esc(name))(values::AbstractVector{<:Real}) = $(esc(name))(values, sum(values))
+        $(esc(:weightstype))(::Type{<:$(esc(name))}) = $(esc(name))
```
nalimilan, line 17: There's probably a simpler way of doing this but I couldn't figure it out...

bkamins: out of curiosity - why do we need this macro? I usually try to avoid such things as they make debugging later harder.

src/weights.jl
```diff
@@ -35,7 +36,33 @@ end
...
 
-Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values), sum(wv))
+Base.getindex(wv::AbstractWeights, ::Colon) = copy(wv)
+
+Base.copy(wv::W) where {W <: AbstractWeights} =
+    weightstype(W)(copy(wv.values), sum(wv))
+
+@propagate_inbounds function Base.view(wv::W, inds...) where {S <: Real, W <: AbstractWeights{S}}
+    @boundscheck checkbounds(wv, inds...)
+    @inbounds v = invoke(view, Tuple{AbstractArray, Vararg{Any}}, wv, inds...)
```
bkamins, line 46: why it is better to make such a view rather than just `v = view(wv.values, inds...)`? I understand the reason is that later you check if in the new object `values` are view of `AbstractWeights` to know you must recompute `sum`?

src/weights.jl
```diff
@@ -35,7 +36,33 @@ end
...
+    # Sum is not actually used but compute the right type for clarity
+    weightstype(W)(v, zero(S))
+end
+
+# This method is implemented for backward compatibility
```
bkamins, line 51: I understand the reason is that this 0-dimensional view must not be `AbstractWeights` as weights must be exactly 1-dimensional.

src/weights.jl
```diff
@@ -35,7 +36,33 @@ end
...
+@propagate_inbounds function Base.view(wv::W, inds::Union{Integer, CartesianIndex}...) where
+    {S <: Real, W <: AbstractWeights{S}}
+    @boundscheck checkbounds(wv, inds...)
+    @inbounds invoke(view, Tuple{AbstractArray, Vararg{Any}}, wv, inds...)
+end
+
+# Always recompute the sum for views of AbstractWeights, as we cannot know whether
+# the parent array has been mutated
+Base.sum(wv::AbstractWeights{S, T, <:SubArray{T, <:Any, <:AbstractWeights}}) where
```
bkamins, line 60: Are we sure that in the code `:sum` field is never accessed directly, but always via the `sum` function?

src/weights.jl
```diff
@@ -35,7 +36,33 @@ end
...
+    {S<:Real, T<:Real} = sum(wv.values)
+
+Base.copy(wv::W) where
+    {S<:Real, T<:Real, W<:AbstractWeights{S, T, <:SubArray{T, <:Any, <:AbstractWeights}}} =
+    weightstype(W)(copy(view(parent(wv.values).values, parentindices(wv.values)...)), wv.sum)
```
bkamins, line 65: for example here the call seems to be incorrect. You either should call `sum(wv)` or not pass the sum at all and force its computation in the constructor. Am I missing something here?

src/weights.jl
```diff
@@ -337,8 +364,12 @@ end
 
 for w in (AnalyticWeights, FrequencyWeights, ProbabilityWeights, Weights)
     @eval begin
-        Base.isequal(x::$w, y::$w) = isequal(x.sum, y.sum) && isequal(x.values, y.values)
-        Base.:(==)(x::$w, y::$w)   = (x.sum == y.sum) && (x.values == y.values)
+        Base.isequal(x::$w, y::$w) =
+            (x.values isa SubArray || y.values isa SubArray || isequal(x.values, y.values)) &&
```
bkamins, line 368: shouldn't `isequal(x.values, y.values)` be `sum(x) == sum(y)` here?

src/weights.jl
```diff
@@ -337,8 +364,12 @@ end
...
+                isequal(x.values, y.values)
+        Base.:(==)(x::$w, y::$w)   =
+            (x.values isa SubArray || y.values isa SubArray || x.sum == y.sum) &&
```
bkamins, line 371: this is the kind of use of `x.sum` that I understand we should avoid (in this case it is OK, but I would assume that we should always use `sum(x)` etc. for safety since sometimes we do not have computed the `:sum` field correctly). Right?

bkamins (changes requested):
I have left some comments.

src/weights.jl
nalimilan: The macro just avoids some duplication as we define 4 almost identical types. We could equally well use `@eval` in a loop.

src/weights.jl
nalimilan: This is to avoid the problem we discussed on Slack: if we merely make a view of `wv.values`, mutating the view will corrupt the parent as the sum will be out of sync with its contents.

(It's not needed to detect whether we must recompute the sum, as we could do that when `wv.values isa SubArray`, whatever the parent type.)

src/weights.jl
bkamins: I know. I just say we save some 10 lines of code at the expense of making it less clear for normal users. Of course it can stay from a technical standpoint.

src/weights.jl
nalimilan: Right. This is just to avoid breaking the cases you spotted above (even though in practice they are unlikely to be used).

src/weights.jl
bkamins: Ah - right!

src/weights.jl
nalimilan: No. :-D But at least it's private.

For additional safety, I considered storing `NaN`, `missing` or `nothing` in the `sum` field. But that would affect the `S` type parameter, which has no practical consequences but could be a bit weird for users (as `sum` would return a different type). Not sure.

At least by storing a zero sum, results should be obviously wrong if people access the field directly. For example, `mean` returns `Inf` and `var` returns `NaN` when the sum is zero. But I agree that's not 100% safe.

nalimilan: We could also override `getproperty` for views of weights if we really wanted to prevent any incorrect behaviors. That would require using `getfield` internally in some cases.

src/weights.jl
nalimilan: The idea is just to pass a `zero` of the right type (as it determines the return type). I could use `zero(typeof(wv.sum))` for clarity if you prefer, or add a comment.

src/weights.jl
nalimilan: Woops. Good catch. Though this is one of the few places where I think accessing the field directly is clearer, as we certainly don't want `sum(x)` to recompute the sum given that it's just an optimization to return `false` quickly if sums are different.
```suggestion
            (x.values isa SubArray || y.values isa SubArray || isequal(x.sum, y.sum)) &&
```

src/weights.jl
nalimilan: Same as above.

src/weights.jl
bkamins: But that is a point that I think it is incorrect, the reason is that when you write `copy(view(parent(wv.values).values, parentindices(wv.values)...))` it is not a view any more (`copy` is applied as the outermost function) so it must have a correct `sum` supplied. At least this is what I understand should happen.

src/weights.jl
bkamins: ok

src/weights.jl
bkamins: we can leave things as they are but I still think https://github.com/JuliaStats/StatsBase.jl/pull/723#discussion_r745115242 is a problem

src/weights.jl
nalimilan: Woops, right. So that also needs to be tested...

bkamins: Yes + the point is that this kind of things is tricky to detect when you are having a special treatment of `sum` field. Maybe we should also add a comment in the docs (or at least in the source code) that this is a private field?

nalimilan: Fixed and tests added.

src/weights.jl
nalimilan: I've changed more uses of `x.sum` to `sum(x)` as they were not justified.

src/weights.jl
nalimilan: I've added a commit which uses `missing` instead of zero for views. That way any invalid access to the `sum` field will be caught (including invalid equality tests). It turns out it doesn't make the code more complex.

nalimilan:
I've adapted testsets so that they loop over views of weights for all tests that cover weights. This makes the code more complex but that's probably worth it if we want to fully support views of weights.

nalimilan:
Ah, `UnitWeights` are not tested and don't work...

bkamins:
> Ah, `UnitWeights` are not tested and don't work...

Do you want me to review now, or should I wait until you work on this?

nalimilan:
As you prefer. The change to support `UnitWeights` will probably be distinct so you can start reviewing the current code if you want.

src/weights.jl
```diff
@@ -35,7 +40,33 @@ end
...
+Base.getindex(wv::AbstractWeights, ::Colon) = copy(wv)
+
+Base.copy(wv::W) where {W <: AbstractWeights} =
+    weightstype(W)(copy(wv.values), sum(wv))
+
+@propagate_inbounds function Base.view(wv::W, inds...) where
+    {S <: Real, W <: AbstractWeights{S}}
+    @boundscheck checkbounds(wv, inds...)
+    @inbounds v = invoke(view, Tuple{AbstractArray, Vararg{Any}}, wv, inds...)
+    weightstype(W){S, eltype(wv), typeof(v)}(v, missing)
```
bkamins, line 52:
```suggestion
    return weightstype(W){S, eltype(wv), typeof(v)}(v, missing)
```

src/weights.jl
```diff
@@ -35,7 +40,33 @@ end
...
+end
+
+# This method is implemented for backward compatibility
+@propagate_inbounds function Base.view(wv::W, inds::Union{Integer, CartesianIndex}...) where
+    {S <: Real, W <: AbstractWeights{S}}
+    @boundscheck checkbounds(wv, inds...)
+    @inbounds invoke(view, Tuple{AbstractArray, Vararg{Any}}, wv, inds...)
+end
+
+# Always recompute the sum for views of AbstractWeights, as we cannot know whether
```
bkamins, line 62: maybe move the definitions of `sum` and `copy` into one place to make sure the reader can see both definitions side by side?

docs/src/weights.md
```diff
@@ -8,6 +8,7 @@ In statistical applications, it is not uncommon to assign weights to samples. To
 !!! note
     - The weight vector is a light-weight wrapper of the input vector. The input vector is NOT copied during construction.
     - The weight vector maintains the sum of weights, which is computed upon construction. If the value of the sum is pre-computed, one can supply it as the second argument to the constructor and save the time of computing the sum again.
+    - Views of weight vectors are also `AbstractWeights`, but their sum has to be recomputed by each call to `sum` as weights in the parent vector may have been mutated.
```
bkamins, line 11: not only views of `AbstractWeights` but with the current implementation if weights is created from a view (even of normal array) we will also recompute - right?

bkamins:
OK. I have commented. I think it is a good idea to use a separate `missing` value when `sum` is missing.

bkamins:
Maybe the only comment is that it would be even safer and a bit faster (but it would complicate code so I am not sure it is worth it) to instead of having `sum::Union{S, Missing}` have `s::S2` where `S2` would be `S` for non-view and `Missing` for view.

nalimilan:
Actually, I wonder whether it wouldn't be better to keep returning `SubArray`s of `AbstractWeights` and change dispatch to allow `Union{AbstractWeights, SubArray{..., <: AbstractWeights}}` instead. This would avoid adding tricky code and AFAICT it would give the same result since `sum` has to recompute the sum on every call for views anyway. The only drawback would be that dispatching on `AbstractWeights` would not be enough (a bit like what we have with `CatArrOrSub` in CategoricalArrays).

bkamins:
As you prefer. Still we would need to make sure the view is one dimensional, as currently one can do:
```
julia> x = Weights(1:3)
3-element Weights{Int64, Int64, UnitRange{Int64}}:
 1
 2
 3

julia> view(x, 1)
0-dimensional view(::Weights{Int64, Int64, UnitRange{Int64}}, 1) with eltype Int64:
1

julia> view(x, 1:1, 1:1)
1×1 view(reshape(::Weights{Int64, Int64, UnitRange{Int64}}, 3, 1), 1:1, 1:1) with eltype Int64:
 1
```

nalimilan:
Yes, we would only change signatures to accept `Union{AbstractWeights, SubArray{<:Real, 1, AbstractWeights}}`.

bkamins:
Yes, then `Union{AbstractWeights, SubArray{<:Real, 1, AbstractWeights}}` should probably be defined with some intuitive name and exported as this should be the signature used by packages.

PR #727, opened 2021-10-23
Make sum and wsum doc consistent with code

knuesel:
First thought there was type piracy going on with

```
?sum

[...]

  ────────────────────────────────────────────────

  sum(v::AbstractArray, w::AbstractVector{<:Real}; [dims])

  Compute the weighted sum of an array v with weights w, optionally over the dimension dims.
```

but it's the documentation that was wrong :)


nalimilan:
Thanks!

PR #730, opened 2021-11-05
Fix uplo function

bkamins:
`copy` is incorrect here. `collect` must be used, as `copy` does not guarantee that the resulting object is mutable:
```
julia> winsor(1:10)
ERROR: setindex! not defined for UnitRange{Int64}
```

### DIFF
--- a/src/robust.jl
+++ b/src/robust.jl
@@ -22,9 +22,8 @@ function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
 
     # indices for lowest count values
     x2 = collect(x)
-    lo = partialsort!(x2, 1:count+1)[end]
-    # indices for largest count values
-    up = partialsort!(x2, n-count:n)[1]
+    lo = partialsort!(x2, count+1)
+    up = partialsort!(x2, n-count)
 
     up, lo
 end
--- a/src/robust.jl
+++ b/src/robust.jl
@@ -21,7 +21,7 @@ function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
     end
 
     # indices for lowest count values
-    x2 = collect(x)
+    x2 = Base.copymutable(x)
     lo = partialsort!(x2, count+1)
     up = partialsort!(x2, n-count)
 
--- a/src/robust.jl
+++ b/src/robust.jl
@@ -21,7 +21,7 @@ function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
     end
 
     # indices for lowest count values
-    x2 = copy(x)
+    x2 = collect(x)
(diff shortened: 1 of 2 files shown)


src/robust.jl
```diff
@@ -21,7 +21,7 @@ function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
     end
 
     # indices for lowest count values
-    x2 = copy(x)
+    x2 = collect(x)
```
ararslan, line 24: Perhaps we should use `copymutable` here?

src/robust.jl
```diff
@@ -21,7 +21,7 @@ function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
...
     lo = partialsort!(x2, 1:count+1)[end]
     # indices for largest count values
     up = partialsort!(x2, n-count:n)[1]
```
bkamins, line 27:
```suggestion
    lo = partialsort!(x2, count+1)
    up = partialsort!(x2, n-count)
```

src/robust.jl
```diff
@@ -21,10 +21,9 @@ function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
     end
 
     # indices for lowest count values
-    x2 = copy(x)
-    lo = partialsort!(x2, 1:count+1)[end]
-    # indices for largest count values
-    up = partialsort!(x2, n-count:n)[1]
+    x2 = collect(x)
```
bkamins, line 24:
```suggestion
    x2 = Base.copymutable(x)
```

src/robust.jl
bkamins: I have also changed the call to `partialsort!` as just requesting `count+1` and `n-count` positions should be enough and it is a bit faster.

ararslan (approved):

nalimilan:
Can you add a test?

src/robust.jl
```diff
@@ -21,10 +21,9 @@ function uplo(x::AbstractVector; prop::Real=0.0, count::Integer=0)
     end
 
     # indices for lowest count values
-    x2 = copy(x)
-    lo = partialsort!(x2, 1:count+1)[end]
-    # indices for largest count values
-    up = partialsort!(x2, n-count:n)[1]
+    x2 = Base.copymutable(x)
+    lo = partialsort!(x2, count+1)
```
nalimilan, line 25: And what if we just did this?
```suggestion
    lo = partialsort(x2, count+1)
```

nalimilan: Sorry that doesn't work.

bkamins:
test added

nalimilan (approved):

PR #732, opened 2021-11-11
Fix typo in doc strings for r2

awadell1:
Variant is `:MacFadden` in the docs, but `:McFadden` in the if else branch. Googling / Wikipedia confirms that "McFadden" is correct


nalimilan:
Thanks!

PR #733, opened 2021-11-12
add percentileofscore() function

AugustoCL:
```
percentileofscore(a::RealVector, score)
percentileofscore(a::RealVector, score::RealVector)
```
The function return the **percentile-position** (0-100) of `score` relative to the array `a`. 

Changes:
- Add `percentileofscore()` function in **src/scalarstats.jl**
- Add function to export in **src/StatsBase.jl**
- Add tests to function in **test/scalarstats.jl**
- Update Documentation in **docs/src/scalarstats.md**

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.33.12"
+version = "0.33.13"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/src/common.jl
+++ b/src/common.jl
@@ -23,9 +23,9 @@ const RealFP = Union{Float32, Float64}
 # A convenient typealias for deprecating default corrected Bool
 const DepBool = Union{Bool, Nothing}
 
-function depcheck(fname::Symbol, b::DepBool)
-    if b == nothing
-        msg = "$fname will default to corrected=true in the future. Use corrected=false for previous behaviour."
+function depcheck(fname::Symbol, varname::Symbol, b::DepBool)
+    if b === nothing
+        msg = "$fname will default to $varname=true in the future. Use $varname=false for previous behaviour."
         Base.depwarn(msg, fname)
         false
     else
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -92,11 +92,11 @@ _scattermatm(x::DenseMatrix, wv::AbstractWeights, mean, dims::Int) =
 ## weighted cov
 covm(x::DenseMatrix, mean, w::AbstractWeights, dims::Int=1;
      corrected::DepBool=nothing) =
-    rmul!(scattermat(x, w, mean=mean, dims=dims), varcorrection(w, depcheck(:covm, corrected)))
+    rmul!(scattermat(x, w, mean=mean, dims=dims), varcorrection(w, depcheck(:covm, :corrected, corrected)))
 
 
 cov(x::DenseMatrix, w::AbstractWeights, dims::Int=1; corrected::DepBool=nothing) =
-    covm(x, mean(x, w, dims=dims), w, dims; corrected=depcheck(:cov, corrected))
+    covm(x, mean(x, w, dims=dims), w, dims; corrected=depcheck(:cov, :corrected, corrected))
 
 function corm(x::DenseMatrix, mean, w::AbstractWeights, vardim::Int=1)
     c = covm(x, mean, w, vardim; corrected=false)
@@ -120,7 +120,7 @@ end
 function mean_and_cov(x::DenseMatrix, wv::AbstractWeights, dims::Int=1;
                       corrected::DepBool=nothing)
     m = mean(x, wv, dims=dims)
-    return m, cov(x, wv, dims; corrected=depcheck(:mean_and_cov, corrected))
+    return m, cov(x, wv, dims; corrected=depcheck(:mean_and_cov, :corrected, corrected))
 end
 
 """
--- a/src/moments.jl
+++ b/src/moments.jl
@@ -19,7 +19,7 @@ the population variance is computed by replacing
 * `Weights`: `ArgumentError` (bias correction not supported)
 """
 varm(v::RealArray, w::AbstractWeights, m::Real; corrected::DepBool=nothing) =
-    _moment2(v, w, m; corrected=depcheck(:varm, corrected))
+    _moment2(v, w, m; corrected=depcheck(:varm, :corrected, corrected))
 
 """
     var(x::AbstractArray, w::AbstractWeights, [dim]; mean=nothing, corrected=false)
@@ -40,7 +40,7 @@ replacing ``\\frac{1}{\\sum{w}}`` with a factor dependent on the type of weights
 """
 function var(v::RealArray, w::AbstractWeights; mean=nothing,
                   corrected::DepBool=nothing)
-    corrected = depcheck(:var, corrected)
+    corrected = depcheck(:var, :corrected, corrected)
 
     if mean == nothing
         varm(v, w, Statistics.mean(v, w); corrected=corrected)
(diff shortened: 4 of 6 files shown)


mschauer:
What is this? An empirical distribution function with a particular interpolation rule?

AugustoCL:
> 
> 
> What is this? An empirical distribution function with a particular interpolation rule?

Yeah, I can see this definition in this function.

The function answer the question "what percentile this `score` belong based in the array `a`?" (one score or an array of it). And the implementation have 4 methods to that. 

Basically, I was searching for an implementation of this task and I finded in the scipy, then I saw the source-code of it and I decided to make the same for Julia. When I saw the docs for StatsBase.jl, I thought that a good place for the function could be inside the scalarstats.jl functions. 

The Scipy implementation i mentioned: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.percentileofscore.html

I also mentioned the scipy code in the note in the docstring of the function. 

And I'm open to all suggestions and reviews about it. =)

AugustoCL:
Also, if you have any suggestions for a better name for the function, we can change it =]

AugustoCL:
Looking how the `quantile()` and `percentile()` function relate to each other, I try to recreat that by adding the **`quantileofscore()`** function.

Which is, in a quick explanation, `quantileofscore(...) = percentileofscore(...) / 100`.

mschauer:
@davibarreira

davibarreira:
> @davibarreira

Thanks for the ping. But this implementation is 1D, and it's already covered in Distributions.jl. The package I've implemented is for multivariate. I'm working on a fast algorithm for percentile and cdf stuff on multivariate. There are some current papers on the subject.

AugustoCL:
@davibarreira Could you please show me the implementation covered in the Distributions.jl? I couldn't find.

davibarreira:
@AugustoCL , I was not talking about your implementation per se, I was just pointing out that your implementation was for 1D distributions, which is `DiscreteNonParametric` in distributions.
Although, with `DiscreteNonParametric` you can get "similar" behavior to this `percentileofscore`, but it's more verbose and does not have the `kind` variation.

Here is how you can do something similar.
```julia
using Distributions
d = DiscreteNonParametric([1,2,3,4,5], fill(1/5,5))
cdf(d, 3)
```

AugustoCL:
Interesting.

Thinking about it, I got some points that I want to share.
- I think the cdf solution it's more generic and customizable because it contains cases with non-uniforms probability vectors (which make sense to use it in Distributions.jl only) and the `percentileofscore` it's equivalent to a cases with an uniform probability vector.
- Since the `StatsBase.jl` it's a "base" package, I understand that it's not recommended add features that call other pkgs (like the cdf example).

So, since the uniform probability vector case (`percentileofscore`) it's more commonly used by non-experts in probability (could also be considered more "base"), doesn't depend on any pkg and got different `methods` that could be easily interpreted by looking in the source-code and docstring, I think it's reasonable have the `percentileofscore` and `quantileofscore` in the `StatsBase.jl` instead in `Distributions.jl`. 

What do you guys think about it? Please, tell me if I'm forgetting something important.

davibarreira:
It makes sense to me. But I'm not very acquainted with StatsBase and it's guidelines on what should be in and what should not. So I guess a maintainer can give better feedback.

nalimilan:
Does this function exist in other stats libraries than SciPy (e.g. R packages)? In which contexts is it typically used?

We should probably find a better name, as "score" isn't a very good term for what can be any value, and "percentile" is just a particular case of "quantile". (Note that historically SciPy didn't have `quantile`, but only `scoreatpercentile` which is now `percentile`.)

AugustoCL:
> Does this function exist in other stats libraries than SciPy (e.g. R packages)? In which contexts is it typically used?
> 
> We should probably find a better name, as "score" isn't a very good term for what can be any value, and "percentile" is just a particular case of "quantile". (Note that historically SciPy didn't have `quantile`, but only `scoreatpercentile` which is now `percentile`.)

yes, I search and find equivalent functions in R, LibreOffice and Excel.

- LibreOffice: https://wiki.documentfoundation.org/Documentation/Calc_Functions/PERCENTRANK
- Excel: https://support.microsoft.com/en-us/office/percentrank-function-f1b5836c-9619-4847-9fc9-080ec9024442
- R dplyr (but doesnt accept a new value, return all vector with percents): https://stackoverflow.com/questions/60793656/make-an-equivalent-to-excels-percentrank-exc-function-in-r
- An old question for percentrank in R: https://stats.stackexchange.com/questions/11924/computing-percentile-rank-in-r
- Other old question where the answer was the cdf solution: https://stat.ethz.ch/pipermail/r-help/2011-May/279167.html

The context who brought me here, it was because my friend, who works making simulations in the mining sector, was looking for that function because he use in a simulation to check the discrepancy both in percentage terms and in order ("percentile") of some coefficients simulated. He asked for that function in Julia, I finded only in scipy, and then reproduce in Julia.

about the name, I'm open to all suggestion. Since in R, Libreoffice and Excel, they used percentrank, I think that would be fine use `percentilerank()` and `quantilerank()`. What you guys suggests?

I saw that you make the merge in commit `1678fd1`. Then I opened the commit, but didn't find the code I wrote in the commit and also this PR doesn't have any file in the "File Changed" section. 
I didn't understand what happened here, could you please explain to me?

nalimilan:
OK, thanks! `percentilerank` and `quantilerank` would be reasonable names I guess. Regarding the arguments it would take, AFAICT the variants that can be chosen via `method` correspond to [the various ranking functions](https://juliastats.org/StatsBase.jl/latest/ranking/) we support (`ordinalrank`, `denserank`...), right? It would be good to use the same terminology

> I saw that you make the merge in commit `1678fd1`. Then I opened the commit, but didn't find the code I wrote in the commit and also this PR doesn't have any file in the "File Changed" section.
> I didn't understand what happened here, could you please explain to me?

I've never seen this before. It looks like a bot called [pull](https://github.com/apps/pull) did that. Are you sure it's not enabled on your account? I'd try disabling it and then force-pushing commit 5c9bf12 to this branch again.

AugustoCL:
#> I've never seen this before. It looks like a bot called [pull](https://github.com/apps/pull) did that. Are you sure it's not enabled on your account? I'd try disabling it and then force-pushing commit [5c9bf12](https://github.com/JuliaStats/StatsBase.jl/commit/5c9bf120499b99afa61dcafb0aac3ee0f208a286) to this branch again.

That it was funny haha. Sorry for my mess. Now, I disable the pull bot and I think we could continue the PR analysis. 
@nalimilan, If you think it would be better to create another PR mentioning this one, I can do that. 

> Regarding the arguments it would take, AFAICT the variants that can be chosen via method correspond to the various ranking functions we support (ordinalrank, denserank...), right? It would be good to use the same terminology

I made a research and thought about  #it, but couldn't find a direct relationship of them. To me it seems that one is different methods for a direct ranking of an array and the other it's some methods to calculate the "percentile rank" measure (which is also mentioned in ranking wikipedia page as a different ranking, like "Statistics Ranking").

Also, in the wikipedia of the "Percentile Rank" [here](https://en.wikipedia.org/wiki/Percentile_rank), there is a small discussion in the text that mentioned the variations of the method (variantions that could be related to the different methods of the `percentileofscore` function). See what I mentioned in the image below

percentilerank

Also in that picture, there is a formula for `percentilerank` that is resulted from the generic formula presented in the section 3.2 of the book [Fundamental Statistics for the Behavioral Sciences](http://www.bryanburnham.net/wp-content/uploads/2014/07/Fundamental-Statistics-for-the-Behavioral-Sciences-v2.0.pdf#page=57). 

When I compare the 4 methods presented in the function of this PR, I only didn't fully understand the `:rank` method (because of the if-else that sum 0 or 1. So, I think that would be better to replace that method with the formula presented in the book above, because we have a reference for that. **What do you think?**

nalimilan:
> I made a research and thought about #it, but couldn't find a direct relationship of them. To me it seems that one is different methods for a direct ranking of an array and the other it's some methods to calculate the "percentile rank" measure (which is also mentioned in ranking wikipedia page as a different ranking, like "Statistics Ranking").

Actually I was confused, I think `method` rather corresponds to the `alpha` and `beta` argument to `quantile`. For example, in Excel `PERCENTILE.INC` corresponds to `PERCENTRANK.INC` and `PERCENTILE.EXC` corresponds to `PERCENTRANK.EXC`. Could you check whether we could replace `method` with `alpha` and `beta` instead, so that we have a consistent API? If necessary, for simplicity we could support only some combinations of these two arguments, which would match the current possible choices for `method` (and therefore keep the same `if` structure). See `?quantile` for details.

> When I compare the 4 methods presented in the function of this PR, I only didn't fully understand the `:rank` method (because of the if-else that sum 0 or 1. So, I think that would be better to replace that method with the formula presented in the book above, because we have a reference for that. **What do you think?**

Makes sense. Using Wikipedia formulas is often a good idea as they are the most commonly used ones.

PR #735, opened 2021-11-21
Reexport functions from Statistics

nalimilan:
It does not make any sense to avoid exporting functions from Statistics to which we add methods. These are documented in the manual but currently not available without typing `using Statistics`.
The only method from Statistics which we do not overload is `middle`, which this commit does not reexport for now.
Also don't reexport `stdm` and `varm` as they are currently not mentioned in the StatsBase manual (and should probably be deprecated, see https://github.com/JuliaStats/StatsBase.jl/pull/736).

It would even make sense to also include docstrings from Statistics in the StatsBase manual so that one can get them from a single place.

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -36,9 +36,7 @@ export
     quantile,
     quantile!,
     std,
-    stdm,
     var,
-    varm,
 
     ## weights
     AbstractWeights,    # abstract type to represent any weight vector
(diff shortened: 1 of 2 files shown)


ararslan:
I've oftened wondered why we don't just `@reexport using Statistics` in StatsBase.

nalimilan:
I was tempted to do that, but as noted at https://github.com/JuliaStats/StatsBase.jl/pull/736 `stdm` and `varm` are not clearly documented in StatsBase so I'd rather not export them.

bkamins (approved):

PR #736, opened 2021-11-21
Deprecate `varm` and `stdm`

nalimilan:
`varm` and `stdm` are redundant with `var(..., mean=m)` and `std(..., mean=m)`. Moreover, their status is not clear: they are technically exported by Statistics (but not by StatsBase itself), but their docstrings are not included in the manual. They were only tested indirectly, so add tests to ensure deprecations work.

The status of `varm!` is even weirder as it's imported from Statistics, but Statistics does not export it, and it has no docstrings. Keep a deprecation just in case.

### DIFF
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -39,3 +39,10 @@ end
 ### Deprecated September 2019
 @deprecate sum(A::AbstractArray, w::AbstractWeights, dims::Int) sum(A, w, dims=dims)
 @deprecate values(wv::AbstractWeights) convert(Vector, wv)
+
+### Deprecated November 2021
+@deprecate stdm(x::RealArray, w::AbstractWeights, m::Real; corrected::DepBool=nothing) std(x, w, mean=m, corrected=corrected) false
+@deprecate varm(x::RealArray, w::AbstractWeights, m::Real; corrected::DepBool=nothing) var(x, w, mean=m, corrected=corrected) false
+@deprecate stdm(x::RealArray, w::AbstractWeights, m::RealArray, dim::Int; corrected::DepBool=nothing) std(x, w, dim, mean=m, corrected=corrected) false
+@deprecate varm(x::RealArray, w::AbstractWeights, m::RealArray, dim::Int; corrected::DepBool=nothing) var(x, w, dim, mean=m, corrected=corrected) false
+@deprecate varm!(R::AbstractArray, x::RealArray, w::AbstractWeights, m::RealArray, dim::Int; corrected::DepBool=nothing) var!(R, x, w, dim, mean=m, corrected=corrected) false
--- a/src/moments.jl
+++ b/src/moments.jl
@@ -2,26 +2,6 @@
 
 ## var
 """
-    varm(x::AbstractArray, w::AbstractWeights, m, [dim]; corrected=false)
-
-Compute the variance of a real-valued array `x` with a known mean `m`, optionally
-over a dimension `dim`. Observations in `x` are weighted using weight vector `w`.
-The uncorrected (when `corrected=false`) sample variance is defined as:
-```math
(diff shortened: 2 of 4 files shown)


ararslan (approved):

src/deprecates.jl
```diff
@@ -39,3 +39,10 @@ end
 ### Deprecated September 2019
 @deprecate sum(A::AbstractArray, w::AbstractWeights, dims::Int) sum(A, w, dims=dims)
 @deprecate values(wv::AbstractWeights) convert(Vector, wv)
+
+### Deprecated November 2021
+@deprecate stdm(x::RealArray, w::AbstractWeights, m::Real; corrected::DepBool=nothing) std(x, w, mean=m, corrected=corrected) false
```
bkamins, line 44: why do we disable export? This is breaking I think

src/moments.jl
```diff
@@ -43,30 +23,22 @@ function var(v::RealArray, w::AbstractWeights; mean=nothing,
     corrected = depcheck(:var, :corrected, corrected)
 
     if mean == nothing
-        varm(v, w, Statistics.mean(v, w); corrected=corrected)
+        _moment2(v, w, Statistics.mean(v, w); corrected=corrected)
```
bkamins, line 26: why not `var` with kwarg?

src/moments.jl
```diff
@@ -43,30 +23,22 @@ function var(v::RealArray, w::AbstractWeights; mean=nothing,
...
 function var!(R::AbstractArray, A::RealArray, w::AbstractWeights, dims::Int;
               mean=nothing, corrected::DepBool=nothing)
     corrected = depcheck(:var!, :corrected, corrected)
 
     if mean == 0
-        varm!(R, A, w, Base.reducedim_initarray(A, dims, 0, eltype(R)), dims;
-                   corrected=corrected)
-    elseif mean == nothing
-        varm!(R, A, w, Statistics.mean(A, w, dims=dims), dims; corrected=corrected)
+        mean = Base.reducedim_initarray(A, dims, 0, eltype(R))
```
bkamins, line 39: why is this not tested. Also the same as above - why is it better than just calling the default replacement?

test/moments.jl
```diff
@@ -20,11 +20,13 @@ w = [3.84, 2.70, 8.29, 8.91, 9.71, 0.0]
     @testset "Variance" begin
         @test var(x, wv; corrected=false)           ≈ expected_var
         @test var(x, wv; mean=m, corrected=false)   ≈ expected_var
+        @test varm(x, wv, m; corrected=false)   ≈ expected_var
```
bkamins, line 23: maybe we should make the code alignment consistent everywhere?

src/deprecates.jl
nalimilan: No, because it's not exported currently. We still overload the function defined in Statistics though to it can be used after `using Statistics`.

src/moments.jl
nalimilan: We're already in `var`. Calling ourselves with the computed mean would work but do you think it would really be cleaner?

(I'd like to deprecate `_moment2` and similar functions at some point to simplify the API and implementation...)

src/deprecates.jl
bkamins: Ah - OK. I mixed it up with the fact that Statistics exports it.

src/moments.jl
bkamins: :)

bkamins (approved):

src/moments.jl
nalimilan: Turns out it's not tested on master already. This is an undocumented feature of `var` (which calls `var!`). I wish it wasn't allowed, but I guess we have keep supporting it now (and add tests for it)...

PR #737, opened 2021-11-21

aplavin:
update to https://github.com/JuliaStats/StatsBase.jl/pull/658
cannot edit that one

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -380,9 +380,7 @@ Base.:(==)(x::AbstractWeights, y::AbstractWeights)   = false
 
 Compute the weighted sum of an array `v` with weights `w`, optionally over the dimension `dim`.
 """
-wsum(v::AbstractVector, w::AbstractVector) = dot(v, w)
-wsum(v::AbstractArray, w::AbstractVector) = dot(vec(v), w)
-wsum(v::AbstractArray, w::AbstractVector, dims::Colon) = wsum(v, w)
+wsum(v::AbstractArray, w::AbstractVector, dims::Colon=:) = w' * vec(v)
 
 ## wsum along dimension
 #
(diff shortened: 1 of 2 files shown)


src/weights.jl
```diff
@@ -380,9 +380,7 @@ Base.:(==)(x::AbstractWeights, y::AbstractWeights)   = false
 
 Compute the weighted sum of an array `v` with weights `w`, optionally over the dimension `dim`.
 """
-wsum(v::AbstractVector, w::AbstractVector) = dot(v, w)
-wsum(v::AbstractArray, w::AbstractVector) = dot(vec(v), w)
-wsum(v::AbstractArray, w::AbstractVector, dims::Colon) = wsum(v, w)
+wsum(v::AbstractArray, w::AbstractVector, dims::Colon=:) = w' * vec(v)
```
nalimilan, line 383: Don't you want this? `'` calls `adjoint`, which AFAICT is what you want to avoid for complex numbers, right?
```suggestion
wsum(v::AbstractArray, w::AbstractVector, dims::Colon=:) = vec(v)' * w
```

aplavin: I don't understand... Your suggestion also has `'`, right? Mine only applies `'` to weights `w`, while yours to values `v`. The main point of my PR is to avoid applying `'` (adjoint, then conjugate) to `v`.

nalimilan: Sorry, that `v` letter keeps confusing me as I interpret it as meaning "vector". We should change it to `x` or `A` at some point.

Anyway, since `w` is not guaranteed to contain only real values (though we should probably have applied this restriction), let's favor safety over the beauty of the syntax: using `transpose` is  more appropriate here.
```suggestion
wsum(v::AbstractArray, w::AbstractVector, dims::Colon=:) = transpose(w) * vec(v)
```

nalimilan:
Thanks!

PR #742, opened 2021-11-30
Complex signalcorr

kagalenko-m-b:
This PR aims to fix the issue #459. It mostly replaces functions' input type signatures by the less restrictive ones. To produce test vectors for autocovariance and autocorrelation, I added the file ```rcall_test.jl``` that defines test functions, relying on R via RCall.jl.  

I also fixed ```durbin!()``` function so that it is now valid for complex inputs, and avoids the need for repeated invocation by keeping the results for the succession of solved linear sstems.


src/signalcorr.jl
```diff
@@ -61,34 +60,32 @@ where each column in the result will correspond to a column in `x`.
...
+    demean::Bool=true
+    )
     lx = length(x)
     m = length(lags)
     length(r) == m || throw(DimensionMismatch())
     check_lags(lx, lags)
-
     T = typeof(zero(eltype(x)) / 1)
-    z::Vector{T} = demean ? x .- mean(x) : x
+    z .= x
```
mschauer, line 75: This now also allocates `z` if `demean=false`

kagalenko-m-b: Old version allocates when demean=true, though. That is the default case, so presumably more common.

test/signalcorr.jl
```diff
@@ -9,88 +9,76 @@ using Test
 
 # random data for testing
 
-x = [-2.133252557240862    -.7445937365828654;
```
mschauer, line 12: We should still test the real case?

test/signalcorr.jl
```diff
@@ -109,28 +97,41 @@ c11 = crosscor(x1, x1)
...
+           -5.427443903358727e-01 + 3.146715147305132e-01im]
+
+@test pacf(x1, 1:4) ≈ pacf_ls[1:4]
+
+## pacf Yule-Walker
+
+function yulewalker_qr(v::AbstractVector)
+    A = toeplitz(v)
+    b = v[2:end]
+    x =-A\b
```
mschauer, line 119:
```suggestion
    x = -A\b
```

test/signalcorr.jl
```diff
@@ -109,28 +97,41 @@ c11 = crosscor(x1, x1)
...
+end
+function toeplitz(v::AbstractVector{T}) where T
+    N=length(v)
+    A = zeros(T, N - 1, N - 1)
+    for n in 1:N-1
+        A[n, n + 1:end] = conj(v[2:N-n])
```
mschauer, line 125:
```suggestion
        A[n, n+1:end] = conj(v[2:N-n])
```

test/signalcorr.jl
```diff
@@ -109,28 +97,41 @@ c11 = crosscor(x1, x1)
...
+        A[n, 1:n] = reverse(v[1:n])
+    end
+    return A
+end
+# durbin solver 
+acf = autocor(x1)
+p = [yulewalker_qr(acf[1:n])[n-1] for n in 2:length(acf)]
+@test p ≈ StatsBase.durbin(acf[2:end])
+
+pacfy = [];
```
mschauer, line 135: ?

test/signalcorr.jl
kagalenko-m-b: Should we? It would amount to testing the real case in ```LinearAlgebra.dot()```

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,30 @@
...
         α /= β
         for j = 1:div(k,2)
             tmp = y[j]
-            y[j] += α*y[k-j+1]
-            y[k-j+1] += α*tmp
+            y[j] += α*conj(y[k-j+1])
+            y[k-j+1] += α*conj(tmp)
         end
-        if isodd(k) y[div(k,2)+1] *= one(T) + α end
+        if isodd(k) y[div(k,2)+1] += α*conj(y[div(k,2)+1]) end
```
mschauer, line 21:
```suggestion
        if isodd(k)
            y[div(k,2)+1] += α*conj(y[div(k,2)+1]) 
        end
```

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,30 @@
 # Symmetric Toeplitz solver
-function durbin!(r::AbstractVector{T}, y::AbstractVector{T}) where T<:BlasReal
+function durbin!(r::AbstractVector{T}, p::AbstractVector{T}, y::AbstractVector{T}) where T
```
mschauer, line 2: Can you add a docstring if you understand this function? That would be nice.

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,30 @@
...
+            y[j] += α*conj(y[k-j+1])
+            y[k-j+1] += α*conj(tmp)
         end
-        if isodd(k) y[div(k,2)+1] *= one(T) + α end
+        if isodd(k) y[div(k,2)+1] += α*conj(y[div(k,2)+1]) end
         y[k+1] = α
+        p[k+1] = α
     end
-    return y
+    return p
```
mschauer, line 25: What's `p` here?

src/signalcorr.jl
mschauer: Just to avoid changing code and test simultaneously.

test/signalcorr.jl
mschauer: Just to avoid changing code and test simultaneously.

test/signalcorr.jl
```diff
@@ -9,88 +9,76 @@ using Test
...
+     -0.7024909626802256 - 0.2915841698461933im   0.8432078641246675 - 1.373657703019187im
+     -0.5684618031428246 + 0.3729428437838556im  -0.18853209977543475 - 0.9698921354142658im
+     1.2992421672413557 - 0.17943637636017656im  -1.7801079183259816 - 1.3063359194617465im
+     -0.26481424468298875 - 1.1636864449113618im -0.2605299632484854 - 0.7356290016171836im]
 
 x1 = view(x, :, 1)
 x2 = view(x, :, 2)
-realx = convert(AbstractMatrix{Real}, x)
-realx1 = convert(AbstractVector{Real}, x1)
-realx2 = convert(AbstractVector{Real}, x2)
```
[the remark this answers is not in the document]
mschauer: @devmotion You'll know, c.f. #567

src/signalcorr.jl
```diff
@@ -43,8 +43,7 @@ end
 
 default_autolags(lx::Int) = 0 : default_laglen(lx)
 
-_autodot(x::AbstractVector{<:RealFP}, lx::Int, l::Int) = dot(x, 1:(lx-l), x, (1+l):lx)
```
devmotion, line 46: This could/should be kept, I think? It avoids the `view` in the method below.

src/signalcorr.jl
```diff
@@ -61,34 +60,31 @@ where each column in the result will correspond to a column in `x`.
 
 The output is not normalized. See [`autocor!`](@ref) for a method with normalization.
 """
-function autocov!(r::RealVector, x::RealVector, lags::IntegerVector; demean::Bool=true)
+function autocov!(
+    r::AbstractVector,
+    x::AbstractVector,
+    lags::AbstractVector{<:Integer},
+    z=zeros(typeof(zero(eltype(x)) / 1), length(x));
```
devmotion, line 67: What's the reason for `z` here? I assume it is intended to fix the type instability? I think this should be done differently: The implementation here will lead to different containers (always a `Vector`) than the current implementation.

Moreover, it is not used in the non-`demean` case below, and in general does not have to be filled with zeros.

src/signalcorr.jl
```diff
@@ -61,34 +60,31 @@ where each column in the result will correspond to a column in `x`.
...
+    demean::Bool=true
+    )
     lx = length(x)
     m = length(lags)
     length(r) == m || throw(DimensionMismatch())
     check_lags(lx, lags)
-
     T = typeof(zero(eltype(x)) / 1)
-    z::Vector{T} = demean ? x .- mean(x) : x
+    demean  ? z = x :   z .= x .- mean(x)
```
devmotion, line 75: The order is wrong. Moreover, `z` should be modified in both cases - currently a new variable (possibly of a different type) is created in the first branch.


```suggestion
    demean ? z .= x .- mean(x) : copyto!(z, x)
```

Maybe it would be simpler to just change this to

```suggestion
    z = x .- (demean ? mean(x) : zero(eltype(x)) / 1)
```
and not pass `z` as an argument. This will be type stable. Maybe one could check if `x` is already of the correct element type but then it still can't be guaranteed that `x` would be of the same type as the other branch.

src/signalcorr.jl
```diff
@@ -110,18 +106,18 @@ When left unspecified, the lags used are the integers from 0 to
 
 The output is not normalized. See [`autocor`](@ref) for a function with normalization.
 """
-function autocov(x::RealVector, lags::IntegerVector; demean::Bool=true)
+function autocov(x::AbstractVector, lags::AbstractVector{<:Integer}; demean::Bool=true)
     out = Vector{float(eltype(x))}(undef, length(lags))
-    autocov!(out, x, lags; demean=demean)
+    autocov!(out, x, lags; demean)
```
devmotion, line 111: This requires Julia >= 1.4


```suggestion
    autocov!(out, x, lags; demean=demean)
```

src/signalcorr.jl
```diff
@@ -110,18 +106,18 @@ When left unspecified, the lags used are the integers from 0 to
...
 end
 
-function autocov(x::RealMatrix, lags::IntegerVector; demean::Bool=true)
+function autocov(x::AbstractMatrix, lags::AbstractVector{<:Integer}; demean::Bool=true)
     out = Matrix{float(eltype(x))}(undef, length(lags), size(x,2))
-    autocov!(out, x, lags; demean=demean)
+    autocov!(out, x, lags; demean)
```
devmotion, line 116:
```suggestion
    autocov!(out, x, lags; demean=demean)
```

src/signalcorr.jl
```diff
@@ -110,18 +106,18 @@ When left unspecified, the lags used are the integers from 0 to
...
 end
 
-autocov(x::AbstractVecOrMat{<:Real}; demean::Bool=true) =
-    autocov(x, default_autolags(size(x,1)); demean=demean)
+autocov(x::AbstractVecOrMat; demean::Bool=true) =
+    autocov(x, default_autolags(size(x,1)); demean)
```
devmotion, line 120:
```suggestion
    autocov(x, default_autolags(size(x,1)); demean=demean)
```

src/signalcorr.jl
```diff
@@ -139,36 +135,27 @@ where each column in the result will correspond to a column in `x`.
...
+    r::AbstractVector,
+    x::AbstractVector,
+    lags::AbstractVector{<:Integer},
+    z=zeros(typeof(zero(eltype(x)) / 1), length(x));
+    demean::Bool=true
+    )
+    autocov!(view(r, 1:1), x, 0:0, z; demean)
+    zz = r[1]
+    autocov!(r, x, lags, z; demean)
+    r ./= zz
```
devmotion, line 148: One could avoid the expensive broadcasting machinery by using eg. `ldiv!`.

src/signalcorr.jl
```diff
@@ -139,36 +135,27 @@ where each column in the result will correspond to a column in `x`.
...
-        for k = 1 : m
-            r[k,j] = _autodot(z, lx, lags[k]) / zz
-        end
+function autocor!(
+    r::AbstractMatrix, x::AbstractMatrix, lags::AbstractVector{<:Integer}; demean::Bool=true
+    )
+    T = typeof(zero(eltype(x))/1)
+    z = Vector{T}(undef, size(x, 1))
+    for n in 1:size(x, 2)
+        autocor!(view(r, :, n), view(x, :, n), lags, z; demean)
```
devmotion, line 158:
```suggestion
        autocor!(view(r, :, n), view(x, :, n), lags, z; demean=demean)
```

src/signalcorr.jl
```diff
@@ -191,18 +178,18 @@ When left unspecified, the lags used are the integers from 0 to
 The output is normalized by the variance of `x`, i.e. so that the lag 0
 autocorrelation is 1. See [`autocov`](@ref) for the unnormalized form.
 """
-function autocor(x::RealVector, lags::IntegerVector; demean::Bool=true)
+function autocor(x::AbstractVector, lags::AbstractVector{<:Integer}; demean::Bool=true)
     out = Vector{float(eltype(x))}(undef, length(lags))
-    autocor!(out, x, lags; demean=demean)
+    autocor!(out, x, lags; demean)
```
devmotion, line 183:
```suggestion
    autocor!(out, x, lags; demean=demean)
```

src/signalcorr.jl
```diff
@@ -191,18 +178,18 @@ When left unspecified, the lags used are the integers from 0 to
...
 end
 
-function autocor(x::RealMatrix, lags::IntegerVector; demean::Bool=true)
+function autocor(x::AbstractMatrix, lags::AbstractVector{<:Integer}; demean::Bool=true)
     out = Matrix{float(eltype(x))}(undef, length(lags), size(x,2))
-    autocor!(out, x, lags; demean=demean)
+    autocor!(out, x, lags; demean)
```
devmotion, line 188:
```suggestion
    autocor!(out, x, lags; demean=demean)
```

src/signalcorr.jl
```diff
@@ -191,18 +178,18 @@ When left unspecified, the lags used are the integers from 0 to
...
 end
 
-autocor(x::AbstractVecOrMat{<:Real}; demean::Bool=true) =
-    autocor(x, default_autolags(size(x,1)); demean=demean)
+autocor(x::AbstractVecOrMat; demean::Bool=true) =
+    autocor(x, default_autolags(size(x,1)); demean)
```
devmotion, line 192:
```suggestion
    autocor(x, default_autolags(size(x,1)); demean=demean)
```

src/signalcorr.jl
```diff
@@ -213,14 +200,7 @@ autocor(x::AbstractVecOrMat{<:Real}; demean::Bool=true) =
 
 default_crosslags(lx::Int) = (l=default_laglen(lx); -l:l)
 
-function _crossdot(x::AbstractVector{T}, y::AbstractVector{T}, lx::Int, l::Int) where {T<:RealFP}
```
devmotion, line 216: This optimization could (should?) be kept, maybe?

devmotion:
I wonder if any of the implementations could benefit from RealDot.jl? It contains optimized implementations of `real(dot(x, y))`.

src/signalcorr.jl
kagalenko-m-b: If ```z``` is not passed as an argument, that will result in repeated allocations [here](https://github.com/JuliaStats/StatsBase.jl/pull/742/files#diff-415d7dec9339a5c81ca5a48be3206acffd88d83bf8dc8ba685fef3e94880fb2fR87)

devmotion: I see, then you should use the first suggestion :slightly_smiling_face: Currently the allocations are not avoided and the implementation is incorrect.

test/signalcorr.jl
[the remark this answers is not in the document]
devmotion: This is required for e.g. https://github.com/TuringLang/MCMCChains.jl/issues/187#issuecomment-593621159. In any case, we should not break or remove existing functionality in this PR and in particular not remove any tests.

kagalenko-m-b:
@devmotion The function ```demean_col()``` should be considered broken and replaced because of [this](https://github.com/JuliaLang/Statistics.jl/issues/22) .

devmotion:
Not sure why you pinged me but this seems unrelated to this PR and not specific to complex inputs? It would be good to keep the PR minimal and just make the changes that are needed to support complex numbers. I would prefer if other issues are discussed or directly fixed in separate PRs. It would be important as well to keep all existing tests to ensure that no existing functionality is broken (as I mentioned also in this comment: https://github.com/JuliaStats/StatsBase.jl/pull/742#discussion_r759381162).

src/signalcorr.jl
```diff
@@ -64,15 +63,15 @@ function autocov!(
...
     demean::Bool=true
     )
     lx = length(x)
     m = length(lags)
     length(r) == m || throw(DimensionMismatch())
     check_lags(lx, lags)
-    T = typeof(zero(eltype(x)) / 1)
-    demean  ? z = x :   z .= x .- mean(x)
+    z .= x
+    demean && z .= z .- mean(z)
```
devmotion, line 74: Simpler and only writing to `z` once and without broadcasting overhead in the second branch:
```suggestion
    demean ? z .= x .- mean(x) : copyto!(z, x) 
```

src/signalcorr.jl
```diff
@@ -43,9 +43,7 @@ end
 
 default_autolags(lx::Int) = 0 : default_laglen(lx)
 
-_autodot(x::AbstractVector{<:RealFP}, lx::Int, l::Int) = dot(x, 1:(lx-l), x, (1+l):lx)
-_autodot(x::RealVector, lx::Int, l::Int) = dot(view(x, 1:(lx-l)), view(x, (1+l):lx))
-
+_autodot(x::AbstractVector, lx::Int, l::Int) = dot(view(x, 1:(lx-l)), view(x, (1+l):lx))
```
devmotion, line 46:
```suggestion
_autodot(x::AbstractVector{<:RealFP}, lx::Int, l::Int) = dot(x, 1:(lx-l), x, (1+l):lx)
_autodot(x::AbstractVector, lx::Int, l::Int) = dot(view(x, 1:(lx-l)), view(x, (1+l):lx))
```

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
 
 The output is not normalized. See [`autocor!`](@ref) for a method with normalization.
 """
-function autocov!(r::RealVector, x::RealVector, lags::IntegerVector; demean::Bool=true)
+function autocov!(
+    r::AbstractVector,
+    x::AbstractVector,
+    lags::AbstractVector{<:Integer},
```
devmotion, line 65: Is this the same as `IntegerVector`?

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
...
+    z=Vector{typeof(zero(eltype(x)) / 1)}(undef, size(x, 1));
```
devmotion, line 66:
```suggestion
    z::Vector=Vector{typeof(zero(eltype(x)) / 1)}(undef, length(x));
```

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
...
+    demean::Bool=true
+    )
```
devmotion, line 68:
```suggestion
)
```

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
...
-function autocov!(r::RealMatrix, x::RealMatrix, lags::IntegerVector; demean::Bool=true)
-    lx = size(x, 1)
-    ns = size(x, 2)
-    m = length(lags)
-    size(r) == (m, ns) || throw(DimensionMismatch())
-    check_lags(lx, lags)
-
+function autocov!(
+    r::AbstractMatrix, x::AbstractMatrix, lags::AbstractVector{<:Integer}; demean::Bool=true
+    )
```
devmotion, line 82:
```suggestion
)
```

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
...
+function autocov!(
+    r::AbstractVector,
+    x::AbstractVector,
+    lags::AbstractVector{<:Integer},
+    z=Vector{typeof(zero(eltype(x)) / 1)}(undef, size(x, 1));
+    demean::Bool=true
+    )
     lx = length(x)
     m = length(lags)
     length(r) == m || throw(DimensionMismatch())
```
devmotion, line 71: Maybe add a check of `length(z)` as well?

devmotion:
```suggestion
    length(r) == length(z) == m || throw(DimensionMismatch())
```

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
...
     check_lags(lx, lags)
-
-    T = typeof(zero(eltype(x)) / 1)
-    z::Vector{T} = demean ? x .- mean(x) : x
+    demean ? z .= x .- mean(x) : copyto!(z, x)
     for k = 1 : m  # foreach lag value
         r[k] = _autodot(z, lx, lags[k]) / lx
     end
```
devmotion, line 76:
```suggestion
    @inbounds for (k, lags_k) in zip(eachindex(r), lags) # foreach lag value
        r[k] = _autodot(z, lx, lags_k) / lx
    end
```

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
...
-    z = Vector{T}(undef, lx)
-    for j = 1 : ns
-        demean_col!(z, x, j, demean)
-        for k = 1 : m
-            r[k,j] = _autodot(z, lx, lags[k]) / lx
-        end
+    z = Vector{T}(undef, size(x, 1))
+    for n in 1:size(x, 2)
+        autocov!(view(r, :, n), view(x, :, n), lags, z; demean)
     end
```
devmotion, line 87: Why did you replace the old version? It seems fine and also did not cause unnecessary allocations - vector `z` was allocated up-front and reused for every column.

src/signalcorr.jl
```diff
@@ -139,36 +134,27 @@ where each column in the result will correspond to a column in `x`.
...
-    end
+function autocor!(
+    r::AbstractVector,
+    x::AbstractVector,
+    lags::AbstractVector{<:Integer},
+    z=zeros(typeof(zero(eltype(x)) / 1), length(x));
+    demean::Bool=true
+    )
+    autocov!(view(r, 1:1), x, 0:0, z; demean)
+    zz = r[1]
```
devmotion, line 145: This seems strange. I wonder if we could just update the method in the same way as `autocov!` above?

src/signalcorr.jl
```diff
@@ -139,36 +134,27 @@ where each column in the result will correspond to a column in `x`.
...
-        for k = 1 : m
-            r[k,j] = _autodot(z, lx, lags[k]) / zz
-        end
+function autocor!(
+    r::AbstractMatrix, x::AbstractMatrix, lags::AbstractVector{<:Integer}; demean::Bool=true
+    )
+    T = typeof(zero(eltype(x))/1)
+    z = Vector{T}(undef, size(x, 1))
+    for n in 1:size(x, 2)
+        autocor!(view(r, :, n), view(x, :, n), lags, z; demean=demean)
```
devmotion, line 157: Again, the current version seems fine?

devmotion (commented):
I had another look and my impression is that apart from the changes of the input types most changes should not be needed. The existing code for the matrix case is already efficient and reuses a cache, so allocations should not be reduced by this PR (in contrast, for the vector case it will always allocate an array now but doesn't currently).

kagalenko-m-b:
> I had another look and my impression is that apart from the changes of the input types most changes should not be needed. The existing code for the matrix case is already efficient and reuses a cache, so allocations should not be reduced by this PR (in contrast, for the vector case it will always allocate an array now but doesn't currently).

The reasoning here was not to improve efficiency, but to reduce the code duplication.

kagalenko-m-b:
> Is this the same as IntegerVector

Yes but, from the file where IntegerVector is defined:
>  These could be removed when the Base supports covariant type notation, i.e.  AbstractVector{<:Real}

src/signalcorr.jl
```diff
@@ -61,34 +59,31 @@ where each column in the result will correspond to a column in `x`.
 
 The output is not normalized. See [`autocor!`](@ref) for a method with normalization.
 """
-function autocov!(r::RealVector, x::RealVector, lags::IntegerVector; demean::Bool=true)
+function autocov!(
+    r::AbstractVector,
+    x::AbstractVector,
+    lags::AbstractVector{<:Integer},
+    z::Vector=Vector{typeof(zero(eltype(x)) / 1)}(undef, length(x));
```
devmotion, line 66: @kagalenko-m-b What do you think about adding an internal `_autocov` function that requires a `z` vector and could be used by the different `autocov!` and `autocor!` methods? Then the `z` cache would not leak to the user-facing API and still code duplications could be reduced.

src/signalcorr.jl
```diff
@@ -458,7 +458,7 @@ function crosscor!(r::RealMatrix, x::RealVector, y::RealMatrix, lags::IntegerVec
     return r
 end
 
-function crosscor!(r::AbstractArray{<:Real,3}, x::RealMatrix, y::RealMatrix, lags::IntegerVector; demean::Bool=true)
+function crosscor!(r::AbstractArray{U,3} where U, x::AbstractMatrix, y::AbstractMatrix, lags::AbstractVector{<:Integer}; demean::Bool=true)
```
nalimilan, line 461:
```suggestion
function crosscor!(r::AbstractArray{<:Any,3}, x::AbstractMatrix, y::AbstractMatrix, lags::AbstractVector{<:Integer}; demean::Bool=true)
```

test/signalcorr_complex.jl
```diff
@@ -0,0 +1,153 @@
+# Test corr.jl
```
nalimilan, line 1: Can you put this in signalcorr.jl instead?

test/rcall_test.jl
```diff
@@ -0,0 +1,29 @@
+using RCall
```
nalimilan, line 1: I'd rather not have our tests depend on R. Hardcoding test values known to be the same as R is enough.

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
-# Symmetric Toeplitz solver
-function durbin!(r::AbstractVector{T}, y::AbstractVector{T}) where T<:BlasReal
+"""
+    durbin!(r::AbstractVector, y::AbstractVector, p::AbstractVector) -> Vector
+
+Solve Yule-Walker equations using Durbin algorithm.
+
+Solution of NxN system is obtained iteratively, by solving 1x1, 2x2, ...
+in succession. For use in computing partial autocorrelation matrix,
+return the last elements of the successive solutions in vector p[].
```
nalimilan, line 8:
```suggestion
return the last elements of the successive solutions in vector `p`.
```

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
-# Symmetric Toeplitz solver
-function durbin!(r::AbstractVector{T}, y::AbstractVector{T}) where T<:BlasReal
+"""
+    durbin!(r::AbstractVector, y::AbstractVector, p::AbstractVector) -> Vector
+
+Solve Yule-Walker equations using Durbin algorithm.
+
+Solution of NxN system is obtained iteratively, by solving 1x1, 2x2, ...
```
nalimilan, line 6:
```suggestion
Solution of N×N system is obtained iteratively, by solving 1×1, 2×2, ...
```

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
...
+in succession. For use in computing partial autocorrelation matrix,
+return the last elements of the successive solutions in vector p[].
+
+The section 4.7 of Golub,VanLoan "Matrix Computations," 4th ed.,
```
nalimilan, line 10: Can you give the full reference?

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
...
+discusses the algorithm in detail.
+
+# Arguments
+- `r::AbstractVector`: first column of a herimitian positive definite
```
nalimilan, line 14:
```suggestion
- `r::AbstractVector`: first column of a Hermitian positive definite
```

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
...
+    Toeplitz matrix, excluding the diagonal element (equal to one).
+- `y::AbstractVector`: work vector for solution, should have the same length
+    as r[]
```
nalimilan, line 17:
```suggestion
    as `r`
```

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
...
+- `p::AbstractVector`: last elements of the successive solutions, should
+     have the same length  as r[]
```
nalimilan, line 19:
```suggestion
     have the same length  as `r`
```

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
...
+
+# Returns
+- `y::AbstractVector`: same as the second argument
```
nalimilan, line 22: The convention we use is rather to say "return a vector..." in the first sentence.

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
...
             tmp = y[j]
-            y[j] += α*y[k-j+1]
-            y[k-j+1] += α*tmp
+            y[j] += α*conj(y[k-j+1])
+            y[k-j+1] += α*conj(tmp)
+        end
+        if isodd(k)
+            y[div(k,2)+1] += α*conj(y[div(k,2)+1]) 
         end
-        if isodd(k) y[div(k,2)+1] *= one(T) + α end
```
nalimilan, line 20: Any reason to change this?

src/toeplitzsolvers.jl
```diff
@@ -1,28 +1,54 @@
...
         y[k+1] = α
+        p[k+1] = α
     end
     return y
 end
-durbin(r::AbstractVector{T}) where {T<:BlasReal} = durbin!(r, zeros(T, length(r)))
+durbin(r::AbstractVector{T}) where T = durbin!(r, zeros(T, length(r)), zeros(T, length(r)))
```
nalimilan, line 51:
```suggestion
durbin(r::AbstractVector{T}) where {T} = durbin!(r, zeros(T, length(r)), zeros(T, length(r)))
```

test/signalcorr_complex.jl
```diff
@@ -0,0 +1,153 @@
...
+pacf_ls = [-1.598495044296996e-03 - 2.915104118351207e-01im
+           -5.560162016912027e-01 + 2.950837739894279e-01im
+           -2.547001916363494e-02 + 2.326084658014266e-01im
+           -5.427443903358727e-01 + 3.146715147305132e-01im]
+
+@test pacf(x1, 1:4) ≈ pacf_ls[1:4]
+
+## pacf Yule-Walker
+
+function yulewalker_qr(v::AbstractVector)
```
nalimilan, line 131: Can you add a comment explaining what these do? Why don't we check this for reals too?

test/signalcorr_complex.jl
```diff
@@ -0,0 +1,153 @@
...
+    A = toeplitz(v)
+    b = v[2:end]
+    x = -A\b
+end
+function toeplitz(v::AbstractVector{T}) where T
+    N=length(v)
```
nalimilan, line 137:
```suggestion
    N = length(v)
```

nalimilan (commented):
Thanks!

Could you add tests for when the input array contains `missing`? Previously this was an error, but now that the signature is less strict we should ensure we don't return misleading results.

nalimilan:
Thanks, but you don't seem to have answered all my questions/comments.

kagalenko-m-b:
> Thanks, but you don't seem to have answered all my questions/comments.

I have not resolved [this](https://github.com/JuliaStats/StatsBase.jl/pull/742#discussion_r762436508) and [this](https://github.com/JuliaStats/StatsBase.jl/pull/742#discussion_r762437629) conversations, because I am not sure how to proceed. Those are similar - I added functions to generate complex test vectors. On one hand, those functions are not necessary to run the tests, they can be replaced by their outputs, and the one with ```RCall``` isn't used at all. On the other hand, they document the origin of the test vectors. For the [second](https://github.com/JuliaStats/StatsBase.jl/pull/742#discussion_r762437629) of those, it is present only in the complex case, because the real case already has test vectors. So, as a maintainer, please indicate your preference for the best way to resolve this. It can be as simple as deleting those parts from PR, because they are really a kind of documentation and need not to be run.

kagalenko-m-b:
> Thanks, but you don't seem to have answered all my questions/comments.

In the latest revision I have removed all parts that caused questions. Now this PR is streamlined - it does nothing other than correct function signatures and revise the Durbin algorithm to make it valid for complex inputs. The tests with the nightly Julia are failing, but not on the parts that I touched.

kagalenko-m-b:
Rebased on the latest master, any chance of having this merged?

PR #744, opened 2021-12-04
Use function definitions in StatsAPI

rafaqz:
Companion PR to:
https://github.com/JuliaStats/StatsAPI.jl/pull/4


src/statmodels.jl
```diff
@@ -224,7 +7,7 @@ Coefficient of determination (R-squared).
 For a linear model, the R² is defined as ``ESS/TSS``, with ``ESS`` the explained sum of squares
 and ``TSS`` the total sum of squares.
 """
-function r2(model::StatisticalModel)
+function StatsAPI.r2(model::StatisticalModel)
```
nalimilan, line 10: Not needed since you `import`ed it.
```suggestion
function r2(model::StatisticalModel)
```

src/statmodels.jl
```diff
@@ -253,7 +36,7 @@ In the above formulas, ``L`` is the likelihood of the model,
 The Cox-Snell and the deviance ratio variants both match the classical definition of R²
 for linear models.
 """
-function r2(model::StatisticalModel, variant::Symbol)
+function StatsAPI.r2(model::StatisticalModel, variant::Symbol)
```
nalimilan, line 39:
```suggestion
function r2(model::StatisticalModel, variant::Symbol)
```

src/statmodels.jl
```diff
@@ -302,7 +73,7 @@ In these formulas, ``L`` is the likelihood of the model, ``L0`` that of the null
 ``D_0`` is the deviance of the null model, ``n`` is the number of observations (given by [`nobs`](@ref)) and
 ``k`` is the number of consumed degrees of freedom of the model (as returned by [`dof`](@ref)).
 """
-function adjr2(model::StatisticalModel, variant::Symbol)
+function StatsAPI.adjr2(model::StatisticalModel, variant::Symbol)
```
nalimilan, line 76:
```suggestion
function adjr2(model::StatisticalModel, variant::Symbol)
```

src/statmodels.jl
```diff
@@ -276,18 +59,6 @@ end
 
 const r² = r2
```
nalimilan, line 61: No longer needed since that's handled by StatsAPI, right?
```suggestion
```

src/StatsBase.jl
```diff
@@ -20,7 +20,15 @@ import LinearAlgebra: BlasReal, BlasFloat
 import Statistics: mean, mean!, var, varm, varm!, std, stdm, cov, covm,
                    cor, corm, cov2cor!, unscaled_covzm, quantile, sqrt!,
                    median, middle
-import StatsAPI: pairwise, pairwise!
+import StatsAPI: StatisticalModel, RegressionModel
+import StatsAPI: pairwise, pairwise!, params, params!,
```
nalimilan, line 24:
```suggestion
using StatsAPI: StatisticalModel, RegressionModel
import StatsAPI: pairwise, pairwise!, params, params!,
```

src/statmodels.jl
rafaqz: I normally add these for clarity, not because you have to. It just makes it clear the method is not defined here.

nalimilan: We don't do this in the rest of StatsBase (e.g. for functions imported from Statistics), so better be consistent.

nalimilan:
Ah, building the docs fails as docstrings are missing now. Maybe we could remove this line:
https://github.com/JuliaStats/StatsBase.jl/blob/b0ca0b06cdc78f68d1fa2af5de4524b79494ff9e/docs/make.jl#L10
That will also include functions from Statistics.jl, but that may actually be a good thing to centralize docs in a single place.

rafaqz:
Adding `StatsAPI` to modules seems to work fine. 

I've mentioned in the docs that the types and methods are actually defined in StatsAPI.jl

docs/src/statmodels.md
```diff
@@ -1,6 +1,7 @@
 # Abstraction for Statistical Models
 
-This package defines an abstract type `StatisticalModel`, and an abstract subtype `RegressionModel`.
+[StatsAPI.jl](https://github.com/JuliaStats/StatsAPI.jl) defines an abstract type `StatisticalModel`,
+and an abstract subtype `RegressionModel`. They are both extented by StatsBase, and documented here.
```
nalimilan, line 4: I wouldn't mention that StatsBase extends these types since that's the case only for a pair of functions (which should probably be implemented in StatsAPI too):
```suggestion
[StatsAPI.jl](https://github.com/JuliaStats/StatsAPI.jl) defines an abstract type `StatisticalModel`,
and an abstract subtype `RegressionModel`. Functions defined for these types in StatsAPI
are reexported by StatsBase and documented here for convenience.
```

docs/src/statmodels.md
```diff
@@ -30,7 +31,6 @@ rss
 score
 stderror
 vcov
-weights(::StatisticalModel)
```
nalimilan, line 33: Why not keep this?

src/statmodels.jl
```diff
@@ -1,220 +1,3 @@
...
-Bayesian Information Criterion, defined as ``-2 \\log L + k \\log n``, with ``L``
-the likelihood of the model,  ``k`` its number of consumed degrees of freedom
-(as returned by [`dof`](@ref)), and ``n`` the number of observations
-(as returned by [`nobs`](@ref)).
-"""
-bic(model::StatisticalModel) = -2loglikelihood(model) + dof(model)*log(nobs(model))
-
 """
     r2(model::StatisticalModel)
     r²(model::StatisticalModel)
```
nalimilan, line 3: Actually I hadn't realized that we haven't moved these to StatsAPI with the rest of functions. I admit these are not mere stubs, but implementations are trivial and lightweight enough that I think it's OK to define them in StatsAPI. Then the situation would be clearer. See https://github.com/JuliaStats/StatsAPI.jl/pull/7.

nalimilan:
Cool. I think you also need to add StatsAPI to docs/Project.toml to get the docs to build correctly.

rafaqz:
Yep, unfortunately that doesn't cause an error locally.

docs/src/statmodels.md
rafaqz: I was wondering about moving those as well, since we have the others now...

docs/src/statmodels.md
rafaqz: There is no longer a specific definition in StatsAPI - its just `function weights end`.

`weight` cant stay here either because it's already included in weights.md, so you get a duplicate entry error

src/statmodels.jl
rafaqz: Ok, I'll remove them from this PR as well when your StatsAPI PR is moved and bumped.

docs/src/statmodels.md
nalimilan: Ah, indeed. That's annoying. I guess the only solution is to include just `weights` here, but make the signatures in weights.md more specific.

src/statmodels.jl
nalimilan: I've tagged the new release with that change.

docs/src/statmodels.md
rafaqz: Ok I've reshuffled this so it can still be in both, with the specific signature in weights.md

src/statmodels.jl
rafaqz: This is updated to match now.

docs/src/statmodels.md
nalimilan, line 4:
```suggestion
and an abstract subtype `RegressionModel`. They are both extended by StatsBase, and documented here.
```

nalimilan:
Thanks!

PR #757, opened 2022-01-18
Improve weights docstrings

nalimilan:
`AnalyticWeights` have a precise definition, on which we rely in several functions.
Also make docstrings more consistent across types.

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -96,8 +96,8 @@ end
 Construct an `AnalyticWeights` vector with weight values `vs`.
 A precomputed sum may be provided as `wsum`.
 
-Analytic weights describe a non-random relative importance (usually between 0 and 1)
-for each observation. These weights may also be referred to as reliability weights,
+Analytic weights represent the inverse of the variance for each case.
+These weights may also be referred to as reliability weights,
 precision weights or inverse variance weights. These are typically used when the observations
 being weighted are aggregate values (e.g., averages) with differing variances.
 """ AnalyticWeights
@@ -136,7 +136,7 @@ end
 Construct a `FrequencyWeights` vector with weight values `vs`.
 A precomputed sum may be provided as `wsum`.
 
-Frequency weights describe the number of times (or frequency) each observation
+Frequency weights represent the number of times (or frequency) each case
 was observed. These weights may also be referred to as case weights or repeat weights.
 """ FrequencyWeights
 
@@ -173,9 +173,9 @@ end
 Construct a `ProbabilityWeights` vector with weight values `vs`.
 A precomputed sum may be provided as `wsum`.
 
-Probability weights represent the inverse of the sampling probability for each observation,
+Probability weights represent the inverse of the sampling probability for each case,
 providing a correction mechanism for under- or over-sampling certain population groups.
-These weights may also be referred to as sampling weights.
+These weights may also be referred to as sampling weights or survey weights.
 """ ProbabilityWeights
 
 """


ParadaCarleton:
I think we should make an important note for both `AnalyticWeights` and `FrequencyWeights` that the scale of the weights matters, i.e. `f(x, weights)` will usually not equal `f(x, 2 .* weights)`.

It would also help to be very specific about the following, to avoid people making the mistake of trying to use normalized weights:
For `AnalyticWeights`, `w[i]` must be equal to `1/var(x[i])`, and `var(x[i])` must be known ahead of time.
For `FrequencyWeights`, `w[i]` must be equal to the number of observations for `x[i]`.

I believe the docstring for `ProbabilityWeights` is already clear enough, because probability weights are scale-invariant. (For now; in theory we might want to make a distinction between self-normalized and unnormalized weights in the future, because unnormalized weights can give unbiased estimators, but ATM we only use self-normalized weights).

src/weights.jl
```diff
@@ -96,8 +96,8 @@ end
 Construct an `AnalyticWeights` vector with weight values `vs`.
 A precomputed sum may be provided as `wsum`.
 
-Analytic weights describe a non-random relative importance (usually between 0 and 1)
-for each observation. These weights may also be referred to as reliability weights,
+Analytic weights represent the inverse of the variance for each case.
```
rofinn, line 99: `inverse of the variance` feels kinda redundant here since we also mention that they could also be called inverse variance weights.

src/weights.jl
```diff
@@ -96,8 +96,8 @@ end
...
+These weights may also be referred to as reliability weights,
 precision weights or inverse variance weights. These are typically used when the observations
```
rofinn, line 101:
```suggestion
precision weights or regression weights. These are typically used when the observations
```

src/weights.jl
```diff
@@ -136,7 +136,7 @@ end
 Construct a `FrequencyWeights` vector with weight values `vs`.
 A precomputed sum may be provided as `wsum`.
 
-Frequency weights describe the number of times (or frequency) each observation
+Frequency weights represent the number of times (or frequency) each case
```
rofinn, line 139: Change from `observation` to `case` seem good 👍🏻

rofinn (approved):
LGTM. We already mentioned inverse variance weights, so maybe rewording that docstring to be less redundant would be good?

src/weights.jl
nalimilan: Yeah maybe we can avoid repeating it later. But I think this should be mentioned here as the definition of the weights.

src/weights.jl
nalimilan: Are you sure the expression "regression weights" is used commonly? I'm afraid it could confuse users, as you can use any kind of weights in regression (e.g. Stata supports all three types).

nalimilan:
> I think we should make an important note for both `AnalyticWeights` and `FrequencyWeights` that the scale of the weights matters, i.e. `f(x, weights)` will usually not equal `f(x, 2 .* weights)`.
> 
> It would also help to be very specific about the following, to avoid people making the mistake of trying to use normalized weights: For `AnalyticWeights`, `w[i]` must be equal to `1/var(x[i])`, and `var(x[i])` must be known ahead of time. For `FrequencyWeights`, `w[i]` must be equal to the number of observations for `x[i]`.

I've added mentions regarding scale-invariance of frequency and probability weights. For analytic weights I'd rather wait until https://github.com/JuliaStats/StatsBase.jl/issues/758 is settled as we may want to adjust the definition a bit (maybe in the next breaking release). AFAICT currently they are scale-invariant, right?

src/weights.jl
rofinn: I think SAS uses that term for them, but I don't know how common it is elsewhere. I'm also fine with just removing the redundant `inverse variance` and only listing two other common names (e.g., `reliability weights`, `precision weights`).

https://blogs.sas.com/content/iml/2017/10/02/weight-variables-in-statistics-sas.html

nalimilan: OK, let's do that.

nalimilan:
I've updated the description of analytic weights in the light of https://github.com/JuliaStats/StatsBase.jl/issues/758. Does that sound correct?

ParadaCarleton:
> I've updated the description of analytic weights in the light of #758. Does that sound correct?

I think we should hold off until #758 is resolved.

ParadaCarleton:
Oh, brief note that I think could be useful for users -- currently, all our methods for `ProbabilityWeights` normalize the weights before calculating an estimator. This is probably the best default, but sometimes it's useful to use the Horvitz-Thompson estimators for means, variances, etc.; these estimators use the unnormalized weights, which makes them unbiased (but _usually_ results in higher variance). The behavior of these estimators can be replicated by setting `sum=1`, in which case the weights won't be normalized.

nalimilan:
@ParadaCarleton Don't you think that this PR is a strict improvement over the current situation, even if we decide to split `AnalyticWeights` into several types?

ParadaCarleton:
I think it's an improvement, yeah, but I'd clarify that the weights:
1. Refer specifically to sample sizes for each observation.
2. I'd add a warning about `std` doing something different from what you'd expect.

PR #760, opened 2022-02-09
add a fast binnedStatistic function to match SciPy functionality

kirklong:
Disclaimer: This is my first pull request to a "real" project ever, so apologies if I've done anything wrong and appreciate your help!

I recently decided to switch a project I was working on from Python to Julia for performance reasons, but one thing I needed that did not appear to exist yet in Julia was an analogue to the [SciPy `binned_statistic`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic.html) function, so I figured this would be a fun challenge to write my own! With some [help from the community](https://discourse.julialang.org/t/weighted-histogram-2x-as-slow-in-julia-vs-python/74953) I was able to make it even faster than the SciPy implementation.

This is a generalization of the histogram function where a statistic is computed on each bin. I only needed the sum use case, but before submitting this pull request I've generalized the code to compute other common statistics at each bin -- in this implementation the statistics options are sum, mean, median, standard deviation, and variance. There is a long doc-string I wrote up before the function in the file that has more info + usage examples.

Technically this can be accomplished (albeit much slower) via existing functionality in StatsBase, i.e. the following bit of code would produce a "summed histogram": 

```julia
function histSum(x::Array,y::Array,bins::Int=100)
    h = fit(Histogram, vec(x), aweights(y), nbins=bins)
    h.edges, h.weights
end
```

However this is prohibitively slow, both compared to the Python version and my new implementation, so I thought it might be nice to have this as a function that all Julia/StatsBase users could benefit from if they ever need something similar! In testing on my machine the speed-up between my new version and the way that uses the existing functionality for the sum statistic is ~15-20x. The only slow statistic (but still faster than existing functionality) is my current implementation of median, which is pretty hacky/lazy but I was trying to closer match the SciPy functionality. Happy to write up corresponding documentation for this and hope to continue to improve performance + available options if this is well-received and actually helpful!

I added this new function to the bottom of `hist.jl` as that seemed like the best place for it, but of course happy for it to be moved around / turned into a different file / whatever.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -2,6 +2,7 @@ using Base.Cartesian
 
 import Base: show, ==, push!, append!, float
 import LinearAlgebra: norm, normalize, normalize!
+import Statistics: median
 
 
 ## Fast getindex function for multiple arrays, returns a tuple of array elements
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks for proposing this! However, I'm hesitant to add something like this to StatsBase as it seems like a very specific function do perform operations which can normally be achieved using more generic functions in other packages. In general, we try to keep the API minimal, with generic building blocks that can be combined to achieve all operations. This is possible as Julia is fast, contrary to SciPy which needs to provide special functions written in a low-level language to get good performance.

There seem to be two different use cases here:
- A weighted histogram, which is what you originally needed IIUC. You say the code is super slow, but there's no reason we can't improve performance under the hood while keeping the current API.
- Arbitrary statistics computed after grouping values. Grouped operations are already supported in DataFrames.jl and SplitApplyCombine.jl, with optimized implementations for reductions such as `sum` (in DataFrames). What is lacking is the ability to transform a vector of values into a vector of integers on which to group, but this can be fixed separately. That would also allow choosing between different grouping strategies (equal-width bins, quantiles, various algorithms such as Jenks...).

kirklong:
This all makes sense and I see your points here, and of course this does not have to be added to this package as I understand it's a relatively specific thing and you prefer to keep the API minimal...However I'm wondering if you might know a better place for it? 

In academia (I'm an astrophysics PhD student) this kind of thing is used *a lot* and I know a big barrier to entry to Julia for many of my colleagues is that what they consider "basic" tools like this are missing from Julia vs. Python. Anecdotally I've heard from several people in my department that this exact kind of problem (simple function that's a one line call in Python but in Julia requires you to write from scratch) is one of the main reasons they are "waiting for Julia to mature" before trying it out. It's true that obviously this was not very hard to implement this in Julia using pretty generic code (really mostly just the Julia Base) but it seems silly to force people to always write everything from scratch just because Julia is fast and it *can* be done if that makes sense? I've fixed my problem, sure, but I'm sure other scientists would also benefit from this kind of thing being importable from a common module without us all having to write our own versions from scratch -- that seems like a pretty big waste of time for operations like this that could be standardized. 

I love using Julia and want more people to join me, so am trying to think if there's a place where these kinds of simple, commonly used routines might go and welcome all of your suggestions as you are much more knowledgeable than me! Thanks for taking the time to respond and no worries about it not fitting into StatsBase as you envision it.

bkamins:
I think a modified version of this PR could go to OnlineStats.jl. This would additionally ensure that a streaming version of the functionality would be available (which I think would be useful for your use case in general). The only limitation would be that probably `median` would be problematic (as there is no O(1) on-line algorithm for median - I am not sure how important this statistics is for you).

On the other hand - if you wanted to follow your implementation (you could even start it in your package featuring only this function) then here are some suggestions:
* instead of hard-coding functions  as symbols, I think it is better to just pass functions "as is". This is a more extensible API with a benefit that the default path could still perform any aggregation that would be asked for (just slower)
* I think that it would be better to try thinking of something along the lines of `fit(BinnedStatistic, grouping_var, statistic_var, aggregation_function, bins)` (I am not saying it is best API, but it would seem more consistent with what StatsBase.jl offers)

Other than that I think that if you indeed feel that this function is quite often needed it is worth to implement it somewhere. In particular, indeed, such a custom function will be faster than more general functions that we currently have since such function can take advantage of special grouping structure you assume.

kirklong:
Thanks for these excellent suggestions @bkamins! I like both of your style comments and will incorporate them, and especially agree that there really is no need to hard-code the functions as symbols. I have started creating my own package just for "fun" (good learning experience) but will also ask over in OnlineStats.jl if they are interested in a modified version of the PR, but without the median functionality as you pointed out. I'll close this issue here -- thanks again all!

bkamins:
@kirklong - you can ping me in whatever repo you would end up submitting this and I can review.

nalimilan:
> In academia (I'm an astrophysics PhD student) this kind of thing is used _a lot_ and I know a big barrier to entry to Julia for many of my colleagues is that what they consider "basic" tools like this are missing from Julia vs. Python. Anecdotally I've heard from several people in my department that this exact kind of problem (simple function that's a one line call in Python but in Julia requires you to write from scratch) is one of the main reasons they are "waiting for Julia to mature" before trying it out. It's true that obviously this was not very hard to implement in Julia using pretty generic code (really mostly just the Julia Base) but it seems silly to force people to always write everything from scratch just because Julia is fast and it _can_ be done if that makes sense? I've fixed my problem, sure, but I'm sure other scientists would also benefit from this kind of thing being importable from a common module without us all having to write our own versions from scratch -- that seems like a pretty big waste of time for operations like this that could be standardized.

To be clear, I wasn't suggesting that people should write this code from scratch, rather that we should find simple but more generic ways of doing this operation.

Could you describe the use cases you see for such a function, apart from weighted histogram you mentioned in the description? It's difficult to know what people need in different disciplinary fields.

PR #767, opened 2022-02-19
Add tests for StatsAPI modeling functions

nalimilan:
Ensure that we don't accidentally stop exporting some functions and that
fallbacks defined in StatsAPI work.
Stop importing `params` and `params!` as they are neither used nor reexported

### DIFF
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -21,7 +21,7 @@ import Statistics: mean, mean!, var, varm, varm!, std, stdm, cov, covm,
                    cor, corm, cov2cor!, unscaled_covzm, quantile, sqrt!,
                    median, middle
 using StatsAPI: StatisticalModel, RegressionModel
-import StatsAPI: pairwise, pairwise!,
+import StatsAPI: pairwise, pairwise!, params, params!,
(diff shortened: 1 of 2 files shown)


bkamins (approved):

src/StatsBase.jl
```diff
@@ -21,7 +21,7 @@ import Statistics: mean, mean!, var, varm, varm!, std, stdm, cov, covm,
                    cor, corm, cov2cor!, unscaled_covzm, quantile, sqrt!,
                    median, middle
 using StatsAPI: StatisticalModel, RegressionModel
-import StatsAPI: pairwise, pairwise!, params, params!,
```
bkamins, line 24: why `params` is not tested?

nalimilan: It's not actually (and has never been) reexported, so if we tested something it would be that it's *not* defined. :-)

nalimilan: Woops, actually the fact that we don't export it doesn't mean it's fine to stop importing it. Packages probably rely on StatsBase defining it. It was added by https://github.com/JuliaStats/StatsBase.jl/pull/274. I've added a commit to fix this.

bkamins (approved):

Issue #766, opened 2022-02-18
Histogram dropping values when dealing with signed zero


mtanneau:
I encountered this issue through `UnicodePlots` (https://github.com/JuliaPlots/UnicodePlots.jl/issues/229), but I believe the root cause lies in `StatsBase`.

It appears that `histogram` will drop some entries when the left-end of the data is a _signed_ zero `-0.0`.
From the original issue
```julia
julia> histogram([0.0, 1.0])
              ┌                                        ┐ 
   [0.0, 0.5) ┤█████████████████████████████████████  1  
   [0.5, 1.0) ┤  0                                       
   [1.0, 1.5) ┤█████████████████████████████████████  1  
              └                                        ┘ 
                               Frequency

julia> histogram([-0.0, 1.0])
              ┌                                        ┐ 
   [0.0, 0.5) ┤  0                                         # where did that zero go?
   [0.5, 1.0) ┤  0                                       
   [1.0, 1.5) ┤█████████████████████████████████████  1  
              └                                        ┘ 
                               Frequency

julia> histogram([0.0, -0.0])
              ┌                                        ┐ 
   [0.0, 1.0) ┤█████████████████████████████████████  1    # only one value?
              └                                        ┘ 
                               Frequency
```

On the one hand, I can understand that `-0.0` would technically fall in a separate bin than `+0.0`, because of floating-point ordering.
However, it was surprising to see those entries dropped from the total count.

t-bltg:
Thanks for reporting. I would say at first glance that the issue lies in `StatsBase`:

```julia
julia> using StatsBase
julia> fit(Histogram, [-0.0, 1.0])
Histogram{Int64, 1, Tuple{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}}}
edges:
  0.0:0.5:1.5
weights: [0, 0, 1]
closed: left
isdensity: false
julia> fit(Histogram, [-0.0, 1.0], closed=:right)
Histogram{Int64, 1, Tuple{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}}}
edges:
  -0.5:0.5:1.0
weights: [1, 0, 1]
closed: right
isdensity: false
```

Can you please open an issue in https://github.com/JuliaStats/StatsBase.jl ?

t-bltg:
I think this will be more representative in `StatsBase`:

```julia
julia> using StatsBase
[...]

PR #768, opened 2022-02-20
Fix handling of -0.0 in histograms

nalimilan:
`searchsortedfirst` and `searchsortedlast` use `isless` for comparisons and therefore consider `-0.0` to be different from `0.0`. This means that these two values do not end up in the same bin when an edge is 0.
This does not make much sense statistically, but even worse is that when an extreme edge is 0, `-0.0` is not counted at all.

Fix this by replacing `-0.0` with `0.0` before the search.
Closes https://github.com/JuliaStats/StatsBase.jl/issues/766.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -194,11 +194,10 @@ mutable struct Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
         # We do not handle -0.0 in ranges correctly in `binindex` for performance
         # Constructing ranges starting or ending with -0.0 is very hard,
         # and ranges containing -0.0 elsewhere virtually impossible,
-        # so check the former just in case as it is cheap
+        # but check this just in case as it is cheap
         foreach(edges) do e
-            e isa AbstractRange &&
-                (isequal(first(e), -0.0) || isequal(last(e), -0.0)) &&
-                throw(ArgumentError("ranges starting or ending with -0.0 not allowed in edges"))
+            e isa AbstractRange && any(isequal(-0.0), e) &&
+                throw(ArgumentError("ranges containing -0.0 not allowed in edges"))
         end
         new{T,N,E}(edges,weights,closed,isdensity)
     end
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -191,6 +191,15 @@ mutable struct Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
         closed == :right || closed == :left || error("closed must :left or :right")
         isdensity && !(T <: AbstractFloat) && error("Density histogram must have float-type weights")
         _edges_nbins(edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
+        # We do not handle -0.0 in ranges correctly in `binindex` for performance
+        # Constructing ranges starting or ending with -0.0 is very hard,
+        # and ranges containing -0.0 elsewhere virtually impossible,
+        # so check the former just in case as it is cheap
+        foreach(edges) do e
+            e isa AbstractRange &&
+                (isequal(first(e), -0.0) || isequal(last(e), -0.0)) &&
+                throw(ArgumentError("ranges starting or ending with -0.0 not allowed in edges"))
+        end
         new{T,N,E}(edges,weights,closed,isdensity)
     end
 end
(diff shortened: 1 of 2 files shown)


nalimilan:
Cc: @Moelf, @oschulz

Moelf:
is `-0.0` and `0.0` the only edge case? I think so? the only other weird thing come to mind is `-NaN` and `NaN` but they can't be bin edge.

src/hist.jl
```diff
@@ -226,11 +226,17 @@ binindex(h::AbstractHistogram{T,1}, x::Real) where {T} = binindex(h, (x,))[1]
 binindex(h::Histogram{T,N}, xs::NTuple{N,Real}) where {T,N} =
     map((edge, x) -> _edge_binindex(edge, h.closed, x), h.edges, xs)
 
+_normalize_zero(x::AbstractFloat) = isequal(x, -0.0) ? oftype(x, 0.0) : x
```
Moelf, line 229:
```suggestion
_normalize_zero(x::AbstractFloat) = ifelse(isequal(x, -0.0), oftype(x, 0.0), x)
```

for the performance? also I hope it's automatically inlined.

src/hist.jl
```diff
@@ -226,11 +226,17 @@ binindex(h::AbstractHistogram{T,1}, x::Real) where {T} = binindex(h, (x,))[1]
...
+_normalize_zero(x::Any) = x
```
Moelf, line 230:
```suggestion
_normalize_zero(x) = x
```

Moelf:
I thought we need to
```julia
T = eltype(edge)
ifelse(isequal(x, -0.0), zero(T), x)
```

for type stability?

nalimilan:
`oftype(x, 0.0)` is equivalent to `zero(x)`, but yeah the latter is probably nicer.

Do you think the spirit of this PR is right? Have you encountered this problem in FHist.jl too?

src/hist.jl
nalimilan: AFAICT LLVM is smart enough that this gives exactly the same code.

nalimilan:
> is `-0.0` and `0.0` the only edge case? I think so? the only other weird thing come to mind is `-NaN` and `NaN` but they can't be bin edge.

Yes I think zero is the only tricky case apart from `NaN`. Currently `NaN`s in the data trigger an error if you don't specify edges manually, and if you do they are just ignored, which is dangerous and differs from what we do elsewhere. They are also accepted as edges, which seems absurd. Maybe we should disallow this in another PR.

nalimilan:
Unfortunately, passing `by=_normalize_zero` seems to incur a 30% performance penalty on a random `Vector{Float64}`. It would be better to check edges in the `Histogram` constructor, but given that the edges vector is passed by the caller, mutating it to replace `-0.0` with `0.0` could be problematic. Maybe we should just print a deprecation warning if edges contain `-0.0`, and later we will throw an error telling the caller to fix this. This is a relatively unlikely case I guess.

Moelf:
>Do you think the spirit of this PR is right?

yeah totally. For physics we don't run into 0.0 much and also we don't care losing a few events... (they are weighted to 0.0001) or something.

>Unfortunately, passing by=_normalize_zero seems to incur a 30% performance penalty on a random 

yeah I don't completely understand the performance model of `searchsorted*`, and also the way we have histogram implemented as recursive pushing and `map()` to find dimensions etc may also be a problem when everything is compiled together: I'd expect performance to be much better if we simply have a for loop with `searchsorted*` inlined together with the `by = ` in the loop-body.

oschulz:
Thanks for this!

nalimilan:
I've found a solution though it's unfortunately a bit complex. It turns out that the overhead of normalizing `-0.0` to `0.0` only appears when edges are a range, as passing `by` to `searchsortedfirst`/`searchsortedlast` forces using the fallback `AbstractVector` method. So I figured we can avoid normalizing zeros when edges are a range.

Luckily it's almost impossible to construct a range which includes `-0.0`. Even `-0.0:1.0` starts with `0.0` rather than `-0.0`. So we never build a range which contains `-0.0` when edges are omitted. Neither `range`, `x:y` nor `x:y:z` seem to allow creating a range containing `-0.0`: the only way I've found is to do `UnitRange(-0.0, 1.0)` or `LinRange(-1.0, -0.0, 2)`. This is so unlikely that I've added a check in the constructor to throw an error if we encounter such a case so that people can fix it or report it instead of silently getting incorrect results. For standard range types, we could automatically use a new range after replacing `-0.0` with `0.0`, but I'm not sure it's worth it.

src/hist.jl
```diff
@@ -191,6 +191,15 @@ mutable struct Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
         closed == :right || closed == :left || error("closed must :left or :right")
         isdensity && !(T <: AbstractFloat) && error("Density histogram must have float-type weights")
         _edges_nbins(edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
+        # We do not handle -0.0 in ranges correctly in `binindex` for performance
+        # Constructing ranges starting or ending with -0.0 is very hard,
+        # and ranges containing -0.0 elsewhere virtually impossible,
+        # so check the former just in case as it is cheap
+        foreach(edges) do e
+            e isa AbstractRange &&
+                (isequal(first(e), -0.0) || isequal(last(e), -0.0)) &&
```
bkamins, line 200: why not `any(isequal(-0.0), e)` as this would be a bit safer and cost is negligible I think?

nalimilan: Yeah, I guess in normal use the number of bins is much lower than the number of observations so checking for all of them has a negligible cost. I've pushed a commit to do that.

bkamins (approved):

Issue #771, opened 2022-02-27
Failed type assertion in `pairwise`


yha:
```
using HypothesisTests, StatsBase
function test(x)
	pairwise(eachindex(x); symmetric=true) do j,i
		pval = pvalue(SignedRankTest(x[i], x[j]))
		pval, rand()
	end
end

x = [rand(1000) for _=1:10]

test(x)
```
```
ERROR: TypeError: in typeassert, expected Matrix{<:Tuple{Union{Nothing, Float64, Int64}, Float64}}, got a value of type Matrix{Tuple{Real, Float64}}
Stacktrace:
 [1] _pairwise(#unused#::Val{:none}, f::var"#29#30"{Vector{Vector{Float64}}}, x::Base.OneTo{Int64}, y::Base.OneTo{Int64}, symmetric::Bool)
   @ StatsBase C:\Users\sternlab\.julia\packages\StatsBase\pJqvO\src\pairwise.jl:155
 [2] pairwise(f::Function, x::Base.OneTo{Int64}, y::Base.OneTo{Int64}; symmetric::Bool, skipmissing::Symbol)
   @ StatsBase C:\Users\sternlab\.julia\packages\StatsBase\pJqvO\src\pairwise.jl:288
 [3] test(x::Vector{Vector{Float64}})
   @ Main .\Untitled-9:3
 [4] top-level scope
   @ Untitled-9:11
```
Package versions:
```
  [09f84164] HypothesisTests v0.10.6
  [2913bbd2] StatsBase v0.33.16
```
(this doesn't happen if the `pairwise` call is in global scope rather than a function)

nalimilan:
Good catch. This is a very tricky case indeed. https://github.com/JuliaStats/StatsBase.jl/pull/772 fixes this. The fact that inference gives such an imprecise type in the first place could also be improved in HypothesisTests, see https://github.com/JuliaStats/HypothesisTests.jl/issues/263.

PR #772, opened 2022-03-04
Fix pairwise for type-unstable corner case function

nalimilan:
`promote_type` is not a completely correct way of computing an upper bound for the return eltype. Use the same strategy as `map` and `broadcast` in Base instead (https://github.com/JuliaLang/julia/pull/30485), but with `promote_eltype` rather than `promote_typejoin`. We can still use `typejoin_union_tuple` since promotion does not happen inside tuple types.

On Julia versions before 1.6 we would have to copy the full definition of `typejoin_union_tuple`, which is quite complex, so instead fall back to inferring eltype `Any` (inference wasn't so good anyway on these versions).

Fixes https://github.com/JuliaStats/StatsBase.jl/issues/771.

### DIFF
--- a/src/pairwise.jl
+++ b/src/pairwise.jl
@@ -129,23 +129,24 @@ if VERSION >= v"1.6.0-DEV"
     else
         using Base.Broadcast: typejoin_union_tuple
     end
-    # Identical to `Base.promote_typejoin` except that it uses `promote_type`
-    # instead of `typejoin`
-    function promote_type_union(::Type{T}) where T
-        if T === Union{}
-            return Union{}
-        elseif T isa UnionAll
-            return Any # TODO: compute more precise bounds
-        elseif T isa Union
-            return promote_type(promote_type_union(T.a), promote_type_union(T.b))
-        elseif T <: Tuple
-            return typejoin_union_tuple(T)
-        else
-            return T
-        end
-    end
 else
-    promote_type_union(::Type) = Any
+    typejoin_union_tuple(::Type) = Any
+end
+
+# Identical to `Base.promote_typejoin` except that it uses `promote_type`
+# instead of `typejoin` to combine members of `Union` types
+function promote_type_union(::Type{T}) where T
+    if T === Union{}
+        return Union{}
+    elseif T isa UnionAll
+        return Any # TODO: compute more precise bounds
+    elseif T isa Union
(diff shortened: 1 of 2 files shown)


src/pairwise.jl
```diff
@@ -122,6 +122,32 @@ function _pairwise!(f, dest::AbstractMatrix, x, y;
...
+        elseif T isa Union
+            return promote_type(promote_type_union(T.a), promote_type_union(T.b))
+        elseif T <: Tuple
+            return typejoin_union_tuple(T)
+        else
+            return T
+        end
+    end
+else
+    promote_type_union(::Type) = Any
```
bkamins, line 148: do we need to be that defensive? If passed type is not a union then I guess we can return it - right?

src/pairwise.jl
```diff
@@ -122,6 +122,32 @@ function _pairwise!(f, dest::AbstractMatrix, x, y;
...
+    if isdefined(Base, :typejoin_union_tuple)
+        using Base: typejoin_union_tuple
+    else
+        using Base.Broadcast: typejoin_union_tuple
+    end
+    # Identical to `Base.promote_typejoin` except that it uses `promote_type`
+    # instead of `typejoin`
+    function promote_type_union(::Type{T}) where T
+        if T === Union{}
+            return Union{}
```
bkamins, line 136: I would make sure we cover all cases with tests.

src/pairwise.jl
nalimilan: Good point. Actually the only case that would be hard to support is `Tuple` types. I've refactored the code to improve this.

src/pairwise.jl
```diff
@@ -122,6 +122,33 @@ function _pairwise!(f, dest::AbstractMatrix, x, y;
...
+    if T === Union{}
+        return Union{}
+    elseif T isa UnionAll
+        return Any # TODO: compute more precise bounds
+    elseif T isa Union
+        return promote_type(promote_type_union(T.a), promote_type_union(T.b))
+    elseif T <: Tuple
+        return typejoin_union_tuple(T)
+    else
+        return T
```
bkamins, line 148: Is line 148 covered in tests?

nalimilan: It's covered as `promote_type_union` calls itself recursively, but I've added two tests for clarity.

bkamins (approved):

Issue #775, opened 2022-03-22
Slow weighted sum/dot etc.


DNF2:
`dot(w, x)`, `sum(x, w)` etc. seem to use a slow fallback implementation of dot:
```
1.7.2> using StatsBase

1.7.2> x = rand(1000); w = weights(rand(length(x)));

1.7.2> x = rand(1000); w = weights(rand(1000));

1.7.2> @btime dot($w, $x)
  975.000 ns (0 allocations: 0 bytes)
244.6881117358608

1.7.2> @btime dot($w.values, $x)
  85.151 ns (0 allocations: 0 bytes)
244.68811173586084
```
The same goes for calling `sum(x, w)` or `wsum(x, w)`.

This is independent of the number of BLAS threads. 

This is a pretty dramatic performance difference. Should StatsBase call `BLAS.dot` or is this something that should be improved in the fallback `dot` routine in Base?

nalimilan:
Good catch. We need to use the `values` field instead of passing `AbstractWeights` objects directly. I'm not sure we should implement `dot` for `AbstractWeights` objects, as the number of methods to implement would be very large if we start trying to use the fastest methods everywhere. A more reasonable goal is to ensure good performance only for methods which are specifically designed to take weights.

See https://github.com/JuliaStats/StatsBase.jl/pull/778.

PR #778, opened 2022-03-26
Improve performance of weighted sum

nalimilan:
The current code is calling the `AbstractArray` matrix multiplication fallback, which is slower than BLAS.

Fixes https://github.com/JuliaStats/StatsBase.jl/issues/775.

```julia
julia> using BenchmarkTools

julia> using StatsBase

julia> x = rand(1000); w = weights(rand(1000));

# master
julia> @btime wsum(x, w);
  1.029 μs (1 allocation: 16 bytes)

# This PR
julia> @btime wsum(x, w);
  97.572 ns (1 allocation: 16 bytes)
```

Tests appear to cover all weights types and `AbstractVector` already.

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -388,7 +388,11 @@ for W in (AnalyticWeights, FrequencyWeights, ProbabilityWeights, Weights)
         wsum(v::AbstractArray, w::$W, dims::Colon) = transpose(w.values) * vec(v)
     end
 end
-wsum(v::AbstractArray, w::UnitWeights, dims::Colon) = sum(v)
+
+function wsum(A::AbstractArray, w::UnitWeights, dims::Colon)
+    length(A) != length(w) && throw(DimensionMismatch("Inconsistent array dimension."))
+    return sum(A)
+end
 
 ## wsum along dimension
 #
@@ -612,12 +616,6 @@ optionally over the dimension `dims`.
 """
 Base.sum(A::AbstractArray, w::AbstractWeights{<:Real}; dims::Union{Colon,Int}=:) =
     wsum(A, w, dims)
-
-function Base.sum(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
-    a = (dims === :) ? length(A) : size(A, dims)
-    a != length(w) && throw(DimensionMismatch("Inconsistent array dimension."))
-    return sum(A, dims=dims)
-end
 
 ##### Weighted means #####
 
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -382,6 +382,14 @@ Compute the weighted sum of an array `v` with weights `w`, optionally over the d
 """
 wsum(v::AbstractArray, w::AbstractVector, dims::Colon=:) = transpose(w) * vec(v)
 
+# Optimized methods (to ensure we use BLAS when possible)
+for W in (AnalyticWeights, FrequencyWeights, ProbabilityWeights, Weights)
+    @eval begin
+        wsum(v::AbstractArray, w::$W, dims::Colon) = transpose(w.values) * vec(v)
+    end
+end
+wsum(v::AbstractArray, w::UnitWeights, dims::Colon) = sum(v)
+
 ## wsum along dimension
 #
 #  Brief explanation of the algorithm:
--- a/test/weights.jl
+++ b/test/weights.jl
@@ -476,7 +476,7 @@ end
 @testset "Sum, mean, quantiles and variance for unit weights" begin
     wt = uweights(Float64, 3)
 
-    @test sum([1.0, 2.0, 3.0], wt) ≈ 6.0
+    @test sum([1.0, 2.0, 3.0], wt) ≈ wsum([1.0, 2.0, 3.0], wt) ≈ 6.0
     @test mean([1.0, 2.0, 3.0], wt) ≈ 2.0
 
     @test sum(a, wt, dims=1) ≈ sum(a, dims=1)


rofinn:
Seems like a reasonable extension in this case. Looks like there isn't a test for the `UnitWeights` dispatch.

rofinn (approved):

nalimilan:
AFAICT unit weights are tested here, right? https://github.com/JuliaStats/StatsBase.jl/blob/26947bc42ae68224655093c9eac17aba651bb044/test/weights.jl#L479

mschauer:
The CI error seems to be related to sampling?
```
* wsampling.jl ...
[172](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:172)
┌ Warning: Assignment to `a` in soft scope is ambiguous because a global variable by the same name exists: `a` will be treated as a new local. Disambiguate by using `local a` to suppress this warning or `global a` to assign to the existing global variable.
[173](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:173)
└ @ ~/work/StatsBase.jl/StatsBase.jl/test/wsampling.jl:47
[174](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:174)
ERROR: LoadError: MethodError: no method matching direct_sample!(::UnitRange{Int64}, ::Vector{Float64}, ::Matrix{Int64})
[175](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:175)
Closest candidates are:
[176](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:176)
  direct_sample!(::UnitRange, ::AbstractArray) at ~/work/StatsBase.jl/StatsBase.jl/src/sampling.jl:26
[177](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:177)
  direct_sample!(::AbstractRNG, ::AbstractArray, ::AbstractArray) at ~/work/StatsBase.jl/StatsBase.jl/src/sampling.jl:36
[178](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:178)
  direct_sample!(::AbstractArray, ::AbstractWeights, ::AbstractArray) at ~/work/StatsBase.jl/StatsBase.jl/src/sampling.jl:584
[179](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:179)
  ...
[180](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:180)
Stacktrace:
[181](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:181)
 [1] top-level scope
[182](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:182)
   @ ~/work/StatsBase.jl/StatsBase.jl/test/wsampling.jl:47
[183](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:183)
 [2] include(fname::String)
[184](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:184)
   @ Base.MainInclude ./client.jl:476
[185](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:185)
 [3] top-level scope
[186](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:186)
   @ ~/work/StatsBase.jl/StatsBase.jl/test/runtests.jl:34
[187](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:187)
 [4] include(fname::String)
[188](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:188)
   @ Base.MainInclude ./client.jl:476
[189](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:189)
 [5] top-level scope
[190](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:190)
   @ none:6
[191](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:191)
in expression starting at /home/runner/work/StatsBase.jl/StatsBase.jl/test/wsampling.jl:40
[192](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:192)
in expression starting at /home/runner/work/StatsBase.jl/StatsBase.jl/test/runtests.jl:31
[193](https://github.com/JuliaStats/StatsBase.jl/runs/5703799652?check_suite_focus=true#step:6:193)
ERROR: Package StatsBase errored during testing
```

nalimilan:
Yeah these also happen on master: https://github.com/JuliaStats/StatsBase.jl/actions/runs/2044798742

rofinn:
> AFAICT unit weights are tested here, right?

Hmm, codecov says your new dispatch for `wsum` at line 391 isn't getting hit though :/

rofinn:
Alright, I figured out why that method wasn't being hit. The specific line you posted calls a `sum` dispatch which never calls the `wsum` call.

https://github.com/JuliaStats/StatsBase.jl/blob/nl/wsum/src/weights.jl#L616

The `wsum` tests in that block all pass an integer and not `Colon`, so it hits a different dispatch.

https://github.com/JuliaStats/StatsBase.jl/blob/nl/wsum/src/weights.jl#L588

nalimilan:
Indeed. I've added a commit which hopefully simplifies the dispatch logic without breaking anything.

nalimilan:
Good to go?

src/weights.jl
```diff
@@ -382,6 +382,18 @@ Compute the weighted sum of an array `v` with weights `w`, optionally over the d
 """
 wsum(v::AbstractArray, w::AbstractVector, dims::Colon=:) = transpose(w) * vec(v)
 
+# Optimized methods (to ensure we use BLAS when possible)
+for W in (AnalyticWeights, FrequencyWeights, ProbabilityWeights, Weights)
+    @eval begin
+        wsum(v::AbstractArray, w::$W, dims::Colon) = transpose(w.values) * vec(v)
```
devmotion, line 388: Did you compare it with
```suggestion
        wsum(v::AbstractArray, w::$W, dims::Colon) = dot(w.values, v)
```
? I'm not sure if that would be faster or slower.

nalimilan: Long story. :-) https://github.com/JuliaStats/StatsBase.jl/pull/737

devmotion: Is this relevant here? The weights are real numbers and hence transpose and conjugate transpose are the same.

devmotion: Ah I didn't know that. One could still use `dot` for `AbstractArray{<:Number}` if it is more performant.

nalimilan: IIRC I benchmarked it for https://github.com/JuliaLang/Statistics.jl/pull/94 and it wasn't faster, otherwise I would have kept a specialized method for reals. As I noted there, these are equivalent anyway in common cases:
> Use `adjoint(y) * x` instead of `dot(y, x)`, as the former falls back to the latter for arrays of `Number`s but to `sum(uu*vv for (uu, vv) in zip(u, v))` for other types.

devmotion (approved):

PR #790, opened 2022-05-16
move to modern indexing style

Moelf:


Moelf:
check  #722 too

src/deviation.jl
```diff
@@ -9,13 +9,10 @@ Count the number of indices at which the elements of the arrays
...
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
+    length(a) == length(b) || throw(DimensionMismatch("Inconsistent lengths."))
     c = 0
-    for i = 1:n
-        @inbounds if a[i] == b[i]
-            c += 1
-        end
+    for i = eachindex(a, b)
+        c += (a[i] == b[i])
```
devmotion, line 15: We could just use
```suggestion
    for (ai, bi) in zip(a, b)
        c += (ai == bi)
```
?

src/deviation.jl
```diff
@@ -28,13 +25,10 @@ Count the number of indices at which the elements of the arrays
...
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
+    length(a) == length(b) || throw(DimensionMismatch("Inconsistent lengths."))
     c = 0
-    for i = 1:n
-        @inbounds if a[i] != b[i]
-            c += 1
-        end
+    for i = eachindex(a, b)
+        c += (a[i] != b[i])
```
devmotion, line 31:
```suggestion
    for (ai, bi) in zip(a, b)
        c += (ai != bi)
```

src/deviation.jl
```diff
@@ -47,11 +41,10 @@ Compute the squared L2 distance between two arrays: ``\\sum_{i=1}^n |a_i - b_i|^
...
 """
 function sqL2dist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
+    length(a) == length(b) || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
-    for i = 1:n
-        @inbounds r += abs2(a[i] - b[i])
+    for i = eachindex(a, b)
+        r += abs2(a[i] - b[i])
```
devmotion, line 47:
```suggestion
    for (ai, bi) in zip(a, b)
        r += abs2(ai - bi)
```

src/deviation.jl
```diff
@@ -75,11 +68,10 @@ Compute the L1 distance between two arrays: ``\\sum_{i=1}^n |a_i - b_i|``.
...
 """
 function L1dist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
+    length(a) == length(b) || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
-    for i = 1:n
-        @inbounds r += abs(a[i] - b[i])
+    for i = eachindex(a, b)
+        r += abs(a[i] - b[i])
```
devmotion, line 74:
```suggestion
    for (ai, bi) in zip(a, b)
        r += abs(ai - bi)
```

src/deviation.jl
```diff
@@ -94,11 +86,10 @@ two arrays: ``\\max_{i\\in1:n} |a_i - b_i|``.
...
 """
 function Linfdist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
+    length(a) == length(b) || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
-    for i = 1:n
-        @inbounds v = abs(a[i] - b[i])
+    for i = eachindex(a, b)
+        v = abs(a[i] - b[i])
```
devmotion, line 92:
```suggestion
    for (ai, bi) in zip(a, b)
        v = abs(ai - bi)
```

src/deviation.jl
```diff
@@ -118,9 +109,9 @@ Efficient equivalent of `sum(a*log(a/b)-a+b)`.
 function gkldiv(a::AbstractArray{T}, b::AbstractArray{T}) where T<:AbstractFloat
     n = length(a)
     r = 0.0
-    for i = 1:n
-        @inbounds ai = a[i]
-        @inbounds bi = b[i]
+    for i = eachindex(a, b)
+        ai = a[i]
+        bi = b[i]
```
devmotion, line 114:
```suggestion
    for (ai, bi) in zip(a, b)
```

devmotion (commented):
Generally, I think there's no reason to drop `@inbounds` if we use `eachindex` and hence ensure that indexing is correct. However, it seems here we could also just use `zip` without explicit `eachindex`?

As in #722, it would be good to add tests with OffsetArrays.

As an alternative, that would also allow to fix the promotion to `Float64` and allow us to use pairwise summation we could use e.g.
```julia
function sqL2dist(a::AbstractArray{<:Number}, b::AbstractArray{<:Number})
    length(a) == length(b) || throw(DimensionMismatch("length of inputs incompatible"))
    r = sum(Broadcast.instantiate(Broadcast.broadcasted(vec(a), vec(b)) do ai, bi
        return abs2(ai - bi)
    end))
    return r
end
```

src/deviation.jl
```diff
@@ -75,11 +68,10 @@ Compute the L1 distance between two arrays: ``\\sum_{i=1}^n |a_i - b_i|``.
 Efficient equivalent of `sum(abs, a - b)`.
 """
 function L1dist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
+    length(a) == length(b) || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
```
mcabbott, line 72: Shouldn't this check axes not length? Although `eachindex(a, b)` performs such a check, in which case probably better to delete it:
```
julia> eachindex(OffsetArray([1,2,3], 1), OffsetArray([1,2,3], 2))
ERROR: DimensionMismatch: all inputs to eachindex must have the same indices, got OffsetArrays.IdOffsetRange(values=2:4, indices=2:4) and OffsetArrays.IdOffsetRange(values=3:5, indices=3:5)
```
Also, why 0.0 not `zero(T)`? (And why restrict to matching eltypes, `T==T`?)

devmotion: > Shouldn't this check axes not length?

No, arrays don't have to have matching sizes, just the same length. Enforcing this would be breaking.

> Also, why 0.0 not zero(T)? (And why restrict to matching eltypes, T==T?)

That's ancient code. I think it would be good to keep the PR focused on indexing, then it's easier and probably faster to review, merge, and release.

mcabbott: I guess I was going by this:
```
Efficient equivalent of `sum(abs, a - b)`
```
I now notice zip actually enforces this anyway:
```
julia> using OffsetArrays

julia> collect(zip(OffsetArray([1,2,3], 1), OffsetArray([1,2,3], 2)))
ERROR: DimensionMismatch: dimensions must match: a has dims (OffsetArrays.IdOffsetRange(values=2:4, indices=2:4),), b has dims (OffsetArrays.IdOffsetRange(values=3:5, indices=3:5),), mismatch at 1
Stacktrace:
```

devmotion: `zip` doesn't enforce any axes (not even the `length` and hence the check is necessary):
```julia
julia> using OffsetArrays

julia> foreach(println, zip(OffsetArray([1,2,3], 1), OffsetArray([1,2,3], 2)))
(1, 1)
(2, 2)
(3, 3)

julia> foreach(println, zip(OffsetArray([1 2 3], 1, 1), OffsetArray([1,2,3], 2)))
(1, 1)
(2, 2)
(3, 3)

julia> foreach(println, zip(OffsetArray([1,2,3], 1), OffsetArray([1,2,3,4], 2)))
(1, 1)
(2, 2)
(3, 3)

julia> foreach(println, zip(OffsetArray([1 2 3], 1, 1), OffsetArray([1,2,3,4], 2)))
(1, 1)
(2, 2)
(3, 3)
```

mcabbott: Indeed, it seems like you can iterate it, even when `axes` would throw:
```
julia> axes(zip([1 2; 3 4], [1 2 3 4]))
ERROR: DimensionMismatch: dimensions must match: a has dims (Base.OneTo(2), Base.OneTo(2)), b has dims (Base.OneTo(1), Base.OneTo(4)), mismatch at 1
```

Moelf:
Compiler should I get inbounds when using each index

Moelf:
after some testing, `zip()` is as slow as `eachindex(a, b)` without `@inbounds`

I suggest keep using `eachindex(a, b)` but also add `@inbounds`

src/deviation.jl
mcabbott: Maybe this can be marked resolved.

Moelf:
on the master branch of Julia, the `@inbounds` is inferred as expected.

src/deviation.jl
```diff
@@ -9,13 +9,10 @@ Count the number of indices at which the elements of the arrays
...
 function counteq(a::AbstractArray, b::AbstractArray)
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
+    length(a) == length(b) || throw(DimensionMismatch("Inconsistent lengths."))
     c = 0
-    for i = 1:n
-        @inbounds if a[i] == b[i]
-            c += 1
-        end
+    @inbounds for i = eachindex(a, b)
```
devmotion, line 14: This is breaking, as e.g. the following example shows (similar to https://github.com/JuliaStats/StatsBase.jl/pull/790/files#r874144995 without OffsetArrays which was unsupported):
```julia
julia> a = rand(3, 5);

julia> b = view(rand(10, 7), 1:2:10, 1:2:6);

julia> s = 0.0;

julia> for i in eachindex(a, b)
           s += (a[i] != b[i])
       end
ERROR: DimensionMismatch("all inputs to eachindex must have the same axes, got (Base.OneTo(3), Base.OneTo(5)) and (Base.OneTo(5), Base.OneTo(3))")
Stacktrace:
 [1] throw_eachindex_mismatch_indices(::IndexCartesian, ::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, ::Vararg{Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}})
   @ Base ./abstractarray.jl:286
 [2] eachindex
   @ ./multidimensional.jl:372 [inlined]
 [3] eachindex(A::Matrix{Float64}, B::SubArray{Float64, 2, Matrix{Float64}, Tuple{StepRange{Int64, Int64}, StepRange{Int64, Int64}}, false})
   @ Base ./abstractarray.jl:328
 [4] top-level scope
   @ ./REPL[9]:1
```
So either we should use `for (ai, bi) in zip(a, b)` or `for (i, j) in zip(eachindex(a), eachindex(b))`. In any case it seems wrong to assume that `a` and `b` have compatible axes and we can use `eachindex(a, b)`.

Moelf: that's really unfortunate.... can we use `firstindex():lastindex()`? `zip` is really slow

devmotion: What exactly do you have in mind? In general, I guess they might be different for `a` and `b` (and I'm not sure if one may assume that the range yields only valid indices).

devmotion: Distances uses a check and otherwise falls back to `zip`: https://github.com/JuliaStats/Distances.jl/blob/91f51b543ea6c54936d3e6183acdf7da50bf1f9e/src/metrics.jl#L251

devmotion: Did you benchmark the `sum(Broadcast.instantiate(Broadcast.broadcasted((ai, bi) -> ai != bi, vec(a), vec(b))))` alternative as well?

mcabbott: Timing these, the Broadcast version is fast for big arrays, but slower on small ones. Especially when `vec` causes an allocation: 
```
julia> let x = rand(1000), y = rand(1000) 
         @btime sqL2dist($x, $y)
         @btime sqL2dist_bc($x, $y)
         @btime sum(abs2(ai - bi) for (ai, bi) in zip($x, $y))
       end;
  1.454 μs (0 allocations: 0 bytes)
  450.716 ns (0 allocations: 0 bytes)
  1.462 μs (0 allocations: 0 bytes)

julia> let x = rand(1,5), y = rand(5)
         @btime sqL2dist($x, $y)
         @btime sqL2dist_bc($x, $y)
         @btime sum(abs2(ai - bi) for (ai, bi) in zip($x, $y))
       end;
  7.132 ns (0 allocations: 0 bytes)
  57.334 ns (2 allocations: 80 bytes)
  8.133 ns (0 allocations: 0 bytes)
```

devmotion: The main advantage of the broadcast version is that it uses pairwise summation whereas the other approaches don't.

mcabbott: Without the reshape it's within error, I guess. Or you can mess with ReshapedArray, same speed:
```
julia> let x = rand(5,1), y = rand(5)
         @btime sqL2dist($x, $y)
         @btime sqL2dist_bc(Base.ReshapedArray($x, (5,), ()), $y)
         @btime sum(abs2(ai - bi) for (ai, bi) in zip($x, $y))
       end;
  7.132 ns (0 allocations: 0 bytes)
  9.676 ns (0 allocations: 0 bytes)
  8.133 ns (0 allocations: 0 bytes)
```
However, `vec` does not do the same as zip for offset arrays. So my guess is that this broadcast ought to apply only when the axes agree, surely the common case, and fall back to zip? Write your own `_mapreduce` with this non-standard behaviour for this file?

devmotion: Generally, I think the behaviour of `zip(a, b)` is what's implied by the existing code (the current code works eg also for the example above for which `eachindex(a, b)` fails) and what's most consistent with other functions in StatsBase (and similar packages such as Distances, see above).

mcabbott: I wish I was surprised that no strange shapes are tested. That, and clarifying the docstring, would be good to do at the same time, IMO.

src/deviation.jl
```diff
@@ -94,11 +86,10 @@ two arrays: ``\\max_{i\\in1:n} |a_i - b_i|``.
...
 """
 function Linfdist(a::AbstractArray{T}, b::AbstractArray{T}) where T<:Number
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Input dimension mismatch"))
+    length(a) == length(b) || throw(DimensionMismatch("Input dimension mismatch"))
     r = 0.0
-    for i = 1:n
-        @inbounds v = abs(a[i] - b[i])
+    for (ai, bi) = zip(a, b)
+        v = abs(a[i] - b[i])
```
KristofferC, line 92:
```suggestion
        v = abs(ai - bi)
```

devmotion:
Seems like the PR is mainly missing tests, e.g., with OffsetArrays? (Of course, there are other possible improvements discussed in some comments above but in my opinion they could go into separate PRs since this PR is already a clear improvement for arrays with non-standard indices.)

src/deviation.jl
```diff
@@ -9,13 +9,10 @@ Count the number of indices at which the elements of the arrays
...
 function counteq(a::AbstractArray, b::AbstractArray)
-    n = length(a)
-    length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
+    length(a) == length(b) || throw(DimensionMismatch("Inconsistent lengths."))
     c = 0
-    for i = 1:n
-        @inbounds if a[i] == b[i]
-            c += 1
-        end
+    for (ai, bi) = zip(a, b)
```
bkamins, line 14: Using `zip` here is incorrect given the contract of `counteq` (same comments below).

The point is that the contract states:

> Count the number of indices

and `zip` will ignore indices but instead lead to comparing values using their iteration order.

devmotion: IMO that's consistent with the current behaviour though, regardless of the docstring. The current implementation already only cares about whether the first, second, third, etc elements match but doesn't care about shapes or cartesian indices of the compared arrays.

devmotion: Another nice consequence of using `zip` is that it is consistent with how Distances handles input arrays.

nalimilan: I agree with @bkamins that the docstring implies that values at the same indices are compared. So we need to check that indices are the same (or that starting indices are equal in addition to the current check that length are equal). We can keep a tolerance when mixing vectors and matrices by only check that linear indices are equal to avoid breakage.

What do you mean about Distances? Doesn't it check that axes are the same or that at least linear indices are the same?

devmotion: Since the package does neither support nor test arrays with non-standard or non-linear indices, my interpretation is that the docstring just refers to iteration order but it was not intended to come up with a more general design decision. Maybe there's some information in the original PR.

nalimilan: Yes we're free to reinterpret and adapt the docstring as we deem appropriate. But when passing arrays with different axes, it would find it dangerous to silently discard indices. If people use `OffsetArray`s there must be a reason. That's also a safer approach as if people find it too inconvenient, we can relax this requirement and allow mismatched indices. OTC if we allow them now we won't be able to change this later.

bkamins: I think it is clear from the source codes that the original design did not take into account non-1-based indices. Then - naturally - docstrings do not reflect this case.

I think we need to make a general decision for the whole package:
- if we allow non-1-based indexing (I feel from the discussions that we want to allow them)
- if yes - how they should be treated (and this should be made consistent across all methods in the package)
- what is our position about mixing arrays of different dimensionalities (this point is related as higher-dimensional arrays usually support linear indexing, which is 1-based) and when we want to accept only vectors.

After these decisions are made and documented the respective PRs should be done to reflect them. Otherwise we risk a situation when different methods in the package will take different assumptions.

nalimilan: A simple solution could be to adopt the semantics of `broadcast`, as discussed above at https://github.com/JuliaStats/StatsBase.jl/pull/790#discussion_r874217136. This would avoid forcing users to understand yet another rule and it would be easier to implement for us (in practice we could use a different implementation under the hood if needed for performance).

ParadaCarleton:
@Moelf could we wrap up this PR? It looks almost ready to merge and I want to end all the negative press from StatsBase's frequent `@inbounds` errors...

(By the way, does `eachindex` still not automatically elide bounds checking?)

Moelf:
what's the desired action? do we want to use `zip` or `@inbounds`?

ParadaCarleton:
I'm happy with `@inbounds`.

Moelf:
oh, but then it's already fixed on master, closing now

Issue #667, opened 2021-03-04
StatsBase.fit(StatsBase.Histogram) produces wrong output


montyvesselinov:
```
StatsBase.fit(StatsBase.Histogram, [0.101, 0.101, 0.101])
```

produces:

```
StatsBase.Histogram{Int64,1,Tuple{StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}}
edges:
  1.101:2.0:3.101
weights: [0]
closed: left
isdensity: false
```

whiich is obviously wrong.

```
StatsBase.fit(StatsBase.Histogram, [10.1, 10.1, 10.1])
```

also produces wrong results. However,

```
StatsBase.fit(StatsBase.Histogram, [0.1, 0.1, 0.2])
StatsBase.fit(StatsBase.Histogram, [10., 10., 10.])
```

works ok

yurivish:
This might be the same issue as https://github.com/JuliaStats/StatsBase.jl/issues/616 which I came across last November.

PR #792, opened 2022-05-18
Replace Base.floatrange with simplified implementation

metab0t:
Base.floatrange has bugs and is not public API anymore
See https://github.com/JuliaLang/julia/issues/45336
Use a simplified implementation instead
Fixes https://github.com/JuliaStats/StatsBase.jl/issues/667 and https://github.com/JuliaStats/StatsBase.jl/issues/616

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -39,15 +39,6 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where 
     histrange(F(lo), F(hi), n, closed)
 end
 
-# Base.floatrange has bugs and is not public API anymore
-# See https://github.com/JuliaLang/julia/issues/45336
-# Use a simplified implementation instead
-function _floatrange(start::F, step::F, len::Integer, divisor::F) where F
-    start = Base.TwicePrecision{Float64}((start, divisor))
-    step = Base.TwicePrecision{Float64}((step, divisor))
-    StepRangeLen(start, step, len)
-end
-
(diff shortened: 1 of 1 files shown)


getzze:
Thanks for find a workaround for this worrying bug!
`Base.TwicePrecision` is also not public API, so it would be better not to use it either, but instead to rework the `histrange` to simplify it in the first place. What do you think?

ararslan:
It's true that `TwicePrecision` is technically not public but there are a lot of packages that use it (and have done so for a long time). I think it's okay to use it here.

src/hist.jl
```diff
@@ -96,7 +105,7 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
             len += one(F)
         end
     end
-    Base.floatrange(start,step,Int(len),divisor)
+    _floatrange(start,step,Int(len),divisor)
```
ararslan, line 108: Since this is only used in one place, we could opt to inline the definition:
```suggestion
    StepRangeLen(Base.TwicePrecision{Float64}((start, divisor)),
                 Base.TwicePrecision{Float64}((stop, divisor)),
                 Int(len))
```

metab0t:
> It's true that `TwicePrecision` is technically not public but there are a lot of packages that use it (and have done so for a long time). I think it's okay to use it here.

I agree. Depending on `TwicePrecision` is more reliable because of its clear semantics and IIRC it is designed to use with `StepRangeLen`.

ararslan (approved):
The failures on Julia nightly are unrelated. Thanks for the contribution, much appreciated!

PR #799, opened 2022-06-10
Update moments.jl

yangcht:
the current function `mean_and_var` and `mean_and_std` does not allow tuple type for the kwarg `dim`, as it is `dim::Int`. It should be changed to `dim::Any` to allow tuple type.

### DIFF
--- a/src/moments.jl
+++ b/src/moments.jl
@@ -132,25 +132,25 @@ function mean_and_std(x::RealArray, w::AbstractWeights; corrected::DepBool=nothi
 end
 
 
-function mean_and_var(x::RealArray, dim::Int; corrected::Bool=true)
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks! Could you add tests for non-integer values?

BTW, if you feel like it, we should really make `dims` a keyword argument and deprecate the positional one.

yangcht:
Hi! I am a bit unfamiliar with the test code. Could you help with that? Thanks!

yangcht:
any updates?

ararslan:
For the tests, you'd just want to find where those functions are used, copy one of the individual `@test`s, and modify it to exercise the functionality added here.

Issue #802, opened 2022-06-18
Two miss type settings in signalcorr.jl


phasetr:
I found two miss type settings in [L258](https://github.com/JuliaStats/StatsBase.jl/blob/6768a16b0169fcc8a7adc9098d9f994a322b1b36/src/signalcorr.jl#L258) and [L409](https://github.com/JuliaStats/StatsBase.jl/blob/6768a16b0169fcc8a7adc9098d9f994a322b1b36/src/signalcorr.jl#L409).
Here we have `zy::Vector{T}`, but it should be `zy::Vector{S}`.

PR #805, opened 2022-06-20
Fix vector eltype in `crosscov`

nalimilan:
Introduced by https://github.com/JuliaStats/StatsBase.jl/pull/567. Fixes https://github.com/JuliaStats/StatsBase.jl/issues/802.

It doesn't seem easy to find a case where this would make a difference visible from the outside.


nalimilan:
Ah maybe something like this? On current release.
```julia
julia> crosscov(Float16[1, 10], [34566.5345, 3466.4566])
3-element Vector{Float64}:
  Inf
 -Inf
  Inf

julia> crosscov([34566.5345, 3466.4566], Float16[1, 10])
3-element Vector{Float64}:
  34987.5876375
 -69975.17527500002
  34987.58763750001

```

devmotion (approved):

PR #808, opened 2022-06-23
UnitWeights don't have `values` to return from `dataids`

rofinn:
Current fallback can cause errors like this:

```
  type UnitWeights has no field values
  Stacktrace:
    [1] getproperty(x::StatsBase.UnitWeights{Int64}, f::Symbol)
      @ Base ./Base.jl:33
    [2] dataids(wv::StatsBase.UnitWeights{Int64})
      @ StatsBase ~/.julia/packages/StatsBase/n494Y/src/weights.jl:26
    [3] dataids(A::SubArray{Int64, 1, StatsBase.UnitWeights{Int64}, Tuple{Vector{Int64}}, false})
    ```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.33.17"
+version = "0.33.18"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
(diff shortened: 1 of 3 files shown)


nalimilan:
Thanks! `UnitWeights` is indeed always special. Could you add a test?

rofinn:
The issue is that it’s hard to reproduce the convert path without making a bunch of assumptions about Julia’s type inference. I could add an explicit test, but that doesn’t seem very useful?

nalimilan:
A basic test is better than nothing. Anyway we shouldn't add lines that are not covered by tests, to ensure that at least they run without errors.

You could also call `Base.mightalias` (as I did at https://github.com/JuliaStats/StatsBase.jl/pull/793), as IIRC it calls `Base.dataids`.

nalimilan (approved):
Can you tag a new release?

PR #814, opened 2022-07-14
Throw DomainError when sampling with NaN/Inf weights

ruairidhs:


src/sampling.jl
```diff
@@ -551,6 +551,8 @@ Optionally specify a random number generator `rng` as the first argument
 """
 function sample(rng::AbstractRNG, wv::AbstractWeights)
     t = rand(rng) * sum(wv)
+    isfinite(sum(wv)) ||
```
bkamins, line 554: cache `sum(wv)` to improve performance.

src/sampling.jl
```diff
@@ -904,6 +906,8 @@ function sample!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::Abs
         throw(ArgumentError("output array a must not share memory with weights array wv"))
     1 == firstindex(a) == firstindex(wv) == firstindex(x) ||
         throw(ArgumentError("non 1-based arrays are not supported"))
+    isfinite(sum(wv)) ||
```
bkamins, line 909: check if `sum(wv)` can be cached (I have not read the logic enough to do it; maybe we need to pass the sum to internal functions).

bkamins (changes requested):
Looks good, except for performance. I would classify current behavior as a bug.

src/sampling.jl
nalimilan: `AbstractWeights` objects always return the cached sum when `sum` is called.

bkamins: Ah - then it is OK. This should be documented then. It is not now.

bkamins (approved):

src/sampling.jl
bkamins: As an other thought. Maybe then we should not make this change in this PR, but instead require in types implementing `AbstractWeights` constructor that they do this check instead? It would "fix the problem at its root", and we would not have to add this check in every function that uses `AbstractWeights`?

bkamins (changes requested):
Removing approval, as I came to the conclusion that it is better to make this fix in types implementing `AbstractWeights`.

nalimilan:
At https://github.com/JuliaStats/StatsBase.jl/pull/678#issuecomment-814706485 I noted that I wasn't sure whether they can't be legitimate applications where NaN or Inf weights should be allowed, or at least may currently be used without any problems (e.g. if you only work on a subset of the weights vector). This PR is less breaking so it could be good to do it as it fixes a bug. Then we could soft deprecate NaN and Inf weights, without throwing an error right now.

bkamins:
See what I think in https://github.com/JuliaStats/StatsBase.jl/pull/678#issuecomment-1196928410 (in general I think allowing `NaN`/`Inf` is a bug and does not require deprecation period)

bkamins:
> Checking that their sum is finite is probably enough.

I agree that this is what we should do (i.e. even if finite values are passed, but they add up to an infinite value I would throw an error)

ruairidhs:
> Fair enough. Nobody complained on Slack so maybe this corner case isn't used in practice.
> 
> @ruairidhs Sorry for the change in plans, but would you be willing to update the PR so that it throws an error when passing `NaN` or infinite values to `AbstractWeights` constructors and in their `setindex!` method? Checking that their sum is finite is probably enough.

Yep I can do this

ruairidhs:
I added a check when constructing AbstractWeights and also for setindex!.
Let me know what you think, happy to make changes

src/weights.jl
```diff
@@ -41,6 +46,7 @@ end
 Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values), sum(wv))
 
 @propagate_inbounds function Base.setindex!(wv::AbstractWeights, v::Real, i::Int)
+    isfinite(v) || throw(ArgumentError("weights cannot contain Inf or NaN values"))
     s = v - wv[i]
     wv.values[i] = v
     wv.sum += s
```
bkamins, line 52: technically `wv.sum` could become infinite here even if `v` is finite, so another check here might be added just for safety.

ruairidhs: ok makes sense, i swapped it to check the sum

bkamins: I meant to keep the other check also. The reason is that then we fail fast in a more common case without invalidating `wv`.

ruairidhs: Yeah good point to fail before invalidating wv. Is it worth having two checks for each new value? Could do something like below:
```suggestion
    s = v - wv[i]
    new_sum = wv.sum + s
    isfinite(new_sum) || throw(ArgumentError("weights cannot contain Inf or NaN values"))
    wv.values[i] = v
    wv.sum = new_sum
```

bkamins: right - this would be better

bkamins (approved):
Looks good

nalimilan:
Thanks!

Issue #444, opened 2019-01-09
`cov2cor` why must the standard deviations be given as argument?


scheidan:
Why does the user need to pass the standard deviations `s` as argument to `cov2cor`? This information is already contained in the covariance matrix. For example one could define 
```Julia
cov2cor(C) = cov2cor(C, sqrt.(diag(C))
``` 
.Additionally it is unclear, what is calculated if `s != sqrt.(diag(C)`.

See also: https://discourse.julialang.org/t/wrong-cov2cor/19423

nalimilan:
@rofinn Any reason why you required specifying the second argument when you added this function (at https://github.com/JuliaStats/StatsBase.jl/pull/261)?

rofinn:
> Any reason why you required specifying the second argument when you added this function (at #261)?
Because `cov2cor` is currently just a minimal copying version [`Statistics.cov2cor!`](https://github.com/JuliaLang/julia/blob/master/stdlib/Statistics/src/Statistics.jl#L513). We should probably add another method for `cov2cor` and `cov2cor!` that defaults to passing `sqrt.(diag(C))` along as @scheidan suggests.

> Additionally it is unclear, what is calculated if s != sqrt.(diag(C).

Some folks might want to pass in a `std` for more recent observations in a timeseries dataset used to construct `C`. The idea being that folks could use a covariance matrix from a larger timeseries lookback and normalize it using only the more recent `std`.

tpapp:
> Some folks might want to pass in a std for more recent observations in a timeseries dataset used to construct C. The idea being that folks could use a covariance matrix from a larger timeseries lookback and normalize it using only the more recent std.

I think that is a somewhat of a specialized use case, and is just confusing to include it in `cov2cor`.

It may be better to just make `cov2cor` essentially something like
```julia
function cc(C)
    a = inv.(sqrt.(diag(C)))
    C .* a .* permutedims(a)
end
```
possibly optimized for special matrices when applicable.

Incidentally, the current docstring of `StatsBase.cov2cor` refers to 

> `StatsBase.cov2cor!` for an in-place version.

but there is no such thing; it uses `Statistics.cov2cor!` but that is neither exported nor documented, so I guess it is for internal use.

ArnoStrouwen:
This now seems to work for the out-of-place version, but not for the in-place version.
[...]

PR #815, opened 2022-07-14
Let var, std & cor take a CovarianceEstimator

### LINKED ISSUE #734: Let var, std, cor take a CovarianceEstimator
Currently `StatsBase.jl` defines `CovarianceEstimator` [here](https://github.com/JuliaStats/StatsBase.jl/blob/179c533f8ea8c573bac295d04cdbb66f4d3bf9c4/src/cov.jl#L158), and defines methods of `cov` that take one as the first argument.

I think it would be convenient to allow similar dispatches for related methods. These could have fairly simple default implementations, something like the following (untested) sketch:

```julia
var(ce::CovarianceEstimator, x::AbstractVector; kwargs...) = cov(ce, x; kwargs...)

std(ce::CovarianceEstimator, args...; kwargs...) = sqrt(var(ce, args...; kwargs...))

cov2cor!(x::AbstractMatrix) = cov2cor!(x, sqrt.(diag(x)))

cor(ce::CovarianceEstimator, x::AbstractVector, y::AbstractVector) = cov2cor!(cov(ce, x, y))
cor(ce::CovarianceEstimator, X::AbstractMatrix; kwargs...) = cov2cor!(cov(ce, X; kwargs...))
cor(ce::CovarianceEstimator, X::AbstractMatrix, w::AbstractWeights; kwargs...) = cov2cor!(cov(ce, X, w; kwargs...))
```

Any thoughts or objections to this? I'm happy to make a PR for more concrete discussion if this seems reasonable :)

tpgillam:
Closes #734

### DIFF
--- a/src/cov.jl
+++ b/src/cov.jl
@@ -50,7 +50,7 @@ function scattermat end
 
 
 """
-    cov(X, w::AbstractWeights, vardim=1; mean=nothing,  corrected=false)
+    cov(X, w::AbstractWeights, vardim=1; mean=nothing, corrected=false)
 
 Compute the weighted covariance matrix. Similar to `var` and `std` the biased covariance
 matrix (`corrected=false`) is computed by multiplying `scattermat(X, w)` by
@@ -155,6 +155,15 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
 end
 
 """
+    cov2cor!(C)
+
+Compute the correlation matrix from the covariance matrix `C`, in-place.
+
+The leading diagonal is used to determine the standard deviations by which to normalise.
+"""
+cov2cor!(C::AbstractMatrix) = cov2cor!(C, sqrt.(diag(C)))
+
+"""
     CovarianceEstimator
 
 Abstract type for covariance estimators.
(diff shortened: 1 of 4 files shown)


src/cov.jl
```diff
@@ -154,6 +154,15 @@ function cor2cov!(C::AbstractMatrix, s::AbstractArray)
...
 end
 
+"""
+    cov2cor!(C)
+
+Compute the correlation matrix from the covariance matrix `C`, in-place.
+
+The leading diagonal is used to determine the standard deviations by which to normalise.
+"""
+cov2cor!(C::AbstractMatrix) = cov2cor!(C, sqrt.(diag(C)))
```
rofinn, line 164: This should probably extend the existing `Statistics.cov2cor!`, but that'd be type piracy. Maybe we can leave this out or call it `_cov2cor!` here and add it as a method to the Statistics.jl stdlib?

rofinn (approved):
Seems mostly fine. Conflict with `Statistics.cov2cor!` is probably worth considering though.

src/cov.jl
tpgillam: Yep good point! There's some relevant discussion here it seems: https://github.com/JuliaStats/StatsBase.jl/issues/444

For now I'll just rename to be a private method, to avoid getting bogged down in this issue.

tpgillam: (done)

src/cov.jl
```diff
@@ -198,6 +207,60 @@ cov(ce::CovarianceEstimator, X::AbstractMatrix; mean=nothing, dims::Int=1) =
...
+* a precalculated mean in which case it is subtracted from the data `X`.
+  Assuming `size(X)` is `(N,M)`, `mean` can either be:
+  * when `dims=1`, an `AbstractMatrix` of size `(1,M)`,
+  * when `dims=2`, an `AbstractVector` of length `N` or an `AbstractMatrix`
+    of size `(N,1)`.
+"""
+function cor(ce::CovarianceEstimator, X::AbstractMatrix; kwargs...)
+    return _cov2cor!(cov(ce, X; kwargs...))
+end
+function cor(ce::CovarianceEstimator, X::AbstractMatrix, w::AbstractWeights; kwargs...)
```
nalimilan, line 260: This method isn't tested at all. Could you make a PR to add a test?

tpgillam: yep - see #820

PR #818, opened 2022-07-27
dims in mean_and_var and mean_and_std as tuples

yangcht:
Change dim::Int & dims::Int to dim::Any & dims::Any to allow dims/dim to be tuples.
(sorry, the previous pull request https://github.com/JuliaStats/StatsBase.jl/pull/799 was deleted by mistake)


yangcht:
update to a new PR in https://github.com/JuliaStats/StatsBase.jl/pull/819

PR #819, opened 2022-07-27
allow dim/dims to be tuple

yangcht:
Change `dim::Int` & `dims::Int` to `dim::Any` & `dims::Any` to allow dims/dim to be tuples, fixed for the function `var`, `mean_and_std` and `mean_and_var`.


yangcht:
close due to some mistake

Issue #646, opened 2021-01-25
The majority of sampling methods are unsafe and incorrect in the presence of offset axes


yurivish:
These methods are documented [here](https://juliastats.org/StatsBase.jl/stable/sampling/) and all contain unsafe uses of `@inbounds`. Not all sampling methods suffer from this issue, but a number of them do. When I filed my [previous issue](https://github.com/JuliaStats/StatsBase.jl/issues/645) about `knuths_sample!` I had not realized how widespread this problem was. I believe the list below is the full list of sampling methods in `sampling.jl` with this problem (together with Fisher-Yates and Knuth's).

```julia
julia> StatsBase.direct_sample!(OffsetArray(collect(1:11), -5:5), zeros(11))
11-element Array{Float64,1}:
 10.0
  8.0
  0.0
  9.0
 11.0
  4.412732912e9
  7.0
  7.0
  7.0
  4.424086144e9
  4.412732912e9
```

```julia
julia> StatsBase.naive_wsample_norep!(OffsetArray(collect(1:11), -5:5), Weights(ones(11)), zeros(11))
11-element Array{Float64,1}:
     11.0
      4.557992976e9
      7.0
      0.0
 559108.0
      5.0
      8.0
     10.0
      9.0
     11.0
      4.424077344e9
```

```julia
julia> StatsBase.efraimidis_a_wsample_norep!(OffsetArray(collect(1:11), -5:5), Weights(ones(11)), zeros(11))
11-element Array{Float64,1}:
      0.0
     10.0
      9.0
     11.0
      4.557992976e9
      7.0
 559108.0
      0.0
      4.975392768e9
     11.0
      8.0

julia> StatsBase.efraimidis_ares_wsample_norep!(OffsetArray(collect(1:11), -5:5), Weights(ones(11)), zeros(11))
11-element Array{Float64,1}:
      7.0
      0.0
     11.0
     28.0
     10.0
      4.399872352e9
 559108.0
     11.0
      4.557992976e9
      9.0
      8.0

julia> StatsBase.efraimidis_aexpj_wsample_norep!(OffsetArray(collect(1:11), -5:5), Weights(ones(11)), zeros(11))
11-element Array{Float64,1}:
      4.3920316e9
 559108.0
      0.0
     11.0
     10.0
      7.0
      9.0
     11.0
      4.417712592e9
      8.0
      4.557992976e9
```

nalimilan:
[...]

PR #833, opened 2022-09-03
Throw an error for `OffsetArray`s and aliased arrays when sampling

nalimilan:
Add checks to all sampling methods, as they are mentioned in the manual even if not exported. Remove redundant check for aliasing from main method as it is potentially costly.
Also fix existing check (introduced by https://github.com/JuliaStats/StatsBase.jl/pull/793), as it confused the input and output array (the order of argument is not standard).

Fixes https://github.com/JuliaStats/StatsBase.jl/issues/646.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -739,7 +739,7 @@ Noting `k=length(x)` and `n=length(a)`, this algorithm consumes ``O(k)`` random 
 and has overall time complexity ``O(n k)``.
 """
 function naive_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
-                              wv::AbstractWeights, x::AbstractArray)*
+                              wv::AbstractWeights, x::AbstractArray)
     Base.mightalias(a, x) &&
         throw(ArgumentError("output array x must not share memory with input array a"))
     Base.mightalias(x, wv) &&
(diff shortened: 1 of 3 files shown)


src/sampling.jl
```diff
@@ -693,7 +739,13 @@ Noting `k=length(x)` and `n=length(a)`, this algorithm consumes ``O(k)`` random
 and has overall time complexity ``O(n k)``.
 """
 function naive_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
-                              wv::AbstractWeights, x::AbstractArray)
+                              wv::AbstractWeights, x::AbstractArray)*
```
bkamins, line 742:
```suggestion
                              wv::AbstractWeights, x::AbstractArray)
```

bkamins:
> Remove redundant check for aliasing from main method as it is potentially costly.

I have not checked that indeed the check is 100% redundant now, but I assume you checked.

bkamins (approved):
Looks good apart from two comments.

src/sampling.jl
nalimilan: Wow. Any idea why this doesn't trigger an error? :-/

nalimilan:
> I have not checked that indeed the check is 100% redundant now, but I assume you checked.

The tests are indeed supposed to cover this, but it's always good to use a second pair of eyes.

src/sampling.jl
bkamins: I think it is the same as if you have written:
```
function f()
    *
    some other code
end
```
And the operation just references the `*` function without calling it.

bkamins:
> but it's always good to use a second pair of eyes.

I have checked it and it seems it is covered.

nalimilan:
Thanks!

PR #834, opened 2022-09-03
Error when negative weights or zero sum are used when sampling

nalimilan:
These can give misleading results. Checking the sum is cheap, but checking for negative weights is relatively costly. Therefore, compute this information the first time it is requested, and store it in the weights vector like the sum.

`efraimidis_ares_wsample_norep!` and `efraimidis_aexpj_wsample_norep!` already checked these, but throwing different exception types. Harmonize exceptions across algorithms as they can all be called by `sample`.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -584,7 +584,10 @@ Optionally specify a random number generator `rng` as the first argument
 function sample(rng::AbstractRNG, wv::AbstractWeights)
     1 == firstindex(wv) ||
         throw(ArgumentError("non 1-based arrays are not supported"))
-    t = rand(rng) * sum(wv)
+    all(>=(0), wv) || throw(ArgumentError("negative weights are not allowed"))
+    s = sum(wv)
+    s > 0 || throw(ArgumentError("sum of weights must be greater than 0"))
+    t = rand(rng) * s
     n = length(wv)
     i = 1
     cw = wv[1]
@@ -621,6 +624,8 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray,
         throw(ArgumentError("non 1-based arrays are not supported"))
     n = length(a)
     length(wv) == n || throw(DimensionMismatch("Inconsistent lengths."))
+    all(>=(0), wv) || throw(ArgumentError("negative weights are not allowed"))
+    sum(wv) > 0 || throw(ArgumentError("sum of weights must be greater than 0"))
     for i = 1:length(x)
         x[i] = a[sample(rng, wv)]
     end
@@ -710,6 +715,8 @@ function alias_sample!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, 
         throw(ArgumentError("non 1-based arrays are not supported"))
     n = length(a)
     length(wv) == n || throw(DimensionMismatch("Inconsistent lengths."))
+    all(>=(0), wv) || throw(ArgumentError("negative weights are not allowed"))
+    sum(wv) > 0 || throw(ArgumentError("sum of weights must be greater than 0"))
 
     # create alias table
     ap = Vector{Float64}(undef, n)
@@ -749,6 +756,8 @@ function naive_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
     n = length(a)
     length(wv) == n || throw(DimensionMismatch("Inconsistent lengths."))
     k = length(x)
+    all(>=(0), wv) || throw(ArgumentError("negative weights are not allowed"))
+    sum(wv) > 0 || throw(ArgumentError("sum of weights must be greater than 0"))
 
     w = Vector{Float64}(undef, n)
     copyto!(w, wv)
@@ -795,6 +804,8 @@ function efraimidis_a_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
     n = length(a)
     length(wv) == n || throw(DimensionMismatch("a and wv must be of same length (got $n and $(length(wv)))."))
     k = length(x)
+    all(>=(0), wv) || throw(ArgumentError("negative weights are not allowed"))
+    sum(wv) > 0 || throw(ArgumentError("sum of weights must be greater than 0"))
 
     # calculate keys for all items
     keys = randexp(rng, n)
@@ -845,14 +856,14 @@ function efraimidis_ares_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
     @inbounds for _s in 1:n
         s = _s
         w = wv.values[s]
-        w < 0 && error("Negative weight found in weight vector at index $s")
+        w < 0 && throw(ArgumentError("Negative weight found in weight vector at index $s"))
         if w > 0
             i += 1
             pq[i] = (w/randexp(rng) => s)
         end
         i >= k && break
     end
-    i < k && throw(DimensionMismatch("wv must have at least $k strictly positive entries (got $i)"))
+    i < k && throw(ArgumentError("wv must have at least $k strictly positive entries (got $i)"))
     heapify!(pq)
 
     # set threshold
@@ -860,7 +871,7 @@ function efraimidis_ares_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
 
     @inbounds for i in s+1:n
         w = wv.values[i]
-        w < 0 && error("Negative weight found in weight vector at index $i")
+        w < 0 && throw(ArgumentError("Negative weight found in weight vector at index $i"))
         w > 0 || continue
         key = w/randexp(rng)
 
@@ -918,14 +929,14 @@ function efraimidis_aexpj_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
     @inbounds for _s in 1:n
         s = _s
         w = wv.values[s]
-        w < 0 && error("Negative weight found in weight vector at index $s")
+        w < 0 && throw(ArgumentError("Negative weight found in weight vector at index $s"))
         if w > 0
             i += 1
             pq[i] = (w/randexp(rng) => s)
         end
         i >= k && break
     end
-    i < k && throw(DimensionMismatch("wv must have at least $k strictly positive entries (got $i)"))
+    i < k && throw(ArgumentError("wv must have at least $k strictly positive entries (got $i)"))
     heapify!(pq)
 
     # set threshold
@@ -934,7 +945,7 @@ function efraimidis_aexpj_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
 
     @inbounds for i in s+1:n
         w = wv.values[i]
-        w < 0 && error("Negative weight found in weight vector at index $i")
+        w < 0 && throw(ArgumentError("Negative weight found in weight vector at index $i"))
         w > 0 || continue
         X -= w
         X <= 0 || continue
@@ -991,7 +1002,7 @@ function sample!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, x::Abs
             end
         end
     else
-        k <= n || error("Cannot draw $k samples from $n samples without replacement.")
+        k <= n || throw(ArgumentError("Cannot draw $k samples from $n samples without replacement."))
         efraimidis_aexpj_wsample_norep!(rng, a, wv, x; ordered=ordered)
     end
     return x
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -5,20 +5,24 @@ abstract type AbstractWeights{S<:Real, T<:Real, V<:AbstractVector{T}} <: Abstrac
     @weights name
 
 Generates a new generic weight type with specified `name`, which subtypes `AbstractWeights`
-and stores the `values` (`V<:RealVector`) and `sum` (`S<:Real`).
+and stores the `values` (`V<:RealVector`), the pre-computed `sum` (`S<:Real`) and
+whether all values are `positive`.
 """
 macro weights(name)
     return quote
         mutable struct $name{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractWeights{S, T, V}
             values::V
             sum::S
-            function $(esc(name)){S, T, V}(values, sum) where {S<:Real, T<:Real, V<:AbstractVector{T}}
+            positive::Union{Bool, Missing}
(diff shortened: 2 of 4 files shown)


src/weights.jl
```diff
@@ -53,9 +57,34 @@ Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values),
     isfinite(sum) || throw(ArgumentError("weights cannot contain Inf or NaN values"))
     wv.values[i] = v
     wv.sum = sum
+    wv.positive = missing
     v
 end
 
+function Base.all(f::Base.Fix2{typeof(>=)}, wv::AbstractWeights)
```
nalimilan, line 64: This may be overkill, but unfortunately that's the standard way of checking whether all entries in a vector in Julia, so that's the only solution if we want external code to be able to use this feature, without exporting a new `ispositive` function. One advantage of defining this is that if we rework the API to take a `weights` keyword argument, sampling functions will be able to allow any `AbstractArray` and code will automatically work.

src/weights.jl
```diff
@@ -5,20 +5,24 @@ abstract type AbstractWeights{S<:Real, T<:Real, V<:AbstractVector{T}} <: Abstrac
...
+and stores the `values` (`V<:RealVector`), the pre-computed `sum` (`S<:Real`) and
+whether all values are `positive`.
 """
 macro weights(name)
     return quote
         mutable struct $name{S<:Real, T<:Real, V<:AbstractVector{T}} <: AbstractWeights{S, T, V}
             values::V
             sum::S
-            function $(esc(name)){S, T, V}(values, sum) where {S<:Real, T<:Real, V<:AbstractVector{T}}
+            positive::Union{Bool, Missing}
```
devmotion, line 16: It seems it's actually used to store whether all elements are non-negative rather than positive?

src/weights.jl
```diff
@@ -53,9 +57,34 @@ Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values),
     isfinite(sum) || throw(ArgumentError("weights cannot contain Inf or NaN values"))
     wv.values[i] = v
     wv.sum = sum
+    wv.positive = missing
```
devmotion, line 60: I guess it would be more efficient to use
```suggestion
    wv.positive = wv.positive === missing ? missing : wv.positive && v >= zero(v)
```
here?

(Again, IMO it's a bit strange to call it `positive` since it rather checks non-negativity.)

src/weights.jl
devmotion: Hmm. It's easy to not hit this function, e.g., when using `all(x -> x >= 0, wv)` etc. And, of course, it covers also things like `all(>=(2), wv)`. One might also want to check non-negativity by e.g. `!any(<(0), wv)`.

So I think a dedicated separate function would be cleaner and less ambiguous. If one wants to support `AbstractArray`s one could also at some point just define a fallback
```julia
isnonneg(x::AbstractArray{<:Real})  = all(>=(0), x)
```

src/weights.jl
```diff
@@ -53,9 +57,34 @@ Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values),
...
     v
 end
 
+function Base.all(f::Base.Fix2{typeof(>=)}, wv::AbstractWeights)
+    if iszero(f.x)
+        if ismissing(wv.positive)
```
devmotion, line 66: In my experience, in particular with `nothing`, it is often better to use `===` instead of `isnothing`/`ismissing` (see e.g. https://github.com/JuliaLang/julia/issues/35585#issuecomment-619053871):
```suggestion
        if wv.positive === missing
```

src/weights.jl
```diff
@@ -53,9 +57,34 @@ Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values),
...
+        end
+        return wv.positive
+    else
+        return all(f, wv.values)
+    end
+end
+
+function Base.any(f::Base.Fix2{typeof(<)}, wv::AbstractWeights)
+    if iszero(f.x)
+        if ismissing(wv.positive)
```
devmotion, line 78:
```suggestion
        if wv.positive === missing
```

src/sampling.jl
```diff
@@ -845,22 +856,22 @@ function efraimidis_ares_wsample_norep!(rng::AbstractRNG, a::AbstractArray,
     @inbounds for _s in 1:n
         s = _s
         w = wv.values[s]
-        w < 0 && error("Negative weight found in weight vector at index $s")
+        w < 0 && throw(ArgumentError("Negative weight found in weight vector at index $s"))
```
devmotion, line 859: I think these efraimidis functions were actually one of my first contributions to Julia packages. Fun times but I'm not surprisied that some things can be improved and made more consistent :smile:

test/weights.jl
```diff
@@ -125,6 +124,42 @@ end
...
+        @test all(>=(0), f([-0.0, 0.0]))
+        @test !all(>=(0), f([1, -2]))
+        @test !any(<(0), f([1, 2]))
+        @test !any(<(0), f([-0.0, 0.0]))
+        @test any(<(0), f([1, -2]))
+
+        @test all(>=(1), [2, 3, 4])
+        @test !all(>=(1), [0, 1, 2])
+        @test any(<(3), [2, 3, 4])
+        @test !any(<(1), [1, 2, 3])
```
nalimilan, line 139:
```suggestion
        @test all(>=(1), f([2, 3, 4]))
        @test !all(>=(1), f([0, 1, 2]))
        @test any(<(3), f([2, 3, 4]))
        @test !any(<(1), f([1, 2, 3]))
```

src/weights.jl
nalimilan: Well, yeah, that's "positive", not "strictly positive". But maybe "nonnegative" is more explicit. It's kind of ugly though. :-)

src/weights.jl
nalimilan: Right. Though that's supposed to be better in recent releases, even if not completely conclusive (see https://github.com/JuliaLang/julia/pull/44407). Anyway that shouldn't matter here as we're not in a loop. But why not.

src/weights.jl
nalimilan: What annoys me is that there's no reason why this very basic function would live in and be exported by StatsBase. And anyway if users are not aware of the fast path (be it `all(>=(0), x)` or `nonneg(x)`), they won't use it and get the slow one, so I'm not sure choosing one syntax or the other makes a difference.

We could keep this internal for now -- though defining an internal function wouldn't be better than `all(>=(0), wv)` and `!any(<(0), wv)` as we would be sure users wouldn't be able to use it. ;-)

src/weights.jl
nalimilan: Indeed, good trick!

src/weights.jl
bkamins: It is internal so I would prefer a precise name.

src/weights.jl
```diff
@@ -53,9 +57,34 @@ Base.getindex(wv::W, ::Colon) where {W <: AbstractWeights} = W(copy(wv.values),
...
+            # sum is significantly faster than all when no entries are negative
+            wv.positive = sum(<(0), wv.values) == 0
+        end
+        return !wv.positive
+    else
+        return any(f, wv.values)
```
bkamins, line 84: add a test to cover this case.

bkamins:
@nalimilan - do we have benchmarks showing the performance impact?

Also - given the changes we do I think it would be good in https://juliastats.org/StatsBase.jl/stable/weights/#Weight-Vectors-1 to add a recommendation to create weights vector once if it is reused later. The point is that the code like:
```
[sample(Weights(x)) for i in 1:10^6]
```
is inefficient as it should be replaced by:
```
w = Weights(x)
[sample(w) for i in 1:10^6]
```
and I commonly see such patterns since users are unaware of the cost of `Weights` creation.

src/weights.jl
nalimilan: This is https://github.com/JuliaStats/StatsBase.jl/pull/834/files#r962207803. ;-)

src/weights.jl
nalimilan: Actually that doesn't work, as when `positive == false` and `v` is positive, we don't know whether `v` has just replaced the last negative value. So we have to store `missing` rather than `false`. I've pushed something a bit more involved.

src/weights.jl
nalimilan: OK. Actually to avoid the double negation I propose simply calling this `negative` and reversing the logic.

nalimilan:
Here's a simple benchmark for the worst case where the weights vector is used for the first time (so that negative entries have to be checked). I show both timings excluding and including the creation of the weights vector. (This covers the `alias_sample!` method.)

Unfortunately the overhead isn't completely negligible in this situation (up to 40% when sampling a single value from a 10,000 vector), which isn't surprising. So I've tried another approach: checking for negative weights directly inside sampling methods, when iterating over them. This actually gives a much lower overhead (up to 10%), see the last series of benchmarks for an illustration of what this gives for `make_alias_table!`. I haven't checked whether it would be fast for other methods too, but that's probably the case.

So we have three options:
- consider that the overhead of the current state of the PR is acceptable: probably not
- first check whether `wv.negative === missing`, and only check that weights are positive in that case: doable though a bit complex
- always check that weights are positive, and drop the new field from `AbstractWeights` types: this is simpler but imposes the (small) overhead even for repeated sampling with the same weights vector

What I don't understand is why the benchmarks where the cost of creating `weights(y)` is included show a very large regression compared with master, much larger than when excluding that cost. And it persists with the third approach. This doesn't make any sense to me.

```julia
# master

julia> x = rand(100);

julia> y = rand(100);

julia> @btime sample(x, w) setup=(w=weights(y));
  84.847 ns (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  1.855 μs (5 allocations: 4.38 KiB)

julia> @btime sample(x, weights(y));
  105.413 ns (2 allocations: 48 bytes)

julia> @btime sample(x, weights(y), 100);
  1.870 μs (6 allocations: 4.41 KiB)

julia> x = rand(10_000);

julia> y = rand(10_000);

julia> @btime sample(x, w) setup=(w=weights(y));
  1.384 μs (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  112.887 μs (9 allocations: 313.56 KiB)

julia> @btime sample(x, weights(y));
  2.530 μs (2 allocations: 48 bytes)

julia> @btime sample(x, weights(y), 100);
  114.043 μs (10 allocations: 313.59 KiB)

# This PR

julia> x = rand(100);

julia> y = rand(100);

julia> @btime sample(x, w) setup=(w=weights(y));
  94.256 ns (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  2.249 μs (5 allocations: 4.38 KiB)

julia> @btime sample(x, weights(y));
  157.160 ns (2 allocations: 48 bytes)

julia> @btime sample(x, weights(y), 100);
  2.298 μs (6 allocations: 4.41 KiB)

julia> x = rand(10_000);

julia> y = rand(10_000);

julia> @btime sample(x, w) setup=(w=weights(y));
  1.882 μs (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  126.571 μs (9 allocations: 313.56 KiB)

julia> @btime sample(x, weights(y));
  4.723 μs (2 allocations: 48 bytes)

julia> @btime sample(x, weights(y), 100);
  131.106 μs (10 allocations: 313.59 KiB)

# Checking negative weights in make_alias_table! loop

julia> x = rand(100);

julia> y = rand(100);

julia> @btime sample(x, w) setup=(w=weights(y));
  93.747 ns (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  1.944 μs (5 allocations: 4.38 KiB)

julia> @btime sample(x, weights(y));
  157.153 ns (2 allocations: 48 bytes)

julia> @btime sample(x, weights(y), 100);
  2.115 μs (6 allocations: 4.41 KiB)

julia> x = rand(10_000);

julia> y = rand(10_000);

julia> @btime sample(x, w) setup=(w=weights(y));
  1.365 μs (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  120.666 μs (9 allocations: 313.56 KiB)

julia> @btime sample(x, weights(y));
  4.906 μs (2 allocations: 48 bytes)

julia> @btime sample(x, weights(y), 100);
  122.024 μs (10 allocations: 313.59 KiB)
```

src/weights.jl
bkamins: So why does coverage show that it is not covered?

nalimilan: Because it was not applied before the last commit.

bkamins:
@nalimilan - I do not understand this difference in timing:
```
# master

julia> x = rand(100);

julia> y = rand(100);

julia> @btime sample(x, w) setup=(w=weights(y));
  84.847 ns (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  1.855 μs (5 allocations: 4.38 KiB)

# This PR

julia> x = rand(100);

julia> y = rand(100);

julia> @btime sample(x, w) setup=(w=weights(y));
  94.256 ns (1 allocation: 16 bytes)

julia> @btime sample(x, w, 100) setup=(w=weights(y));
  2.249 μs (5 allocations: 4.38 KiB)
```
What is the reason that first time the overhead is 10ns, and in the second it is 0.4μs. I would think the difference should be the same?

src/weights.jl
devmotion: Ah good catch!

src/weights.jl
devmotion: Yes, I would suggest only defining an internal function - that seems sufficient as it's only used in the argument checks internally. Something like `isnnonneg` seems to focus on what we actually want to check whereas defining `all` or `any` catches also other, in principle undesired, cases and hence requires a more complicated implementation with multiple checks.

nalimilan:
Good catch. This seems to be due to `direct_sample!` calling `sample(rng, wv)` for each requested value, which means we check aliasing, sum and negativity of weights 100 times. The code was maybe efficient when it was written, but we've progressively added sanity checks and now it's really absurd. Looks like by copying the code I can almost eliminate the overhead. I'll do more investigations.

PR #835, opened 2022-09-25
Add mvue and mom estimators of N

jbytecode:
This PR implements two methods for estimating $N$ using a sample of integer IDs:

- Minimum-variance unbiased estimator (mvue)
- Methods of moments estimator (mom)

using a simple dispatcher function `nestimate()`.

The problem of estimating $N$ has a famous and well-known use case, the German Tank Problem, is also referenced in the doc string of the method.

Basic tests are added. 

Functions for confidence intervals are not yet implemented. 

In my opinion, it would be nice to see these basic functions in StatsBase as they are the basics of their family of estimators. 

I am so sorry If I am bothering the development team or doing something inconvenient.

Thank you in advance.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -22,7 +22,8 @@ makedocs(
              "multivariate.md",
              "misc.md",
              "statmodels.md",
-             "transformations.md"],
+             "transformations.md",
+             "nestimate.md"],
     strict=true,
     checkdocs=:exports
 )
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -12,7 +12,7 @@ end
 
 
 ```@contents
-Pages = ["weights.md", "scalarstats.md", "robust.md", "deviation.md", "cov.md", "counts.md", "ranking.md", "sampling.md", "empirical.md", "signalcorr.md", "misc.md", "statmodels.md", "transformations.md"]
+Pages = ["weights.md", "scalarstats.md", "robust.md", "deviation.md", "cov.md", "counts.md", "ranking.md", "sampling.md", "empirical.md", "signalcorr.md", "misc.md", "statmodels.md", "transformations.md", "nestimate.md"]
 Depth = 2
 ```
 
--- a/docs/src/nestimate.md
+++ b/docs/src/nestimate.md
@@ -1,0 +1,6 @@
+# Estimate of N
+
+
+```@docs
+nestimate
+```
--- a/src/StatsBase.jl
+++ b/src/StatsBase.jl
@@ -240,7 +240,10 @@ export
 
     # reliability
     CronbachAlpha,        # the type to represent Cronbach's alpha scores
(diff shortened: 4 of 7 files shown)


bkamins:
@jbytecode - these estimators are essentially doing estimation of parameter of `DiscreteUniform` with fixed lower bound to `1`. We currently have a naive estimator for `DiscreteUniform` with both bounds free in https://juliastats.org/Distributions.jl/stable/fit/#Distribution-Fitting.

I think the contribution is better suited there. I think we would need to add parameters to `fit` for `DiscreteUniform`:
* passing which estimator should be used (there are several)
* passing information which parameters of the distribution are known (currently both are assumed to be unknown, in general apart from this case only lower can be known - as in your case, or only upper)

What do you think?

jbytecode:
@bkamins - Thank you for the response and the different point of view you provided. As you mentioned above, the German Tank problem can be transformed into estimating the upper bound parameter of a `DiscreteUniform(1, b)`. Since 

$$
\bar{x} = \frac{a + b}{2}
$$

and the Method of Moments estimate of b is 

$$
\hat{b} = 2\bar{x} - 1
$$

This approach can be generalized with the lower bound parameter so 

$$
\hat{a} = 2\bar{x} - b
$$

where b is known. The mom estimates fit your point of view and can be well integrated with the `fit` function in `Distributions.jl`.

This approach has two problems:

- The mom estimates of this parameters are not efficient (by means of the variance of the estimator), so the minimum variance unbiased estimator is used instead
- It is not guaranteed the a is non-negative (this may cause non-sense result, but by using a min() function the negative result can be cast to 1.

The second problem is

$$
\hat{b} = \max(x) + \frac{\max(x)}{n} - 1
$$

is the minimum variance unbiased estimate of `b`, but I am not even sure if it can be transformed into 


$$
\hat{a} = \min(x) - \frac{\min(x)}{n} + 1
$$

or 

$$
\hat{a} = \max(x) - \frac{\max(x)}{n} + 1
$$

when b is known and the a is of interest.

In short, I am totally agreed with your thoughts. If the problem is handled as an estimation problem of distribution parameters in a `DiscreateUniform(a, b)`, it can be implemented in `Distributions.jl` but a little bit investigation is required, otherwise, if the problem is basically a German Tank problem, a function `germantank()` fits well with the `StatBase.jl`. 

The decision should be taken by the core developers like you.

Thank you for consuming your time and providing your beautiful thoughts.

bkamins:
> The decision should be taken by the core developers like you.

I think it should live in Distributions.jl. However, I think it is OK to provide just methods for `a=1` (or whatever methods you find easy to implement). The point is that in StatsBase.jl we try to limit the number of functionalities provided, especially that in some future it might be merged with Statistics.jl. Functions for estimation of parameters of distributions should live in Distributions.jl.

PR #836, opened 2022-10-05
Fix `mean_and_var` and `mean_and_std` docstring signature

bowenszhu:
The default value for the argument `corrected` in the docstring signature of `mean_and_var` and `mean_and_std` is `false`, but it is actually `true` in the code.

https://github.com/JuliaStats/StatsBase.jl/blob/a0e6f1e807a84a09b5f74431bb0099f4aaed5ae0/src/moments.jl#L153-L165
https://github.com/JuliaStats/StatsBase.jl/blob/a0e6f1e807a84a09b5f74431bb0099f4aaed5ae0/src/moments.jl#L167-L180

### DIFF
--- a/src/moments.jl
+++ b/src/moments.jl
@@ -90,7 +90,7 @@ std(v::RealArray, w::AbstractWeights, dim::Int;
 
 ##### Fused statistics
 """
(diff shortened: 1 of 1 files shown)


nalimilan (approved):
Thanks! The default was changed a long time ago and it looks like we missed some occurrences.

PR #840, opened 2022-11-21
get rid of old typealiases

KristofferC:
As the comment in `common.jl` says, these are obsolete now with the terse covariant syntax.

Interestingly, there is apparently a whole package just for these type alises: https://github.com/AP6YC/NumericalTypeAliases.jl/blob/develop/README.md#aliases.


docs/src/counts.md
```diff
@@ -7,7 +7,7 @@ The package provides functions to count the occurrences of distinct values.
 ```@docs
 counts
 proportions
-addcounts!(r::AbstractArray, x::StatsBase.IntegerArray, levels::StatsBase.IntUnitRange)
+addcounts!(r::AbstractArray, x::StatsBase.AbstractArray{<:Integer}, levels::StatsBase.IntUnitRange)
```
ararslan, line 10:
```suggestion
addcounts!(r::AbstractArray, x::AbstractArray{<:Integer}, levels::StatsBase.IntUnitRange)
```
I guess `IntUnitRange` could be similarly removed, as it's an alias for `UnitRange{<:Integer}`.

docs/src/scalarstats.md
```diff
@@ -69,7 +69,7 @@ percentile
 iqr
 nquantile
 quantile
-Statistics.median(v::StatsBase.RealVector, w::AbstractWeights{<:Real})
+Statistics.median(v::StatsBase.AbstractVector{<:Real}, w::AbstractWeights{<:Real})
```
ararslan, line 72:
```suggestion
Statistics.median(v::AbstractVector{<:Real}, w::AbstractWeights{<:Real})
```

ararslan (commented):
Thank you for doing this! Though clearly the proper fix would have been to take a dependency on NumericalTypeAliases. :trollface:

KristofferC:
Updated

PR #853, opened 2023-02-27
Fix `inverse_rle` with negative lengths

### LINKED ISSUE #851: Buffer overflow using `inverse_rle`
https://github.com/JuliaStats/StatsBase.jl/blob/42feefb84aa81d9f8918500ecf159dbee5bcb91f/src/misc.jl#L71-L86

If `lens` contains a negative element, the allocated array `r` will be smaller than needed and buffer overflow will occur. This will sometimes result in an access violation.

```julia
julia> using StatsBase

julia> inverse_rle(1:3, [-2,1,3])
2-element
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x1ee9e7b -- jl_gc_pool_alloc at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:1217
in expression starting at none:0       
jl_gc_pool_alloc at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:1217
jl_gc_alloc_ at /cygdrive/c/buildbot/worker/package_win64/build/src\julia_internal.h:339 [inlined]
jl_gc_alloc at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:3290  
_new_array_ at /cygdrive/c/buildbot/worker/package_win64/build/src\array.c:127 [inlined]
_new_array at /cygdrive/c/buildbot/worker/package_win64/build/src\array.c:193 
[inlined]
jl_alloc_array_1d at /cygdrive/c/buildbot/worker/package_win64/build/src\array.c:441
[...]
```

Simple fix could be to add a check for negative elements.

ararslan:
Fixes #851.

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -60,8 +60,10 @@ run lengths.
 function inverse_rle(vals::AbstractVector{T}, lens::AbstractVector{<:Integer}) where T
     m = length(vals)
     length(lens) == m || raise_dimerror()
+    n = sum(lens)
+    n >= 0 || throw(ArgumentError("lengths must be non-negative"))
 
-    r = Vector{T}(undef, sum(lens))
+    r = Vector{T}(undef, n)
     p = 0
     @inbounds for i = 1 : m
         j = lens[i]
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -60,11 +60,11 @@ run lengths.
 function inverse_rle(vals::AbstractVector{T}, lens::AbstractVector{<:Integer}) where T
     m = length(vals)
     length(lens) == m || raise_dimerror()
-    all(>=(0), lens) || throw(ArgumentError("lengths must be non-negative"))
     r = Vector{T}(undef, sum(lens))
     p = 0
     @inbounds for i = 1 : m
         j = lens[i]
+        j >= 0 || throw(ArgumentError("lengths must be non-negative"))
         v = vals[i]
         while j > 0
             r[p+=1] = v
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -60,6 +60,7 @@ run lengths.
 function inverse_rle(vals::AbstractVector{T}, lens::AbstractVector{<:Integer}) where T
     m = length(vals)
     length(lens) == m || raise_dimerror()
+
     r = Vector{T}(undef, sum(lens))
     p = 0
     @inbounds for i = 1 : m
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -60,7 +60,7 @@ run lengths.
 function inverse_rle(vals::AbstractVector{T}, lens::AbstractVector{<:Integer}) where T
     m = length(vals)
     length(lens) == m || raise_dimerror()
-
+    all(>=(0), lens) || throw(ArgumentError("lengths must be non-negative"))
     r = Vector{T}(undef, sum(lens))
     p = 0
     @inbounds for i = 1 : m
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.33.21"
+version = "0.33.22"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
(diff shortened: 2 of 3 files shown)


src/misc.jl
```diff
@@ -60,7 +60,7 @@ run lengths.
 function inverse_rle(vals::AbstractVector{T}, lens::AbstractVector{<:Integer}) where T
     m = length(vals)
     length(lens) == m || raise_dimerror()
-
+    all(>=(0), lens) || throw(ArgumentError("lengths must be non-negative"))
```
devmotion, line 63: Is there a specific reason to support `0`? I would have assumed that a valid run-length encoding does not contain values with length `0`.
```suggestion
    all(>(0), lens) || throw(ArgumentError("lengths must be positive"))
```

I'm also a bit surprised that tests do not error on Julia 1.0 since
```julia
help?> >=(0)
  >=(x)

  Create a function that compares its argument to x using >=, i.e. a function equivalent to y -> y >= x. The returned function is of type
  Base.Fix2{typeof(>=)}, which can be used to implement specialized methods.

  │ Julia 1.2
  │
  │  This functionality requires at least Julia 1.2.

help?> >(0)
  >(x)

  Create a function that compares its argument to x using >, i.e. a function equivalent to y -> y > x. The returned function is of type
  Base.Fix2{typeof(>)}, which can be used to implement specialized methods.

  │ Julia 1.2
  │
  │  This functionality requires at least Julia 1.2.
````

Comment (ararslan) [reply src/misc.jl:63]: > Is there a specific reason to support 0?

Not really but it doesn't hurt to support IMO and may possibly be convenient to someone at some point ¯\\\_(ツ)\_/¯

> I'm also a bit surprised that tests do not error on Julia 1.0 since [...]

Huh yeah, me too. Maybe it existed in 1.0 but was undocumented and/or didn't return a `Fix2`? I don't have 1.0 handy to check.

Comment (kritpetra) [reply src/misc.jl:63]: > Is there a specific reason to support `0`? I would have assumed that a valid run-length encoding does not contain values with length `0`.

I can definitely see how someone would want to programmatically create lists where some values can be skipped, e.g.
```julia
julia> r = 3
julia> startidx = [2,5,3]
julia> inverse_rle.(Ref(["A", "B", "C"]), ([i-1, r, max(0,8-i-r)] for i in startidx))
3-element Vector{Vector{String}}:
 ["A", "B", "B", "B", "C", "C", "C"]
 ["A", "A", "A", "A", "B", "B", "B"]
 ["A", "A", "B", "B", "B", "C", "C"]
```

nalimilan: I agree allowing zeros can be useful.

I've checked Julia 1.0, `>(0)` and `>=(0)` work thanks to Compat.jl.

src/misc.jl
```diff
@@ -60,7 +60,7 @@ run lengths.
...
     r = Vector{T}(undef, sum(lens))
     p = 0
     @inbounds for i = 1 : m
```
nalimilan, line 66: How about doing `j >= 0 || throw(ArgumentError("lengths must be positive"))` here instead, to avoid going over the vector twice?

ararslan: I guess the tradeoff is iterate twice or iterate once but always allocate a (potentially large) array and branch at each iteration. I suspect iterating twice would be more efficient but I haven't actually benchmarked or anything like that.

nalimilan: Allocating before throwing an error doesn't  sound like a problem to me as it shouldn't happen in valid code. As for branching on each iteration, given that there's already a branch I imagine it won't make the loop slower. But not a big deal anyway.

nalimilan (approved):

src/misc.jl
ararslan: > Allocating before throwing an error doesn't sound like a problem to me as it shouldn't happen in valid code.

Ha yeah, fair point.

> As for branching on each iteration, given that there's already a branch I imagine it won't make the loop slower.

Oh you're right, for some reason the `while` loop didn't register to me as branching.

I'm happy to change if you prefer, I don't really care either way.

nalimilan: Maybe change it if you feel like it, otherwise it's fine.

src/misc.jl
```diff
@@ -65,6 +65,7 @@ function inverse_rle(vals::AbstractVector{T}, lens::AbstractVector{<:Integer}) w
     p = 0
```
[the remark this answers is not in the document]
ararslan: Added a check on the sum per your suggestion since we need to compute the sum anyway in order to initialize the output array.

PR #854, opened 2023-03-02
Avoid define `>` and `>=`.

N5N3:
`Base` has a general fallback which map `>=` to `<=` (and `>` to `<`).

### DIFF
--- a/src/statmodels.jl
+++ b/src/statmodels.jl
@@ -85,7 +85,7 @@ TestStat(x::TestStat) = x
 
 float(x::Union{TestStat, PValue}) = float(x.v)
 
-for op in [:(==), :<, :(isless), :(isequal)] # isless and < to place nice with NaN
+for op in [:(==), :<, :≤, :(isless), :(isequal)] # isless and < to place nice with NaN
     @eval begin
         Base.$op(x::Union{TestStat, PValue}, y::Real) = $op(x.v, y)
         Base.$op(y::Real, x::Union{TestStat, PValue}) = $op(y, x.v)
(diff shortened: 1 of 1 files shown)


src/statmodels.jl
```diff
@@ -85,7 +85,7 @@ TestStat(x::TestStat) = x
 
 float(x::Union{TestStat, PValue}) = float(x.v)
 
-for op in [:(==), :<, :≤, :>, :≥, :(isless), :(isequal)] # isless and < to place nice with NaN
+for op in [:(==), :<, :≤, #==:>, :≥,==# :(isless), :(isequal)] # isless and < to place nice with NaN
```
ararslan, line 88:
```suggestion
for op in [:(==), :<, :(isless), :(isequal)] # isless and < to place nice with NaN
```
Might as well remove rather than comment out. Also we don't need to define `<=` since having `==` and `<` suffices.

N5N3: `<=` is good to have as it has many optimized specializations. (e.g. for comparsion between floats and ints)

src/statmodels.jl
```diff
@@ -85,7 +85,7 @@ TestStat(x::TestStat) = x
 
 float(x::Union{TestStat, PValue}) = float(x.v)
 
-for op in [:(==), :<, :≤, :>, :≥, :(isless), :(isequal)] # isless and < to place nice with NaN
+for op in [:(==), :<, :(isless), :(isequal)] # isless and < to place nice with NaN
```
N5N3, line 88:
```suggestion
for op in [:(==), :<, :≤, :(isless), :(isequal)] # isless and < to place nice with NaN
```

PR #855, opened 2023-03-04
proportionmap accepts iterators

### LINKED ISSUE #842: `proportionmap` should accept any iterator
```
julia> x = [1, 2, 3]
3-element Array{Int64,1}:
 1
 2
 3

julia> proportionmap(skipmissing(x))
ERROR: MethodError: no method matching proportionmap(::Base.SkipMissing{Vector{Int64}})
Closest candidates are:
  proportionmap(::AbstractArray) at ~/.julia/packages/StatsBase/XgjIN/src/counts.jl:455
  proportionmap(::AbstractArray, ::AbstractWeights) at ~/.julia/packages/StatsBase/XgjIN/src/counts.jl:456
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1
```

Ref. https://github.com/JuliaStats/StatsBase.jl/issues/581 for `countmap`.

ArunSanganal:
closes #842.

### DIFF
--- a/src/counts.jl
+++ b/src/counts.jl
@@ -450,5 +450,5 @@ Return a dictionary mapping each unique value in `x` to its proportion in `x`.
 If a vector of weights `wv` is provided, the proportion of weights is computed rather
 than the proportion of raw counts.
 """
-proportionmap(x::AbstractArray) = _normalize_countmap(countmap(x), length(x))
+proportionmap(x) = _normalize_countmap(countmap(x), length(collect(x)))
 proportionmap(x::AbstractArray, wv::AbstractWeights) = _normalize_countmap(countmap(x, wv), sum(wv))
--- a/test/counts.jl
+++ b/test/counts.jl
@@ -209,3 +209,9 @@ if VERSION >= v"1.9.0-DEV"
         # countmap and proportionmap only support the :dict algorithm for weighted sums.
     end
 end
+
+@testset "proportionmap with iterator" begin
+    a = [1, 2, 3, 4]
+    b =[true, true ,false, false, true, false]
+    @test proportionmap(skipmissing(a)) == Dict(1 => 0.25, 2 => 0.25, 3 => 0.25, 4 => 0.25)
+    @test proportionmap(skipmissing(b)) == Dict(true => 0.5, false => 0.5)


src/counts.jl
```diff
@@ -450,5 +450,5 @@ Return a dictionary mapping each unique value in `x` to its proportion in `x`.
 If a vector of weights `wv` is provided, the proportion of weights is computed rather
 than the proportion of raw counts.
 """
-proportionmap(x::AbstractArray) = _normalize_countmap(countmap(x), length(x))
+proportionmap(x) = _normalize_countmap(countmap(x), length(collect(x)))
```
devmotion, line 453: Could we just count the total number of elements when building the `countmap`? It seems inefficient to materialize `x` only to obtain its length if we already iterate through it anyway.

ArunSanganal: Something around sum(values(count(x))?

devmotion: No, I thought counting directly inside of `countmap`. But probably `sum(values, countmap(x))` would still be more efficient than using `collect(x)` if `x` is an iterator with a large number of elements.

ArunSanganal: `` julia> @btime proportionmap(skipmissing(a))
  8.625 μs (27 allocations: 146.67 KiB)
Dict{Int64, Float64} with 4 entries:
  4 => 0.25
  2 => 0.25
  3 => 0.25
  1 => 0.25 ``
  
  `` julia> @btime proportionmap(skipmissing(a))
  316.667 ns (9 allocations: 1.08 KiB)
Dict{Int64, Float64} with 4 entries:
  4 => 0.25
  2 => 0.25
  3 => 0.25
  1 => 0.25 ``
  
  Looks like a significant improvement 🧐

src/counts.jl
```diff
@@ -450,5 +450,13 @@ Return a dictionary mapping each unique value in `x` to its proportion in `x`.
...
 than the proportion of raw counts.
 """
-proportionmap(x::AbstractArray) = _normalize_countmap(countmap(x), length(x))
+function proportionmap(x)
+    countm = Dict{eltype(x), Int}()
+    n = 0
+    for y in x
+        countm[y] = get(countm, y, 0) + 1
+        n += 1
+    end
```
nalimilan, line 459: This reinvents `countmap`. Better make `countmap` allow iterators instead, so that both functions benefit.

ArunSanganal: `countmap` already accepts iterators; I did that to keep a count of `n` while iterating.

nalimilan: OK. The problem is that `countmap` uses different algorithms under the hood for performance. By using a `Dict` here, you lose the benefit of the fast radix sort and count sort algorithms.

I see two solutions:
- do `n = Base.IteratorSize(x) isa Union{HasLength, HasShape} ? length(x) : sum(values(countm))`
- adjust all `_addcounts!` methods to return the number of elements (this should be cheap so not a big deal if it's not used by `addcounts`)

tylerjthomas9: I am looking to help get this across the line. Is this your first proposed solution?

```julia
function proportionmap(x)
    countm = countmap(x)
    n = Base.IteratorSize(x) isa Union{Base.HasLength, Base.HasShape} ? length(x) : sum(values(countm))
    _normalize_countmap(countm, n)
end
```

Issue #777, opened 2022-03-25
variation() has no optional argument corrected=false / true


mauricelanghinrichs:
Hey all!

I noticed that `variation()` and potentially other derived methods in StatsBase don't seem to have the "corrected" optional keyword argument. `variation()` calls `std()` I guess, where one can optionally switch off correcting for degrees of freedom (division by N or N-1). I wondered if it makes sense to add this option also for `variation()`, and maybe other methods that rely on `std()` or similar, too.

In applications I find this particularly relevant when computing coefficients of variations of single-element arrays (as it may happen when array length is itself probabilistic). For example:

```julia
using StatsBase

a = [1.0]
variation(a) # NaN
variation(a, corrected=true) # not possible, but should also give NaN
variation(a, corrected=false) # not possible, but should give 0.0
```

My version is `StatsBase v0.33.16`.

Best

nalimilan:
Makes sense. Would you make a pull request to make `variation` pass `corrected` when calling `mean_and_std`?

PR #856, opened 2023-03-16
Optional argument for `variation()`

itsdebartha:
- Added `corrected=true/false` for variation
- Added test for the above addition

This commit addresses #777

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -404,13 +404,16 @@ span(x) = ((a, b) = extrema(x); a:b)
 
 # Variation coefficient: std / mean
 """
-    variation(x, m=mean(x))
+    variation(x, m=mean(x); corrected=true)
 
 Return the coefficient of variation of collection `x`, optionally specifying
-a precomputed mean `m`. The coefficient of variation is the ratio of the
-standard deviation to the mean.
+a precomputed mean `m`, and the optional correction parameter `corrected`.
+The coefficient of variation is the ratio of the
+standard deviation to the mean. If `corrected` is `false`,
+then `std` is calculated with denominator `n`. Else, the `std` is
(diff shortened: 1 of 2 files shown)


ParadaCarleton:
@nalimilan do you know why the docs are failing?

src/scalarstats.jl
```diff
@@ -411,7 +411,7 @@ a precomputed mean `m`. The coefficient of variation is the ratio of the
 standard deviation to the mean.
 """
 variation(x, m) = stdm(x, m) / m
```
devmotion, line 413: This should be changed as well, I assume:
```suggestion
variation(x, m; corrected::Bool=true) = stdm(x, m; corrected=corrected) / m
```

devmotion (commented):
Can you also update the docstring?

src/scalarstats.jl
itsdebartha: I found `stdm` to be in `deprecetes.jl` as of November 2021. So I was wondering if I need to change it or not...

devmotion: Only specific methods of `stdm` with weights are deprecated. The function that is called here is defined in the Statistics (https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.stdm) and not (yet) deprecated there (https://github.com/JuliaStats/Statistics.jl/issues/5).

itsdebartha: Alright. I have made the required changes. Thanks.

itsdebartha:
I also added tests that would use the method of `stdm` (i.e. specified mean) which were previously not there

ParadaCarleton:
This looks good, thanks!

CarloLucibello:
tag a new release? it has been a while...

Issue #693, opened 2021-05-20
[Feature Request] Add Standard Deviation to summarystats()?


noamsgl:
Feature Request: Add STD to summarystats()...

https://github.com/JuliaStats/StatsBase.jl/blob/a0e6f1e807a84a09b5f74431bb0099f4aaed5ae0/src/scalarstats.jl#L555

nalimilan:
I agree that it's surprising not to print the standard deviation here. Do you feel like making a pull request to add it?

itsdebartha:
@nalimilan is this a dormant issue or still open for contribution?

I was thinking how about a `summarystat()` just like the `summary` function in R. It'll return different values for string/factor types just like R does...

nalimilan:
We need more opinions to decide what's best. In DataFrames (https://github.com/JuliaData/DataFrames.jl/pull/2459), we decided not to report standard deviations and quartiles by default so that the output fits in the screen width: one needs to do `describe(df, :detailed)` to get them. Here screen width isn't a problem and we already report quartiles, so maybe we could print the standard deviation too.

@bkamins @pdeffebach What do you think?

bkamins:
In general I almost always want to see `std`, so I would like to have this change. The only issue is that it would be breaking. I am not sure what decision would be best. Maybe we can consider it to be mildly breaking and go for it?

pdeffebach:
I agree I almost always want `std`. I would maybe call it mildly breaking? Its really only useful in interactive work.

nalimilan:
That would only change the printing, so that's considered non-breaking I think?

nalimilan:
Do we consider that adding a new field to an object is breaking though? That sounds quite restrictive.

bkamins:
OK - let us add it.

itsdebartha:
Created #858

PR #858, opened 2023-04-21
summarystats improvements

itsdebartha:
This commit addresses issue #693 
- Add standard deviation to summarystats
- Update documentation
- Add tests

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -857,6 +857,7 @@ kldivergence(p::AbstractArray{<:Real}, q::AbstractArray{<:Real}, b::Real) =
 
 struct SummaryStats{T<:Union{AbstractFloat,Missing}}
     mean::T
+    sd::T
     min::T
     q25::T
     median::T
@@ -871,14 +872,16 @@ end
     summarystats(a)
 
 Compute summary statistics for a real-valued array `a`. Returns a
-`SummaryStats` object containing the mean, minimum, 25th percentile,
-median, 75th percentile, and maxmimum.
+`SummaryStats` object containing the number of observations,
+number of missing observations, standard deviation, mean, minimum,
+25th percentile, median, 75th percentile, and maxmimum.
 """
 function summarystats(a::AbstractArray{T}) where T<:Union{Real,Missing}
     # `mean` doesn't fail on empty input but rather returns `NaN`, so we can use the
     # return type to populate the `SummaryStats` structure.
     s = T >: Missing ? collect(skipmissing(a)) : a
     m = mean(s)
+    stdev = std(s)
     R = typeof(m)
     n = length(a)
     ns = length(s)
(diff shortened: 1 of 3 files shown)


itsdebartha:
`sem` returns `ArgumentError` but expects `MethodError` in CI. However, in my machine, `sem` returns `MethodError`, and passes the tests.

ParadaCarleton:
The tests are failing on Julia nightly, which is fine (it usually means there's a bug in the nightly build, which is pre-alpha, rather than a bug in our code).

itsdebartha:
Yeah, seems like `MethodError` has changed to `ArgumentError` for some of the results...

src/scalarstats.jl
```diff
@@ -871,14 +872,16 @@ end
     summarystats(a)
 
 Compute summary statistics for a real-valued array `a`. Returns a
-`SummaryStats` object containing the mean, minimum, 25th percentile,
-median, 75th percentile, and maxmimum.
+`SummaryStats` object containing the number of observations,
+number of missing observations, standard deviation, mean, minimum,
+25th percentile, median, 75th percentile, and maxmimum.
```
nalimilan, line 877: While we're at it:
```suggestion
25th percentile, median, 75th percentile, and maximum.
```

test/scalarstats.jl
```diff
@@ -332,30 +332,39 @@ s = summarystats(1:5)
 @test isa(s, StatsBase.SummaryStats)
 @test s.min == 1.0
 @test s.max == 5.0
+@test s.nobs   == 5
+@test s.nmiss  == 0
 @test s.mean   ≈ 3.0
 @test s.median ≈ 3.0
 @test s.q25    ≈ 2.0
 @test s.q75    ≈ 4.0
+@test s.sd     ≈ 1.5 atol=1e-1
```
nalimilan, line 341: Use a lower tolerance? I guess you should be able to keep the default tolerance if you copy the actual value.

nalimilan (commented):
Thanks! Sorry for the delay.

src/scalarstats.jl
itsdebartha: Thanks for pointing out this mistake...

test/scalarstats.jl
itsdebartha: Yeah, it's updated now...

src/deviation.jl
```diff
@@ -12,7 +12,7 @@ function counteq(a::AbstractArray, b::AbstractArray)
     n = length(a)
     length(b) == n || throw(DimensionMismatch("Inconsistent lengths."))
     c = 0
-    for i = 1:n
+    for i in eachindex(a, b)
```
nalimilan, line 15: Thanks for doing this (it's been long due!), but can you move this to another PR?

itsdebartha: Yup, I have done it...

itsdebartha:
@nalimilan I have made the corrections. Hopefully, it is ok now...

src/scalarstats.jl
```diff
@@ -871,14 +872,16 @@ end
...
+`SummaryStats` object containing the number of observations,
+number of missing observations, standard deviation, mean, minimum,
+25th percentile, median, 75th percentile, and maximum.
 """
 function summarystats(a::AbstractArray{T}) where T<:Union{Real,Missing}
     # `mean` doesn't fail on empty input but rather returns `NaN`, so we can use the
     # return type to populate the `SummaryStats` structure.
     s = T >: Missing ? collect(skipmissing(a)) : a
     m = mean(s)
+    stdev = std(s)
```
nalimilan, line 884: Actually this should be a bit more efficient:
```suggestion
    stdev = std(s, mean=m)
```

PR #859, opened 2023-04-26
Set version to 0.34.0

ararslan:
See https://github.com/JuliaRegistries/General/pull/82386

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.33.22"
+version = "0.34.0"
 
 [deps]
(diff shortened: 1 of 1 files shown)


andreasnoack (approved):

3f6a:
Why not 1.0 ?

ararslan:
We discussed 1.0 and decided to go with 0.34 for now. There are social implications to calling something 1.0, particularly the general expectation of stability regardless of whether said stability exists in practice, and I don't think StatsBase is quite at the point at which that makes sense yet. Generally speaking, the package is quite stable; the only reason a breaking release is being made at all is for a change in edge case behavior with weights and the removal of some non-public internals that a ton of other packages were relying on. The maintainers haven't had the broader discussion over what we're willing to commit to as a v1.0 in terms of interface, what functionality should live where (cf. Statistics stdlib, StatsAPI, StatsModels), and all that, and until that's all resolved, I wouldn't want to block releases for bug fixes and compatibility tweaks. So 0.34 seems like a sensible path forward in the meantime.

PR #863, opened 2023-05-03
Fix tests on Julia nightly

ararslan:
The tests for specific exception types in `sem` are relying on the exceptions thrown from deeper in the call stack, outside of this package's control. We do still expect these cases to throw, but since we can't actually control the type of the exception, we can instead test that any exception is thrown.

The test of the number of allocations from calling `mad` on a generator fails because there are nearly 10x more allocations than the upper bound used in the test. We can mark this test broken with Julia v1.10, and the test will fail if the allocations for this case are improved, be it upstream or in the implementation of `mad` (if applicable).

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -4,4 +4,4 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
 [compat]
-Documenter = "0.24"
+Documenter = "0.27"
--- a/test/scalarstats.jl
+++ b/test/scalarstats.jl
@@ -201,8 +201,8 @@ x = sort!(vcat([5:-1:i for i in 1:5]...))
 @test @inferred(isnan(sem(Int[], ProbabilityWeights(Int[]); mean=0f0)))
 
 @test @inferred(isnan(sem(skipmissing(Union{Int,Missing}[missing, missing]))))
-@test_throws MethodError sem(Any[])
-@test_throws MethodError sem(skipmissing([missing]))
+@test_throws Exception sem(Any[])
+@test_throws Exception sem(skipmissing([missing]))
 
 @test mad(1:5; center=3, normalize=true) ≈ 1.4826022185056018
 @test mad(skipmissing([missing; 1:5; missing]); center=3, normalize=true) ≈ 1.4826022185056018
@@ -221,7 +221,14 @@ x = sort!(vcat([5:-1:i for i in 1:5]...))
 @test_throws ArgumentError mad(Int[], normalize = true)
 @test mad(Iterators.repeated(4, 10)) == 0
 @test mad(Integer[1,2,3,4]) === mad(1:4)
-@test (@benchmark mad((i for i in 1:10000))).allocs < 200
+let itr = (i for i in 1:10000)
+    if VERSION >= v"1.10.0-"
+        # FIXME: Allocations are closer to 10x this on 1.10
+        @test_broken (@benchmark mad($itr)).allocs < 200
+    else
+        @test (@benchmark mad($itr)).allocs < 200
+    end
+end
 
 # Issue 197
 @test mad(1:2, normalize=true) ≈ 0.7413011092528009


ararslan:
> ```
> Error: 1 docstring not included in the manual:
>
>     StatsBase.addcounts! :: Tuple{AbstractArray, AbstractArray{<:Integer}, UnitRange{<:Integer}}
>
> These are docstrings in the checked modules (configured with the modules keyword)
> that are not included in @docs or @autodocs blocks.
>
> Populate: populating indices.
> ERROR: LoadError: `makedocs` encountered an error (:missing_docs). Terminating build before rendering.
> ```

This is so weird. I can reproduce it locally _only_ if there is an existing `docs/build` directory.
```
$ git clean -fdx
Removing Manifest.toml
Removing docs/Manifest.toml
Removing docs/build/
$ julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(; path=pwd())); Pkg.instantiate()'
[...]
$ julia --project=docs/ docs/make.jl
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
[ Info: HTMLWriter: rendering HTML pages.
$ julia --project=docs/ docs/make.jl
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
┌ Error: 1 docstring not included in the manual:
│
│     StatsBase.addcounts! :: Tuple{AbstractArray, AbstractArray{<:Integer}, UnitRange{<:Integer}}
│
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in @docs or @autodocs blocks.
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/H5y27/src/Utilities/Utilities.jl:32
[ Info: Populate: populating indices.
ERROR: LoadError: `makedocs` encountered an error (:missing_docs). Terminating build before rendering.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] runner(#unused#::Type{Documenter.Builder.RenderDocument}, doc::Documenter.Documents.Document)
   @ Documenter.Builder ~/.julia/packages/Documenter/H5y27/src/Builder.jl:256
 [3] dispatch(#unused#::Type{Documenter.Builder.DocumentPipeline}, x::Documenter.Documents.Document)
   @ Documenter.Utilities.Selectors ~/.julia/packages/Documenter/H5y27/src/Utilities/Selectors.jl:170
 [4] #2
   @ ~/.julia/packages/Documenter/H5y27/src/Documenter.jl:273 [inlined]
 [5] cd(f::Documenter.var"#2#3"{Documenter.Documents.Document}, dir::String)
   @ Base.Filesystem ./file.jl:112
 [6] makedocs(; debug::Bool, format::Documenter.Writers.HTMLWriter.HTML, kwargs::Base.Pairs{Symbol, Any, NTuple{5, Symbol}, NamedTuple{(:sitename, :modules, :pages, :strict, :checkdocs), Tuple{String, Vector{Module}, Vector{String}, Bool, Symbol}}})
   @ Documenter ~/.julia/packages/Documenter/H5y27/src/Documenter.jl:272
 [7] top-level scope
   @ ~/Projects/julia-packages/StatsBase.jl/docs/make.jl:8
in expression starting at /Users/aarslan/Projects/julia-packages/StatsBase.jl/docs/make.jl:8
```

devmotion:
It seems the Documenter issue was fixed upstream (but is not available in 1.9 yet), so this shouldn't hold back the PR, should it?

ararslan:
> It seems the Documenter issue was fixed upstream (but is not available in 1.9 yet)

It sounds like the upstream change that fixed the issue with Documenter will not go into 1.9 (https://github.com/JuliaLang/julia/issues/49975#issuecomment-1569840642).

> this shouldn't hold back the PR, should it?

Yeah I suppose not. I had forgotten about this PR but I believe at the time I figured I'd tackle CI issues together but the Documenter thing turned out to really be A Thing.

PR #865, opened 2023-05-31
entropy with isprobvec check

### LINKED ISSUE #769: Entropy calculation with non-probability vectors
The current (v0.33.16) implementation of `entropy(p)`
https://github.com/JuliaStats/StatsBase.jl/blob/071d10a110d756fa15c6a6878936eee27e4c2cb6/src/scalarstats.jl#L760-L768

returns weird results if the argument isn't a probability-like vector (i.e. `sum(p) != 1`)

```julia
julia> entropy([0.5,0.5],2)       # correct
1.0
julia> entropy([0.5,0.25],2)      # this should throw an error?
1.0
julia> entropy([0.25,0.25],2)     # this too, or at least some notion of normalization of p?
1.0
julia> entropy([0.125,0.125],2)   # incorrect, no idea what's happening here
0.75

### PR DESCRIPTION
**Pull request by milankl**
fixes #769 docstrings added, tests missing

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -726,21 +726,26 @@ zscore(X::AbstractArray{<:Real}, dim::Int) = ((μ, σ) = mean_and_std(X, dim); z
 #############################
 
 """
-    entropy(p, [b])
+    isprobvec(p::AbstractVector{<:Real}) -> Bool
+
+Checks whether `p` is a probability vector, i.e. p[i] >= 0 for each index i, and sum(p) ≈ 1.
+Taken from `Distributions.isprobvec`.""" 
+isprobvec(p::AbstractVector{<:Real}) = all(x -> x ≥ zero(x), p) && isapprox(sum(p), one(eltype(p)))
+
+"""
+    entropy(p::AbstractVector{<:Real}, [b]; check::Bool=true)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by devmotion — COMMENTED** [review]
I think silently incorrect results should be avoided as much as possible. I want to mention though that based on my experience in Distributions, this PR trades off safety versus performance and convenience: Due to numerical inaccuracies the check can fail even if the user computes the input vector in such a way that in non-floating point math it would be normalized.

A simple benchmark:

master:
```julia
julia> using StatsBase, Zygote, BenchmarkTools

julia> @btime entropy($(fill(1e-5, 10^5)));
  472.918 μs (0 allocations: 0 bytes)

julia> _, pb = Zygote.pullback(entropy, fill(1e-5, 10^5));

julia> @btime $pb(1.0);
  74.084 μs (16 allocations: 781.58 KiB)
```

This PR:
```julia
julia> using StatsBase, Zygote, BenchmarkTools

julia> @btime entropy($(fill(1e-5, 10^5)));
  512.763 μs (0 allocations: 0 bytes)

julia> _, pb = Zygote.pullback(entropy, fill(1e-5, 10^5));

julia> @btime $pb(1.0);
  79.138 μs (21 allocations: 781.73 KiB)
```

A more general comment: Can you add tests?

PR #873, opened 2023-06-30
Fix `sample` with wide integer types

### LINKED ISSUE #872: `sample` throws error when used on `UInt64`
In Julia 1.9.1 with `StatsBase` v0.34.0, the following code 
```
T=UInt64;   
StatsBase.sample(T(1):T(10), T(1), replace=false,ordered=false)
StatsBase.sample(T(1):T(10), T(3), replace=false,ordered=false)
```
runs fine, but the code 
```StatsBase.sample(T(1):T(10), T(2), replace=false,ordered=false)```
returns 

> ERROR: MethodError: no method matching samplepair(::Random._GLOBAL_RNG, ::UInt64)
> 
> Closest candidates are:
>   samplepair(::AbstractRNG, ::Int64)
>    @ StatsBase ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:108
>   samplepair(::AbstractRNG, ::AbstractArray)
>    @ StatsBase ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:123
> 
> Stacktrace:
>  [1] samplepair(rng::Random._GLOBAL_RNG, a::UnitRange{UInt64})
>    @ StatsBase ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:124
>  [2] sample!(rng::Random._GLOBAL_RNG, a::UnitRange{UInt64}, x::Vector{UInt64}; replace::Bool, ordered::Bool)
>    @ StatsBase ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:514
>  [3] sample!
>    @ ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:486 [inlined]
>  [4] #sample#209
>    @ ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:543 [inlined]
>  [5] sample
>    @ ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:541 [inlined]
>  [6] #sample#210
>    @ ~/.julia/packages/StatsBase/iMkPf/src/sampling.jl:545 [inlined]
>  [7] top-level scope
>    @ REPL[79]:1
>

### PR DESCRIPTION
**Pull request by ararslan**
As observed in #872, an unordered sample without replacement with a type 64 bits or wider (notably excluding `Int64`, at least on a 64-bit system) hits a `MethodError` in `samplepair`, which had been restricted to `Int` inputs. Relaxing this restriction and adjusting the `samplepair` implementation accordingly allows these cases to succeed.

Fixes #872.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -105,12 +105,12 @@ Draw a pair of distinct integers between 1 and `n` without replacement.
 Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
 """
-function samplepair(rng::AbstractRNG, n::Int)
-    i1 = rand(rng, 1:n)
-    i2 = rand(rng, 1:n-1)
+function samplepair(rng::AbstractRNG, n::Integer)
+    i1 = rand(rng, Base.OneTo(n))
+    i2 = rand(rng, Base.OneTo(n - one(n)))
     return (i1, ifelse(i2 == i1, n, i2))
 end
-samplepair(n::Int) = samplepair(Random.GLOBAL_RNG, n)
+samplepair(n::Integer) = samplepair(Random.GLOBAL_RNG, n)
 
 """
     samplepair([rng], a)
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -106,8 +106,8 @@ Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
 """
 function samplepair(rng::AbstractRNG, n::Integer)
-    i1 = rand(rng, Base.OneTo(n))
-    i2 = rand(rng, Base.OneTo(n - one(n)))
+    i1 = rand(rng, one(n):n)
+    i2 = rand(rng, one(n):(n - one(n)))
     return (i1, ifelse(i2 == i1, n, i2))
 end
 samplepair(n::Integer) = samplepair(Random.GLOBAL_RNG, n)
--- a/test/sampling.jl
+++ b/test/sampling.jl
@@ -92,7 +92,9 @@ test_rng_use(sample, 1:10, 10)
     @test samplepair(rng, [3, 4, 2, 6, 8]) === (3, 8)
     @test samplepair(rng, [1, 2])          === (1, 2)
 
-    @test extrema(samplepair(rng, UInt128(2))) == UInt128.((1, 2))
+    onetwo = samplepair(rng, UInt128(2))
+    @test extrema(onetwo) == (1, 2)
+    @test eltype(onetwo) === UInt128
 end
 
 test_rng_use(samplepair, 1000)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: test/sampling.jl [hunk 1]
```diff
@@ -265,4 +267,15 @@ test_same(replace=false, ordered=false)
...
     end
-end
\ No newline at end of file
+end
+
+@testset "issue #872" begin
+    for T in [Int8, Int16, Int32, Int64, Int128, BigInt], f in [identity, unsigned]
+        T == BigInt && f == unsigned && continue
+        T = f(T)
+        samp = sample(T(1):T(10), T(2); replace=false, ordered=false)
```
devmotion, line 276: Does
```suggestion
        samp = sample(T(1):T(10), 2; replace=false, ordered=false)
```
also return samples of type `T`? I would have assumed that the type of the second argument does not affect the output type.

test/sampling.jl
```diff
@@ -91,6 +91,8 @@ test_rng_use(sample, 1:10, 10)
 
     @test samplepair(rng, [3, 4, 2, 6, 8]) === (3, 8)
     @test samplepair(rng, [1, 2])          === (1, 2)
+
+    @test extrema(samplepair(rng, UInt128(2))) == UInt128.((1, 2))
```
nalimilan, line 95: The `UInt128` on the RHS has no effect to check the return type.
```suggestion
    @test extrema(samplepair(rng, UInt128(2))) == (1, 2)
    @test eltype(samplepair(rng, UInt128(2))) === UInt128
```

src/sampling.jl
```diff
@@ -105,12 +105,12 @@ Draw a pair of distinct integers between 1 and `n` without replacement.
 Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
 """
-function samplepair(rng::AbstractRNG, n::Int)
-    i1 = rand(rng, 1:n)
-    i2 = rand(rng, 1:n-1)
+function samplepair(rng::AbstractRNG, n::Integer)
+    i1 = rand(rng, Base.OneTo(n))
+    i2 = rand(rng, Base.OneTo(n - one(n)))
```
nalimilan, line 110: Isn't this equivalent? I find it more standard and readable:
```suggestion
    i1 = rand(rng, 1:n)
    i2 = rand(rng, 1:(n - one(n)))
```

test/sampling.jl
nalimilan: Indeed. Could also test `sample(1:10, T(2); replace=false, ordered=false)`.

src/sampling.jl
ararslan: It'd have to be `one(n):(n - one(n))` to avoid promoting smaller types to `Int`. I find that a bit less clear than `Base.OneTo(n - one(n))` but don't feel strongly.

test/sampling.jl
ararslan: Good call, tests for both cases added!

devmotion (approved):
LGTM :+1:

PR #878, opened 2023-07-04
Use `default_rng()` instead of `GLOBAL_RNG` on Julia >= 1.3

devmotion:
I noticed that StatsBase still uses `Random.GLOBAL_RNG` as the default RNG. This PR changes it to `Random.default_rng()` on Julia >= 1.3 (it was introduced in Julia 1.3) which is the new and [documented](https://github.com/JuliaLang/julia/pull/44733) API. On newer Julia versions `Random.GLOBAL_RNG` still exists for backward compatibility but [it has been observed that `Random.GLOBAL_RNG` can lead to reduced performance](https://github.com/JuliaLang/julia/issues/48683).

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -5,7 +5,13 @@
 #
 ###########################################################
 
-using Random: Sampler, Random.GLOBAL_RNG
+using Random: Sampler
+
+@static if VERSION < v"1.3"
+    default_rng() = Random.GLOBAL_RNG
+else
+    using Random: default_rng
+end
 
 ### Algorithms for sampling with replacement
 
@@ -25,7 +31,7 @@ function direct_sample!(rng::AbstractRNG, a::UnitRange, x::AbstractArray)
     end
     return x
 end
-direct_sample!(a::UnitRange, x::AbstractArray) = direct_sample!(Random.GLOBAL_RNG, a, x)
+direct_sample!(a::UnitRange, x::AbstractArray) = direct_sample!(default_rng(), a, x)
 
 """
     direct_sample!([rng], a::AbstractArray, x::AbstractArray)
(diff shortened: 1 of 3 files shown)


src/sampling.jl
```diff
@@ -5,7 +5,13 @@
 #
 ###########################################################
 
-using Random: Sampler, Random.GLOBAL_RNG
+using Random: Sampler
+
+@static if VERSION < v"1.3"
```
ararslan, line 10:
```suggestion
if VERSION < v"1.3.0-DEV.565"
```
`@static` isn't necessary at the top level. Also I tracked down the commit that defined `default_rng` and got its version for maximum ~~pedantry~~ accuracy. Alternatively, we could use `isdefined(Random, :default_rng)` in place of a version check.

devmotion: Ah, my bad. I read a while ago about unnecessary `@static`s but still sometimes add them a bit too freely 😅 And thanks for looking up the commit, I had decided to just not care about the 1.3 dev versions 😉

test/sampling.jl
```diff
@@ -17,12 +18,17 @@ function test_rng_use(func, non_rng_args...)
...
     @test func(MersenneTwister(1), deepcopy(non_rng_args)...) ==
           func(MersenneTwister(1), deepcopy(non_rng_args)...)
-    # default RNG is Random.GLOBAL_RNG
+    # default RNG is Random.GLOBAL_RNG/Random.default_rng()
     Random.seed!(47)
     x = func(deepcopy(non_rng_args)...)
     Random.seed!(47)
     y = func(Random.GLOBAL_RNG, deepcopy(non_rng_args)...)
     @test x == y
+    if VERSION >= v"1.3"
```
devmotion, line 27:
```suggestion
    if VERSION >= v"1.3.0-DEV.565"
```

src/sampling.jl
ararslan: AFAIK it doesn't hurt to have it there, it just doesn't have any effect.

> thanks for looking up the commit

Any chance I can get to use Julia's `contrib/commit-name.sh`, I take

ararslan (approved):

PR #885, opened 2023-07-27
Make SparseArrays.jl a weak dependency

dkarrasch:
This is a little weird: one could get a sparse matrix just by using a specific keyword argument, without ever calling or loading SparseArrays.jl. So I'm not sure to which extent this might be considered breaking. In terms of future perspective, this is very desirable for downstream packages to not load SparseArrays.jl unconditionally, especially on v1.10+.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,12 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
+[weakdeps]
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[extensions]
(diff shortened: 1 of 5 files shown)


devmotion:
This seems very breaking to me: It breaks `indicatormat(...; sparse=true)` for everyone that has not loaded SparseArrays (im- or explicitly). It seems generally weird to me that this PR means that one branch of `indicatormat` would be broken by default.

So I think SparseArrays should be kept as a regular dependency or the design of `indicatormat` should be changed (e.g., it could always return a dense indicator matrix and the for the sparse version a separate function `spindicatormat` would be defined, with a stump in the main package and a proper implementation in the extension (which arguably is a bit weird and not the prototypical dispatch-based example for an extension - maybe a dispatchable type as optional argument would be better?)).

dkarrasch:
Ok, I'll close this PR then, but not without re-iterating that a dependency on SparseArrays.jl should be, in the future, reviewed very carefully, especially for fundamental packages like StatsBase.jl. 😉

Issue #868, opened 2023-06-20
mean(f, x, weights) not supported


jariji:
This method is not supported. Could it be?

```jl
julia> mean(==(1), [1,2,3])
0.3333333333333333

julia> mean([1,2,3], weights([10,20,30]))
2.3333333333333335

julia> mean(==(1), [1,2,3], weights([10,20,30]))
ERROR: MethodError: no method matching mean(::Base.Fix2{typeof(==), Int64}, ::Vector{Int64}, ::Weights{Int64, Int64, Vector{Int64}})
```

Issue #748, opened 2021-12-10
Array Weights


ParadaCarleton:
Currently, weights can only be vectors, but there's no reason to rule out array-variate weights. I came across this problem while trying to generalize MCMCChains.jl to allow for weighted samples, but having weighted matrices isn't that uncommon when working with panel data, where the probability of follow-up can differ over time. What would we need to get weighted matrices working?

rofinn:
How would the stats functions typically work for weight matrices? Would it be comparable to broadcasting the weighted stats function across both weights and data (either colwise or rowwise)?

ParadaCarleton:
> How would the stats functions typically work for weight matrices? Would it be comparable to broadcasting the weighted stats function across both weights and data (either colwise or rowwise)?

Yes; the mean across, say, the first row would just be the same as taking the weighted mean of that row, using the weights in the first row of the weight matrix.

nalimilan:
Could you show a concrete example where you would use weight matrices?

In general I think it could make sense to allow any weight arrays, but that would require carefully checking that all functions taking `AbstractWeights` are correctly written to handle them. Notably, they should check that data and weights have identical shapes to avoid incorrect or surprising behaviors (e.g. with `data=[1, 2, 3, 4]` and `weights=[1 2; 3 4]`, which should probably not be allowed).

Also note that currently weight vectors are automatically broadcasted when computing e.g. the mean over dimensions:
```julia
julia> mean(rand(3, 4), weights(rand(4)), dims=2)
3×1 Matrix{Float64}:
 0.6384569311824672
 0.3961733566525447
 0.6022251466078608
```
If we allow weight matrices, we will have to make sure they don't interfere with this behavior.

ParadaCarleton:
> Could you show a concrete example where you would use weight matrices?

The current use case is MCMC samples where I have a set of weights that can vary across multiple dimensions (namely I want `weight[2,4]` to represent the weight of the second sample of the 4th chain in a collection). 

The more general use case is any study design where you have unequal probabilities of sampling the same person across time for followup -- for instance, we might oversample participants of one group by a factor of 2 at the start, but follow up with them increasingly often as time goes on (if we expect this group to have higher dropout rates, requiring us to sample them more frequently to get enough data).

briochemc:
My 2 cents: One fairly common case where this would be useful is for spatial data, e.g., variables of a climate model.

[...]

PR #886, opened 2023-08-01
Weighted mean with function

itsdebartha:
This PR addresses #868 . Adds the method for a weighted mean of elements transformed by a function.

- Added `mean(f, itr, weights)`
- Added tests for the method

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -697,10 +697,7 @@ w = rand(n)
 mean(√, x, weights(w))
 ```
 """
-mean(f, A::AbstractArray, w::AbstractWeights) =
-_funcweightedmean(f, A, w)
-
-function _funcweightedmean(f, A::AbstractArray, w::AbstractWeights)
+function mean(f, A::AbstractArray, w::AbstractWeights)
     return sum(Broadcast.instantiate(Broadcast.broadcasted(A, w) do a_i, wg
         return f(a_i) * wg
     end)) / sum(w)


src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+n = 20
+x = rand(n)
+w = rand(n)
+mean(√, x, weights(w))
+```
+"""
+mean(f, A::AbstractArray, w::AbstractWeights; dims::Union{Colon,Int}=:) =
+    _mean(f.(A), w, dims)
+
+function mean(f, A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
```
ParadaCarleton, line 704: This is a slightly more generic version of the same thing, which is less likely to require maintenance in the future (if we add additional keyword arguments to `mean`). I suggest using this pattern when you can.
```suggestion
function mean(f, A::AbstractArray, w::UnitWeights; kwargs...)
```

src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+    a = (dims === :) ? length(A) : size(A, dims)
+    a != length(w) && throw(DimensionMismatch("Inconsistent array dimension."))
+    return mean(f.(A), dims=dims)
```
ParadaCarleton, line 707: Avoid memory allocation by using `mean(f, A)` instead of `mean(f.(A))`. Remember that `f.(A)` creates an extra array, which is slow. Memory access is usually the biggest bottleneck on modern CPUs. `mean(f.(A))` is 2 separate operations: The first one creates a new array, `f.(A)`, and the second calculates its mean. `mean(f, A)` calculates the mean of `(f(x) for x in A)` directly, as one operation, without creating a new array.
```suggestion
    return mean(f, A; dims)
```
I'd also suggest making it slightly more generic, as 
```suggestion
    return mean(f, A; kwargs...)
```

src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
     return mean(A, dims=dims)
 end
 
+"""
+    mean(f, A::AbstractArray, w::AbstractWeights[, dims::Int])
```
ParadaCarleton, line 686:
```suggestion
    mean(f, A::AbstractArray, w::AbstractWeights[; dims])
```
`dims` shouldn't be required to be an integer.

src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+# Examples
+```julia
+n = 20
+x = rand(n)
+w = rand(n)
+mean(√, x, weights(w))
+```
+"""
+mean(f, A::AbstractArray, w::AbstractWeights; dims::Union{Colon,Int}=:) =
+    _mean(f.(A), w, dims)
```
ParadaCarleton, line 702: Avoid memory allocation by using `mean(f, A)` instead of `mean(f.(A))`. Remember that `f.(A)` creates an extra array, which is slow. Memory access is usually the biggest bottleneck on modern CPUs. `mean(f.(A))` is 2 separate operations: The first one creates a new array, `f.(A)`, and the second calculates its mean. `mean(f, A)` calculates the mean of `(f(x) for x in A)` directly, as one operation, without creating a new array.
```suggestion
    _mean(f, A; dims)
```
I'd also suggest making it slightly more generic, as 
```suggestion
    _mean(f, A; kwargs...)
```
(See below for more details.)

test/weights.jl
```diff
@@ -270,6 +270,27 @@ end
         @test mean(a, f(wt), dims=3) ≈ sum(a.*reshape(wt, 1, 1, length(wt)), dims=3)/sum(wt)
         @test_throws ErrorException mean(a, f(wt), dims=4)
     end
+
+    @test mean(√, [1:3;], f([1.0, 1.0, 0.5])) ≈ 1.3120956
+    @test mean(√, 1:3, f([1.0, 1.0, 0.5]))    ≈ 1.3120956
+    @test mean(√, [1 + 2im, 4 + 5im], f([1.0, 0.5])) ≈ 1.60824421 + 0.88948688im
```
ParadaCarleton, line 276: Should try this with multiple functions, other than the square root, and an integer array as well.

src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+
+function mean(f, A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
+    a = (dims === :) ? length(A) : size(A, dims)
+    a != length(w) && throw(DimensionMismatch("Inconsistent array dimension."))
```
ParadaCarleton, line 706: No need to special-case here:
```suggestion
    if size(A, dims) != length(w)
        throw(DimensionMismatch("Inconsistent array dimension.")) 
    end
```

test/weights.jl
```diff
@@ -270,6 +270,27 @@ end
...
+
+    for wt in ([1.0, 1.0, 1.0], [1.0, 0.2, 0.0], [0.2, 0.0, 1.0])
+        @test mean(√, a, f(wt), dims=1) ≈ sum(sqrt.(a).*reshape(wt, length(wt), 1, 1), dims=1)/sum(wt)
+        @test mean(√, a, f(wt), dims=2) ≈ sum(sqrt.(a).*reshape(wt, 1, length(wt), 1), dims=2)/sum(wt)
+        @test mean(√, a, f(wt), dims=3) ≈ sum(sqrt.(a).*reshape(wt, 1, 1, length(wt)), dims=3)/sum(wt)
+        @test_throws ErrorException mean(√, a, f(wt), dims=4)
+    end
```
ParadaCarleton, line 283: Besides the above comments, there's the question of whether we want to support this; I don't think we've settled on an interface for #748.

src/weights.jl
itsdebartha: Removing the special case will return errors, as `size` doesn't accept `:` as a dimension type. I think it's best to keep it as it is...

test/weights.jl
itsdebartha: In that case, I think it's best to follow this same format, for now, and add possibly more tests, with other functions...

test/weights.jl
itsdebartha: Yep, I agree...

src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+
+# Examples
+```julia
+n = 20
+x = rand(n)
+w = rand(n)
+mean(√, x, weights(w))
+```
+"""
+mean(f, A::AbstractArray, w::AbstractWeights; kwargs...) =
```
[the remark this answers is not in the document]
itsdebartha: Alright...

src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+    mean(collect(Iterators.map(f, A)), w; kwargs...)
```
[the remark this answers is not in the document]
itsdebartha: I will run into type ambiguity with `mean(f, ::AbstractArray)`. As suggested by you, I am going with `broadcast` instead of `Iterators.map`...

src/weights.jl
```diff
@@ -682,6 +682,31 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+# Examples
+```julia
+n = 20
+x = rand(n)
+w = rand(n)
+mean(√, x, weights(w))
+```
+"""
+mean(f, A, w::AbstractWeights; kwargs...) =
+    mean(broadcast(f, A), w; kwargs...)
```
ParadaCarleton, line 702: Like I said, check [the method for weighted `sem`](https://github.com/JuliaStats/StatsBase.jl/blob/0e8e4aaaafdbacf17424e93b48e00c03674885f4/src/scalarstats.jl#L509) to see how to use `Broadcasting.broadcasted` correctly. `broadcast` is not the same thing as `Broadcasting.broadcasted`. The latter is lazy.

ParadaCarleton (changes requested):

src/weights.jl
```diff
@@ -682,6 +682,35 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+mean(√, x, weights(w))
+```
+"""
+mean(f, A::AbstractArray, w::AbstractWeights) =
+_funcweightedmean(f, A, w)
+
+function _funcweightedmean(f, A::AbstractArray, w::AbstractWeights)
+    return sum(Broadcast.broadcasted(f, A, w) do f, a_i, wg
+        return f(a_i) * wg
+    end) / sum(w)
```
devmotion, line 706: You should use `Broadcast.instantiate`. Moreover, to me it seems this should be changed to
```suggestion
    return sum(Broadcast.instantiate(Broadcast.broadcasted(A, w) do a_i, wg
        return f(a_i) * wg
    end)) / sum(w)
```

ParadaCarleton: Using `Broadcast.instantiate` causes extra allocations, no?

devmotion: I haven't tried. But without `Broadcast.instantiate` it will be slow and fall back to Cartesian indexing. See, e.g., https://github.com/JuliaLang/julia/pull/31020 ("we require Broadcast.instantiate for fast reduce").

src/weights.jl
```diff
@@ -682,6 +682,35 @@ function mean(A::AbstractArray, w::UnitWeights; dims::Union{Colon,Int}=:)
...
+# Examples
+```julia
+n = 20
+x = rand(n)
+w = rand(n)
+mean(√, x, weights(w))
+```
+"""
+mean(f, A::AbstractArray, w::AbstractWeights) =
+_funcweightedmean(f, A, w)
```
devmotion, line 701: I don't remember, was there any particular reason for introducing `_funcweightedmean` instead of implementing two methods for `mean`?

itsdebartha: No there wasn't any such reason. I am removing this function, and implementing it as a method for `mean`

itsdebartha:
@devmotion @ParadaCarleton gentle ping, is this PR ready to merge?

PR #892, opened 2023-09-16
Add more aqua tests

charleskawczynski:
This PR adds and fixes more Aqua tests.

### DIFF
--- a/test/aqua.jl
+++ b/test/aqua.jl
@@ -1,0 +1,36 @@
+using Test
+using StatsBase
+using Aqua
+
+@testset "Aqua tests (performance)" begin
+    # This tests that we don't accidentally run into
+    # https://github.com/JuliaLang/julia/issues/29393
+    # Aqua.test_unbound_args(StatsBase)
+    ua = Aqua.detect_unbound_args_recursively(StatsBase)
+    @test length(ua) == 0
+
+    # See: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1750
+    # Test that we're not introducing method ambiguities across deps
+    ambs = Aqua.detect_ambiguities(StatsBase; recursive = true)
+    # Uncomment for debugging:
+    # for method_ambiguity in ambs
+    #     @show method_ambiguity
+    # end
+    @test length(ambs) ≤ 5
+    potentially_pirated_methods = Aqua.Piracy.hunt(StatsBase)
+    # Uncomment for debugging:
+    # for pirate in potentially_pirated_methods
+    #     @show pirate
+    # end
+    @test length(potentially_pirated_methods) ≤ 27
+end
+
+@testset "Aqua tests (additional)" begin
+    Aqua.test_undefined_exports(StatsBase)
+    Aqua.test_stale_deps(StatsBase)
+    Aqua.test_deps_compat(StatsBase)
+    Aqua.test_project_extras(StatsBase)
+    Aqua.test_project_toml_formatting(StatsBase)
+end
+
+nothing
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3,7 +3,7 @@ using Dates
 using LinearAlgebra
 using Random
 
-tests = ["ambiguous",
+tests = ["aqua",
          "weights",
          "moments",
          "scalarstats",
--- a/Project.toml
+++ b/Project.toml
@@ -26,6 +26,7 @@ StatsAPI = "1.2"
 julia = "1"
 
 [extras]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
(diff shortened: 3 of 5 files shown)


devmotion:
I still think we should go with https://github.com/JuliaStats/StatsBase.jl/pull/866. But it seems not everyone agrees.

charleskawczynski:
> I still think we should go with #866. But it seems not everyone agrees.

Yeah, that’s exactly why I thought it would be good to add tests in this way. My hope is that this is not controversial, and the benefit is that things don’t get worse

charleskawczynski:
I have no idea what’s going on with the test though

test/runtests.jl
```diff
@@ -3,7 +3,7 @@ using Dates
 using LinearAlgebra
 using Random
```
nalimilan, line 4: Do failures go away if you add this here?
```julia
using Statistics
```

test/aqua.jl
```diff
@@ -0,0 +1,36 @@
...
+    @test length(ua) == 0
+
+    # See: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1750
+    # Test that we're not introducing method ambiguities across deps
+    ambs = Aqua.detect_ambiguities(StatsBase; recursive = true)
+    # Uncomment for debugging:
+    # for method_ambiguity in ambs
+    #     @show method_ambiguity
+    # end
+    @test length(ambs) ≤ 5
```
nalimilan, line 19: Is there a simple way to check which functions are ambiguous?

test/runtests.jl
charleskawczynski: No, the number of failures stays the same.

test/aqua.jl
charleskawczynski: We can use `Aqua.test_ambiguities(StatsBase; recursive = true, exclude = [StatsBase.TestStat, Base.:(==)])`. However, this doesn't test the number of ambiguities, only the functions.. Testing the number of ambiguities is nice, IMO, because we can even write this as:
```julia
    @test length(ambs) ≤ 5
    @test_broken length(ambs) < 5
```
so that we're alerted that it's fixed. It's also just easier to spread across the ecosystem than to collect all the functions after printing them all out..

charleskawczynski: I'm happy to change it if it means getting the PR merged.

test/aqua.jl
```diff
@@ -0,0 +1,36 @@
...
+    potentially_pirated_methods = Aqua.Piracy.hunt(StatsBase)
+    # Uncomment for debugging:
+    # for pirate in potentially_pirated_methods
+    #     @show pirate
+    # end
+    @test length(potentially_pirated_methods) ≤ 27
+end
+
+@testset "Aqua tests (additional)" begin
+    Aqua.test_undefined_exports(StatsBase)
```
devmotion, line 29: Just run `Aqua.test_all` and disable the ones you're handling separately? Then we won't miss out on new tests potentially added in the future.

test/aqua.jl
```diff
@@ -0,0 +1,36 @@
...
+    Aqua.test_stale_deps(StatsBase)
+    Aqua.test_deps_compat(StatsBase)
+    Aqua.test_project_extras(StatsBase)
+    Aqua.test_project_toml_formatting(StatsBase)
```
devmotion, line 33: There's a bug currently and hence this should not be tested generally (see https://github.com/JuliaTesting/Aqua.jl/issues/105). Hence I tend to use
```suggestion
Aqua.test_all(
    StatsBase;
    project_toml_formatting=VERSION >= v"1.7" || !haskey(ENV, "GITHUB_ACTIONS"),
)
    Aqua.test_project_toml_formatting(StatsBase; )
```

test/aqua.jl
```diff
@@ -0,0 +1,36 @@
...
+@testset "Aqua tests (performance)" begin
+    # This tests that we don't accidentally run into
+    # https://github.com/JuliaLang/julia/issues/29393
+    # Aqua.test_unbound_args(StatsBase)
+    ua = Aqua.detect_unbound_args_recursively(StatsBase)
+    @test length(ua) == 0
+
+    # See: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1750
+    # Test that we're not introducing method ambiguities across deps
+    ambs = Aqua.detect_ambiguities(StatsBase; recursive = true)
```
devmotion, line 14: Why do we want to test ambiguities recursively? It's generally argued against that in https://github.com/JuliaTesting/Aqua.jl/issues/77 (and against the default setting of testing `Base` and `Core` as well in the Aqua test suite).

test/aqua.jl
```diff
@@ -0,0 +1,36 @@
+using Test
+using StatsBase
+using Aqua
+
+@testset "Aqua tests (performance)" begin
+    # This tests that we don't accidentally run into
+    # https://github.com/JuliaLang/julia/issues/29393
+    # Aqua.test_unbound_args(StatsBase)
+    ua = Aqua.detect_unbound_args_recursively(StatsBase)
+    @test length(ua) == 0
```
devmotion, line 10: Can we just run the full Aqua test suite (see also my other comment below)? It's already included there by default: https://github.com/JuliaTesting/Aqua.jl/blob/3d5ed9fa2c915bbb06b2ef9037d57029d18bc8b5/src/unbound_args.jl#L36

test/aqua.jl
charleskawczynski: While I do like the idea of not missing out on new tests in the future, my issue with `test_all` is that, if something does fail, then the simplest fix is often to call the individual test functions.

test/aqua.jl
charleskawczynski: Alternatively, I'm fine with commenting this out.

test/aqua.jl
charleskawczynski: I figured that it was more thorough.. I'm happy to change this to `ambs = Aqua.detect_ambiguities(StatsBase)`

test/aqua.jl
charleskawczynski: Yeah, I guess we can change it to the test function. I have my own gripe with that interface, tbh. https://github.com/JuliaTesting/Aqua.jl/issues/180

charleskawczynski:
This doesn't seem to have any traction. Closing.

PR #894, opened 2023-09-25
Implement `Base.allequal` and `Base.allunique` for weight vectors

devmotion:
In this PR I propose to define `allequal` and `allunique` for weight vectors by falling back on `allequal(weights.values)` and `allunique(weights.values)` (this means optimizations for types of `weights.values` will be exploited instead of falling back to the generic iterative algorithm) and implement the optimizations for `UnitWeights`.

In a project I wanted to check `allequal(weights)` to avoid an expensive computation for this special case. However, to make this performant for `UnitWeights` currently I have to write `weights isa UnitWeights || allequal(weights)`.

### DIFF
--- a/src/weights.jl
+++ b/src/weights.jl
@@ -28,12 +28,6 @@ isempty(wv::AbstractWeights) = isempty(wv.values)
 size(wv::AbstractWeights) = size(wv.values)
 Base.axes(wv::AbstractWeights) = Base.axes(wv.values)
 
-# https://github.com/JuliaLang/julia/pull/43354
-if VERSION >= v"1.8.0-DEV.1494" # 98e60ffb11ee431e462b092b48a31a1204bd263d
-    Base.allequal(wv::AbstractWeights) = allequal(wv.values)
-end
-Base.allunique(wv::AbstractWeights) = allunique(wv.values)
-
 Base.IndexStyle(::Type{<:AbstractWeights{S,T,V}}) where {S,T,V} = IndexStyle(V)
 
 Base.dataids(wv::AbstractWeights) = Base.dataids(wv.values)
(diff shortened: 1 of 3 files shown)


src/weights.jl
```diff
@@ -313,6 +319,12 @@ length(wv::UnitWeights) = wv.len
 size(wv::UnitWeights) = tuple(length(wv))
 Base.axes(wv::UnitWeights) = tuple(Base.OneTo(length(wv)))
 
+# https://github.com/JuliaLang/julia/pull/43354
+if VERSION >= v"1.8.0-DEV.1494" # 98e60ffb11ee431e462b092b48a31a1204bd263d
```
ararslan, line 323: Not a big deal at all but it might be nice to combine the two conditionals so that there's only one place that needs editing once we require 1.8 or later.

src/weights.jl
```diff
@@ -28,6 +28,12 @@ isempty(wv::AbstractWeights) = isempty(wv.values)
 size(wv::AbstractWeights) = size(wv.values)
 Base.axes(wv::AbstractWeights) = Base.axes(wv.values)
 
+# https://github.com/JuliaLang/julia/pull/43354
+if VERSION >= v"1.8.0-DEV.1494" # 98e60ffb11ee431e462b092b48a31a1204bd263d
+    Base.allequal(wv::AbstractWeights) = allequal(wv.values)
+end
+Base.allunique(wv::AbstractWeights) = allunique(wv.values)
```
ararslan, line 35: I agree that `allequal` would be a good addition but I'm a bit ambivalent about `allunique`. Doesn't seem bad to have but I can't really think of a situation where it'd be useful. Do you have one in mind?

devmotion: No, I don't have a use-case in mind, I think it won't be used a lot. I mainly added it since it complements `allequal` (and while the fallback would be sufficient, as for `allequal`, one could potentially benefit if an optimied method exists for `weights.values`).

ararslan (approved):

src/weights.jl
devmotion: Fixed in https://github.com/JuliaStats/StatsBase.jl/pull/894/commits/f2421a46af446f90bdb9bad3173f6dfd249f3f45.

PR #900, opened 2023-10-23
rewrite _momentX methods in more functional style

kalmarek:
depending on the width of vector instruction this is up to 4× faster (N=100), the old implementation matches for k = 6;

@devmotion

### DIFF
--- a/src/moments.jl
+++ b/src/moments.jl
@@ -161,86 +161,56 @@ end
 
 ##### General central moment
 function _moment2(v::AbstractArray{<:Real}, m::Real; corrected=false)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += z * z
-    end
-    varcorrection(n, corrected) * s
+    s = sum(x->abs2(x-m), v, init=zero(m))
+    return varcorrection(length(v), corrected) * s
 end
 
 function _moment2(v::AbstractArray{<:Real}, wv::AbstractWeights, m::Real; corrected=false)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += (z * z) * wv[i]
-    end
-
-    varcorrection(wv, corrected) * s
+    s = sum(i -> (@inbounds abs2(v[i] - m) * wv[i]), eachindex(v), init=zero(m))
+    return varcorrection(wv, corrected) * s
 end
 
 function _moment3(v::AbstractArray{<:Real}, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += z * z * z
-    end
-    s / n
+    s = sum(x->(x-m)^3, v, init=zero(m))
+    return s/length(v)
 end
 
 function _moment3(v::AbstractArray{<:Real}, wv::AbstractWeights, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += (z * z * z) * wv[i]
-    end
-    s / sum(wv)
+    s = sum(
+        i -> (@inbounds (z = (v[i] - m); z * z * z * wv[i])),
+        eachindex(v),
+        init=zero(m),
+    )
+    return s/sum(wv)
 end
 
 function _moment4(v::AbstractArray{<:Real}, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += abs2(z * z)
-    end
-    s / n
+    s = sum(x-> (z = x-m; abs2(z*z)), v, init=zero(m))
+    return s/length(v)
 end
 
 function _moment4(v::AbstractArray{<:Real}, wv::AbstractWeights, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += abs2(z * z) * wv[i]
-    end
-    s / sum(wv)
+    s = sum(
+        i -> (@inbounds (z = (v[i] - m); abs2(z * z) * wv[i])),
+        eachindex(v),
+        init=zero(m),
+    )
+    return s/sum(wv)
 end
 
 function _momentk(v::AbstractArray{<:Real}, k::Int, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += (z ^ k)
-    end
-    s / n
+    s = sum(x -> (x - m)^k, v, init=zero(m))
+    return s/length(v)
 end
 
 function _momentk(v::AbstractArray{<:Real}, k::Int, wv::AbstractWeights, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += (z ^ k) * wv[i]
-    end
-    s / sum(wv)
+    s = sum(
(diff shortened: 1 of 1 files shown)


andreasnoack:
I think this is a good change. It requires the minimum Julia version to be 1.6 but I think that is okay.

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
 ##### General central moment
 function _moment2(v::AbstractArray{<:Real}, m::Real; corrected=false)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += z * z
-    end
-    varcorrection(n, corrected) * s
+    s = sum(x->abs2(x-m), v, init=zero(m))
```
devmotion, line 164: I think we should use an improved `init` value if possible - in many cases the result of `sum(...)` won't be of the same type as `zero(m)`.

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
 function _moment2(v::AbstractArray{<:Real}, wv::AbstractWeights, m::Real; corrected=false)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += (z * z) * wv[i]
-    end
-
-    varcorrection(wv, corrected) * s
+    s = sum(i -> (@inbounds abs2(v[i] - m) * wv[i]), eachindex(v), init=zero(m))
```
devmotion, line 169: Similarly, I think we need a different `init` here. Additionally, `@inbounds` is unsafe here as `i` might not be an actual index of `wv`.

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
 
 function _moment3(v::AbstractArray{<:Real}, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += z * z * z
-    end
-    s / n
+    s = sum(x->(x-m)^3, v, init=zero(m))
```
devmotion, line 174: Here as well.

src/moments.jl
devmotion: For instance, we could use
```suggestion
    init = (zero(eltype(v)) - zero(m))^2
    s = sum(x->(x-m)^2, v; init=init)
```

src/moments.jl
devmotion:
```suggestion
    init = (zero(eltype(v)) - zero(m))^2 * zero(eltype(wv))
    s = sum(i -> (v[i] - m)^2 * wv[i], eachindex(v, wv); init=init)
```

src/moments.jl
devmotion:
```suggestion
    init = (zero(eltype(v)) - zero(m))^3
    s = sum(x->(x-m)^3, v; init=init)
```

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += (z * z * z) * wv[i]
-    end
-    s / sum(wv)
+    s = sum(
+        i -> (@inbounds (z = (v[i] - m); z * z * z * wv[i])),
+        eachindex(v),
+        init=zero(m),
+    )
```
devmotion, line 183: Same comments as above:
```suggestion
    init = (zero(eltype(v)) - zero(m))^3 * zero(eltype(wv))
    s = sum(
        i -> (v[i] - m)^3 * wv[i],
        eachindex(v, wv);
        init=init,
    )
```

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
 
 function _moment4(v::AbstractArray{<:Real}, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += abs2(z * z)
-    end
-    s / n
+    s = sum(x-> (z = x-m; abs2(z*z)), v, init=zero(m))
```
devmotion, line 188:
```suggestion
    init = (zero(eltype(v)) - zero(m))^4
    s = sum(x->(x-m)^4, v; init=init)
```

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += abs2(z * z) * wv[i]
-    end
-    s / sum(wv)
+    s = sum(
+        i -> (@inbounds (z = (v[i] - m); abs2(z * z) * wv[i])),
+        eachindex(v),
+        init=zero(m),
+    )
```
devmotion, line 197:
```suggestion
    init = (zero(eltype(v)) - zero(m))^4 * zero(eltype(wv))
    s = sum(
        i -> (v[i] - m)^4 * wv[i],
        eachindex(v, wv);
        init=init,
    )
```

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
 
 function _momentk(v::AbstractArray{<:Real}, k::Int, m::Real)
-    n = length(v)
-    s = 0.0
-    for i = 1:n
-        @inbounds z = v[i] - m
-        s += (z ^ k)
-    end
-    s / n
+    s = sum(x -> (x - m)^k, v, init=zero(m))
```
devmotion, line 202:
```suggestion
    init = (zero(eltype(v)) - zero(m))^k
    s = sum(x -> (x - m)^k, v; init=init)
```

src/moments.jl
```diff
@@ -161,87 +161,57 @@ end
...
-    for i = 1:n
-        @inbounds z = v[i] - m
-        @inbounds s += (z ^ k) * wv[i]
-    end
-    s / sum(wv)
+    s = sum(
+        i -> (@inbounds (z = (v[i] - m); z^k * wv[i])),
+        eachindex(v),
+        init=zero(m),
+    )
```
devmotion, line 211:
```suggestion
    init = (zero(eltype(v)) - zero(m))^k * zero(eltype(wv))
    s = sum(
        i -> (v[i] - m)^k * wv[i],
        eachindex(v, wv);
        init=init,
    )
```

devmotion:
Since it's unspecified whether `init` is used at all for non-empty collections, an alternative might be to use the suggested `init` values only for empty collections and use `sum` without `init` keyword argument for non-empty collections (as a side effect this would not even require to bump the Julia version it seems).

src/moments.jl
kalmarek: @devmotion if `v` and `wv` do not have the same indices `eachindex(v, wv)` will throw and therefore it seems that `@inbounds` here should be safe. How about this:
```suggestion
    s = if isempty(v)
         (zero(eltype(v)) - zero(m))^2*zero(eltype(wv))
    elseif eachindex(v) == eachindex(wv)
        sum(i -> (@inbounds (v[i] - m)^2 * wv[i]), eachindex(v, wv))
    else
        sum(i -> ((v_ - m)^2 * wv_) for (v_, wv_) in zip(v, wv))
    end
```

kalmarek:
> Since it's unspecified whether `init` is used at all for non-empty collections, an alternative might be to use the suggested `init` values only for empty collections and use `sum` without `init` keyword argument for non-empty collections (as a side effect this would not even require to bump the Julia version it seems).

I'd proceed with this version then. let me know what do you think about the suggestion I made above

src/moments.jl
andreasnoack: @devmotion bump

kalmarek: @andreasnoack I don't want to be negative (I'm really writing this with a smile :D ), but given the pace we had with github.com/JuliaStats/HypothesisTests.jl/pull/124 I expect this to be merged ~ 2030 :D

But by any means I'm not in a hurry.

PR #902, opened 2023-11-09
Addition of Yule-Walker Analysis Method

josemanuel22:
This pull request proposes the addition of a method to perform Yule-Walker analysis in Julia, analogous to the yule_walker method available in Python's statsmodels.regression.linear_model. The aim is to extend Julia's statistical analysis capabilities, specifically for autoregressive processes, by implementing a method that estimates the parameters of an AR model using the Yule-Walker equations.

https://stackoverflow.com/questions/77107822/how-to-do-a-yule-walker-analysis-in-julia/77117406#77117406

Details:

1. The new method will calculate the autoregressive coefficients and the variance of the error term.
2. The implementation will focus on efficiency and accuracy, leveraging Julia's strengths in handling mathematical operations.
3. Unit tests will be added to ensure the method's reliability and to maintain the integrity of the package.
I look forward to feedback and suggestions for improvements.

### DIFF
--- a/src/signalcorr.jl
+++ b/src/signalcorr.jl
@@ -629,3 +629,77 @@ end
 function pacf(x::AbstractVector{<:Real}, lags::AbstractVector{<:Integer}; method::Symbol=:regression)
     vec(pacf(reshape(x, length(x), 1), lags, method=method))
 end
+
+
+"""
+    yule_walker(x::Vector{<:Real};
+                order::Int64=1,
+                method="adjusted",
+                df::Union{Nothing,Int64}=nothing,
+                inv=false,
+                demean=true,
+    )
+
+Estimate AutoRegressive (AR) parameters using the Yule-Walker equations. This function
+estimates AR parameters using the Yule-Walker equations. It supports different ACF
+estimation methods (adjusted or maximum likelihood) and can optionally return the inverse
+of the Toeplitz matrix.
+
+Example:
+```julia
+data = [0.1, 0.2, 0.3, 0.4, 0.5]
+order = 2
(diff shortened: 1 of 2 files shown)


josemanuel22:
@andreasnoack Thank you  very much for your comment. I had not seen it, but it seems that this function is not documented, nor is it exported. It is for internal use, maybe it could be documented and exported. What do you think?

andreasnoack:
Sorry for being a bit brief here. I made the comment a bit in a hurry. The point was mostly that you can compute the Yule-Walker estimates efficiently by reusing existing functions, i.e.

## Yule-Walker
```julia
julia> -StatsBase.durbin(autocor(x, 1:4))
4-element Vector{Float64}:
  1.2600631466569998
 -0.4558455606978694
 -0.17590708455408563
  0.011391047373061776
```

## Least-squares
```julia
julia> [ones(length(x) - k) [x[k - j + i] for i in 1:length(x)-k, j in 1:k]]\x[k+1:end]
5-element Vector{Float64}:
 28.267780791960924
  1.2864783021078277
 -0.48686426630545226
 -0.16910686502522418
  0.017892893292120397
```

Thinking a bit more about it, I think a function for estimating coefficients might also be better located in a dedicated time series package. The `acf` and `pacf` functions are a bit more like summary statistics but if they were added today, I guess they might also go in to a dedicated time series package instead of this one.

josemanuel22:
Perfect, thank you for the clarification. I am not aware of any time series package in Julia that has this functionality, as it seems they resort to `StatsBase.jl`. I'm not sure if it would make sense to open a PR there. With this, I believe we can close this PR.

PR #903, opened 2023-11-20
Don't call `raise_dimerror`, which doesn't exist

yha:


devmotion:
Fixed by https://github.com/JuliaStats/StatsBase.jl/pull/922.

Issue #910, opened 2023-12-28
[feature request] allow `transform` to avoid Z-score transforming when sigma=0


SimonEnsemble:
my design matrix `Φ` has a column of one's to handle the intercept. so anytime I Z-score transform, I get `NaN`s after dividing by a zero variance. would be nice if `transform` had an option to handle this, ie. not transform when sigma=0.0.

```
feature_transform = fit(ZScoreTransform, Φ, dims=1)
Φ̂ = StatsBase.transform(feature_transform, Φ)
```

SimonEnsemble:
eg. in scikit-learn's `StandardScaler`: "If a variance is zero, we can’t achieve unit variance, and the data is left as-is, giving a scaling factor of 1."

https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler

PR #911, opened 2023-12-29
allow `transform` to avoid Z-score transforming when `sigma=0`

tylerjthomas9:
Ref: https://github.com/JuliaStats/StatsBase.jl/issues/910

Here is my attempt to prevent NaNs when transforming data with 0 variance.

Here is the result:

```julia
julia> X = ones(5,8)
5×8 Matrix{Float64}:
 1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0

julia> feature_transform = fit(ZScoreTransform, X, dims=1)
ZScoreTransform{Float64, Vector{Float64}}(8, 1, [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0], [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0])

julia> StatsBase.transform(feature_transform, X)
5×8 Matrix{Float64}:
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
```

### DIFF
--- a/src/transformations.jl
+++ b/src/transformations.jl
@@ -123,9 +123,21 @@ function fit(::Type{ZScoreTransform}, X::AbstractMatrix{<:Real};
     else
         throw(DomainError(dims, "fit only accept dims to be 1 or 2."))
     end
-    return ZScoreTransform(l, dims, (center ? vec(m) : similar(m, 0)),
-                                    (scale ? vec(s) : similar(s, 0)))
+    if scale
+        s_vec = vec(s)
+        # avoid z-score transforming when sigma=0
+        if any(s_vec .== 0.0)
+            zero_variance_indices = s_vec .== 0.0
+            s_vec[zero_variance_indices] .= 1.0
+        end
+    else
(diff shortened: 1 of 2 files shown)


andreasnoack:
I don't like the discontinuity that this introduces. You'd end up with a huge difference between the transform of `X` and the smallest perturbation of `X.

PR #913, opened 2024-01-01
Faster algorithm for ordered sampling with replacement

Tortar:
This is based on a classical result for example described here https://stats.stackexchange.com/questions/348358/a-fast-uniform-order-statistic-generator (and in the reference of the most upvoted answer). I wasn't able to find a reference describing its modification for sampling a finite population, but I adapted it to such a case.

In particular, the performance increase is substantial, more than 5 times faster when the performance increase stabilize, e.g.

This PR:

```julia
julia> using StatsBase, BenchmarkTools

julia> a = [1:1000;];

julia> @btime sample($a, 10^2, ordered=true);
  453.017 ns (2 allocations: 1.75 KiB)

julia> @btime sample($a, 10^6, ordered=true);
  3.491 ms (4 allocations: 15.26 MiB)
```

Main:

```julia
julia> using StatsBase, BenchmarkTools

julia> a = [1:1000;];

julia> @btime sample($a, 10^2, ordered=true);
  1.564 μs (2 allocations: 1.75 KiB)

julia> @btime sample($a, 10^6, ordered=true);
  21.273 ms (4 allocations: 15.26 MiB)
```

the switching point between this algorithm and the one implemented in main is set at `k=10` because I found that empirically at that point the timings were almost equal.

Numerically it should be stable enough, but let me know what you think

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -522,10 +522,10 @@ function sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray;
 
     if replace  # with replacement
         if ordered
-            if k < 10
-                sample_ordered!(direct_sample!, rng, a, x)
-            else
+            if k > 10
                 uniform_orderstat_sample!(rng, a, x)
+            else
+                sample_ordered!(direct_sample!, rng, a, x)
             end
         else
             direct_sample!(rng, a, x)
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -101,6 +101,29 @@ sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray,
         sampler!(rng, a, wv, x)
     end
 
+"""
+This algorithm generates a sorted sample with replacement by
+adapting the classic result that the cumulative sum of n+1 
+exponentially-distributed random numbers divided by the overall sum 
+(dropping the last) is a sorted sample from a uniform[0,1]
+"""
+function uniform_orderstat_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
+    1 == firstindex(a) == firstindex(x) ||
+        throw(ArgumentError("non 1-based arrays are not supported"))
+    Base.mightalias(a, x) &&
+        throw(ArgumentError("output array x must not share memory with input array a"))
+    n = length(a)
+    k = length(x)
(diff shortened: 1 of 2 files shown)


Tortar:
Test that is failing is using the previous algorithm at lines 83-84 of sampling.jl

```julia
83    aa = Int.(sample(r, 10; ordered=true))
84    check_sample_wrep(aa, (3, 12), 0; ordered=true, rev=rev)
```

so I think it is unrelated right?

Tortar:
Swapping the tests make them work, but not totally sure what was the cause actually

Tortar:
Actually it seems to me that the way tests are written is not very good because no rng is set, indeed trying to run those tests 100 times, sometimes gets to failure anyway (before and after this pr), should we set an rng?

Tortar:
closing because I want to do same more experimentation with the algorithm before proposing it, you can find them here:https://github.com/Tortar/SortedRands.jl

Tortar:
gentle bump, since #927 was merged, what about taking a look at another speed-up? :D

src/sampling.jl
```diff
@@ -101,6 +101,29 @@ sample_ordered!(sampler!, rng::AbstractRNG, a::AbstractArray,
...
+"""
+function uniform_orderstat_sample!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
+    1 == firstindex(a) == firstindex(x) ||
+        throw(ArgumentError("non 1-based arrays are not supported"))
+    Base.mightalias(a, x) &&
+        throw(ArgumentError("output array x must not share memory with input array a"))
+    n = length(a)
+    k = length(x)
+    exp_rands = randexp(rng, k)
+    sorted_rands = accumulate!(+, exp_rands, exp_rands)
```
devmotion, line 118: This is not safe. From the docstring of `accumulate!`:
```julia
  │ Warning
  │
  │  Behavior can be unexpected when any mutated argument shares memory with any other argument.
```

Tortar: right, changed in `cumsum`, it is more or less the same performance even with it, I updated the timing in the main comment

Tortar:
do you have any more review comments @devmotion? :-)

Tortar:
Bump

Tortar:
Bump2

Issue #904, opened 2023-11-29
Inf and NaN in weights


aplavin:
Currently, `weights` throws an error whenever it encounters non-finite values: `ArgumentError: weights cannot contain Inf or NaN values`. That's highly surprising IMO, basically all other operations propagate NaNs and allow Infs – both basic arithmetics and aggregations.
```julia
# totally sensible and unambiguous results:
julia> median([1,2,3,Inf])
2.5
julia> mean([1,2,3,Inf])
Inf
julia> mean([1,2,3,Inf], weights([1,1,1,1]))
Inf

julia> mean([1,2,3,NaN])
NaN
julia> median([1,2,3,NaN])
NaN

# and suddenly:
julia> mean([1,2,3,4], weights([1,1,Inf,1]))
ERROR: ArgumentError: weights cannot contain Inf or NaN values
julia> mean([1,2,3,4], weights([1,1,NaN,1]))
ERROR: ArgumentError: weights cannot contain Inf or NaN values
```

I looked up as found that this error was introduced after the issue https://github.com/JuliaStats/StatsBase.jl/issues/671. I think throwing this exception denies perfectly sensible behavior, so that weighted aggregations worked exactly as their unweighted counterparts – by propagating NaNs. And the throw in `weights()` should be removed, maybe adding it in specific aggregations where Infs/NaNs really do not make any sense (eg, mean and median should work).

Numpy does propagate, of course:
```python
In [2]: np.average([1,2,3], weights=[1,np.nan,3])
Out[2]: nan
```

devmotion:
I wasn't involved in the previous discussions in #671 and https://github.com/JuliaStats/StatsBase.jl/pull/814, but my impression was that the errors are intentional and the previous behaviour (no errors) was considered a bug. I think I tend to agree with this view since weights are supposed to sum to 1 (or at least it should be able to normalize them to such weights), which is not possible if a weight is `NaN` or `Inf`.

devmotion:
I don't see a clear inconsistency, there's no restriction on the values and `Inf`/`NaN` propagate in the same way for weighted operations:
```julia
julia> using StatsBase

julia> mean([1,2,3,Inf])
Inf

julia> mean([1,2,3,Inf], weights([1,1,1,1]))
Inf

julia> mean([1,2,3,NaN])
NaN

julia> mean([1,2,3,NaN], weights([1,1,1,1]))
NaN
```
The restriction only applies to the weights themselves because they have to sum up to 1.

PR #914, opened 2024-01-21
functions that only support finite weights now throw errors for non-finites

aplavin:
AbstractWeights can contain any values, including NaNs – and many StatsBase functions work with them just fine. But not all, for some like `sample` it doesn't make sense. So, here I make these functions throw an exception instead of returning a potentially incorrect result.

Some docstrings, like the one for `quantile`, even explicitly say that

> An error is raised if `w` contains any `NaN` values.

but this wasn't the case without this PR.

Ideally, NaNs should also be supported by concrete weights types included into StatsBase (https://github.com/JuliaStats/StatsBase.jl/issues/904), but this PR is independent of that.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -729,7 +729,7 @@ function alias_sample!(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights, 
     # sampling
     s = Sampler(rng, 1:n)
     for i = 1:length(x)
-        j = rand(rng, wsum)
+        j = rand(rng, s)
         x[i] = rand(rng) < ap[j] ? a[j] : a[alias[j]]
     end
     return x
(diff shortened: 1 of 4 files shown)


src/sampling.jl
```diff
@@ -590,6 +590,7 @@ Optionally specify a random number generator `rng` as the first argument
 function sample(rng::AbstractRNG, wv::AbstractWeights)
     1 == firstindex(wv) ||
         throw(ArgumentError("non 1-based arrays are not supported"))
+    isfinite(sum(wv)) || throw(ArgumentError("only finite weights are supported"))
```
devmotion, line 593: A general comment: Could we avoid re-computing `sum(wv)` below? It's not guaranteed that `sum(wv)` is cheap for any `AbstractWeights` type.

devmotion: Isn't the sparse documentation an additional argument for avoiding re-evaluating `sum(wv)` and not relying on such fast paths?

devmotion (approved):

devmotion:
The PR can't be merged? "This branch has conflicts that must be resolved"

andreasnoack (approved):

PR #920, opened 2024-03-08
Codecov: set `fail_ci_if_error` to `false` iff the PR is made from a fork

DilumAluthge:


DilumAluthge:
@devmotion What do you think of this approach?

devmotion:
I think this might be useful given the current instability of codecov. I'd like to avoid surprising coverage drops in PRs from forks though - maybe if someone with admin permission additionally makes codecov a mandatory check, this would not be possible? On the other hand, of course, this would require that at least one of the codecov uploads succeeds.

DilumAluthge:
You could set up GitHub Merge Queue on this repo, and then make Codecov a required status check. The GitHub Merge Queue jobs (`merge_group` jobs) will have access to the `CODECOV_TOKEN`.

andreasnoack:
Hasn't the Codecov action stabilized so much at this point that it's not worth adding extra complexity to our script?

Issue #849, opened 2023-02-16
Change corkendall to be multi-threaded and handle missing values


PGS62:
Back in Feb 2021, and with help from @nalimilan, I re-wrote `corkendall` to improve its speed by a factor of about seven (see [# 634](https://github.com/JuliaStats/StatsBase.jl/issues/634)).

More recently I have been working on a multi-threaded version, and have released a version which I'd like to propose should replace the current version here in StatsBase. As well as being multi-threaded, the new version has support for missing values, though it does not do this via `pairwise`.

The speed up is broadly equal to the number of cores available. No speed up is achievable if both `x` and `y` are vectors.

Could I request that a StatsBase maintainer looks at the code and lets me know their thoughts?

https://github.com/PGS62/KendallTau.jl 

Philip

nalimilan:
Sorry for the delay. Sounds interesting. A few questions/remarks:
- How does this interact with https://github.com/JuliaStats/StatsBase.jl/pull/850? AFAICT the code mentioned here does not benefit from this optimization, meaning it's slower when there's a single thread, right?
- Isn't there a slowdown when the number of rows or columns is small? Generally, spawning tasks has a significant overhead.
- I'd rather avoid adding a `skipmissing` argument to `corspearman` and `corkendall`, as then for consistency we would need to add it to `cor`, `cov`, `var`, etc. Instead I think you can define specialized methods for `pairwise(::typeof(corspearman), ...)` and so on. Anyway, it's probably better to handle this in a separate PR to keep things simple.

PGS62:
Thanks for your questions.

- There's basically no interaction between this issue and #850 . I noticed that `corspearman` (for which there's no handling of missing values and no use of threads) is inefficient since it makes more calls than needs be to `tiedrank`. Fixing that, while being careful to maintain the desired handling of `nan`s, was straightforward, and I think performance is improved for all possible inputs. The amended `corspearman` does not use threads or handle `missing`s.
[...]

PR #923, opened 2024-03-31
Make `pairwise` multi-threaded and make `corkendall` and `corspearman` wrap `pairwise`.

PGS62:
This pull request follows up on #849. In that discussion I envisaged amending `corkendall` to be multi-threaded. Instead I have done the following:

1. Made `pairwise` and `pairwise!` multi-threaded, and in the case of `skipmissing = :pairwise`, greatly reduced allocations which is needed for good multi-threaded performance. A nice illustration of the performance improvements this leads to is obtained with `f = LinearAlgebra.dot`. The example [here](https://github.com/PGS62/KendallTau.jl?tab=readme-ov-file#pairwise-performance) shows a 31.5 times speedup on a 12 core PC.
2. `corkendall` now simply wraps `pairwise` but with added methods of `_pairwise!` for `typeof(corkendall)` to retain the speedups to `corkendall` that were a result of #634.
3. `corspearman` also wraps `pairwise` with added methods of `_pairwise!` for `typeof(corspearman)`. The purpose of these added methods is to reduce the number of calls to `tiedrank` from `2nm` to `2(n+m)` when `n` and `m` are respectively the number of columns in the arguments `x` and `y` to the call to `corspearman(x,y...`.

Since `corkendall` and `corspearman` now wrap `pairwise` they have a `skipmissing` argument, which introduces an additional inconsistency between `corkandall`/`corspearman ` and `cov`/`cor`. 

I developed the new code in my package [KendallTau](https://github.com/PGS62/KendallTau.jl). Please see the README file for further examples of the performance improvements achieved, which for all but the smallest input data exceed the number of cores on the PC (thanks to reduced allocations).

I have amended the tests as needed, and I believe that test coverage is as close to 100% as possible.

I apologise that there is quite a lot for maintainers to review, and I'll be happy to answer questions.


test/pairwise.jl
```diff
@@ -23,24 +29,35 @@ arbitrary_fun(x, y) = cor(x, y)
...
         @test res isa Matrix{Float64}
 
         @inferred pairwise(f, x, y)
-
-        @test_throws Union{ArgumentError,MethodError} pairwise(f, [Int[]], [Int[]])
-        @test_throws Union{ArgumentError,MethodError} pairwise!(f, zeros(1, 1), [Int[]], [Int[]])
+        if throwsforzerolengthinput
+            # @test_throws Union{ArgumentError,MethodError} pairwise(f, [Int[]], [Int[]])
+            @test_throws CompositeException pairwise(f, [Int[]], [Int[]])
+            #@test_throws Union{ArgumentError,MethodError} pairwise!(f, zeros(1, 1), [Int[]], [Int[]])
```
andreasnoack, line 58: Is this commented out intentionally? There are a few other tests that are commented out.

src/rankcorr.jl
```diff
@@ -11,209 +11,624 @@
...
-    yrank = tiedrank(y)
-    for j = 1:n
-        Xj = view(X, :, j)
-        if any(isnan, Xj)
-            C[j,1] = NaN
-        else
-            Xjrank = tiedrank(Xj)
-            C[j,1] = cor(Xjrank, yrank)
+function corspearman(x::AbstractMatrix, y::AbstractVector; skipmissing::Symbol=:none)
+    return corspearman(x, reshape(y, (length(y), 1)); skipmissing)
```
andreasnoack, line 41: I think the 1.0 tests are failing because of this way of passing the keyword argument. I'd be okay with bumping the minimum version for StatsBase since 1.0 is ancient but maybe it's better to just make it `skipmissing=skipmissing` here for now.

andreasnoack (commented):
This is really great. Thanks for contributing such a PR. I'm happy with it but lets keep if open for a couple of days to see if anybody else would like to comment.

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
+        dest .= transpose(dest′)
+        return dest
     end
-    if symmetric
-        m, n = size(dest)
-        @inbounds for j in 1:n, i in (j+1):m
-            dest[i, j] = dest[j, i]
+
+    #cor and friends are special-cased.
+    iscor = (f in (corkendall, corspearman, cor))
```
devmotion, line 15: It is a bit unfortunate that this means other functions can't benefit from the improvements (?) since it is not extendable. Maybe an additional argument would be better?

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
-            dest[i, j] = f(xi, yj)
-        end
+function _pairwise!(::Val{:none}, f, dest::AbstractMatrix{V}, x, y,
+    symmetric::Bool) where {V}
+
+    nr, nc = size(dest)
+
+    # Swap x and y for more efficient threaded loop.
+    if nr < nc
+        dest′ = reshape(dest, size(dest, 2), size(dest, 1))
```
devmotion, line 8: As a minor comment, maybe it would be better to use a different name here - on my mobile phone `dest′` looked identical to `dest'`.

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
+        _pairwise!(Val(:none), f, dest′, y, x, symmetric)
+        dest .= transpose(dest′)
```
devmotion, line 10: Is this safe even though `dest′` and `dest` point to the same memory? For instance, even `LinearAlgebra.transpose!` warns about and does not support if the source and destination array have overlapping memory regions.

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
-    if symmetric
-        m, n = size(dest)
-        @inbounds for j in 1:n, i in (j+1):m
-            dest[i, j] = dest[j, i]
+
+    #cor and friends are special-cased.
+    iscor = (f in (corkendall, corspearman, cor))
+    (iscor || f == cov) && (symmetric = x === y)
+    #cov(x) is faster than cov(x, x)
+    (f == cov) && (f = ((x, y) -> x === y ? cov(x) : cov(x, y)))
```
devmotion, line 18: This seems like an optimization that maybe should be performed on the caller side?

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
+
+    Threads.@threads for subset in equal_sum_subsets(nr, Threads.nthreads())
```
devmotion, line 20: Is it clear that this way to split the iteration is (somewhat) optimal? For instance, I would assume that it would be beneficial if each iteration/task would operate on contiguous memory regions which doesn't seem to be the case here.

I wonder if it is even beneficial to pre-partition the iteration space here in smaller subsets or whether one would achieve the same performance gains by only performing a threaded loop over the upper-triangular indices.

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
+        for j in subset
+            for i = 1:(symmetric ? j : nc)
```
devmotion, line 22: Typically, it is more performant to iterate over columns in the outer loop and rows in the inner loop.

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
+                # For performance, diagonal is special-cased
+                if iscor && i==j && y[i] === x[j] && V !== Union{}
+                    dest[j, i] = V === Missing ? missing : 1.0
```
devmotion, line 25: I think this optimization might hide errors, e.g. in the case of a 1x1 matrix where the destination has element type `Missing` but `f` would actually return a non-missing value. So I think the only safe optimization might be
```suggestion
                    dest[j, i] = oneunit(f_mn)
```
where `f_mn` is the value of some (probably off-diagonal) `f(x[m], y[n])`.

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
+                else
+                    dest[j, i] = f(x[j], y[i])
+                end
+                symmetric && (dest[i, j] = dest[j, i])
```
devmotion, line 29: I wonder if it would be better to use `LinearAlgebra.copytri!` outside of the loop to avoid accessing vastly different memory regions.

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
 end
 
-function check_vectors(x, y, skipmissing::Symbol)
+#Input validation for both pairwise and pairwise!
+function check_vectors(x, y, skipmissing::Symbol, symmetric::Bool)
+
+    if symmetric && x !== y
+        throw(ArgumentError("symmetric=true only makes sense passing " *
+                            "a single set of variables (x === y)"))
+    end
```
devmotion, line 42: Why? In principle, it would also be correct to pass `y = copy(x)`, wouldn't it?

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
-                    # TODO: float() will not be needed after JuliaLang/Statistics.jl#61
-                    dest[i, j] = float(cor(xi))
+    nr, nc = size(dest)
+    m = length(x) == 0 ? 0 : length(first(x))
+
+    # Swap x and y for more efficient threaded loop.
+    if nr < nc
+        dest′ = reshape(dest, size(dest, 2), size(dest, 1))
+        _pairwise!(Val(:pairwise), f, dest′, y, x, symmetric)
+        dest .= transpose(dest′)
```
devmotion, line 86: Same comments as above (variable name + transposing with overlapping memory regions).

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
+        return dest
+    end
+
+    #cor and friends are special-cased.
+    iscor = (f in (corkendall, corspearman, cor))
+    (iscor || f == cov) && (symmetric = x === y)
```
devmotion, line 92: Again, unfortunately this can't be extended to other functions.

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
+    #cov(x) is faster than cov(x, x)
+    (f == cov) && (f = ((x, y) -> x === y ? cov(x) : cov(x, y)))
```
devmotion, line 94: This optimization should be performed by `cov`, shouldn't it?

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
+
+    nmtx = promoted_nmtype(x)[]
+    nmty = promoted_nmtype(y)[]
+
+    Threads.@threads for subset in equal_sum_subsets(nr, Threads.nthreads())
+        for j in subset
```
devmotion, line 100: Same comments as above.

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
+            scratch_fx = task_local_vector(:scratch_fx, nmtx, m)
+            scratch_fy = task_local_vector(:scratch_fy, nmty, m)
```
devmotion, line 102: Why is a task-local vector needed here? Isn't it sufficient to allocate the vectors outside of the `for j in subset` loop?

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
+            for i = 1:(symmetric ? j : nc)
+                if iscor && i == j && y[i] === x[j] && V !== Union{} && V!== Missing
+                    dest[j, i] = 1.0
```
devmotion, line 105: Also here, I think it would be more correct to use something like
```suggestion
                    dest[j, i] = oneunit(f_mn)
```
(which would also allow to remove the checks for `V !== Union{}` and `V !== Missing`)

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
+    return _pairwise!(Val(:none), f, dest, handle_listwise(x, y)..., symmetric)
 end
 
-function _pairwise!(f, dest::AbstractMatrix, x, y;
-                    symmetric::Bool=false, skipmissing::Symbol=:none)
-    if !(skipmissing in (:none, :pairwise, :listwise))
-        throw(ArgumentError("skipmissing must be one of :none, :pairwise or :listwise"))
-    end
+function _pairwise!(f, dest::AbstractMatrix, x, y; symmetric::Bool=false,
+    skipmissing::Symbol=:none)
```
devmotion, line 121: Unrelated formatting change?

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
 
-    x′ = x isa Union{AbstractArray, Tuple, NamedTuple} ? x : collect(x)
-    y′ = y isa Union{AbstractArray, Tuple, NamedTuple} ? y : collect(y)
+    x′ = x isa Union{AbstractArray,Tuple,NamedTuple} ? x : collect(x)
+    y′ = y isa Union{AbstractArray,Tuple,NamedTuple} ? y : collect(y)
```
devmotion, line 124:
```suggestion
    x′ = x isa Union{AbstractArray, Tuple, NamedTuple} ? x : collect(x)
    y′ = y isa Union{AbstractArray, Tuple, NamedTuple} ? y : collect(y)
```

src/pairwise.jl
```diff
@@ -46,76 +74,59 @@ function check_vectors(x, y, skipmissing::Symbol)
...
     m = length(x′)
     n = length(y′)
 
     size(dest) != (m, n) &&
         throw(DimensionMismatch("dest has dimensions $(size(dest)) but expected ($m, $n)"))
-
```
devmotion, line 118:
```suggestion

```

src/pairwise.jl
```diff
@@ -134,7 +145,7 @@ end
 
 # Identical to `Base.promote_typejoin` except that it uses `promote_type`
 # instead of `typejoin` to combine members of `Union` types
-function promote_type_union(::Type{T}) where T
+function promote_type_union(::Type{T}) where {T}
```
devmotion, line 148:
```suggestion
function promote_type_union(::Type{T}) where T
```

src/pairwise.jl
```diff
@@ -178,7 +189,7 @@ function _pairwise(::Val{skipmissing}, f, x, y, symmetric::Bool) where {skipmiss
         # but using `promote_type` rather than `promote_typejoin`)
         U = mapreduce(typeof, promote_type, dest)
         # V is inferred (contrary to U), but it only gives an upper bound for U
-        V = promote_type_union(Union{T, Tsm})
+        V = promote_type_union(Union{T,Tsm})
```
devmotion, line 192:
```suggestion
        V = promote_type_union(Union{T, Tsm})
```

src/pairwise.jl
```diff
@@ -149,14 +160,14 @@ function promote_type_union(::Type{T}) where T
 end
 
 function _pairwise(::Val{skipmissing}, f, x, y, symmetric::Bool) where {skipmissing}
-    x′ = x isa Union{AbstractArray, Tuple, NamedTuple} ? x : collect(x)
-    y′ = y isa Union{AbstractArray, Tuple, NamedTuple} ? y : collect(y)
+    x′ = x isa Union{AbstractArray,Tuple,NamedTuple} ? x : collect(x)
+    y′ = y isa Union{AbstractArray,Tuple,NamedTuple} ? y : collect(y)
```
devmotion, line 164:
```suggestion
    x′ = x isa Union{AbstractArray, Tuple, NamedTuple} ? x : collect(x)
    y′ = y isa Union{AbstractArray, Tuple, NamedTuple} ? y : collect(y)
```

src/pairwise.jl
```diff
@@ -149,14 +160,14 @@ function promote_type_union(::Type{T}) where T
...
     m = length(x′)
     n = length(y′)
 
-    T = Core.Compiler.return_type(f, Tuple{eltype(x′), eltype(y′)})
-    Tsm = Core.Compiler.return_type((x, y) -> f(disallowmissing(x), disallowmissing(y)),
-                                     Tuple{eltype(x′), eltype(y′)})
+    T = Core.Compiler.return_type(f, Tuple{eltype(x′),eltype(y′)})
```
devmotion, line 168:
```suggestion
    T = Core.Compiler.return_type(f, Tuple{eltype(x′), eltype(y′)})
```

src/pairwise.jl
```diff
@@ -149,14 +160,14 @@ function promote_type_union(::Type{T}) where T
...
+    Tsm = Core.Compiler.return_type((x, y) -> f(handle_pairwise(x, y)...),
+        Tuple{eltype(x′),eltype(y′)})
```
devmotion, line 170:
```suggestion
                                     Tuple{eltype(x′),eltype(y′)})
```

src/pairwise.jl
```diff
@@ -307,34 +314,130 @@ julia> pairwise(cor, eachcol(y), skipmissing=:pairwise)
...
+Return a pair `(a,b)`, filtered copies of `(x,y)`, in which elements `x[i]` and
+`y[i]` are excluded if  `ismissing(x[i])||ismissing(y[i])`.
+"""
+function handle_pairwise(x::AbstractVector, y::AbstractVector;
+    scratch_fx::AbstractVector=similar(x, nonmissingtype(eltype(x))),
+    scratch_fy::AbstractVector=similar(y, nonmissingtype(eltype(y))))
+
+    axes(x) == axes(y) || throw(DimensionMismatch("x and y have inconsistent dimensions"))
+    lb = first(axes(x, 1))
+    j = lb - 1
```
devmotion, line 384: This seems a bit dangerous since `j` will be indexed into `scratch_fx` and `scratch_fy` but not `x` (or `y`) and initially `j` will be an invalid index (could be resolved be reordering `j += 1` in the loop).

src/pairwise.jl
```diff
@@ -307,34 +314,130 @@ julia> pairwise(cor, eachcol(y), skipmissing=:pairwise)
...
+julia> StatsBase.equal_sum_subsets(30,5)
+5-element Vector{Vector{Int64}}:
+ [30, 21, 20, 11, 10, 1]
+ [29, 22, 19, 12, 9, 2]
+ [28, 23, 18, 13, 8, 3]
+ [27, 24, 17, 14, 7, 4]
+ [26, 25, 16, 15, 6, 5]
+```
+"""
+function equal_sum_subsets(n::Int, num_subsets::Int)::Vector{Vector{Int}}
```
devmotion, line 416: In case this function is kept (see comments above), is it necessary to return a vector of vectors or could one return an iterator of vectors instead?

test/pairwise.jl
PGS62: The commented-out test at line 56 was replaced by the test at line 57 (necessary since errors in threaded code burble up as `CompositeException`). So I will delete the commented out test at 56 and similar occurrences.

src/pairwise.jl
PGS62: No problem, I will avoid using `′`.

src/pairwise.jl
PGS62: But `iscor` is only `true` for a small set of functions for which we want to set the on-diagonal elements to 1.0, so your scenario "`f` would actually return a non-missing value" cannot apply given that `iscor` is `true`.

src/pairwise.jl
PGS62: Thanks for the suggestion. I will have a go at incorporating it.

src/pairwise.jl
PGS62: Here I'm following the behaviour of `pairwise` and `pairwise!` since they were added to StatsBase in May 2021. See [here](https://github.com/JuliaStats/StatsBase.jl/blob/4c9e55929eb8409de0e3dd1b03d22b7db62e0cda/src/pairwise.jl#L249) and [here](https://github.com/JuliaStats/StatsBase.jl/blob/4c9e55929eb8409de0e3dd1b03d22b7db62e0cda/src/pairwise.jl#L311).

I suspect that the justification for writing line 39 as is rather than `if symmetric && x != y` relates to the (presumably) better speed of testing `x===y` versus testing `x==y`.

src/pairwise.jl
PGS62: As I remarked on line 17, this is not in fact new behaviour, though it's implemented in a different way from the current code.

src/pairwise.jl
PGS62: I'm pretty sure that a task-local vector is needed, but looking at the code I think you are correct that lines 101 and 102 can be moved outside that loop. I will check.

src/pairwise.jl
PGS62: turns out I also needed to change the last line of the function which is now:

```
function handle_pairwise(x::AbstractVector, y::AbstractVector;
    scratch_fx::AbstractVector=similar(x, nonmissingtype(eltype(x))),
    scratch_fy::AbstractVector=similar(y, nonmissingtype(eltype(y))))

    axes(x) == axes(y) || throw(DimensionMismatch("x and y have inconsistent dimensions"))
    lb = first(axes(x, 1))
    j = lb
    @inbounds for i in eachindex(x)
        if !(ismissing(x[i]) || ismissing(y[i]))
            scratch_fx[j] = x[i]
            scratch_fy[j] = y[i]
            j += 1
        end
    end

    return view(scratch_fx, lb:(j-1)), view(scratch_fy, lb:(j-1))
end
```

src/pairwise.jl
PGS62: I'm reasonably sure that `equal_sum_subsets` works well in the sense of doing a good job of load balancing. The only reason it currently returns a vector of vectors is that's what I know how to do, so I would need to read up on iterators.

src/pairwise.jl
PGS62: For whitespace I'm largely relying on the Julia extension for VSCode with the command F1 -> Format Document, and that gives no space after the comma in this example. Is there an alternative approach to automatic code formatting that's preferred for StatsBase?

src/pairwise.jl
PGS62: Again this is VS Code Format Document doing its stuff.

src/pairwise.jl
PGS62: A result of my preference for VSCode "Format Document" - see remarks above.

src/pairwise.jl
PGS62: A result of my preference for VSCode "Format Document" - see remarks above.

src/pairwise.jl
PGS62: Again, this is a result of my enthusiasm for the VSCode Format Document command.

src/pairwise.jl
PGS62: I agree, but the practice of using `_cov` in place of `cov` for improved performance when calculating the diagonal terms predates my changes - in StatsBase 0.34.2 it happens [here](https://github.com/JuliaStats/StatsBase.jl/blob/4c9e55929eb8409de0e3dd1b03d22b7db62e0cda/src/pairwise.jl#L318), so I would prefer to leave as is.

src/pairwise.jl
PGS62: `dest′` has now become `dest2`, but the `′` character also appears is also used for variables `x′` and `y′` and I'd rather keep that as is.

src/pairwise.jl
PGS62: The situation I have in mind is when `f` is expensive to calculate (as is the case for `corkendall` and `corspearman`) and so the important thing is to ensure that each thread makes the same number of calls to `f`, so as to have all threads finish their work near simultaneously. I was able to see the benefit of `equal_sum_subsets`  when calculating a 5,000 x 5,000 kendall correlation matrices simply by observing the loading of cores in Windows task manager. My prior approach had a nasty tail of only one or two cores doing any work. `equal_sum_subsets` solved that with about a 25% speedup (if I recall correctly).

I also experimented with with using [ChunkSplitters](https://github.com/JuliaFolds2/ChunkSplitters.jl) with `split = : scatter` but that approach does not do quite as good a job of equal division between threads of calls to `f`. 

Surely considerations of memory layout would only come into play if `f` is a _really_ fast function? `LinearAlgebra.dot` might be a good example and I was very happy with the performance in that case.

In conclusion I doubt that there exists a significantly faster approach.

src/pairwise.jl
PGS62: The situation I have in mind is when `f` is expensive to calculate (as is the case for `corkendall` and `corspearman`) and so the important thing is to ensure that each thread makes the same number of calls to `f`, so as to have all threads finish their work near simultaneously. I was able to see the benefit of `equal_sum_subsets`  when calculating a 5,000 x 5,000 kendall correlation matrices simply by observing the loading of cores in Windows task manager. My prior approach had a nasty tail of only one or two cores doing any work. `equal_sum_subsets` solved that with about a 25% speedup (if I recall correctly).

I also experimented with with using [ChunkSplitters](https://github.com/JuliaFolds2/ChunkSplitters.jl) with `split = : scatter` but that approach does not do quite as good a job of equal division between threads of calls to `f`. 

Surely considerations of memory layout would only come into play if `f` is a _really_ fast function? `LinearAlgebra.dot` might be a good example and I was very happy with the performance in that case.

In conclusion I doubt that there exists a significantly faster approach.

src/pairwise.jl
PGS62: Yes, that worked, and is committed.

src/pairwise.jl
PGS62: I agree that this looks suspect, and perhaps it's a bit strange that all tests pass with this potential "unsafeness". I will replace the call to `reshape` with one to `similar` which fixes things I think.

PGS62: Fixed in commit baf193b:

test/pairwise.jl
PGS62: Fixed in commit 1a4d064

.vscode/settings.json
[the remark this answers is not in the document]
PGS62: Fixed in commit 893d4b6

src/pairwise.jl
PGS62: Fixed in commit e2c9ef5

src/pairwise.jl
PGS62: Fixed in commit 2119804

src/pairwise.jl
PGS62: For the threaded loops in pairwise.jl this is fixed in commit 5da5943. I fought shy of restructuring the threaded loops in rankcorr.jl but I did swap the `i` and `j` loop variables so that `i` is now the row index and `j` the column index. The code was weirdly hard to understand when those were the wrong way around.

src/pairwise.jl
PGS62: Fixed in commit e2c9ef5

src/pairwise.jl
PGS62: In commit a1d9ae9 I have replaced the function `equal_sum_subsets` with an iterator `EqualSumSubsets` whose element type is another iterator `TwoStepRange`. It seems to work well and makes a small reduction in allocations. But please review as this is the first time I've written code to create an iterator.

PGS62: Ah, didn't work on 32 bit. Maybe I need to replace use of Int64 with Int?

src/rankcorr.jl
PGS62: I've done some more investigation:
Other popular packages (the compat section of Project.toml) have the following minimum versions:
1.9.4 for Statistics
1.6 for Plots, DataFrames, Revise, Optim, ForwardDiff, OhMyREPL, CSV 
1.3 for Distributions
1 for StatsBase
So upping the minimum version for StatsBase seems sensible.

But as I remarked above, `eachcol` is crucial to how `corkendall` and `corspearman` are mere wrappers to `pairwise`. `eachcol` was introduced in Julia 1.1, so I imagined that upping the minimum Julia version for StatsBase to 1.1 would suffice. But no - the code appears to require Julia 1.9.4, and as of now does not work with Julia 1.8.5. For example the method `check_vectors` fails because:

```
# Julia 1.9.4
julia> keys((eachcol([1 2;3 4]))[2])
2-element LinearIndices{1, Tuple{Base.OneTo{Int64}}}:
 1
 2
```
```
# Julia 1.8.5
julia> keys((eachcol([1 2;3 4]))[2])
ERROR: MethodError: no method matching getindex(::Base.Generator{Base.OneTo{Int64}, Base.var"#242#243"{Matrix{Int64}}}, ::Int64)
Stacktrace:
 [1] top-level scope
   @ REPL[1]:1
```
So the alternatives are:

- A. We increase the minimum version to 1.9.4, which happens to be that for Statistics, but is more modern than the current LTS.
-  B. I find a workaround so that this PR requires either no increase in the minimum version number or an increase to 1.6 which is more in line with other popular packages.

PGS62:
I've now made changes so that `pairwise`, `corkendall` and `corspearman` work with all tests passing on Julia 1.6 or later. I've also changed `ci.yml` to test against 1.6 instead of 1.0 and the compat section of Project.toml now puts the minimum Julia version as 1.6.

Would this meet with approval from the StatsBase maintainers?

src/rankcorr.jl
PGS62: It's now the case that increasing the minimum version to 1.6 would work. See [here](https://github.com/JuliaStats/StatsBase.jl/pull/923#issuecomment-2040711012).

src/pairwise.jl
devmotion: Good point. I opened https://github.com/JuliaStats/Statistics.jl/pull/166.

PGS62:
The failed checks after commit [46d5655] are all in the codecov step, with error:
```
Commit creating failed: {"detail":"Tokenless has reached GitHub rate limit. Please upload using a token: https://docs.codecov.com/docs/adding-the-codecov-token. Expected available in 1549 seconds."}
```
IIUC ([this page](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit)) I need to wait and try again.

PGS62:
It's been a while... May I ask why this PR has not been merged? Is it:
a) An oversight?
b) Concerns about upping the minimum Julia version to 1.6?
c) Something else?

frankier:
I know people have many obligations and PRs/emails get ignored sometimes, but this is a bit of an unfortunate situation. @PGS62 was rebuffed from registering a new package, but now this PR sits languishing.

src/rankcorr.jl
```diff
@@ -11,209 +11,628 @@
...
+            return missing
+        end
+    end
+
+    @inbounds for i in eachindex(y)
+        scratch_py[i] = y[permx[i]]
+    end
+
+    if missing isa T || missing isa V
+        sortedx, permutedy = handle_pairwise(sortedx, scratch_py; scratch_fx=scratch_fx, scratch_fy=scratch_fy)
```
nalimilan, line 592: This assumes pairwise deletion but what about `skipmissing=:listwise`?

src/pairwise.jl
```diff
@@ -1,25 +1,53 @@
...
-        @inbounds for j in 1:n, i in (j+1):m
-            dest[i, j] = dest[j, i]
+
+    #cor and friends are special-cased.
+    diag_is_1 = (f in (corkendall, corspearman, cor))
+    (diag_is_1 || f == cov) && (symmetric = x === y)
+    #cov(x) is faster than cov(x, x)
+    (f == cov) && (f = ((x, y) -> x === y ? cov(x) : cov(x, y)))
+
+    Threads.@threads for subset in EqualSumSubsets(nc, Threads.nthreads())
```
nalimilan, line 20: (Continuing https://github.com/JuliaStats/StatsBase.jl/pull/923#discussion_r1548297436)

This approach is clever, but I also wonder whether it wouldn't be possible to achieve the same result with less code. If you identify a cell in the upper triangle using an integer index, you can just split these equally across threads. Then you can compute back what row and column the integer index corresponds to (e.g. via ` i, j = Tuple(CartesianIndices(dest))`). If that's simpler you could even attribute indices from the full table to tasks, and skip those from the lower triangle from inside tasks. Am I missing something?

It could also make sense to spawn more tasks than `Threads.nthreads` when the input matrix is really large. Indeed, we don't have any guarantee that there are `Threads.nthreads()` idle threads. We may be operating within a context which is already threaded, and for example one thread, or two thirds of the threads may already be busy; another app may also be using cores. So when we're sure that the overhead of spawning threads is small compared to the total cost of the operation it's worth spawning lots of tasks that can then be attributed to idle threads as appropriate. This is what we do in DataFrames.jl, with a threshold so that we spawn a new task for each batch of N rows.

nalimilan (commented):
Sorry for the delay! This is a significant refactoring so there are many things to discuss. My main remarks are (see details in inline comments):
- I think the threading strategy can be simplified a lot. Task local storage shouldn't be needed (it's almost never the case), using `@spawn` is the recommended pattern and it composes with nested threaded calls contrary to `@threads`. Also I wonder whether we can find a simpler solution of splitting indices across tasks than `EqualSumSubsets` (see [here](https://github.com/JuliaStats/StatsBase.jl/pull/923#discussion_r1545855866)).
- I wouldn't add a `skipmissing` argument to `corkendall` and `corspeaman`, as this makes the code significantly more verbose and the signature inconsistent with `cor`. We have `pairwise` for that. (And anyway it's better to make separate PRs for distinct features as in my experience large PRs are more likely to become unmanageable and stall.)
- There are lots of unrelated formatting changes, please revert these so that the diff is cleaner and easier to review. Also in the changed code the style isn't always consistent with StatsBase's (in particular, arguments should be aligned in method definitions, no empty initial lines, space after `#` in comments).

PGS62:
My thanks to @nalimilan for the response! As it happens I've been travelling for the last three weeks with almost no internet (Nepal), but I will reply as soon as I can.

PR #930, opened 2024-05-26
Don't run CI on nightly

andreasnoack:
It makes the CI status unreliable. We can test upcoming Julia releases in separate PRs.


devmotion (commented):
I'm generally fine with this PR. Instead of nightly we could test e.g. the latest tagged version instead: https://github.com/JuliaStats/Distributions.jl/pull/1861#pullrequestreview-2079786947

andreasnoack:
I've updated the PR such that prereleases are now tested

PR #942, opened 2024-10-31
Docs `sample`: clarify that `a` is an array

fonsp:
Hey! Beginner here!

I tried to sample from a distribution and it took a while to get that `sample` does not work on distributions. This should fix that!

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -461,9 +461,9 @@ seqsample_d!(a::AbstractArray, x::AbstractArray) = seqsample_d!(default_rng(), a
 
 ### Interface functions (poly-algorithms)
 """
-    sample([rng], a, [wv::AbstractWeights])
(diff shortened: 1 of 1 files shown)


src/sampling.jl
```diff
@@ -463,7 +463,7 @@ seqsample_d!(a::AbstractArray, x::AbstractArray) = seqsample_d!(default_rng(), a
 """
     sample([rng], a, [wv::AbstractWeights])
 
-Select a single random element of `a`. Sampling probabilities are proportional to
+Select a single random element of array `a`. Sampling probabilities are proportional to
```
andreasnoack, line 466: Alternatively, we should adjust the signature above to match the actual signature

fonsp: you decide!

andreasnoack: Then let's go with that. Would you mind doing it here?

andreasnoack (approved):

PR #946, opened 2024-12-12
Fix the compat entry for `Statistics`

DilumAluthge:
This is necessary on older versions of Julia.


devmotion (approved):

DilumAluthge:
Thanks Andreas and David!

Issue #97, opened 2014-10-27
mad computes estimation of normal std


gummif:
I find it strange that the `mad!` function does not compute the MAD but rather `1.4826*MAD` (estimate of the std of a normal distribution). Is there a reason for this? The doc does not even mention this.

PR #949, opened 2025-01-29
fix #905: use the correct mad() normalization

### LINKED ISSUE #905: StatsBase.mad computes the wrong median absolute deviation
By default, `StatsBase.mad` computes a "normalized" median absolute deviation where the normalization factor is equal to `quantile(Normal(), 3/4)`. The default behavior of this function should just be the mad, not an estimator for the standard deviation. This behavior is unexpected and is further not applicable for most distributions, anyway.

aplavin:
Make `mad()` actually follow the definition of median absolute deviation as it is generally defined (eg https://en.wikipedia.org/wiki/Median_absolute_deviation, scipy, matlab, mathematica, ...).

Obviously breaking. Are there any other breaking fixes that could go into the next StatsBase version?

See https://github.com/JuliaStats/StatsBase.jl/issues/905, https://github.com/JuliaStats/StatsBase.jl/issues/97, https://github.com/JuliaStats/StatsBase.jl/issues/347 for more context.

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -536,7 +536,7 @@ If `normalize` is set to `true`, the MAD is multiplied by
 `1 / quantile(Normal(), 3/4) ≈ 1.4826`, in order to obtain a consistent estimator
 of the standard deviation under the assumption that the data is normally distributed.
 """
-function mad(x; center=nothing, normalize::Union{Bool, Nothing}=nothing)
+function mad(x; center=nothing, normalize::Bool=false)
     mad!(Base.copymutable(x); center=center, normalize=normalize)
 end
 
@@ -552,19 +552,14 @@ of the standard deviation under the assumption that the data is normally distrib
 """
 function mad!(x::AbstractArray;
               center=median!(x),
-              normalize::Union{Bool,Nothing}=false)
+              normalize::Bool=false)
     isempty(x) && throw(ArgumentError("mad is not defined for empty arrays"))
-    c = center === nothing ? median!(x) : center
-    T = promote_type(typeof(c), eltype(x))
+    T = promote_type(typeof(center), eltype(x))
     U = eltype(x)
     x2 = U == T ? x : isconcretetype(U) && isconcretetype(T) && sizeof(U) == sizeof(T) ? reinterpret(T, x) : similar(x, T)
-    x2 .= abs.(x .- c)
+    x2 .= abs.(x .- center)
     m = median!(x2)
-    if normalize === true
-        m * mad_to_std_multiplier
-    else
-        m
-    end
+    normalize ? m * mad_to_std_multiplier : m
 end
 
 # Interquartile range


src/scalarstats.jl
```diff
@@ -536,12 +536,12 @@ If `normalize` is set to `true`, the MAD is multiplied by
 `1 / quantile(Normal(), 3/4) ≈ 1.4826`, in order to obtain a consistent estimator
 of the standard deviation under the assumption that the data is normally distributed.
 """
-function mad(x; center=nothing, normalize::Union{Bool, Nothing}=nothing, constant=nothing)
-    mad!(Base.copymutable(x); center=center, normalize=normalize, constant=constant)
+function mad(x; center=nothing, normalize::Union{Bool, Nothing}=nothing)
```
devmotion, line 539: Since this is a breaking change, I think we should remove support for `normalize::Nothing` since seemingly it's only purpose is to print a deprecation warning.
```suggestion
function mad(x; center=nothing, normalize::Bool=false)
```

src/scalarstats.jl
```diff
@@ -552,24 +552,16 @@ of the standard deviation under the assumption that the data is normally distrib
 """
 function mad!(x::AbstractArray;
               center=median!(x),
-              normalize::Union{Bool,Nothing}=true,
-              constant=nothing)
+              normalize::Union{Bool,Nothing}=false)
     isempty(x) && throw(ArgumentError("mad is not defined for empty arrays"))
     c = center === nothing ? median!(x) : center
```
devmotion, line 557: Is it documented anywhere and actually useful that `center::Nothing` is supported?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
-version = "0.34.4"
+version = "0.35.0"
```
nalimilan, line 4: We should really go to 1.0 next time we make a breaking release.
```suggestion
version = "1.0"
```

It would be good to check whether we want to make other breaking changes though (I don't have ideas off the top of my head).

nalimilan:
The difficulty is that we don't want to tag a major release each time we find a breaking change to make. This change alone doesn't justify a major release IMO. And it doesn't seem we have lots of issues tagged as breaking. But there are probably other changes that would be justified that have not been marked as such.

src/scalarstats.jl
```diff
@@ -524,10 +524,10 @@ function sem(x::AbstractArray, weights::ProbabilityWeights; mean=nothing)
 end
 
 # Median absolute deviation
-@irrational mad_constant 1.4826022185056018 BigFloat("1.482602218505601860547076529360423431326703202590312896536266275245674447622701")
+@irrational mad_to_std_multiplier 1.4826022185056018 BigFloat("1.482602218505601860547076529360423431326703202590312896536266275245674447622701")
```
devmotion: This is type piracy. One must not use `@irrational` outside of Base.Math. The proper way to define irrational constants in packages is to use IrrationalConstants.jl

devmotion:
> It would be safer to throw an error by default, requiring to pass true or false explicitly for a while (knowing that turning an error into something that works is not breaking so it can happen anytime). Otherwise code that works fine currently will suddenly start returning values which have a different meaning without any warning.

IMO one should just enforce deprecation warnings using `force = true` in `Base.depwarn` (and actually I think that should be the default setting and also be configurable in `Base.@deprecate`). IMO this would be better than a breaking release that adds error paths just to remove them in a subsequent non-breaking release - there's always the danger that users only update their environments or compat entries when the error path is removed, in which case they never encounter the error path.

PR #951, opened 2025-03-15
Update docs

abhro:
- Make Pkg manage dependency between StatsBase and its docs
- Add doctests
- Fix up docstrings


src/reliability.jl
```diff
@@ -19,11 +19,11 @@ Calculate Cronbach's alpha (1951) from a covariance matrix `covmatrix` according
 the [formula](https://en.wikipedia.org/wiki/Cronbach%27s_alpha):
 
 ```math
-\\rho = \\frac{k}{k-1} (1 - \\frac{\\sum^k_{i=1} \\sigma^2_i}{\\sum_{i=1}^k \\sum_{j=1}^k \\sigma_{ij}})
+ρ = \\frac{k}{k-1} \\left(1 - \\frac{\\sum^k_{i=1} σ^2_i}{\\sum_{i=1}^k \\sum_{j=1}^k σ_{ij}}\\right)
```
ararslan, line 22: Why change the $\LaTeX$ Greek letters to Unicode? If I recall, I've run into situations in the past where the latter rendered in a way that looked inferior to the former, but I can't think of concrete examples offhand.

ararslan (commented):
Thanks for the contribution! I left a couple of notes/questions but broadly this looks great.

docs/Project.toml
```diff
@@ -5,3 +5,6 @@ StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
 [compat]
 Documenter = "1"
+
+[sources.StatsBase]
+path = ".."
```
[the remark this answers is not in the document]
abhro: Hello! I'm not sure of when it was introduced to Julia/Pkg, but for the current version it's documented at https://pkgdocs.julialang.org/v1/toml-files/#The-[sources]-section. I have not tried to build the docs on an earlier version of Julia, but that said, wouldn't the details of building the docs be entirely separate from making sure StatsBase.jl supports older Julia? It's only the website that uses the docs Project, right?

src/reliability.jl
abhro: Hi! Great question! This is for making it easier to see the equation in a REPL or some environment which doesn't render LaTeX. By default Documenter.jl uses KaTeX, which does support Unicode, and it's also rendered well in MathJAX and some other platforms.

As for render quality, I do agree there are some examples where the Unicode input is a lower quality than a TeX command. However, I've found that it is not the case for **lowercase Greek letters**, which should be the only changes I've made from TeX sequences → Unicode. If using Unicode, the _uppercase_ Greek letters get italicized, and `\hbar` (ℏ, reduced Planck's constant) isn't as neat, but those concerns do not apply here AFAICT. (We could also replace the `\sum` sequence, but I think that does produce a worse result haha)

andreasnoack (approved):

Issue #950, opened 2025-03-10
sample with UnitWeights and replace=false throws "no field values" error


hughselway:
Using sample with no replacements and unit weights throws an error, seeming to assume the weights object will have values.

I'm trying to sample in a way agnostic to whether FrequencyWeights or UnitWeights are passed.

**Minimal reproducible example**
```
using StatsBase
sample(1:5, uweights(5), 3, replace=false)
```

I would expect this to be equivalent to `sample(1:5, 3, replace=false)` but instead it throws this error:
```
ERROR: type UnitWeights has no field values
Stacktrace:
 [1] getproperty
   @ ./Base.jl:49 [inlined]
 [2] efraimidis_aexpj_wsample_norep!(rng::Random.TaskLocalRNG, a::UnitRange{…}, wv::UnitWeights{…}, x::Vector{…}; ordered::Bool)
   @ StatsBase ~/.julia/packages/StatsBase/xgoZ5/src/sampling.jl:869
 [3] efraimidis_aexpj_wsample_norep!
   @ ~/.julia/packages/StatsBase/xgoZ5/src/sampling.jl:848 [inlined]
 [4] sample!(rng::Random.TaskLocalRNG, a::UnitRange{…}, wv::UnitWeights{…}, x::Vector{…}; replace::Bool, ordered::Bool)
   @ StatsBase ~/.julia/packages/StatsBase/xgoZ5/src/sampling.jl:944
 [5] sample!
   @ ~/.julia/packages/StatsBase/xgoZ5/src/sampling.jl:918 [inlined]
 [6] sample
   @ ~/.julia/packages/StatsBase/xgoZ5/src/sampling.jl:952 [inlined]
 [7] #sample#220
   @ ~/.julia/packages/StatsBase/xgoZ5/src/sampling.jl:955 [inlined]
 [8] top-level scope
   @ REPL[2]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

**Current workaround**
I can condition on  `typeof` the weights variable until there's an internal fix.

**Versions**
Julia 1.11.3
StatsBase v0.34.4

PR #963, opened 2025-09-17
Fix sampling with `UnitWeights`

devmotion:
Fixes https://github.com/JuliaStats/StatsBase.jl/issues/950.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -603,6 +603,9 @@ sample(wv::AbstractWeights) = sample(default_rng(), wv)
 sample(rng::AbstractRNG, a::AbstractArray, wv::AbstractWeights) = a[sample(rng, wv)]
 sample(a::AbstractArray, wv::AbstractWeights) = sample(default_rng(), a, wv)
 
+# Specialization for `UnitWeights`
+sample(rng::AbstractRNG, wv::UnitWeights) = rand(rng, 1:length(wv))
+
 """
     direct_sample!([rng], a::AbstractArray, wv::AbstractWeights, x::AbstractArray)
 
@@ -632,6 +635,14 @@ function direct_sample!(rng::AbstractRNG, a::AbstractArray,
 end
 direct_sample!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
     direct_sample!(default_rng(), a, wv, x)
+
+# Specialization for `UnitWeights`
+function direct_sample!(rng::AbstractRNG, a::AbstractArray, wv::UnitWeights, x::AbstractArray)
+    if length(a) != length(wv)
+        throw(DimensionMismatch(lazy"Number of samples ($(length(a))) and sample weights ($(length(wv))) must be equal."))
+    end
+    return direct_sample!(rng, a, x)
+end
 
 """
     alias_sample!([rng], a::AbstractArray, wv::AbstractWeights, x::AbstractArray)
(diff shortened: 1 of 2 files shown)


src/sampling.jl
```diff
@@ -633,6 +636,14 @@ end
 direct_sample!(a::AbstractArray, wv::AbstractWeights, x::AbstractArray) =
     direct_sample!(default_rng(), a, wv, x)
 
+# Specialization for `UnitWeights`
+function direct_sample!(rng::AbstractRNG, a::AbstractArray, wv::UnitWeights, x::AbstractArray)
```
nalimilan: Break lines at 92 chars.

test/sampling.jl
```diff
@@ -297,3 +297,51 @@ end
...
+    Random.seed!(123)
+    xs = direct_sample!(1:100, uweights(100), Vector{Int}(undef, 10))
+    Random.seed!(123)
+    @test xs == direct_sample!(1:100, Vector{Int}(undef, 10))
+
+    # Errors
+    @test_throws DimensionMismatch("Number of samples (100) and sample weights (99) must be equal.") sample(1:100, uweights(99), 10; replace=false)
+    @test_throws DimensionMismatch("Number of samples (80) and sample weights (53) must be equal.") direct_sample!(1:80, uweights(53), Vector{Int}(undef, 10))
+
+    # Custom units don't error and behave the same as sampling with `Weights`
```
nalimilan:
```suggestion
    # Custom unit weights don't error and behave the same as sampling with `Weights`
```

test/sampling.jl
```diff
@@ -297,3 +297,51 @@ end
         end
     end
 end
+
+# Custom unit weights without `values` field
+struct YAUnitWeights <: StatsBase.AbstractWeights{Int, Int, Vector{Int}}
```
nalimilan:
```suggestion
struct YAUnitWeights <: AbstractWeights{Int, Int, Vector{Int}}
```

test/sampling.jl
```diff
@@ -297,3 +297,51 @@ end
         end
     end
 end
+
+# Custom unit weights without `values` field
```
nalimilan:
```suggestion
# Custom weights without `values` field
```
(`UnitWeights` doesn't have a `values` field either so I didn't understand the comment at first.)

nalimilan (commented):
Thanks!

src/sampling.jl
```diff
@@ -633,6 +636,14 @@ end
...
+    if length(a) != length(wv)
+        throw(DimensionMismatch(lazy"Number of samples ($(length(a))) and sample weights ($(length(wv))) must be equal."))
```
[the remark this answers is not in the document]
devmotion: > BTW, does `lazy` really make a difference here? I didn't know that trick, but it looks like it just avoids having a call to `length(x)` (run only in case of error) in the function body?

The use of lazy error messages generally helps with type inference, see eg https://github.com/JuliaArrays/FillArrays.jl/pull/420#issue-3487546977, https://github.com/JuliaLang/julia/pull/58672, and https://github.com/JuliaLang/julia/pull/54737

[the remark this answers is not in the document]
devmotion:
```suggestion
        throw(DimensionMismatch(LazyString(
            "Number of samples (",
            length(a),
            ") and sample weights (",
            length(wv),
            ") must be equal.",
        )))
```

nalimilan (approved):
Indentation is weird in some places but with https://github.com/JuliaStats/StatsBase.jl/pull/989 it shouldn't matter.

PR #970, opened 2025-10-21
Remove all `@inbounds` annotations

devmotion:
Same as https://github.com/JuliaStats/Distributions.jl/pull/2005.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 authors = ["JuliaStats"]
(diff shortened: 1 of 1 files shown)


andreasnoack (approved):

src/sampling.jl
```diff
@@ -302,7 +302,7 @@ function seqsample_a!(rng::AbstractRNG, a::AbstractArray, x::AbstractArray)
             n -= 1
             q *= (n - k) / n
         end
-        @inbounds x[j+=1] = a[i+=1]
+        x[j+=1] = a[i+=1]
```
nalimilan: Same here and in other places where `i` is computed using a complex logic.

PR #981, opened 2025-12-14
Allow non-`Real` eltype with `quantile`

nalimilan:
This is consistent with `Statistics.quantile` and avoids breakage due to #977. It's particularly useful for `Union{T, Missing}`, e.g. a view of nonmissing entries in a vector. This also allows supporting some types such as `Date`, though currently this only works for some values (would need to implement `type=1`).

### DIFF
--- a/test/jet.jl
+++ b/test/jet.jl
@@ -17,6 +17,6 @@ if VERSION < v"1.12-"
         println(res)
         reports = JET.get_reports(res)
         @test_broken isempty(reports)
-        @test length(reports) <= 2
+        @test length(reports) <= 1
     end
 end


src/weights.jl
```diff
@@ -650,21 +650,29 @@ function quantile(v::AbstractVector{V}, w::AbstractWeights{W}, p::AbstractVector
...
+    end
+
     # prepare percentiles
     ppermute = sortperm(p)
     p = p[ppermute]
 
     # prepare out vector
-    out = Vector{typeof(zero(V)/1)}(undef, length(p))
+    v1 = vw[1][1]
+    out = Vector{typeof(v1 + zero(eltype(p))*zero(W)*zero(v1))}(undef, length(p))
```
nalimilan: Ideally I wanted to use `map(p)` and return the result directly to avoid having to compute the type manually, but I couldn't find a way for the return type to be inferred. The problem is that the loop does not just depend on the quantile, we also have to update the sum of weights, current and previous values, which get boxed. Not sure there's a clean solution.

src/weights.jl
```diff
@@ -650,21 +650,29 @@ function quantile(v::AbstractVector{V}, w::AbstractWeights{W}, p::AbstractVector
...
     fill!(out, vw[end][1])
 
-    for x in v
-        isnan(x) && return fill!(out, x)
+    # NaN is always sorted last in the absence of missing
+    # This behavior isn't consistent with Statistics.quantile, but preserve it for backward compatibility
+    if vw[end][1] isa Number && isnan(vw[end][1])
+        return fill(vw[end][1], length(p))
```
```suggestion
        return fill!(out, vw[end][1])
```

src/weights.jl
```diff
@@ -650,21 +650,29 @@ function quantile(v::AbstractVector{V}, w::AbstractWeights{W}, p::AbstractVector
...
 
     # prepare out vector
-    out = Vector{typeof(zero(V)/1)}(undef, length(p))
+    v1 = vw[1][1]
+    out = Vector{typeof(v1 + zero(eltype(p))*zero(W)*zero(v1))}(undef, length(p))
     fill!(out, vw[end][1])
 
-    for x in v
-        isnan(x) && return fill!(out, x)
+    # NaN is always sorted last in the absence of missing
```
devmotion: Same question as above.

test/weights.jl
```diff
@@ -459,6 +459,14 @@ end
     w = [1, 1/3, 1/3, 1/3, 1]
     answer = 6.0
     @test quantile(data[1], f(w), 0.5) ≈ answer atol = 1e-5
+
+    # Test non-Real eltype
+    @test_throws ArgumentError quantile([missing, 1], f([1, 2]), 0.5)
+    @test quantile(Union{Float64, Missing}[1, 2, 3, 4], f([1, 2, 2, 1]), 0.5) ==
+        quantile(Any[1, 2, 3, 4], f([1, 2, 2, 1]), 0.5) ==
+        quantile([1, 2, 3, 4], f([1, 2, 2, 1]), 0.5)
+    @test quantile([Date(2005, 01, 01), Date(2005, 01, 01)], f([1, 1]), 0.5) ==
```
devmotion: Can you also add tests for the other methods modified by the PR (eg `median` and `quantile` with a vector of `p`)?

test/jet.jl
```diff
@@ -11,12 +11,13 @@ if VERSION < v"1.12-"
         # Default error analysis for common problem fails since JET errors on interface definitions
         # The (deprecated) `model_response(::StatisticalModel)` calls the interface
         # function `response(::StatisticalModel)` for which no method exists yet
+        # JET is also confused by any(ismissing, ::AbstractVector)
```
nalimilan: The failure is this:
```julia
 ┌ quantile(v::AbstractVector{…}, w::StatsBase.AbstractWeights{…} where T<:Real, p::AbstractVector{…}) where {…} @ StatsBase D:\a\StatsBase.jl\StatsBase.jl\src\weights.jl:653
│ non-boolean `Nothing`, `Some{Bool}` found in boolean context (2/3 union split): goto %108 if not StatsBase.any(StatsBase.ismissing, v::AbstractVector)::Union{Nothing, Some{Bool}, Bool}
```

I don't understand where this comes from, as the type seems perfectly inferred as `Bool`:
```julia
julia> f(x) = any(ismissing, x)

julia> @code_warntype f(Any[1])
MethodInstance for f(::Vector{Any})
  from f(x) @ Main REPL[3]:1
Arguments
  x::Vector{Any}
Body::Bool
1 ─ %1 = Main.any(Main.ismissing, x)::Bool
└──      return %1
```

nalimilan: I've just checked, that error doesn't happen when I enable JET on Julia 1.12. I'm tempted to enable JET only there, but there are two new errors which are wrong (variables are always defined):
```julia
═════ 3 possible errors found ═════
┌ check_vectors(x::Any, y::Any, skipmissing::Symbol) @ StatsBase /home/milan/.julia/packages/StatsBase/5D4Dh/src/pairwise.jl:43
│ local variable `indsx` may be undefined: indsx::Any
└────────────────────
┌ check_vectors(x::Any, y::Any, skipmissing::Symbol) @ StatsBase /home/milan/.julia/packages/StatsBase/5D4Dh/src/pairwise.jl:43
│ local variable `indsy` may be undefined: indsy::Any
└────────────────────
┌ model_response(obj::StatisticalModel) @ StatsBase ./deprecated.jl:215
│ no matching method found `response(::StatisticalModel)`: response(obj::StatisticalModel)
└────────────────────
```
What do you think? Better use JET on 1.12 anyway?

src/weights.jl
```diff
@@ -650,21 +650,27 @@ function quantile(v::AbstractVector{V}, w::AbstractWeights{W}, p::AbstractVector
     vw = sort!(collect(zip(view(v, nz), view(w, nz))))
     N = length(vw)
 
+    any(ismissing, v) &&
+        throw(ArgumentError("quantiles are undefined in presence of missing values"))
```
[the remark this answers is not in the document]
nalimilan: We can't actually use `eltype(v) >: Missing` if we want to be generic as types other than `Missing` may implement it (e.g. https://github.com/nalimilan/TypedMissings.jl). Anyway this is a relatively simple optimization so I think we can rely on it. It's used all over the place, including [in Base itself](https://github.com/JuliaLang/julia/blob/ca9b6662be434e718dc97778a25c777c4e277a34/base/sort.jl#L594).

devmotion (approved):

nalimilan:
Thanks!

Issue #957, opened 2025-06-07
`StatsBase.mode` can have arbitrarily high bias and variance for continuous distributions


ForceBru:
This function can compute the mode for arbitrary iterables, including vectors of floats:

https://github.com/JuliaStats/StatsBase.jl/blob/dfe8945a90638de0a42d8bc7a8eb6f4328d2a358/src/scalarstats.jl#L110-L135

Indeed, it's called to compute the mode of a vector of floats:

```julia
julia> @which StatsBase.mode([1.,3.,3.141])
mode(a)
     @ StatsBase ~/.julia/packages/StatsBase/s2YJR/src/scalarstats.jl:111
```

However, it has __very high bias and variance__ compared to the half-sample mode (HSM, see [1]), for example. This can be easily verified with simulations:

```
 Row │ nobs     dist                            method     bias          variance     MAE        
     │ Int64    Distribu…                       String     Float64       Float64      Float64    
─────┼───────────────────────────────────────────────────────────────────────────────────────────
   1 │    1000  Normal{Float64}(μ=0.0, σ=1.0)   StatsBase  -0.0105188     1.03791     0.818895
   2 │    1000  Normal{Float64}(μ=0.0, σ=1.0)   HSM        -0.00256086    0.0550354   0.189431
   3 │ 1000000  Normal{Float64}(μ=0.0, σ=1.0)   StatsBase  -0.0242424     1.03698     0.820519
   4 │ 1000000  Normal{Float64}(μ=0.0, σ=1.0)   HSM        -0.00500615    0.00342947  0.0479765
   5 │    1000  Normal{Float64}(μ=-1.0, σ=1.0)  StatsBase   0.00682327    0.956165    0.788644
   6 │    1000  Normal{Float64}(μ=-1.0, σ=1.0)  HSM        -0.00291935    0.0560346   0.189825
   7 │ 1000000  Normal{Float64}(μ=-1.0, σ=1.0)  StatsBase   0.0167116     0.955203    0.778444
   8 │ 1000000  Normal{Float64}(μ=-1.0, σ=1.0)  HSM        -0.00111594    0.00361161  0.0485056
   9 │    1000  TDist{Float64}(ν=0.1)           StatsBase   1.68401e33    2.77662e69  1.68401e33
  10 │    1000  TDist{Float64}(ν=0.1)           HSM         0.00730426    0.0190662   0.110146
[...]

PR #984, opened 2025-12-27
Add hsm_mode: Half-Sample Mode for continuous data

anurag-mds:
<h2>Summary</h2>
This PR adds hsm_mode(), an implementation of the half-sample mode (HSM) which is a robust estimator of the mode for continuous distributions.

It is introduced as a separate function ( not an overload of mode() ) to preserve existing behaviour while providing a statistically meaningful alternative for continuous data 

This addresses and closes issue #957.
<hr>
<h2>Motivation</h2>
StatsBase.mode() is frequency-based and works well for discrete data. For continuous distributions, however, samples are usually unique, which makes frequency counts unstable and highly variable in practice

Issue #957 documents this behaviour, particularly for heavy-tailed distributions, where mode() can show extreme variance. 

This PR provides an estimator designed specifically for continuous data
<hr>
<h2>Approach</h2>
hsm_mode() implements the standard half-sample method described in the literature:

- Non-finite values (NaN, Inf) are filtered

- The data are sorted

- The algorithm repeatedly selects the contiguous half-sample with the smallest width

- Once ≤ 2 points remain, the midpoint of the final interval is returned

- The midpoint may not be a sample value, but provides a stable estimate of the location of highest density.

- Time complexity is dominated by sorting (O(n log n)); space complexity is O(n).

- After sorting, the contraction loop operates on SubArray views to avoid allocations.
<hr>
<h2>API Design</h2>

The estimator is exposed as a new function:

**hsm_mode(x::AbstractVector{T}) where T<:Real**

It is NOT added as an overload of mode() in order to:

- avoid changing existing semantics

- clearly distinguish frequency-based and density-based estimation

- let users choose the appropriate method explicitly

The **return type** is an **AbstractFloat**, promoted from the input element type (e.g. integers → Float64, Float32 → Float32).
<hr>
<h2>Testing and Documentation</h2>

Tests cover basic correctness, edge cases, robustness to outliers, handling of non-finite values, and type behavior. All tests pass.

The docstring explains intended use cases, compares with mode(), documents complexity, provides examples, and cites the relevant literature.
<hr>
<h2>References</h2>

<a href="https://www.tandfonline.com/doi/abs/10.1080/01621459.1974.10480246">Robertson & Cryer (1974), JASA</a>

<a href="https://www.sciencedirect.com/science/article/abs/pii/S0167947305001581">Bickel & Fruehwirth (2006), CSDA</a>
<hr>
<h2>Notes</h2>

This PR is intentionally small and focused. Extensions such as weighted HSM or support for missing values are left for future work.

<h3><u><i>I have attached images showing that the tests are passing, and I would like to know whether I should address the existing warnings in the codebase.</i></u></h3>

<b>Images: </b><br>
image<br>
image<br>
image<br>
image<br>
image<br>
image

_**Feedback on naming or API placement is welcome.**_

### DIFF
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -46,14 +46,67 @@ end
 
 # compute mode, given the range of integer values
 """
-    mode(a, [r])
-    mode(a::AbstractArray, wv::AbstractWeights)
+    mode(a; method=:frequency)
+    mode(a::AbstractArray, wv::AbstractWeights; method=:frequency)
+    mode(a::AbstractArray{T}, r::UnitRange{T}; method=:frequency) where T<:Integer
 
-Return the mode (most common number) of an array, optionally
-over a specified range `r` or weighted via a vector `wv`.
-If several modes exist, the first one (in order of appearance) is returned.
+Return the mode (most common value) of `a`, optionally
+over a specified range `r` or weighted via a vector
+`wv`.
+
+The `method` keyword argument selects the estimation method:
+
+- `:frequency`: Frequency-based mode. Counts occurrences and returns the most common
+  value. If several modes exist, the first one (in order of appearance) is returned.
+  This is appropriate for discrete data.
+
+- `:halfsample`: Half-sample mode (HSM). A robust estimator of the mode for
+  continuous data. Repeatedly finds the contiguous half-sample with the smallest
+  range until at most 2 points remain, then returns their midpoint. The return value
+  is always a floating-point number and may not be an element of `a`.
+  Throws an `ArgumentError` if `a` contains any non-finite values (`NaN`, `Inf`,
+  or `-Inf`).
+  This method currently does not support `r` and `wv` arguments.
+    
+  # References
+- D.R. Bickel, R. Fruehwirth (2006). On a fast, robust estimator of the mode.
+  Computational Statistics & Data Analysis, 50(12), 3500-3530.
+- T. Robertson, J.D. Cryer (1974). An iterative procedure for estimating the mode.
+  Journal of the American Statistical Association, 69(348), 1012-1016.
+
+# Examples
+```julia
+julia> mode([1, 2, 2, 3, 3, 3, 4])
+3
+
+julia> mode([1, 2, 2, 3, 3, 3, 4], method=:frequency)
+3
+
+julia> mode([1.0, 1.1, 1.2, 5.0, 5.1], method=:halfsample)
+1.1
+```
 """
-function mode(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
+function mode(a; method::Symbol=:frequency)
+    if method === :halfsample
+        return _hsm_mode(a)
+    elseif method === :frequency
+        return _frequency_mode(a)
+    else
+        throw(ArgumentError(LazyString("`method` must be `:frequency` or `:halfsample`, got `:", method, "`")))
+    end
+end
+
+function mode(a::AbstractArray{T}, r::UnitRange{T}; method::Symbol=:frequency) where T<:Integer
+    if method === :halfsample
+        throw(ArgumentError("The `:halfsample` method does not support a range argument. Call `mode(a, method=:halfsample)` without a range."))
+    elseif method === :frequency
+        return _frequency_mode(a, r)
+    else
+        throw(ArgumentError(LazyString("`method` must be `:frequency` or `:halfsample`, got `:", method, "`")))
+    end
+end
+
+function _frequency_mode(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
     isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
     len = length(a)
     r0 = r[1]
@@ -107,8 +160,7 @@ function modes(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
     return ms
 end
 
-# compute mode over arbitrary iterable
-function mode(a)
+function _frequency_mode(a)
     isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
     cnts = Dict{eltype(a),Int}()
     # first element
@@ -202,6 +254,41 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
 
     # find values corresponding to maximum counts
     return [x for (x, w) in weights if w == mw]
+end
+
+#Internal implementation of the Half-Sample Mode (HSM) estimator.
+function _hsm_mode(a)
+    isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
+
+    # Filter NaN values and sort
+    if !all(isfinite, a)
+        throw(ArgumentError("mode with `method=:halfsample` is not defined " *
+                            "for collections containing non-finite values"))
+    end
+    filteredv = sort!(collect(a))
+    len = length(filteredv)
+
+    len == 1 && return middle(filteredv[1], filteredv[1])
+    len == 2 && return middle(filteredv[1], filteredv[2])
+
+    # Iteratively find the half-sample with the smallest range
+    filteredv = @view filteredv[1:end]
+    while len > 2
+        half = cld(len, 2)
+        best_i = 1
+        best_width = filteredv[half] - filteredv[1]
+        for i in 2:(len - half + 1)
+            w = filteredv[i + half - 1] - filteredv[i]
+            if w < best_width
+                best_width = w
+                best_i = i
+            end
+        end
+        filteredv = @view filteredv[best_i:(best_i + half - 1)]
+        len = length(filteredv)
+    end
+
+    return middle(filteredv[1], filteredv[len])
 end
 
 #############################
--- a/src/scalarstats.jl
+++ b/src/scalarstats.jl
@@ -75,7 +75,6 @@ The `method` keyword argument selects the estimation method:
   Journal of the American Statistical Association, 69(348), 1012-1016.
 
 # Examples
-```julia
 julia> mode([1, 2, 2, 3, 3, 3, 4])
 3
 
@@ -256,10 +255,13 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
     return [x for (x, w) in weights if w == mw]
 end
 
-#Internal implementation of the Half-Sample Mode (HSM) estimator.
+# Internal implementation of the Half-Sample Mode (HSM) estimator.
 function _hsm_mode(a)
     isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
-
+    if !all(x -> x isa Real, a)
+        throw(ArgumentError("mode with `method=:halfsample` is only defined " *
+                            "for collections containing real numbers"))
+    end
     # Filter NaN values and sort
     if !all(isfinite, a)
         throw(ArgumentError("mode with `method=:halfsample` is not defined " *
--- a/test/scalarstats.jl
+++ b/test/scalarstats.jl
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
 @test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.3]))
 @test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.3]))
 
+## mode with method=:frequency (explicit, same as default)
+@test mode([1, 1, 1, 2, 3, 4], method=:frequency) == 1
+@test mode((x for x in [1, 1, 1, 2, 3, 4]), method=:frequency) == 1
+
+# Check no type inference regression on the default/frequency path
+@test @inferred(mode([1]))::Int == 1
+my_frequency_mode(x) = mode(x, method=:frequency)
+@test @inferred(my_frequency_mode([1]))::Int == 1
+
+# Check type inference on halfsample path
+my_hsm_mode(x) = mode(x, method=:halfsample)
+@test @inferred(my_hsm_mode([1]))::Float64 == 1.0
+
+## mode with method=:halfsample (half-sample mode)
+@test mode([10], method=:halfsample)::Float64 == 10.0
+@test mode([1, 5], method=:halfsample)::Float64 == 3.0         # midpoint of two elements
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample)::Float64 == 4.0
+@test mode([1.0, 1.1, 1.2, 5.0, 5.1], method=:halfsample) == 1.15
+@test mode([1.0, 2.0, 3.0, 4.0, 10.0, 11.0, 12.0, 13.0], method=:halfsample) == 1.5
+@test mode([1.0, 2.0, 10.0, 10.1, 10.2, 10.3], method=:halfsample) == 10.05
+
+# Robustness to outliers
+@test mode([1.0, 1.01, 1.011, 100.0, 200.0], method=:halfsample) == 1.0105
+
+# Non-finite values throw ArgumentError
+@test_throws ArgumentError mode([1.0, NaN, 2.0, 2.0, Inf], method=:halfsample)
+@test_throws ArgumentError mode([1.0, NaN, Inf, -Inf], method=:halfsample)
+@test_throws ArgumentError mode([NaN, Inf, -Inf], method=:halfsample)
+@test_throws ArgumentError mode([Inf, -Inf], method=:halfsample)
+@test_throws ArgumentError mode([NaN], method=:halfsample)
+@test_throws ArgumentError mode([NaN, NaN, NaN], method=:halfsample)
+
+# Edge cases
+@test_throws ArgumentError mode(Float64[], method=:halfsample)
+
+
+# Invalid method throws
+@test_throws ArgumentError mode([1, 2, 3], method=:invalid)
+
+## mode with range argument
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:frequency) == 4
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:halfsample)
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:invalid)
+
+
+
+@test mode([1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 5.0, 5.1, 5.2, 5.3], method=:halfsample) == 1.15
 ## zscores
 
 @test zscore([-3:3;], 1.5, 0.5) == [-9.0:2.0:3.0;]
--- a/test/scalarstats.jl
+++ b/test/scalarstats.jl
@@ -87,6 +87,10 @@ my_hsm_mode(x) = mode(x, method=:halfsample)
 # Robustness to outliers
 @test mode([1.0, 1.01, 1.011, 100.0, 200.0], method=:halfsample) == 1.0105
 
+# Non-real values throw ArgumentError
+@test_throws ArgumentError mode(["a", "b", "c"], method=:halfsample)
+@test_throws ArgumentError mode([1+1im, 1+2im], method=:halfsample)
+
 # Non-finite values throw ArgumentError
 @test_throws ArgumentError mode([1.0, NaN, 2.0, 2.0, Inf], method=:halfsample)
 @test_throws ArgumentError mode([1.0, NaN, Inf, -Inf], method=:halfsample)
@@ -98,7 +102,6 @@ my_hsm_mode(x) = mode(x, method=:halfsample)
 # Edge cases
 @test_throws ArgumentError mode(Float64[], method=:halfsample)
 
-
 # Invalid method throws
 @test_throws ArgumentError mode([1, 2, 3], method=:invalid)
 
@@ -106,8 +109,6 @@ my_hsm_mode(x) = mode(x, method=:halfsample)
 @test mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:frequency) == 4
 @test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:halfsample)
 @test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:invalid)
-
-
 
 @test mode([1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 5.0, 5.1, 5.2, 5.3], method=:halfsample) == 1.15
 ## zscores


devmotion:
Was this PR AI-generated?

devmotion:
Was this PR AI-generated?

devmotion:
Was this PR AI-generated?

anurag-mds:
The implementation, test and commits are mine. I reviewed existing StatsBase prs to match project conventions, and I used AI assistance only for wording clarity in the pr description and for general performance review and bugs made by me. it was not used for algorithm or code . since in ci pipeline documentation issues which I totally forgot was missing I am actively fixing that like using ceil(len/2) as per the definition. I am eager to explain any design or choices in detail

anurag-mds:
I think using frequency-based mode for floats can be misleading.

But making HSM the default for AbstractFloat might break cases where floats are actually categories or rounded values.

Maybe a keyword like discrete=false or a dispatch mode(x, HSM()) works better. I can change the API like that if the maintainers think it’s right.

anurag-mds:
Take a look at this:
image

Standard mode is unstable for continuous data  it returns different values depending on the order of elements. hsm_mode consistently identifies the true peak, as in this example: 1000 samples around 5 mixed with uniform noise. This shows why HSM should remain a separate function instead of replacing mode for floats.

src/scalarstats.jl
```diff
@@ -204,6 +204,99 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+Robertson, T., & Cryer, J. D. (1974). "An iterative procedure for estimating
+the mode." Journal of the American Statistical Association, 69(348), 1012-1016.
+
+Bickel, D. R., & Fruehwirth, R. (2006). "On a fast, robust estimator of the mode:
+Comparisons to other robust estimators with applications."
+Computational Statistics & Data Analysis, 50(12), 3500-3530.
+
+See also: [`mode`](@ref), [`modes`](@ref)
+"""
+function hsm_mode(a::AbstractVector{T}) where T <: Real
```
nalimilan: You could allow any iterator given that `sort` is called immediately anyway.

src/scalarstats.jl
```diff
@@ -204,6 +204,99 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+    a = sort(filter(isfinite, a))
+
+    len = length(a)
+
+    len == 0 && throw(ArgumentError("hsm_mode is not defined for empty vectors"))
+    len == 1 && return a[1]
+
+    while (len > 2)
```
nalimilan: Avoid parentheses in this case -same with `if` below.

src/scalarstats.jl
```diff
@@ -204,6 +204,99 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+        best_start = 1
+
+        for i in 1:(len-win_size+1)
+            width = a[i+win_size-1] - a[i]
+            if (width < best_width)
+                best_width = width
+                best_start = i
+            end
+        end
+        a = collect(view(a, best_start:(best_start+win_size-1)))
```
nalimilan: Is `collect` really needed??

anurag-mds:
@nalimilan Does everything seems good?

src/scalarstats.jl
```diff
@@ -46,14 +46,45 @@ end
 
 # compute mode, given the range of integer values
 """
-    mode(a, [r])
+    mode(a, [r]; method=:default)
     mode(a::AbstractArray, wv::AbstractWeights)
 
 Return the mode (most common number) of an array, optionally
 over a specified range `r` or weighted via a vector `wv`.
 If several modes exist, the first one (in order of appearance) is returned.
```
nalimilan: Does this sentence apply to `method=:halfsample`? Otherwise, move it to the description of `:default`.

src/scalarstats.jl
```diff
@@ -46,14 +46,45 @@ end
...
+  floating-point value. This method is particularly useful for:
+  - Continuous or near-continuous data
+  - Data with outliers (more robust than the mean)
+  - Skewed distributions where you want to find the peak density
+  
+  Note: The `:halfsample` method does not support NaN values and will throw an error if any are present.
+  For weighted mode calculations, only `:default` is supported.
+
+# Examples
+```julia
```
nalimilan: Use ```jldoctest and include `julia>` and the output like in the REPL (see other examples and Documenter.jl manual).

src/scalarstats.jl
```diff
@@ -46,14 +46,45 @@ end
...
+# Weighted mode
+mode([1, 2, 3], weights([0.1, 0.6, 0.3]))  # returns 2
+```
 """
-function mode(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
+function mode(a::AbstractArray{T}, r::UnitRange{T}; method::Symbol=:default) where T<:Integer
+    if method == :halfsample
+        return _hsm_mode(a)
+    elseif method != :default
+        throw(ArgumentError("method must be :default or :halfsample, got :$method"))
```
nalimilan:
```suggestion
        throw(ArgumentError("`method` must be `:default` or `:halfsample`, got `:$method`"))
```

src/scalarstats.jl
```diff
@@ -46,14 +46,45 @@ end
...
+
+# Weighted mode
+mode([1, 2, 3], weights([0.1, 0.6, 0.3]))  # returns 2
+```
 """
-function mode(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
+function mode(a::AbstractArray{T}, r::UnitRange{T}; method::Symbol=:default) where T<:Integer
+    if method == :halfsample
+        return _hsm_mode(a)
+    elseif method != :default
```
nalimilan: Better have two internal methods, `_hsm_mode` and `_frequency_mode`, and call them from here.

src/scalarstats.jl
```diff
@@ -161,7 +197,12 @@ function modes(a)
 end
 
 # Weighted mode of arbitrary vectors of values
-function mode(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
+function mode(a::AbstractVector, wv::AbstractWeights{T}; method::Symbol=:default) where T <: Real
+    if method == :halfsample
+        throw(ArgumentError("method=:halfsample is not supported for weighted mode"))
+    elseif method != :default
+        throw(ArgumentError("method must be :default or :halfsample, got :$method"))
+    end
```
nalimilan: Here you can just avoid allowing the `method` argument. This is reflected in the docstring, and it's not super useful to allow it if only one value is supported.

src/scalarstats.jl
```diff
@@ -183,7 +224,12 @@ function mode(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
     return mv
 end
 
-function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
+function modes(a::AbstractVector, wv::AbstractWeights{T}; method::Symbol=:default) where T <: Real
```
nalimilan: Same here.

src/scalarstats.jl
```diff
@@ -204,6 +250,45 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
     return [x for (x, w) in weights if w == mw]
 end
 
+
+#############################
+#
+#   Mode: Half-Sample Mode
+#
+#############################
+
```
nalimilan: This kind of formatting seems to be used to differentiate different families of functions, but here HSM is just a variant of `mode`.
```suggestion
```

src/scalarstats.jl
```diff
@@ -204,6 +250,45 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+# Helper function for half-sample mode algorithm
+function _hsm_mode(a)
+    # Check for NaN
+    any(isnan, a) && throw(ArgumentError("mode with method=:halfsample does not support NaN values"))
```
nalimilan:
```suggestion
    any(isnan, a) &&
        throw(ArgumentError("`mode` with `method=:halfsample` does not support `NaN` values"))
```

src/scalarstats.jl
```diff
@@ -204,6 +250,45 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+    
+    # Sort the data (allows any iterator)
+    a = sort(collect(a))
```
nalimilan:
```suggestion
    a = sort!(collect(a))
```

src/scalarstats.jl
```diff
@@ -204,6 +250,45 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+    
+    len = length(a)
+    len == 0 && throw(ArgumentError("mode with method=:halfsample is not defined for empty collections"))
```
nalimilan: That's also the case for the other method:
```suggestion
    len == 0 && throw(ArgumentError("mode is not defined for empty collections"))
```

src/scalarstats.jl
```diff
@@ -204,6 +250,45 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+    len == 1 && return float(a[1])
+    len == 2 && return middle(a[1], a[2])
+
+    while len > 2
+        win_size = cld(len, 2)  # Use ceiling division for correct half-sample
```
nalimilan: Not sure this comment adds anything (of course we want to do the correct thing ;-).
```suggestion
        win_size = cld(len, 2)
```

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
 @test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.3]))
 @test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.3]))
 
+## mode with method=:halfsample (half-sample mode)
+
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample) ≈ 4.0
+@test mode([10.0], method=:halfsample) ≈ 10.0
+@test mode([1.0, 2.0, 3.0], method=:halfsample) ≈ 1.5
+@test mode([1.0, 2.0, 3.0, 4.0, 5.0], method=:halfsample) ≈ 1.5
+@test isapprox(mode([1.0097, 1.0054, 1.003212, 1.0231, 1.344, 1.00003], method=:halfsample), 1.00431, atol=1e-4)
```
nalimilan: Can you use a higher precision? Better copy the exact value that is returned, and use `==` too.

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
 @test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.3]))
 @test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.3]))
 
+## mode with method=:halfsample (half-sample mode)
+
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample) ≈ 4.0
```
nalimilan: AFAICT the value will always be the same under IEEE float arithmetic, so you can use `==` here and below?

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test_throws ArgumentError mode([NaN, 2.0, 3.0], method=:halfsample)  # NaN not allowed
+@test mode([Inf, 2.0, 3.0], method=:halfsample) ≈ 2.5  # Inf is allowed
+@test mode([1, 2, 3], method=:halfsample) isa Float64
+@test mode([1.0f0, 2.0f0], method=:halfsample) isa Float32
+
+# Additional edge cases
+@test mode([1.0, 1.0, 1.0], method=:halfsample) ≈ 1.0  # all identical
+@test mode([1, 1, 1, 2, 2, 100], method=:halfsample) ≈ 1.0  # outlier: finds densest region
+@test mode(Int32[1, 2, 3], method=:halfsample) isa Float64  # type promotion
+@test mode(Float32[1, 2, 3], method=:halfsample) isa Float32  # type preservation
```
nalimilan: This one seems redundant with the `Float32` test above. Test `Int32` instead? could also be grouped with other type tests above.

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+
+# Test overflow protection with middle()
+@test mode([typemax(Int64)-1, typemax(Int64)], method=:halfsample) ≈ float(typemax(Int64)) - 0.5
```
nalimilan:
```suggestion
@test mode([typemax(Int64)-1, typemax(Int64)], method=:halfsample) == typemax(Int64) - 0.5
```

src/scalarstats.jl
```diff
@@ -46,14 +46,45 @@ end
...
-    mode(a, [r])
+    mode(a, [r]; method=:default)
     mode(a::AbstractArray, wv::AbstractWeights)
 
 Return the mode (most common number) of an array, optionally
 over a specified range `r` or weighted via a vector `wv`.
 If several modes exist, the first one (in order of appearance) is returned.
+
+The `method` keyword argument allows selecting different mode estimation methods:
+- `:default`: Frequency-based mode (counts occurrences). This is the standard mode
```
nalimilan: Call this `:frequency` or `:counts`? The fact that it's the default doesn't explain what it does.

src/scalarstats.jl
```diff
@@ -46,14 +46,45 @@ end
...
+  that returns the most frequently occurring value in the data.
+- `:halfsample`: Half-sample mode (HSM), a robust density-based estimator for continuous data.
```
nalimilan: Can you cite a reference describing the method in a `# References` section?

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.6, 0.3]), method=:halfsample)
+@test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.6, 0.3]), method=:invalid)
+@test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.6, 0.3]), method=:invalid)
+
+# Test mode with range and method parameter
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 1:5, method=:default) == 4
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 1:5, method=:halfsample) ≈ 4.0  # Test halfsample with range
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 1:5, method=:invalid)
+
+
```
nalimilan: Could you add a test with a longer vector, just to make sure the general case is well tested?

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample) ≈ 4.0
+@test mode([10.0], method=:halfsample) ≈ 10.0
+@test mode([1.0, 2.0, 3.0], method=:halfsample) ≈ 1.5
+@test mode([1.0, 2.0, 3.0, 4.0, 5.0], method=:halfsample) ≈ 1.5
+@test isapprox(mode([1.0097, 1.0054, 1.003212, 1.0231, 1.344, 1.00003], method=:halfsample), 1.00431, atol=1e-4)
+@test mode([1.0, 1.0, 1.0, 10.0], method=:halfsample) ≈ 1.0
+@test mode([-5.0, -3.0, -1.0, 0.0, 1.0, 3.0, 5.0], method=:halfsample) ≈ -0.5
+@test_throws ArgumentError mode(Float64[], method=:halfsample)
+@test mode([1.0, 5.0], method=:halfsample) ≈ 3.0  # two elements: returns midpoint
```
nalimilan: Put these with the test for a single value vector.

nalimilan (commented):
Thanks!

anurag-mds:
Thanks @nalimilan for the thorough review. I agree with the remaining points (doc wording, method naming, references, and test adjustments). I’m addressing them now and will push a final cleanup commit so that
 everything is consolidated.

anurag-mds:
@nalimilan I have made necessary changes if anything to fix, modify or add do let me know!
Thanks a lot for the guidance

anurag-mds:
@nalimilan What do you think?

src/scalarstats.jl
```diff
@@ -46,14 +46,53 @@ end
...
+Robertson, T., & Cryer, J. D. (1974). An iterative procedure for estimating the mode. 
+Journal of the American Statistical Association, 69(348), 1012-1016.
+
+Bickel, D. R., & Frühwirth, R. (2006). On a fast, robust estimator of the mode: 
+Comparisons to other robust estimators with applications. Computational Statistics & 
+Data Analysis, 50(12), 3500-3530.
+"""
+function mode(a::AbstractArray{T}, r::UnitRange{T}; method::Symbol=:default) where T<:Integer
+    if method == :halfsample
+        return _hsm_mode(a)
```
nalimilan, line 90: This isn't correct, `r` shouldn't be ignored. Do you think you could implement it? Otherwise throw an error and document it.

test/scalarstats.jl
```diff
@@ -63,6 +63,55 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test mode([1.0097, 1.0054, 1.003212, 1.0231, 1.344, 1.00003], method=:halfsample) == 1.0043060000000001
+@test mode([1.0, 1.0, 1.0, 10.0], method=:halfsample) == 1.0
+@test mode([-5.0, -3.0, -1.0, 0.0, 1.0, 3.0, 5.0], method=:halfsample) == -0.5
+@test_throws ArgumentError mode([NaN, 2.0, 3.0], method=:halfsample)  # NaN not allowed
+@test mode([Inf, 2.0, 3.0], method=:halfsample) == 2.5  # Inf is allowed
+
+# Type tests
+@test mode([1, 2, 3], method=:halfsample) isa Float64
+@test mode(Int32[1, 2, 3], method=:halfsample) isa Float64  # type promotion
+@test mode([1.0f0, 2.0f0], method=:halfsample) isa Float32
```
nalimilan: Redundant with next one:
```suggestion
```

test/scalarstats.jl
```diff
@@ -63,6 +63,55 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test_throws ArgumentError mode([1, 2, 3], method=:invalid)
+
+# Test weighted mode (no method parameter for weighted)
+@test mode([1, 2, 3], weights([0.1, 0.6, 0.3])) == 2
+@test modes([1, 2, 3], weights([0.1, 0.6, 0.3])) == [2]
+
+# Test mode with range and method parameter
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 1:5, method=:default) == 4
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 1:5, method=:halfsample) == 4.0  # Test halfsample with range
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 1:5, method=:invalid)
```
nalimilan: Test with `2:4` to actually test that argument.

test/scalarstats.jl
```diff
@@ -63,6 +63,55 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+# Test with longer vector
+@test mode([1.0, 1.1, 1.2, 1.3, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0], method=:halfsample) == 1.15
+
+# Test invalid method
+@test_throws ArgumentError mode([1, 2, 3], method=:invalid)
+
+# Test weighted mode (no method parameter for weighted)
+@test mode([1, 2, 3], weights([0.1, 0.6, 0.3])) == 2
+@test modes([1, 2, 3], weights([0.1, 0.6, 0.3])) == [2]
+
```
nalimilan: This must already be tested elsewhere.
```suggestion
```

src/scalarstats.jl
```diff
@@ -204,6 +248,40 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+    # Check for NaN
+    any(isnan, a) &&
+        throw(ArgumentError("`mode` with `method=:halfsample` does not support `NaN` values"))
+    
+    # Sort the data (allows any iterator)
+    a = sort!(collect(a))
+    
+    len = length(a)
+    len == 0 && throw(ArgumentError("mode is not defined for empty collections"))
+    len == 1 && return float(a[1])
```
nalimilan: Actually we have a `middle` method precisely to ensure we return the right type even with a single argument:
```suggestion
    len == 1 && return middle(a[1])
```

test/scalarstats.jl
```diff
@@ -63,6 +63,55 @@ wv = weights([0.1:0.1:0.7; 0.1])
 @test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.3]))
 @test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.3]))
 
+## mode with method=:halfsample (half-sample mode)
+
+@test mode([10.0], method=:halfsample) ≈ 10.0
+@test mode([1.0, 5.0], method=:halfsample) ≈ 3.0  # two elements: returns midpoint
```
nalimilan:
```suggestion
@test mode([10], method=:halfsample)::Float64 == 10
@test mode([1, 5], method=:halfsample)::Float64 = 3  # two elements: returns midpoint
```

test/scalarstats.jl
```diff
@@ -63,6 +63,55 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test_throws ArgumentError mode(Float64[], method=:halfsample)
+
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample) == 4.0
```
nalimilan:
```suggestion
@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample)::Float64 == 4
```

src/scalarstats.jl
```diff
@@ -108,7 +147,12 @@ function modes(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
 end
 
 # compute mode over arbitrary iterable
-function mode(a)
+function mode(a; method::Symbol=:default)
+    if method == :halfsample
+        return _hsm_mode(a)
+    elseif method != :default
+        throw(ArgumentError("`method` must be `:default` or `:halfsample`, got `:$method`"))
+    end
```
nalimilan: Given that the two methods don't return the same type when inputs are integers, we need to inline this function so that the compiler knows which function is called inside the caller function. (Same above BTW.)
```suggestion
@inline function mode(a; method::Symbol=:default)
    if method === :halfsample
        return _hsm_mode(a)
    elseif method === :default
        return _frequency_mode(a)
    else
        throw(ArgumentError("`method` must be `:default` or `:halfsample`, got `:$method`"))
    end
end

# Helper function for default mode algorithm
function _frequency_mode_mode(a)
```
```

File: test/scalarstats.jl
Comment (nalimilan) [inline test/scalarstats.jl:50]: To check that there's no inference regression in the default case:
```suggestion
@test @inferred(mode(d1)) == 3
```

test/scalarstats.jl
```diff
@@ -63,6 +63,55 @@ wv = weights([0.1:0.1:0.7; 0.1])
 @test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.3]))
 @test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.3]))
 
+## mode with method=:halfsample (half-sample mode)
+
```
nalimilan:
```suggestion
@test mode([1], method=:default) == 1
@test mode((x for x in [1]), method=:default) == 1
my_hsm_mode(x) = mode(x, method=:default)
@test inferred(my_hsm_mode([1])) == 1
my_frequency_mode(x) = mode(x, method=:halfsample)
@test inferred(my_frequency_mode([1])) == 1

```

src/scalarstats.jl
nalimilan: Not addressed.

anurag-mds:
Everything Noted.
I will make sure that all issues are properly addressed...

anurag-mds:
Rebasing took more time Is everything good @nalimilan ?

anurag-mds:
Is there anything left @nalimilan ?

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+
+julia> mode([1, 2, 2, 3, 3, 3, 4], method=:frequency)
+3
+
+julia> mode([1.0, 1.1, 1.2, 5.0, 5.1], method=:halfsample)
+1.1
+```
 """
-function mode(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
+@inline function mode(a; method::Symbol=:frequency)
```
devmotion: Why `@inline`?

```suggestion
function mode(a; method::Symbol=:frequency)
```

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+    if method === :halfsample
+        return _hsm_mode(a)
+    elseif method === :frequency
+        return _frequency_mode(a)
+    else
+        throw(ArgumentError("`method` must be `:frequency` or `:halfsample`, got `:$method`"))
```
devmotion:
```suggestion
        throw(ArgumentError(LazyString("`method` must be `:frequency` or `:halfsample`, got `:", method, "`")))
```

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+    end
+end
+
+@inline function mode(a::AbstractArray{T}, r::UnitRange{T}; method::Symbol=:frequency) where T<:Integer
```
devmotion: Is it necessary that the types are identical?
```suggestion
function mode(a::AbstractArray{T}, r::UnitRange{T}; method::Symbol=:frequency) where T<:Integer
```

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+    if method === :halfsample
+        throw(ArgumentError("The `:halfsample` method does not support a range argument. Call `mode(a, method=:halfsample)` without a range."))
+    elseif method === :frequency
+        return _frequency_mode(a, r)
+    else
+        throw(ArgumentError("`method` must be `:frequency` or `:halfsample`, got `:$method`"))
```
devmotion:
```suggestion
        throw(ArgumentError(LazyString("`method` must be `:frequency` or `:halfsample`, got `:", method, "`")))
```

src/scalarstats.jl
```diff
@@ -204,6 +248,43 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
     return [x for (x, w) in weights if w == mw]
 end
 
+
+"""
+    _hsm_mode(a)
+
+Internal implementation of the Half-Sample Mode (HSM) estimator.
+Do not call directly; use `mode(a, method=:halfsample)` instead.
+"""
```
devmotion: IMO internal functions should not have docstrings, only code comments:

```suggestion
# Internal implementation of the Half-Sample Mode (HSM) estimator
```

src/scalarstats.jl
```diff
@@ -204,6 +248,43 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+function _hsm_mode(a)
+    isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
+
+    # Filter non-finite values and sort
+    filtered = sort([x for x in a if isfinite(x)])
```
devmotion:
```suggestion
    filtered = sort!([x for x in a if isfinite(x)])
```

src/scalarstats.jl
```diff
@@ -204,6 +248,43 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+    len = length(filtered)
+
+    len == 0 && throw(ArgumentError("mode is not defined for collections with no finite values"))
```
devmotion: I can't see this restriction in the other mode implementation above. So maybe that is only a limitation of half-sample mode?

anurag-mds: Yes, it is

As you can see for Regular Frequency Mode:

Consider [Nan, Inf, -Inf]
Here,
1) It can still count frequencies
2) Nan appears once, Inf appears once
3) Returns first one

Now, for HSM Mode:

Consider the same array,
Here,
1) Filters out all non-finite values
2) Nothing left to calculate
3) MUST throw error
4) Cannot find a cluster with no data!

anurag-mds:
I've addressed the final performance suggestions (sort! and LazyString). Ready for final review.
@nalimilan @devmotion

anurag-mds:
Can you please review the updated changes now?

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+
+The `method` keyword argument selects the estimation method:
+
+- `:frequency`: Frequency-based mode. Counts occurrences and returns the most common
+  value. If several modes exist, the first one (in order of appearance) is returned.
+  This is appropriate for discrete data.
+
+- `:halfsample`: Half-sample mode (HSM). A robust estimator of the mode for
+  continuous data. Repeatedly finds the contiguous half-sample with the smallest
+  range until ≤ 2 points remain, then returns their midpoint. The return value
```
nalimilan:
```suggestion
  range until at most 2 points remain, then returns their midpoint. The return value
```

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
 """
-    mode(a, [r])
-    mode(a::AbstractArray, wv::AbstractWeights)
+    mode(a; method=:frequency)
+    mode(a::AbstractArray{T}, r::UnitRange{T}; method=:frequency) where T<:Integer
 
-Return the mode (most common number) of an array, optionally
-over a specified range `r` or weighted via a vector `wv`.
-If several modes exist, the first one (in order of appearance) is returned.
+Return the mode (most common value) of `a`.
```
nalimilan:
```suggestion
Return the mode (most common value) of `a`, optionally
over a specified range `r` or weighted via a vector `wv`.
```

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+    if method === :halfsample
+        throw(ArgumentError("The `:halfsample` method does not support a range argument. Call `mode(a, method=:halfsample)` without a range."))
+    elseif method === :frequency
+        return _frequency_mode(a, r)
+    else
+        throw(ArgumentError(LazyString("`method` must be `:frequency` or `:halfsample`, got `:", method, "`")))
+    end
+end
+
+# The original frequency-based implementation, now a private helper
```
nalimilan: Future readers won't care about what existed before this PR.
```suggestion
```

src/scalarstats.jl
```diff
@@ -107,8 +151,8 @@ function modes(a::AbstractArray{T}, r::UnitRange{T}) where T<:Integer
     return ms
 end
 
-# compute mode over arbitrary iterable
-function mode(a)
+# The original frequency-based implementation, now a private helper
```
nalimilan:
```suggestion
```

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+
+- `:frequency`: Frequency-based mode. Counts occurrences and returns the most common
+  value. If several modes exist, the first one (in order of appearance) is returned.
+  This is appropriate for discrete data.
+
+- `:halfsample`: Half-sample mode (HSM). A robust estimator of the mode for
+  continuous data. Repeatedly finds the contiguous half-sample with the smallest
+  range until ≤ 2 points remain, then returns their midpoint. The return value
+  is always a floating-point number and may not be an element of `a`.
+  This is appropriate for continuous data and is robust to outliers.
```
nalimilan: Already said above. And need to say that some arguments are not allowed.
```suggestion
  This method currently does not support `r` and `wv` arguments.
```

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
...
+  See: Bickel (2002), Robertson & Cryer (1974).
```
nalimilan: Give full references, otherwise they are impossible to find.

src/scalarstats.jl
```diff
@@ -46,14 +46,58 @@ end
 
 # compute mode, given the range of integer values
```
nalimilan, line 47:
```suggestion
```

src/scalarstats.jl
nalimilan: Then the error message should be more explicit:
```suggestion
    len == 0 && throw(ArgumentError("mode with `method=:halfsample` is not defined " *
                                    "for collections with no finite values"))
```

src/scalarstats.jl
```diff
@@ -204,6 +248,38 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+    filtered = sort!([x for x in a if isfinite(x)])
+    len = length(filtered)
+
+    len == 0 && throw(ArgumentError("mode is not defined for collections with no finite values"))
+    len == 1 && return middle(filtered[1], filtered[1])
+    len == 2 && return middle(filtered[1], filtered[2])
+
+    # Iteratively find the half-sample with the smallest range
+    while len > 2
+        half = ceil(Int, len / 2)
```
nalimilan:
```suggestion
        half = cld(len, 2)
```
Right?

src/scalarstats.jl
```diff
@@ -204,6 +248,38 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+        best_i = 1
+        best_width = filtered[half] - filtered[1]
+        for i in 2:(len - half + 1)
+            w = filtered[i + half - 1] - filtered[i]
+            if w < best_width
+                best_width = w
+                best_i = i
+            end
+        end
+        filtered = filtered[best_i:(best_i + half - 1)]
```
nalimilan: In previous versions you used a view, why did you stop?

test/scalarstats.jl
```diff
@@ -63,6 +63,50 @@ wv = weights([0.1:0.1:0.7; 0.1])
 @test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.3]))
 @test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.3]))
 
+## mode with method=:frequency (explicit, same as default)
+@test mode([1], method=:frequency) == 1
+@test mode((x for x in [1]), method=:frequency) == 1
+
+# Check no type inference regression on the default/frequency path
+my_frequency_mode(x) = mode(x, method=:frequency)
+@test @inferred(my_frequency_mode([1])) == 1
```
nalimilan:
```suggestion
@test @inferred(mode([1]))::Int == 1

my_frequency_mode(x) = mode(x, method=:frequency)
@test @inferred(my_frequency_mode([1]))::Int == 1
```

test/scalarstats.jl
```diff
@@ -63,6 +63,50 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+
+# Check type inference on halfsample path
+my_hsm_mode(x) = mode(x, method=:halfsample)
+@test @inferred(my_hsm_mode([1]))::Float64 == 1.0
+
+## mode with method=:halfsample (half-sample mode)
+@test mode([10], method=:halfsample)::Float64 == 10.0
+@test mode([1, 5], method=:halfsample)::Float64 == 3.0         # midpoint of two elements
+@test mode([1.0, 5.0], method=:halfsample) ≈ 3.0
```
nalimilan: This is exactly the same as the previous case.
```suggestion
@test mode([1.0, 5.0], method=:halfsample) == 3.0
```

test/scalarstats.jl
```diff
@@ -63,6 +63,50 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample)::Float64 == 4.0
+@test mode([1.0, 1.1, 1.2, 5.0, 5.1], method=:halfsample) ≈ 1.1  atol=0.1
```
nalimilan: `atol=0.1` is enormous. Isn't this enough?
```suggestion
@test mode([1.0, 1.1, 1.2, 5.0, 5.1], method=:halfsample) ≈ 1.1
```

test/scalarstats.jl
```diff
@@ -63,6 +63,50 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+
+# Robustness to outliers
+@test abs(mode([1.0, 1.05, 1.1, 1.15, 100.0], method=:halfsample) - 1.075) < 0.2
```
nalimilan: Better test the exact expected return value.

test/scalarstats.jl
```diff
@@ -63,6 +63,50 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+
+# Non-finite values are filtered
+@test mode([1.0, NaN, 2.0, 2.0, Inf], method=:halfsample) ≈ 2.0
```
nalimilan:
```suggestion
@test mode([1.0, NaN, 2.0, 2.0, Inf], method=:halfsample) == 2.0
```

test/scalarstats.jl
```diff
@@ -63,6 +63,50 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+
+## mode with range argument
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:frequency) == 4
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:halfsample)
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:invalid)
+
+## @inferred check for default mode — no inference regression
+d1 = [1, 2, 3, 3, 4]
+@test @inferred(mode(d1)) == 3
+
```
nalimilan: I propose testing this above instead, with similar checks.
```suggestion
```

nalimilan (commented):
Sorry for the delay. Looks almost ready to me.

@devmotion Do you have better ideas of a name for `method=:frequency`? The definition of the mode implies frequency whatever the estimation method, but I can't find a better name...

anurag-mds:
@nalimilan Done applied the suggested changes
Extremely sorry for such a noisy PR
But I got to know many standards and style julia follows

src/scalarstats.jl
```diff
@@ -46,14 +46,64 @@ end
...
+
+The `method` keyword argument selects the estimation method:
+
+- `:frequency`: Frequency-based mode. Counts occurrences and returns the most common
+  value. If several modes exist, the first one (in order of appearance) is returned.
+  This is appropriate for discrete data.
+
+- `:halfsample`: Half-sample mode (HSM). A robust estimator of the mode for
+  continuous data. Repeatedly finds the contiguous half-sample with the smallest
+  range until atmost 2 points remain, then returns their midpoint. The return value
```
nalimilan:
```suggestion
  range until at most 2 points remain, then returns their midpoint. The return value
```

src/scalarstats.jl
```diff
@@ -46,14 +46,64 @@ end
 
 # compute mode, given the range of integer values
 """
-    mode(a, [r])
-    mode(a::AbstractArray, wv::AbstractWeights)
+    mode(a; method=:frequency)
+    mode(a::AbstractArray{T}, r::UnitRange{T}; method=:frequency) where T<:Integer
```
nalimilan, line 51:
```suggestion
    mode(a::AbstractArray, wv::AbstractWeights; method=:frequency)
    mode(a::AbstractArray{T}, r::UnitRange{T}; method=:frequency) where T<:Integer
```

src/scalarstats.jl
```diff
@@ -204,6 +253,38 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+#Internal implementation of the Half-Sample Mode (HSM) estimator.
+function _hsm_mode(a)
+    isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
+
+    # Filter non-finite values and sort
+    filtered = sort!([x for x in a if isfinite(x)])
+    len = length(filtered)
+
+    len == 0 && throw(ArgumentError("mode with `method=:halfsample` is not defined " *
+                    "for collections with no finite values"))
```
nalimilan:
```suggestion
    len == 0 && throw(ArgumentError("mode with `method=:halfsample` is not defined " *
                                    "for collections with no finite values"))
```

src/scalarstats.jl
```diff
@@ -204,6 +253,38 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+        best_i = 1
+        best_width = filtered[half] - filtered[1]
+        for i in 2:(len - half + 1)
+            w = filtered[i + half - 1] - filtered[i]
+            if w < best_width
+                best_width = w
+                best_i = i
+            end
+        end
+        filtered = @view filtered[best_i:(best_i + half - 1)]
```
nalimilan: For type stability, better use a different name like `filteredv` and do `filteredv = @view filtered[1:end]` before the loop.

test/scalarstats.jl
```diff
@@ -63,6 +63,47 @@ wv = weights([0.1:0.1:0.7; 0.1])
 @test_throws ArgumentError mode([1, 2, 3], weights([0.1, 0.3]))
 @test_throws ArgumentError modes([1, 2, 3], weights([0.1, 0.3]))
 
+## mode with method=:frequency (explicit, same as default)
+@test mode([1], method=:frequency) == 1
+@test mode((x for x in [1]), method=:frequency) == 1
```
nalimilan: Better test with values which would give something different with the other method:
```suggestion
@test mode([1, 1, 1, 2, 3, 4], method=:frequency) == 1
@test mode((x for x in [1, 1, 1, 2, 3, 4]), method=:frequency) == 1
```

test/scalarstats.jl
```diff
@@ -63,6 +63,47 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test_throws ArgumentError mode([1, 2, 3], method=:invalid)
+
+## mode with range argument
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:frequency) == 4
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:halfsample)
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:invalid)
+
+
+
+@test mode([1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 5.0, 5.1, 5.2, 5.3], method=:halfsample) == 1.15
```
nalimilan:
```suggestion
@test mode([1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 5.0, 5.1, 5.2, 5.3], method=:halfsample) == 1.15


```

src/scalarstats.jl
```diff
@@ -204,6 +253,38 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
     return [x for (x, w) in weights if w == mw]
 end
 
+#Internal implementation of the Half-Sample Mode (HSM) estimator.
+function _hsm_mode(a)
+    isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
+
+    # Filter non-finite values and sort
+    filtered = sort!([x for x in a if isfinite(x)])
```
nalimilan: Actually I now realize that StatsBase generally doesn't skip infinite values silently. For example `quantile` throws an error in the presence of NaN. This is less dangerous. The advantage is also that if we change our minds later we can simply turn the error into something that works, while the contrary would be breaking.

`Inf` and `-Inf` could be allowed, right? They are sorted in the right position, and if the two final values are finite, the result will be finite too. It's similar to `quantile([1, Inf], .5)` returning `Inf`.

anurag-mds:
Fixed the remaining points  `at most` typo, exact test value pinned (`1.0750000000000002`), error message updated to `'for collections containing only NaN values'`, `filteredv` view used throughout for type stability, frequency tests updated to use longer vector, docstring signature added.

One change I made proactively based on your comment about:

- `Inf`: changed the filter from `isfinite` to `!isnan`, so `Inf` and `-Inf` are now allowed through.

- `mode([NaN, Inf, -Inf])` returns `NaN` rather than throwing. 

**Happy to revert this if you'd prefer the stricter behaviour.**

nalimilan:
> * `mode([NaN, Inf, -Inf])` returns `NaN` rather than throwing.

Please throw an `ArgumentError` instead in the presence of NaN, this seems more explicit to me and it's easier to change later if we want.

src/scalarstats.jl
```diff
@@ -204,6 +254,40 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+
+    len == 1 && return middle(filtered[1], filtered[1])
+    len == 2 && return middle(filtered[1], filtered[2])
+
+    # Iteratively find the half-sample with the smallest range
+    filteredv = @view filtered[1:end]
+    while len > 2
+        half = cld(len, 2)
+        best_i = 1
+        best_width = filtered[half] - filtered[1]
```
nalimilan: Need to use `filteredv` here and below, right? I'm a bit worried that this wasn't caught by tests. Isn't this covered?

anurag-mds: The reason the tests didn't catch it is that the previous test cases were too simple they didn't 'squeeze' the data enough to show the mistake. I've now updated the code to use filteredv everywhere and added a more complex `test case ([1.0, 2.0, ... 13.0])` that would have failed with the old bug.

I also switched to a strict ArgumentError for any non-finite values as you suggested. Ready for another look!

anurag-mds:
I also wanted to say I'm genuinely sorry for the extra work this PR has caused. When this PR was created it was my very first PR, and I'm honestly a bit embarrassed that I ended up increasing your workload instead of making it easier. I’m learning a lot about the project's standards through this process, and I really appreciate your patience and guidance in helping me get this implementation right.

src/scalarstats.jl
```diff
@@ -46,14 +46,67 @@ end
...
+- D.R. Bickel, R. Fruehwirth (2006). On a fast, robust estimator of the mode.
+  Computational Statistics & Data Analysis, 50(12), 3500-3530.
+- T. Robertson, J.D. Cryer (1974). An iterative procedure for estimating the mode.
+  Journal of the American Statistical Association, 69(348), 1012-1016.
 
-Return the mode (most common number) of an array, optionally
-over a specified range `r` or weighted via a vector `wv`.
-If several modes exist, the first one (in order of appearance) is returned.
+# Examples
+```julia
```
nalimilan:
```suggestion
```jldoctest
```

nalimilan:
Sorry for the lack of reply, I was waiting for @devmotion to comment on the naming of arguments and then I forgot about the PR. I'll check it again but I think it's almost ready.

src/scalarstats.jl
```diff
@@ -204,6 +256,41 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
     return [x for (x, w) in weights if w == mw]
 end
 
+#Internal implementation of the Half-Sample Mode (HSM) estimator.
```
nalimilan:
```suggestion
# Internal implementation of the Half-Sample Mode (HSM) estimator.
```

src/scalarstats.jl
```diff
@@ -204,6 +256,41 @@ function modes(a::AbstractVector, wv::AbstractWeights{T}) where T <: Real
...
+function _hsm_mode(a)
+    isempty(a) && throw(ArgumentError("mode is not defined for empty collections"))
+
```
nalimilan:
```suggestion
    if !all(x -> x isa Real, a)
        throw(ArgumentError("mode with `method=:halfsample` is only defined " *
                            "for collections containing real numbers"))
    end
```

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test mode([10], method=:halfsample)::Float64 == 10.0
+@test mode([1, 5], method=:halfsample)::Float64 == 3.0         # midpoint of two elements
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], method=:halfsample)::Float64 == 4.0
+@test mode([1.0, 1.1, 1.2, 5.0, 5.1], method=:halfsample) == 1.15
+@test mode([1.0, 2.0, 3.0, 4.0, 10.0, 11.0, 12.0, 13.0], method=:halfsample) == 1.5
+@test mode([1.0, 2.0, 10.0, 10.1, 10.2, 10.3], method=:halfsample) == 10.05
+
+# Robustness to outliers
+@test mode([1.0, 1.01, 1.011, 100.0, 200.0], method=:halfsample) == 1.0105
+
```
nalimilan, line 89:
```suggestion

# Non-real values throw ArgumentError
@test_throws ArgumentError mode(["a", "b", "c"], method=:halfsample)
@test_throws ArgumentError mode([1+1im, 1+2im], method=:halfsample)

```

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+@test_throws ArgumentError mode([1.0, NaN, 2.0, 2.0, Inf], method=:halfsample)
+@test_throws ArgumentError mode([1.0, NaN, Inf, -Inf], method=:halfsample)
+@test_throws ArgumentError mode([NaN, Inf, -Inf], method=:halfsample)
+@test_throws ArgumentError mode([Inf, -Inf], method=:halfsample)
+@test_throws ArgumentError mode([NaN], method=:halfsample)
+@test_throws ArgumentError mode([NaN, NaN, NaN], method=:halfsample)
+
+# Edge cases
+@test_throws ArgumentError mode(Float64[], method=:halfsample)
+
```
nalimilan, line 104:
```suggestion
```

test/scalarstats.jl
```diff
@@ -63,6 +63,53 @@ wv = weights([0.1:0.1:0.7; 0.1])
...
+# Invalid method throws
+@test_throws ArgumentError mode([1, 2, 3], method=:invalid)
+
+## mode with range argument
+@test mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:frequency) == 4
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:halfsample)
+@test_throws ArgumentError mode([1, 2, 2, 3, 4, 4, 4, 5], 2:4, method=:invalid)
+
+
```
nalimilan:
```suggestion
```

nalimilan (commented):
Looks good modulo some suggestions I'm making. I'll apply them and see whether tests pass.

nalimilan (approved):

nalimilan:
Thanks!

anurag-mds:
Thank you @nalimilan and @devmotion for the reviews and patience. Really glad this made it in!

PR #991, opened 2026-01-11
Do not use `Base.@irrational`

devmotion:
`Base.@irrational` should not be used outside of Base as it causes type piracy and creates conflicts if another package defines a `Base.@irrational` constant of the same name: https://github.com/JuliaLang/julia/issues/46531

Instead, packages can use IrrationalConstants for defining a custom subtype of `AbstractIrrational` using `IrrationalConstants.@irrational` as a drop-in replacement.

Addresses https://github.com/JuliaStats/StatsBase.jl/pull/949/files#r2649638898

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,12 +1,13 @@
 name = "StatsBase"
 uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+version = "0.34.10"
 authors = ["JuliaStats"]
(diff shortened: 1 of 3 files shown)


nalimilan (approved):
Looks good until we drop this deprecation!

devmotion:
Which deprecation? The constant is required even if #905 is merged.

nalimilan:
Ah good point.

PR #1004, opened 2026-07-15
Fix OOM/hang in histrange when bin width is below floating-point res

### LINKED ISSUE #972: Very slow fit call
I'm not sure if it is a bug, but this happens in our use case.

This runs quickly:

```julia
using StatsBase: fit, Histogram

fit(Histogram, [-5.6,-5.6])
```

This takes a very long time:

```julia
using StatsBase: fit, Histogram

fit(Histogram, [-5.603325961434038e29,-5.603325961434038e29])
```

BjarkeHautop:
`fit(Histogram, x)` threw an `OutOfMemoryError` when all values were identical with a large magnitude. The hardcoded `step = 1` in histrange is below the floating-point spacing at that magnitude, so the endpoint adjustment loop ran for billions of iterations and produced a very large weights array.

The same root cause could also make the `hi != lo` branch hang forever when the computed bin width falls below the resolution of the data.

This PR fixes both cases by ensuring the step is never smaller than the floating-point spacing of the data.

Generated with the help of Claude Code (Fable) and manually adjusted and verified.

Fixes #972

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -43,10 +43,18 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
     if hi == lo
         start = F(hi)
         step = one(F)
+        if start + step == start
+            # step is below the floating-point spacing at `hi`: the endpoint
+            # adjustments below would grow `len` to ~eps(hi)/2
+            step = exp10(ceil(log10(eps(start))))
+        end
         divisor = one(F)
         len = one(F)
     else
         bw = (F(hi) - F(lo)) / n
+        # a bin width below the floating-point spacing of the data would
+        # stall the endpoint adjustments below
+        bw = max(bw, 2*eps(float(max(abs(lo), abs(hi)))))
         lbw = log10(bw)
         if lbw >= 0
             step = exp10(floor(lbw))
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -44,9 +44,8 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
         start = F(hi)
         step = one(F)
         if start + step == start
-            # step is below the floating-point spacing at `hi`: the endpoint
-            # adjustments below would grow `len` to ~eps(hi)/2
-            step = exp10(ceil(log10(eps(start))))
+            # step is below the floating-point spacing at `hi`: bump it up
+            step = eps(start)
         end
         divisor = one(F)
         len = one(F)
@@ -54,7 +53,7 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
         bw = (F(hi) - F(lo)) / n
         # a bin width below the floating-point spacing of the data would
         # stall the endpoint adjustments below
-        bw = max(bw, 2*eps(float(max(abs(lo), abs(hi)))))
+        bw = max(bw, eps(float(max(abs(lo), abs(hi)))))
         lbw = log10(bw)
         if lbw >= 0
             step = exp10(floor(lbw))
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -42,18 +42,15 @@ end
 function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
     if hi == lo
         start = F(hi)
-        step = one(F)
-        if start + step == start
-            # step is below the floating-point spacing at `hi`: bump it up
-            step = eps(start)
-        end
+        # if step is below the floating-point spacing at `hi`, bump it up
+        step = max(one(F), eps(start))
         divisor = one(F)
         len = one(F)
     else
         bw = (F(hi) - F(lo)) / n
         # a bin width below the floating-point spacing of the data would
         # stall the endpoint adjustments below
-        bw = max(bw, eps(float(max(abs(lo), abs(hi)))))
+        bw = max(bw, eps(F(hi)), eps(F(lo)))
         lbw = log10(bw)
         if lbw >= 0
             step = exp10(floor(lbw))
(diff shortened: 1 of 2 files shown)


src/hist.jl
```diff
@@ -43,10 +43,18 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
     if hi == lo
         start = F(hi)
         step = one(F)
+        if start + step == start
+            # step is below the floating-point spacing at `hi`: the endpoint
+            # adjustments below would grow `len` to ~eps(hi)/2
+            step = exp10(ceil(log10(eps(start))))
```
nalimilan: What's the reasoning behind this expression? Why not just `ceil(start)`? That would at least deserve a comment.

src/hist.jl
```diff
@@ -43,10 +43,18 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
...
+        end
         divisor = one(F)
         len = one(F)
     else
         bw = (F(hi) - F(lo)) / n
+        # a bin width below the floating-point spacing of the data would
+        # stall the endpoint adjustments below
+        bw = max(bw, 2*eps(float(max(abs(lo), abs(hi)))))
```
nalimilan: Why `2*`?

test/hist.jl
```diff
@@ -111,6 +114,33 @@ end
     @test StatsBase.histrange([1.0 for i in 1:100], 10, :left) == 1.0:1.0:2.0
     @test StatsBase.histrange([1.05 for i in 1:100], 10, :left) == 1.05:1.0:2.05
 
+    # Issue 972: values whose magnitude is so large that a bin width of
+    # one (or the requested width) is below the floating-point spacing
+    let x = -5.603325961434038e25
+        r = StatsBase.histrange([x, x], 10, :left)
+        @test first(r) == x
+        @test length(r) == 2
+        @test step(r) >= eps(x)
```
nalimilan, line 123: Also test this in the loop below?

test/hist.jl
```diff
@@ -111,6 +114,33 @@ end
     @test StatsBase.histrange([1.0 for i in 1:100], 10, :left) == 1.0:1.0:2.0
     @test StatsBase.histrange([1.05 for i in 1:100], 10, :left) == 1.05:1.0:2.05
 
+    # Issue 972: values whose magnitude is so large that a bin width of
+    # one (or the requested width) is below the floating-point spacing
```
nalimilan: "requested" seems to imply it's requested by the user. It's just computed internally, right?

nalimilan (commented):
Thanks!

src/hist.jl
BjarkeHautop: Gives a much larger bin size than needed:
```julia
start = -5.603325961434038e25

println(ceil(start))
#> -5.603325961434038e25
println(start + ceil(start))
#> -1.1206651922868077e26

# Much smaller bin:
@show step_eps = exp10(ceil(log10(eps(start))))
#> step_eps = exp10(ceil(log10(eps(start)))) = 1.0e10
println(start + step_eps)
#> -5.603325961434037e25

# Looking at it again I could actually just use eps(start) ofc:
@show step_eps = eps(start)
#> step_eps = eps(start) = 8.589934592e9
start + step_eps
#> -5.603325961434037e25
```

<sup>Created on 2026-07-20 with [MinimalWorkingExamples v0.2.2-DEV](https://github.com/BjarkeHautop/MinimalWorkingExamples.jl) using Julia 1.12.6</sup>

<details>
<summary>Environment</summary>

```text
Julia Version 1.12.6
Commit 15346901f00 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 8845HS w/ Radeon 780M Graphics
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver3)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 16 virtual cores)
Environment:
  JULIA_EDITOR = code
  JULIA_VSCODE_REPL = 1
  JULIA_LOAD_PATH = @:@stdlib
```

</details>

Shouldn't matter unless close to overflowing I guess, but since `eps(start)` works
it's probably better to just use that?

src/hist.jl
BjarkeHautop: Shouldn't be needed, your right. Updated.

test/hist.jl
BjarkeHautop: Yes, reworded.

src/hist.jl
nalimilan: Yes better do the simple thing if it works I'd say.

src/hist.jl
```diff
@@ -43,10 +43,17 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
     if hi == lo
         start = F(hi)
         step = one(F)
+        if start + step == start
+            # step is below the floating-point spacing at `hi`: bump it up
+            step = eps(start)
+        end
```
nalimilan: And what about something even simpler?
```suggestion
# if step is below the floating-point spacing at `hi`, bump it up
step = max(one(F), eps(start))
```

src/hist.jl
```diff
@@ -43,10 +43,17 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F
...
         divisor = one(F)
         len = one(F)
     else
         bw = (F(hi) - F(lo)) / n
+        # a bin width below the floating-point spacing of the data would
+        # stall the endpoint adjustments below
+        bw = max(bw, eps(float(max(abs(lo), abs(hi)))))
```
nalimilan:
```suggestion
        bw = max(bw, eps(F(hi)), eps(F(lo))
```

BjarkeHautop: (missing parenthesis, so just commited myself)

nalimilan (approved):

nalimilan:
Thanks!

PR #1007, opened 2026-08-21
Escape pipes when writing a CoefTable as markdown

andreasnoack:
`show(::IO, ::MIME"text/markdown", ::CoefTable)` never escapes `|`, so any cell containing one silently destroys the table. Downstream packages hit this with the R-style `Pr(>|z|)` p-value header (GLM, MixedModels, and others build their coefficient tables with it), which is why it is easy to reproduce but invisible from inside StatsBase — no header or fixture here contains a pipe.

## Reproduction

```julia
julia> ct = CoefTable(Any[[0.1615], [0.0499], [3.24], [0.0012]],
                      ["Coef.", "Std. Error", "z", "Pr(>|z|)"], ["Age"], 4, 3);

julia> md = sprint(show, MIME"text/markdown"(), ct);

julia> only(Markdown.parse(md).content)   # expected a Table
Markdown.Paragraph(...)
```

The header row carries the two extra pipes from `Pr(>|z|)`, so it presents 10 pipe-delimited fields against the alignment row's 8. No markdown parser accepts that as a table; Julia's degrades it to a paragraph and then applies inline typography to the `---` rules, turning them into en-dashes. Rendered through Documenter or any markdown viewer, the whole table appears as one run of pipes and dashes.

## Fix

Escape `|` in row names, column names and string cells — the same rule the Markdown stdlib applies in `plain(::IO, ::Markdown.Table)` — before the column widths are measured, so the widths stay consistent with what is printed.

Added a test covering a pipe in a header, a row name and a string cell, asserting that every row presents the same number of cells and that the output parses as a `Markdown.Table`. It fails without the fix. The two existing golden-output markdown tests are unchanged, since neither contains a pipe.

## Related

#1008 is an alternative that goes further: it drops the hand-rolled writer in favour of building a `Markdown.Table`, which removes this class of bug rather than this instance of it. That one changes the emitted spacing and adds Markdown as a dependency, so it is offered separately rather than folded in here.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -39,9 +39,10 @@ julia = "1.10"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
 JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
+Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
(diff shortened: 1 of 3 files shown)


palday (approved):
@andreasnoack care to patch bump and release this right away? I'll take a look at #1008 when I get some time later in the week

PR #1010, opened 2026-09-08
Use IEEE comparison in binindex

andreasnoack:
## Motivation

`binindex` locates the bin of an observation with `searchsortedfirst`/`searchsortedlast`, which compare with `isless` by default. Under `isless`, `-0.0` sorts strictly below `0.0`, so an observation of `-0.0` would land in a different bin than `0.0`. The current code works around this by passing `by=_normalize_zero` for vector edges, keeping a separate `AbstractRange` method without `by` to avoid losing Base's arithmetic fast path, and rejecting ranges that contain `-0.0` in the `Histogram` constructor since that method could not handle them.

Passing `lt = <` instead gives IEEE semantics directly: `-0.0 < 0.0` is false, so the two zeros are equal and no normalization is needed. NaN compares false with everything, so it still ends up outside the edges and is dropped by `push!` as before. Base's arithmetic range method accepts the resulting `Lt` ordering, so the fast path is kept.

## Changes

- `_edge_binindex` becomes a single method using `lt = <`; `_normalize_zero` and the `AbstractRange` method are removed.
- The constructor check rejecting ranges containing `-0.0` is removed, since its reason no longer applies. Such ranges now bin identically to the corresponding ranges with `0.0`.
- Tests: the two `@test_throws` for `-0.0` ranges are replaced by equality tests, and explicit tests that NaN observations are dropped are added for both `closed` values and both edge types.

`<` is also cheaper than `isless`. `fit(Histogram, v, edges)` on 1e6 Float64 observations with 100 bins, on an M-series Mac:

| edges | master | this PR |
|---|---|---|
| range | 10.4 ms | 8.1 ms |
| range, Float32 data | 10.7 ms | 8.1 ms |
| range, 2D | 23.0 ms | 17.4 ms |
| `Vector{Float64}` | 34.2 ms | 15.7 ms |

## Downstream packages

Bin assignment is unchanged for all inputs except that ranges containing `-0.0` are now accepted instead of throwing. The removed helper `_normalize_zero` was internal. A GitHub code search found no package referencing `_normalize_zero` or the removed `ArgumentError` message.

Consumers of `StatsBase.Histogram` that were checked and go through `fit`/`binindex` unchanged: Plots, StatsPlots, Makie (`hist`, `stephist`, datashader), AlgebraOfGraphics, UnicodePlots, PairPlots, BAT, FHist, RadiationSpectra, LegendSpecFits. None of them depends on the `-0.0` range check or on `_normalize_zero`.

Before merging, any downstream package that would break because of this change must be fixed first. None has been identified so far; if one turns up it should be listed here.

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -197,14 +197,6 @@ mutable struct Histogram{T<:Real,N,E} <: AbstractHistogram{T,N,E}
         closed == :right || closed == :left || error("closed must :left or :right")
         isdensity && !(T <: AbstractFloat) && error("Density histogram must have float-type weights")
         _edges_nbins(edges) == size(weights) || error("Histogram edge vectors must be 1 longer than corresponding weight dimensions")
-        # We do not handle -0.0 in ranges correctly in `binindex` for performance
-        # Constructing ranges starting or ending with -0.0 is very hard,
-        # and ranges containing -0.0 elsewhere virtually impossible,
-        # but check this just in case as it is cheap
-        foreach(edges) do e
-            e isa AbstractRange && any(isequal(-0.0), e) &&
-                throw(ArgumentError("ranges containing -0.0 not allowed in edges"))
-        end
         new{T,N,E}(edges,weights,closed,isdensity)
     end
 end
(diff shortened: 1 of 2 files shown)


nalimilan:
Sounds so simple that it's surprising this approach wasn't used in the first place. Maybe it was slow at the time.

Am I right that passing edges containing both `0.0` and `-0.0` throws an error in the current code? Is this tested?

andreasnoack:
It doesn't error. Neither on master nor this PR. I do think we should require strictly increasing edges, but I'll do that in another PR.

nalimilan (approved):

PR #1011, opened 2026-09-08
Return automatically chosen histogram edges as UniformEdges and fix silently dropped observations

### LINKED ISSUE #1009: Data loss with floating-point `Histogram`s
The function `fit(Histogram, v; nbins, closed)` silently drops observations at the extremes of the data. 

Here's a quick simulation illustrating the bug using 100 batches of measurements, each with 1 to 3 decimal digits. We take readings in batches with 2 to 5 readings per batch and observe that StatsBase `Histogram`s quietly drop observations; 11 of the 100 runs lose at least one reading.

```julia
julia> using StatsBase, Random

julia> Random.seed!(2026); # feel free to tweak this hyperparameter; including for reproducibility

julia> total_collected = total_counted = batches_with_data_loss = 0;

julia> for batch in 1:100
           n = rand(2:5)
           digs = rand(1:3)
           readings = Float32[round(rand(), digits=digs) for _ in 1:n]
           global total_collected += length(readings)

           h = fit(Histogram, readings, nbins=rand(1:12), closed=rand((:left, :right)))
           c = sum(h.weights)
           global total_counted += c
           global batches_with_data_loss += (c != length(readings))
       end

julia> total_collected # total readings collected
364

julia> total_counted # total readings countd
352

julia> readings_dropped = total_collected - total_counted # number of individual data readings that were silently dropped
12

julia> batches_with_data_loss # number of batches with data loss
11
```

## AI-Suggested fix

`binindex` decides membership by `searchsortedlast` / `searchsortedfirst` against the returned range, so the only check that can't drift is one made against that range. Build it, then re-run the endpoint fix-up against `first(r)` / `last(r)`:

```julia
    r = StepRangeLen(Base.TwicePrecision{Float64}((start, divisor)),
                     Base.TwicePrecision{Float64}((step, divisor)),
                     Int(len))
    # The loops above ran in `F` with plain division; the edges the caller
    # gets are evaluated by the TwicePrecision range, and the two can differ
    # by an ulp. Re-check against the range itself, which is what `binindex` uses.
    lo64, hi64 = Float64(lo), Float64(hi)
    if closed == :right
        while lo64 <= first(r); r = StepRangeLen(r.ref - r.step, r.step, length(r) + 1); end
        while last(r) < hi64;   r = StepRangeLen(r.ref, r.step, length(r) + 1); end
    else
        while lo64 < first(r);  r = StepRangeLen(r.ref - r.step, r.step, length(r) + 1); end
        while last(r) <= hi64;  r = StepRangeLen(r.ref, r.step, length(r) + 1); end
    end
    return r
```

This replaces the final `StepRangeLen(...)` expression in `histrange(lo::F, hi::F, n, closed)`. It is type-stable, passes the existing `test/hist.jl` (148 tests), and takes the drop rate to 0 in 40,000-fit sweeps for `Float32`, `Float16` and `Float64` on [0,1) and `Float32` on [0,10). The only visible change is that an affected range gains one bin at the end that was losing data, which is exactly what the existing fix-up loops already do in the cases they catch.

For `Float64` input the `F` arithmetic and the range's arithmetic are still different: the guard evaluates `(start + (len-1)*step)/divisor` with one Float64 division, while the range evaluates `ref + (len-1)*step` in `TwicePrecision`, and for `closed=:right` the latter can come out one ulp lower (`0.28:0.02:0.39999999999999997` for `[0.3, 0.4]`, `nbins=7`). The re-check above covers it with no special-casing.

Two narrower fixes were tried and are not enough:
- Promoting `lo`/`hi` to `Float64` at the top fixes `Float32`/`Float16` but leaves the `Float64` `closed=:right` drops untouched.
- Building the range in `TwicePrecision{F}` still drops `Float32` observations (the `F` guard and the `TwicePrecision{F}` range disagree the same way), leaves `Float64` unchanged, and throws a `BoundsError` for `Float16`.

Regression test: `sum(fit(Histogram, v; nbins=n, closed=c).weights) == length(v)` for every auto-edge call, over `Float16`/`Float32`/`Float64` and both `closed`.

andreasnoack:
Return automatically chosen histogram edges as `UniformEdges` and fix silently dropped observations

Fixes #1009. Builds on #1010.

## Problem

`fit(Histogram, v; nbins)` could silently drop observations at the extremes of the data (#1009). Two independent causes:

- For `Float32` and `Float16` data, `histrange` did its endpoint checks in the element type while returning `Float64` edges. `Float32(0.7)` is `0.69999998` as a `Float64`, so an edge at `0.7` passed the check in `Float32` and then excluded the observation in `binindex`.
- For `Float64` data with `closed=:right`, the last edge was checked as `(start + (len-1)*step)/divisor` but the returned `TwicePrecision` range evaluated it one ulp lower, e.g. `0.19999999999999998` for data `[0.0, 0.2]` with 9 bins.

Both come from the same source: the edges were represented as a float range whose elements are computed by arithmetic that does not reproduce the decimal numbers the edges are meant to be, and the checks were done against something other than what `binindex` compares against.

## Approach

The automatically chosen edges are now the decimal numbers `k * 10^e` for consecutive multiples `k` of a "nice" width (1, 2 or 5 times a power of ten, as before), each rounded to the nearest value of the data's float type `F`, and stored in a `Vector{F}`. There are only two representations involved, integers on the decimal side and `F` on the binary side, with one conversion per edge and no arithmetic on the converted values.

The edges are returned as `UniformEdges{F} <: AbstractVector{F}`, which wraps the stored vector together with the bin width. The type records what `histrange` knows and a plain vector would lose, namely that the bins have equal width and what that width is: `step(edges)` returns it, `binvolume` and `normalize` use it so bin volumes are exact rather than differences of rounded edges, `show` prints the edges as first edge, width, last edge like a range, and `binindex` estimates the bin from the width and corrects against the stored edges, which is faster than both the previous range arithmetic and a binary search. The elements are the stored `F` values, so no arithmetic progression is represented and the numerics do not depend on the type. User-supplied edges are untouched, so a non-range vector of edges continues to mean bins of possibly unequal width.

- Rounding each edge individually means no arithmetic progression has to be represented in `F`, so the same code works for `Float16`, `Float32`, `Float64`, `BigFloat` and any other `AbstractFloat`.
- An observation that is the rounding of the same decimal as an edge compares equal to that edge, so decimal-rounded data such as `round.(x, digits=2)` lands in the bin one expects: `0.7f0` is in the bin starting at `0.7f0`.
- The endpoints are adjusted by comparing `lo` and `hi` against the `F` edges themselves, i.e. against exactly what `binindex` sees, so containment holds by construction for every type.
- The width is clamped to three times the floating-point spacing of the data, which guarantees strictly increasing edges (consecutive decimals then round to distinct floats, also where the last edge lands in the next binade). Data spanning only a few ulps get correspondingly few bins.
- Identical values get a single bin of unit width with decimal edges, e.g. `[1.0, 2.0]` for data equal to `1.05`, instead of `[1.05, 2.05]`.

The conversion of `k * 10^e` to `F` is a single IEEE multiplication or division when `k` and `10^|e|` are exactly representable, which is correctly rounded and covers everything up to about `1e±22` for `Float64` and `1e±10` for `Float32`. Beyond that a correctly rounded power of ten is taken from a small table and at most three roundings occur, so edges are within two ulps of the decimal. Exact conversion for all exponents is the job of a decimal parser and is not attempted here. Note that `F(k // 10^d)` cannot be used: Base performs that division in `F` with rounded operands (JuliaLang/julia#49749).

`binindex` compares with `<` instead of `isless` since #1010, which treats `-0.0` and `0.0` as equal without the `_normalize_zero` workaround and lets NaN fall outside the edges as before.

## Behaviour changes

- `histrange`, and therefore `edges` of histograms fitted with `nbins`, returns `UniformEdges{F}` instead of a `StepRangeLen`. It is an `AbstractVector{F}` with `step`, but not an `AbstractRange`. For `Float32` and `Float16` data the element type is now `F` rather than `Float64`.
- All observations are inside the automatically chosen edges, for every float type.
- Identical values produce decimal edges around the value rather than edges at the value.
- `show(::Histogram)` prints user-supplied edge vectors with `:limit => true`; `UniformEdges` print as first edge, width, last edge.
- `midpoints` of `UniformEdges` are averages of neighbouring edges and may differ from the previous range-based midpoints by an ulp.
- `fit` with `nbins` is faster: on 1e6 `Float64` observations, 13.3 → 10.0 ms with 10 bins and 13.2 → 10.0 ms with 100 bins. `histrange` itself is about 20x faster (0.5 µs for 100 bins).

The docstring of `fit` now states that observations outside supplied `edges` are not counted, which has been the behaviour since 2014 but was undocumented, and that edges should have the element type of the data for decimal-rounded observations and edges to compare as intended.

## Downstream packages

Checked by reading their sources. Consumers that call `step(h.edges[1])` on `nbins` histograms keep working because `UniformEdges` supports `step`: PairPlots, HistTools, EvoId, Octofitter's legacy Makie extension. Would break:

- RadiationSpectra (`@assert isa(h.edges[1], AbstractRange)` on histograms passed to its plot recipes)

Ulp-level output changes only: AlgebraOfGraphics (its `midpoints(::AbstractVector)` method is now taken; its tests use `≈`), Makie's datashader, UnicodePlots labels for `Float32` data. Transparent: Plots, StatsPlots, Makie `hist`/`stephist`, BAT, FHist, LegendSpecFits. Gadfly and the Plotly packages do not use `StatsBase.Histogram`.

Before merging, RadiationSpectra should be fixed to not require an `AbstractRange`, and this should be released as a minor version.

## Tests

- Regression tests for #1009 over `Float16`, `Float32`, `Float64` and `BigFloat`, both `closed` values and 1 to 12 bins: edges have element type `F`, contain the data, and `sum(weights)` equals the number of observations.
- Element type, `step` and exact edges for `Float32`, `Float16` and `BigFloat` input; strictly increasing edges for data spanning 0 to 40 ulps at several magnitudes and up to 1000 requested bins; the two-ulp bound at extreme magnitudes; non-finite data throws; `binindex` on `UniformEdges` agrees with the generic search for all inputs including signed zeros and infinities, and NaN is outside the bins on both paths; `binvolume` equals `step`; `show` of `UniformEdges` and of long user-supplied edge vectors.
- The #1004 test asserting `first(r) == x` for identical values is replaced by containment and strictness checks, since the edge coinciding with the value was an artifact of the old implementation.
- Existing tests comparing `histrange` to ranges with `==` are unchanged and pass, as nearest-rounded edges coincide with the previous values.


test/hist.jl
```diff
@@ -96,6 +96,10 @@ end
     @test StatsBase.histrange([200.0,300.0], 10, :right) == 190.0:10.0:300.0
 
     @test @inferred(StatsBase.histrange(Int64[1:5;], 1, :left)) == 0:5:10
+    @test StatsBase.histrange(Int64[1:5;], 1, :left) isa StatsBase.UniformEdges{Float64}
+    @test step(StatsBase.histrange(Int64[1:5;], 1, :left)) == 5.0
```
nalimilan: Any idea why `Int64` is used here? Same below.

test/hist.jl
```diff
@@ -112,19 +116,101 @@ end
...
+    @test StatsBase.histrange(Float16[0.7, 0.8], 12, :left) isa StatsBase.UniformEdges{Float16}
+    @test StatsBase.histrange(BigFloat[0.7, 0.8], 12, :left) isa StatsBase.UniformEdges{BigFloat}
+    @test step(StatsBase.histrange(Float32[0.7, 0.8], 12, :left)) === 0.01f0
+    @test step(StatsBase.histrange(Float16[0.001, 0.002], 12, :left)) === Float16(0.0001)
+    @test StatsBase.histrange(Float32[0.7, 0.8], 12, :left) == Float32.(0.7:0.01:0.81)
+    @test StatsBase.histrange(Float32[0.7, 0.8], 12, :right) == Float32.(0.69:0.01:0.8)
+    @test StatsBase.histrange(Float16[0.001, 0.002], 12, :left) == Float16.(0.001:0.0001:0.0021)
+    # BigFloat[0.7, 0.8] would convert the Float64 literals, which lie below the decimals
+    @test StatsBase.histrange(parse.(BigFloat, ["0.7", "0.8"]), 12, :left) ==
+        parse.(BigFloat, ["0.7", "0.71", "0.72", "0.73", "0.74", "0.75", "0.76", "0.77", "0.78", "0.79", "0.8", "0.81"])
```
nalimilan:
```suggestion
    @test StatsBase.histrange(BigFloat.(["0.7", "0.8"]), 12, :left) ==
        BigFloat.(["0.7", "0.71", "0.72", "0.73", "0.74", "0.75", "0.76", "0.77", "0.78", "0.79", "0.8", "0.81"])
```

test/hist.jl
```diff
@@ -112,19 +116,101 @@ end
...
+    end
+    @test StatsBase.histrange(Float16[1e-4, 3e-4], 4, :left) == Float16[1e-4, 1.5e-4, 2e-4, 2.5e-4, 3e-4, 3.5e-4]
+
+    # Requested bins finer than the resolution of the data give fewer, strictly increasing edges
+    r = StatsBase.histrange([1e17, 1e17 + 16], 4, :right)
+    @test issorted(r, lt = <=) && first(r) < 1e17 && 1e17 + 16 <= last(r)
+    for F in (Float16, Float32, Float64), closed in (:left, :right), n in (1, 3, 10, 1000)
+        for x in (F(0.7), F(1), prevfloat(F(2)), F(-1000), floatmax(F) / 2), k in (0, 1, 2, 5, 40)
+            r = StatsBase.histrange(x, nextfloat(x, k), n, closed)
+            @test issorted(r, lt = <=)
```
nalimilan, line 158: Also test that `first(r) < x && `nextfloat(x, k) <= last(r)` as above?

test/hist.jl
```diff
@@ -112,19 +116,101 @@ end
...
+                          (Float16[1e-4, 3e-4], Float16[1e-4, 1.5e-4, 2e-4, 2.5e-4, 3e-4, 3.5e-4]))
+        r = StatsBase.histrange(v, 4, :left)
+        @test length(r) == length(expected)
+        @test all(abs(x - y) <= 2 * eps(y) for (x, y) in zip(r, expected))
+    end
+    @test StatsBase.histrange(Float16[1e-4, 3e-4], 4, :left) == Float16[1e-4, 1.5e-4, 2e-4, 2.5e-4, 3e-4, 3.5e-4]
+
+    # Requested bins finer than the resolution of the data give fewer, strictly increasing edges
+    r = StatsBase.histrange([1e17, 1e17 + 16], 4, :right)
+    @test issorted(r, lt = <=) && first(r) < 1e17 && 1e17 + 16 <= last(r)
```
nalimilan, line 152: Test length to check the "fewer" part of the comment?

src/hist.jl
```diff
@@ -32,77 +53,135 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
...
     end
-    StepRangeLen(Base.TwicePrecision{Float64}((start, divisor)),
-                 Base.TwicePrecision{Float64}((step, divisor)),
-                 Int(len))
+    return UniformEdges(F[edge(k) for k in kfirst:m:klast], stepF)
+end
+
+# Integer type for the multiples of the width: they are bounded by 2^precision(F) / 3
+_multiple_type(::Type{<:Union{Float16,Float32,Float64}}) = Int
+_multiple_type(::Type{<:AbstractFloat}) = BigInt
```
nalimilan: Would there be a way to check whether `Int` would be enough for custom float types instead of using `BigInt`? Maybe by comparing `maxintfloat` to `typemax(Int)`?

src/hist.jl
```diff
@@ -32,77 +53,135 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
...
+        # 10^(-e) overflows F although k * 10^e may not: divide by an exact power first
+        a = _maxexp10(F)
+        return F(k) / F(10)^a / _pow10(F, -e - a)
+    end
+end
+
+# 10^d for d ≥ 0 rounded to the nearest F, from tables for the IEEE types (Inf beyond floatmax)
+const _POW10_FLOAT64 = Float64[Float64(big(10)^d) for d in 0:308]
+const _POW10_FLOAT32 = Float32[Float32(big(10)^d) for d in 0:38]
+const _POW10_FLOAT16 = Float16[Float16(big(10)^d) for d in 0:4]
```
nalimilan:
```suggestion
const _POW10_FLOAT64 = [Float64(big(10)^d) for d in 0:308]
const _POW10_FLOAT32 = [Float32(big(10)^d) for d in 0:38]
const _POW10_FLOAT16 = [Float16(big(10)^d) for d in 0:4]
```

src/hist.jl
```diff
@@ -32,77 +53,135 @@ function histrange(v::AbstractArray{T}, n::Integer, closed::Symbol=:left) where
...
+_pow10(::Type{Float64}, d::Integer) = d < length(_POW10_FLOAT64) ? @inbounds(_POW10_FLOAT64[d + 1]) : Inf
+_pow10(::Type{Float32}, d::Integer) = d < length(_POW10_FLOAT32) ? @inbounds(_POW10_FLOAT32[d + 1]) : Inf32
+_pow10(::Type{Float16}, d::Integer) = d < length(_POW10_FLOAT16) ? @inbounds(_POW10_FLOAT16[d + 1]) : Inf16
+_pow10(::Type{F}, d::Integer) where F<:AbstractFloat = F(10)^d
```
nalimilan: Shouldn't this be
```suggestion
_pow10(::Type{F}, d::Integer) where F<:AbstractFloat = F(big(10)^d)
```

src/hist.jl
```diff
@@ -380,13 +478,28 @@ Fit a histogram to `data`.
...
   the edges of the bins along each dimension. If no edges are provided, they are chosen
   so that approximately `nbins` bins of equal width are constructed along each dimension.
 
 !!! note
     In most cases, the number of bins will be `nbins`. However, to ensure that the bins have
-    equal width, more or fewer than `nbins` bins may be used.
+    equal width, more or fewer than `nbins` bins may be used. The automatically chosen bin
+    width is a "nice" decimal number (1, 2 or 5 times a power of ten) and the edges are
+    multiples of it rounded to the floating point type of the data, returned as
+    [`UniformEdges`](@ref), a vector of edges which also records the bin width as `step`.
```
nalimilan: This reference makes the docs fail. You probably need to include this somewhere in the manual.

nalimilan (commented):
I'm not really able to check the core logic, but here are a few comments.

test/hist.jl
andreasnoack: No reason: the neighbouring pre-existing tests used it and I mirrored the style. All changed to plain `Int` now.

_(Reply drafted by Claude Code at my request.)_

test/hist.jl
andreasnoack: Done.

_(Reply drafted by Claude Code at my request.)_

test/hist.jl
andreasnoack: Added for both `closed` values.

_(Reply drafted by Claude Code at my request.)_

test/hist.jl
andreasnoack: Added: three edges, so two bins for the four requested.

_(Reply drafted by Claude Code at my request.)_

src/hist.jl
andreasnoack: Good idea, done. The multiples are bounded by 5/3 of `maxintfloat(F)`, hence the factor 2 in the comparison. It constant-folds for the IEEE types, so the explicit methods could go.

_(Reply drafted by Claude Code at my request.)_

src/hist.jl
andreasnoack: Done.

_(Reply drafted by Claude Code at my request.)_

src/hist.jl
andreasnoack: Yes, done.

_(Reply drafted by Claude Code at my request.)_

src/hist.jl
andreasnoack: Added `StatsBase.UniformEdges` to the manual; the reference has to be qualified since the type is not exported. The docs build passes locally now.

_(Reply drafted by Claude Code at my request.)_

src/hist.jl
```diff
@@ -130,9 +130,8 @@ function histrange(lo::F, hi::F, n::Integer, closed::Symbol=:left) where F<:Abst
     return UniformEdges(F[edge(k) for k in kfirst:m:klast], stepF)
 end
 
-# Integer type for the multiples of the width: they are bounded by 2^precision(F) / 3
-_multiple_type(::Type{<:Union{Float16,Float32,Float64}}) = Int
-_multiple_type(::Type{<:AbstractFloat}) = BigInt
+# Integer type for the multiples of the width, which are bounded by 5/3 * maxintfloat(F)
+_multiple_type(::Type{F}) where F<:AbstractFloat = 2 * maxintfloat(F) <= typemax(Int) ? Int : BigInt
```
nalimilan, line 134: Is this supposed to be `5/3 * maxintfloat(F)`?

src/hist.jl
nalimilan: Maybe needs a test then?

nalimilan (commented):
Looks good to me, but there are some uncovered lines according to Codecov.

